PR #24, opened 2025-05-20
Add documentation for expectations and standards related to model formats/configs

markurtz:
## Summary

Documentation and examples added to standardize the model serialization formats, intended compatibility, and deserialization formats.

## Testing

Self review, PR reviews

### DIFF
--- a/docs/architecture/model_formats.md
+++ b/docs/architecture/model_formats.md
@@ -4,7 +4,7 @@
 
 In this document, you will learn:
 
-1. The structure and organization of Speculators model formats
+1. The structure and organization of the Speculators model formats
 2. How to save, load, and share speculator models across the ecosystem
 3. The design principles that guide Speculators' model format implementation
 
@@ -16,21 +16,21 @@ Before reading this document, you should be familiar with the following concepts
 
 ## Overview
 
-Speculators employs a model format designed with ecosystem compatibility and user experience as top priorities. This format serves as a standard and bridge between various speculative decoding implementations and the broader machine learning ecosystem, particularly focusing on compatibility with Hugging Face's model hub, Speculators' own implementations built on top of Transformers, and other platforms like vLLM. With the Hugging Face format as a foundation, Speculators enables extensibility and usability from practioners consuming existing algorithms to researchers developing new ones.
+Speculators employs a model format designed with ecosystem compatibility and user experience as top priorities. This format serves as a standard and bridge between various speculative decoding implementations and the broader machine learning ecosystem, particularly focusing on compatibility with Hugging Face's model hub, Speculators' implementations built on top of Transformers, and other platforms like vLLM. With the Hugging Face format as a foundation, Speculators enables extensibility and usability for practitioners consuming existing algorithms and researchers developing new ones.
 
-The core components that make up the Speculators model format are the following, which are explained in detail in dedicated sections below:
+The core components that make up the Speculators model format are the following:
 
 - Hugging Face compatible directory structure and file formats
 - Speculators-specific extensions through the `speculators_config` key in the `config.json` file
 - Standardized interfaces for saving, loading, and pushing models to the Hugging Face Hub
 
-Complete examples are provided at the end of this document for popular speculative decoding algorithms.
+At the end of this document, examples of configuration files for popular algorithms are provided.
 
 ## Components
 
 ### Base Model Format
 
-The base model format for Speculators builds directly on the standard Hugging Face model format and represents the specific draft model, ensuring compatability with existing tools, libraries, and workflows. A typical Speculators model directory contains:
+The base model format for Speculators builds directly on the standard Hugging Face model format and represents the specific draft model, ensuring compatibility with existing tools, libraries, and workflows. A typical Speculators model directory contains:
 
 ```
 model_directory/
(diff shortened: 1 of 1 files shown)


docs/architecture/model_formats.md
```diff
@@ -0,0 +1,544 @@
...
+- `"sample"`: Samples from the top-p probability mass of the predicted tokens at each step
+- `"tree"`: Generates a tree of possible tokens for more likely matches at each step
+
+Example usage in a config:
+
+```json
+"proposal_methods": [
+    {
+        "proposal_type": "greedy",
+        "...": "..."
```
markurtz: draft/token length

docs/architecture/model_formats.md
```diff
@@ -0,0 +1,544 @@
...
+    },
+    {
+        "proposal_type": "sample",
+        "...": "..."
```
markurtz: "max_new_tokens": 5,
"temperature": 0.8,
"top_p": 0.5

docs/architecture/model_formats.md
```diff
@@ -0,0 +1,544 @@
...
+    }
+],
```
markurtz: {
    "proposal_type": "tree",
    "max_total_tokens": #,
    "max_depth": #,
    "max_branches_per_node; top_k": #
}

docs/architecture/model_formats.md
```diff
@@ -0,0 +1,544 @@
...
+
+```json
+{
+    "architectures": ["MLPDraftModel"],
+    "model_type": "mlp_draft_model",
+    "torch_dtype": "bfloat16",
+    "transformers_version": "4.35.0",
+    "speculators_version": "0.1.0",
+    "inputs": ["input_ids"],
+    "...": "..."
```
markurtz, line 67: SpecDec -- indepdent original model config.json
MLPSpeculator -- go through more
Transformer Eagle -- inputs: [input_ids, end_net_output (before layernorm)], fc -> LlamaLayer(w/o layernorm) (configurable layer type) -> naive tree decoding
Transformer Eagle 2 -- inputs: [input_ids, end_net_output (before layernorm)], fc -> LlamaLayer(w/o layernorm) (configurable layer type) -> tree decoding
HASS --  inputs: [input_ids, end_net_output (before layernorm)], fc (plus bias) -> LlamaLayer(w/o layernorm) (configurable layer type) -> tree decoding
Transformer Eagle 3 --  inputs: [input_ids, beginning layer outputs, middle layer outputs, end layer outputs], [fc->layer norm] (hidden states) + [input embeddings->LayerNorm] -> 2X LlamaLayer(LayerNorm pull output before to feed to next iteration, total output goes to projection head) (configurable layer type) -> tree decoding

docs/architecture/model_formats.md
```diff
@@ -0,0 +1,690 @@
...
+            },
+            {
+                "proposal_type": "tree",
+                "tree_type": "static",
+                "initial_branching_factor": 4,
+                "branching_factor": 2,
+                "depth": 5
+            }
+        ],
+        "default_proposal_method": "tree",
```
markmc, line 306: Just using this EAGLE example to talk this through ...

Currently, vLLM only supports greedy sampling with EAGLE. See https://github.com/vllm-project/vllm/blob/58738772410c5e0d60b61db39538a9b313d2d7ad/vllm/v1/spec_decode/eagle.py#L182

The discussion in https://github.com/vllm-project/vllm/pull/16899 suggests that if vLLM does support random sampling, that it should follow the sampling params of the target model in order to have the best chance of matching its distribution

Tree decoding is proposed in https://github.com/vllm-project/vllm/pull/17560 - tree depth and `num_spec_expand` are configurable via `--speculative-config`

So ...

What does it mean for the creator of a speculator model to list proposal methods like this?

Is this a recommendation that tree decoding with those parameters is optimal?

Might there be multiple tree decoding "profiles" with different values? How would a user choose between them?

What if the inference engine does not support that proposal method? Can they fall back to greedy?

What does it mean to specify `draft_tokens = 5`? Is this also a recommendation? It should be used unless the user specifies their own value?

HTH

docs/architecture/model_formats.md
```diff
@@ -0,0 +1,690 @@
...
+        "verifier": {
+            "name_or_path": "meta-llama/Llama-3.1-8B-Instruct",
+            "architectures": ["LlamaForCausalLM"],
```
markmc, line 309: Thinking through the implications of this ...

Is this supposed to override the verifier model config (assuming they don't match?)

What would the use case be to override?

I'd imagine it would be quite tricky to make this override work in vLLM - by the time we load this config, we've probably done some setup of the verifier based on its config

If it's not an override, does vLLM just ignore it? That would be a bit weird ...

docs/architecture/model_formats.md
```diff
@@ -0,0 +1,690 @@
...
+#### Algorithm
+
+The `algorithm` field is a required string that specifies the speculative decoding algorithm used by the model. It serves as a primary identifier for the algorithm with which the speculator model was created and intended to be used. It additionally allows the Speculators library and other tools to automatically load the correct implementation and associated utilities when a model is loaded.
+
+This field must match one of the supported algorithms in the Speculators library, such as:
+
+- `"eagle"`, `"eagle_2"`, `"eagle_3"` - Eagle speculator variants based on Transformer architecture for the draft model
+- `"hass"` - Similar to Eagle based on the Transformer architecture for the draft model
+- `"mlp_speculator"` - Based on a multi-layer perceptron (MLP) architecture for the draft model
+- `"specdec"` - An independent speculator model
```
markmc, line 115: It would be super convenient if we could match the existing names used in vLLM

```
+ALGORITHMS = [
+    "eagle",
+    "eagle3",
+    "medusa",
+    "mlp_speculator",
+    "deepseek_mtp",
```

rahul-tuli (changes requested):
Could we update these docs, they seem stale

PR #26, opened 2025-05-22
Add base speculators configuration classes and tests

markurtz:
## Summary

Added the base required configuration classes for speculators and associated unit and integration tests:
- SpeculatorModelConfig (root)
- SpeculatorsConfig (contained in model config)
- TokenProposalConfig (contained in speculators config)
- VerifierConfig (contained in speculators config)

## Testing

Added tests to cover the newly added configurations. Tests will need to be updated in later revisions to test subsequently added functionality such as from_pretrained.

### DIFF
--- a/src/speculators/__init__.py
+++ b/src/speculators/__init__.py
@@ -1,0 +1,35 @@
+"""
+Speculators: A Unified Library for Speculative Decoding Algorithms for LLMs
+
+Speculators provides a standardized framework for creating, representing, and
+storing speculative decoding algorithms for large language model (LLM) inference.
+It enables developers to implement and productize various speculative decoding
+approaches with a consistent interface, making them ready for integration with
+LLM inference servers like vLLM.
+
+Speculative decoding is a technique that can significantly improve LLM inference
+performance by predicting multiple tokens with a smaller, speculative model and
+then verifying the predictions with the original, larger model.
+This approach tradesoff extra computation for reduced latency, making it suitable
+for real-time applications on deployments that are not compute-constrained.
+
+The library offers a modular architecture with components for:
+- Standardized interfaces for working with speculative decoding algorithms that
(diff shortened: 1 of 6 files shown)


rahul-tuli (approved):

src/speculators/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+Speculators provides a standardized framework for creating, representing, and
+storing speculative decoding algorithms for large language model (LLM) inference.
+It enables developers to implement and productize various speculative decoding
+approaches with a consistent interface, making them ready for integration with
+LLM inference servers like vLLM.
+
+Speculative decoding is a technique that can significantly improve LLM inference
+performance by predicting multiple tokens with a smaller, speculative model and
+then verifying the predictions with the original, larger model.
+This approach tradesoff extra computation for reduced latency, making it suitable
```
anmarques:
```suggestion
This approach trades off extra computation for reduced latency, making it suitable
```

src/speculators/config.py
```diff
@@ -0,0 +1,260 @@
...
+    """
+    The base config for a verifier model which defines the parameters that are required
+    to either load the original verifier model or validate compatibility with a new
+    verifier based on the the architecture and tokenizers/processor properties.
+    It provides convenience methods to extract the required parameters from a
+    PretrainedConfig object.
+    """
+
+    @classmethod
+    def from_verifier_config(
```
anmarques, line 67: Is this supposed to take in a HF `PretrainedConfig`?

If so the name is a bit confusing, since we call the class `VerifierConfig` and the method `from_verifier_config`, at first I thought we were basically cloning a `VerifierConfig` object.

Instead I would setup a method called  `from_pretrained` that would take the `name_or_path` and internally load the model config.

markurtz: TODO: wipe out in a follow up diff

PR #37, opened 2025-06-26
feat: Implement Eagle speculator model

rahul-tuli:
# PR 2: Implement Eagle Speculator Model

## Summary
Implements the EagleSpeculator model supporting EAGLE1, EAGLE2, and HASS architectural variants through configuration. This provides a unified implementation for all Eagle-based speculators.

### Model Architecture
- **EagleSpeculator**: Complete PyTorch model implementation
  - Token embeddings with optional post-embedding layernorm
  - Fusion layer combining embeddings and verifier hidden states
  - Single LlamaDecoderLayer for processing
  - Output projection with optional pre-lm-head layernorm
  - Weight tying between embed_tokens and lm_head

#### Initialization
```python
def __init__(self, config: EagleSpeculatorConfig):
    # Core components
    self.embed_tokens = nn.Embedding(vocab_size, hidden_size)
    self.fc = nn.Linear(2 * hidden_size, hidden_size, bias=config.fusion_bias)
    self.layers = nn.ModuleList([LlamaDecoderLayer(...)])
    self.lm_head = nn.Linear(hidden_size, vocab_size, bias=False)
    
    # Optional layernorms (for HASS variant)
    if config.layernorms:
        self.post_embedding_layernorm = LlamaRMSNorm(hidden_size)
        self.pre_lm_head_layernorm = LlamaRMSNorm(hidden_size)
    
    # Weight tying
    self.lm_head.weight = self.embed_tokens.weight
```

#### Forward Pass
1. Embed input tokens
2. Apply optional post-embedding layernorm
3. Fuse embeddings with verifier hidden states
4. Process through decoder layer with rotary embeddings
5. Apply optional pre-lm-head layernorm
6. Project to vocabulary space

### Configuration Support
- **EAGLE1/EAGLE2**: `layernorms=False, fusion_bias=False`
- **HASS standard**: `layernorms=False, fusion_bias=True`
- **HASS with layernorms**: `layernorms=True, fusion_bias=False`

### Technical Decisions
- Single decoder layer (layers[0]) with Identity for input_layernorm when layernorms=False
- Reuses LlamaDecoderLayer from transformers for consistency
- Proper attention mask and position ID handling
- Supports HuggingFace generation methods

### Dependencies
- Added `torch` to pyproject.toml dependencies

## Testing
Model functionality is tested through the converter e2e tests in PR 3, which verify:
- Model instantiation
- Forward pass execution
- Config preservation through save/load cycles

## Review Notes
- The model reuses components from transformers
- All architectural variants use the same code path with config-based branching


src/speculators/models/eagle.py
```diff
@@ -72,3 +92,362 @@ class EagleSpeculatorConfig(SpeculatorModelConfig):
...
+            )
+
+        return super().from_dict(config_dict, **kwargs)
+
+
+class EagleSpeculator(PreTrainedModel, GenerationMixin):
+    """
+    Eagle speculator model for speculative decoding.
+
+    This model implements EAGLE1, EAGLE2, and HASS variants through configuration.
```
brian-dellabetta: nit: just to indicate this is an acronym from research teams that we are integrating (it seems)
```suggestion
    This model implements EAGLE1, EAGLE2, HASS, and TTT variants through configuration.
```

src/speculators/models/eagle.py
```diff
@@ -72,3 +92,362 @@ class EagleSpeculatorConfig(SpeculatorModelConfig):
...
+        """
+        return self.embed_tokens
+
+    def set_input_embeddings(self, value):
+        """
+        Set input embeddings layer.
+
+        :param value: New embedding layer
+        """
+        self.embed_tokens = value
```
brian-dellabetta: nit: might be a good place for a getter/setter:
```suggestion
    @property
    def input_embeddings(self):
        """
        Get input embeddings layer.

        :return: Embedding layer
        """
        return self.embed_tokens

    @input_embedding.setter
    def input_embeddings(self, value):
        """
        Set input embeddings layer.

        :param value: New embedding layer
        """
        self.embed_tokens = value
```

src/speculators/config.py
```diff
@@ -79,7 +79,7 @@ class VerifierConfig(BaseModel):
 
     @classmethod
     def from_verifier_config(
```
rahul-tuli: Could we rename this to from_config?

src/speculators/models/eagle.py
```diff
@@ -112,6 +169,351 @@ def validate_transformer_layer_config(cls, value: Any) -> PretrainedConfig:
...
+
+    Example:
+        ```python
+        from speculators import SpeculatorsConfig, VerifierConfig
+        from speculators.models import EagleSpeculator, EagleSpeculatorConfig
+        from speculators.proposals import GreedyTokenProposalConfig
+        from transformers import AutoConfig, AutoTokenizer
+
+        config = EagleSpeculatorConfig(
+            transformer_layer_config=AutoConfig.from_pretrained("meta-llama/Llama-3.1-8B-Instruct"),
```
rahul-tuli, line 212: misleading?

src/speculators/models/eagle.py
```diff
@@ -112,6 +169,351 @@ def validate_transformer_layer_config(cls, value: Any) -> PretrainedConfig:
...
+            speculators_config=SpeculatorsConfig(
+                algorithm="eagle_v1",
```
rahul-tuli:
```suggestion
                algorithm="eagle",
```

src/speculators/models/eagle.py
```diff
@@ -1,85 +1,135 @@
...
+        from speculators.proposals import GreedyTokenProposalConfig
+        from transformers import AutoConfig
 
-    This unified configuration supports both EAGLE1 and HASS variants through
-    configurable parameters, allowing a single model implementation to handle
-    both architectures.
+        config = EagleSpeculatorConfig(
+            transformer_layer_config=AutoConfig.from_pretrained("meta-llama/Llama-3.1-8B-Instruct"),
+            speculators_config=SpeculatorsConfig(
+                algorithm="eagle_v1",
```
rahul-tuli:
```suggestion
                algorithm="eagle",
```

src/speculators/models/eagle.py
```diff
@@ -1,85 +1,135 @@
...
+        from speculators import SpeculatorsConfig, VerifierConfig
+        from speculators.models import EagleSpeculatorConfig
+        from speculators.proposals import GreedyTokenProposalConfig
+        from transformers import AutoConfig
 
-    This unified configuration supports both EAGLE1 and HASS variants through
-    configurable parameters, allowing a single model implementation to handle
-    both architectures.
+        config = EagleSpeculatorConfig(
+            transformer_layer_config=AutoConfig.from_pretrained("meta-llama/Llama-3.1-8B-Instruct"),
```
rahul-tuli, line 59: Won't this have number of hidden layers set to >1

src/speculators/models/eagle.py
```diff
@@ -112,6 +169,351 @@ def validate_transformer_layer_config(cls, value: Any) -> PretrainedConfig:
...
+                    architectures=["LlamaForCausalLM"],
+                )
         )
+        speculator = EagleSpeculator(
+            config, verifier=verifier, verifier_attachment_mode="full"
+        )
+
+        tokenizer = AutoTokenizer.from_pretrained("meta-llama/Llama-3.1-8B-Instruct")
+        input_ids = tokenizer.encode("The future of AI is", return_tensors="pt")
+        outputs = speculator.generate(
```
anmarques: I think the example is a bit confusing. Is it supposed to show the initialization of a random speculator for training or of a pretrained speculator ready for inference?

It looks to me it's the latter, but in this case shouldn't we use `from_pretrained`?

src/speculators/models/eagle.py
```diff
@@ -112,6 +169,351 @@ def validate_transformer_layer_config(cls, value: Any) -> PretrainedConfig:
...
+            perform generation until a full verifier is attached.
+        :return: The PreTrainedModel instance for the verifier that was attached.
+        """
+        verifier = super().attach_verifier(
+            verifier=verifier,
+            mode=mode,
+        )
+
+        # Extract layers from the verifier model
+        self.embed_tokens = verifier.embed_tokens  # type: ignore[attr-defined]
```
anmarques: This will fail for some models. We can land as is, but we'll need to fix it. For instance, for Mistral this should look like

self.embed_tokens = verifier.model.language_model.embed_tokens
self.rotary_emb = verifier.model.language_model.rotary_emb
self.lm_head = verifier.lm_head

anmarques (commented):
Looks good to land from me. I didn't review the tests.

I added a couple of comments, but no blockers to landing

src/speculators/models/eagle.py
markurtz: What do you mean by misleading here?

src/speculators/models/eagle.py
markurtz: Technically, but given the model definition and the naming, we're only constructing one layer. The only other alternative I see is over complicating a sub selection of the LlamaConfig params where we are restricting them. I'd rather target ease of use in this sense of just setting a config and not needing to worry about selecting specific args. Let me know if you have other ideas

src/speculators/models/eagle.py
markurtz: We don't have any models pushed up yet that would work with the pathways for this example. Once we have one up, we can swap for from_pretrained with a workable stub set. For now, though, I'll remove the tokenizer and generate execution

src/speculators/models/eagle.py
markurtz: yep, there's quite a bit we'll need to do to start generalizing and supporting othe rmodels. This extraction here is part of it, the forward may need to be edited, and as we support more transformer layer types, we'll need to expand those as well

markurtz (approved):
Unblocking for my request changes now

anmarques (approved):

PR #47, opened 2025-07-09
chore: Update pr comment behavior

fynnsu:
Currently every pr commit results in a new comment like https://github.com/neuralmagic/speculators/pull/39#issuecomment-3020058659 which clutters the pr.

This pr changes the behavior so that new builds edit the existing comment instead of creating a new one. Previous comments can be seen by looking at the revision history of the comment.

Note: this uses the existing Github token, instead of the automation bot token but could be updated to use that token instead if that's preferred.

### DIFF
--- a/.github/workflows/development.yml
+++ b/.github/workflows/development.yml
@@ -117,23 +117,21 @@ jobs:
           compression-level: 6
           if-no-files-found: error
           retention-days: 30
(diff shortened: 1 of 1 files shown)


markurtz (approved):
Great, thanks @fynnsu!

PR #49, opened 2025-07-10
feat: Expand Eagle Speculator to Support Multiple Transformer Layer Types

### LINKED ISSUE #44: Feature Request: Expand Eagle Speculator to Support Multiple Transformer Layer Types
### Summary

Expand `EagleSpeculator` and `EagleSpeculatorConfig` in `src/speculators/models/eagle.py` to support creating **transformer layer types beyond just LlamaDecoder layers**, targeting the **top 5–10 popular architectures** currently used in Hugging Face Transformers.

### Context

Currently, EagleSpeculator only utilizes a Llama decoder layer for the speculator architecture. Extending support to additional transformer layer types will enable broader applicability of EagleSpeculator across various model families in production and research workflows.

### Acceptance Criteria

**Implementations**

- Update `EagleSpeculatorConfig` to:
    - Allow specification of transformer layer types beyond LlamaDecoder.
    - Include configuration validation for the newly supported architectures.
- Update `EagleSpeculator` to:
    - Create and initialize transformer layers matching the specified architecture type.
    - Maintain compatibility with existing workflows and ensure seamless integration.

**Targeted Architectures**

Support the top **5–10 most popular architectures**, such as:

- Llama (existing)
- Mistral
- Qwen
- DeepSeek
- Mixtral
- Gemma
- Granite

**Tasks**

- [ ]  Research and identify transformer layer classes within Hugging Face for each targeted architecture.
- [ ]  Update `EagleSpeculatorConfig` to include architecture type selection.
- [ ]  Update `EagleSpeculator` to construct the selected transformer layer type correctly.
- [ ]  Update or create relevant tests in:
    - `tests/unit/models/test_eagle_config.py`
    - `tests/unit/models/test_eagle_model.py`
- [ ]  Ensure compatibility with `SpeculatorModelConfig.from_pretrained` and `SpeculatorModel.from_pretrained`.

### Out of Scope

- Implementing **forward pass**, **generate pass**, or **algorithm factory** unless that functionality has already been developed and merged into `main`.

fynnsu:
Closes #44 

Expands support for other decoder layer types (MistralDecoderLayer, Qwen3DecoderLayer, etc.)
> # Tasks 
> * [x]   Research and identify transformer layer classes within Hugging Face for each targeted architecture.
> * [x]   Update `EagleSpeculatorConfig` to include architecture type selection.

This is handled using the existing `transformer_layer_architecture` field to specify the decoder layer class. 

The `transformer_layer_config` must also match the decoder layer type. i.e. to use `LlamaDecoderLayer` `transformer_layer_config` must be an instance of `LlamaConfig`, for `MistralDecoderLayer` it must be an instance of `MistralConfig`, etc.

> * [x]   Update `EagleSpeculator` to construct the selected transformer layer type correctly.

We find the corresponding decoder layer class (and also layer norm class), using the config class to determine the model type / import path. This generalizes the approach so that we can use **any** decoder layer and config combination in the transformers library.

> * [x]   Update or create relevant tests in:
>     * `tests/unit/models/test_eagle_config.py`
>     * `tests/unit/models/test_eagle_model.py`

Added explicit tests for the architectures listed in #44 (Llama, Mistral, Qwen, DeepSeek, Mistral, Gemma, Granite)

> * [x]   Ensure compatibility with `SpeculatorModelConfig.from_pretrained` and `SpeculatorModel.from_pretrained`.

### DIFF
--- a/src/speculators/models/eagle.py
+++ b/src/speculators/models/eagle.py
@@ -10,6 +10,7 @@ Classes:
     EagleSpeculator: Main model implementation for EAGLE/HASS speculators
 """
 
+import importlib
 import os
 from typing import Any, ClassVar, Literal, Optional, Union
 
@@ -19,26 +20,7 @@ from torch import nn
 from transformers import AutoConfig, PretrainedConfig, PreTrainedModel
 from transformers.modeling_attn_mask_utils import _prepare_4d_causal_attention_mask
 from transformers.modeling_outputs import CausalLMOutputWithPast
-from transformers.models.deepseek_v3.modeling_deepseek_v3 import (
-    DeepseekV3DecoderLayer,
-    DeepseekV3RMSNorm,
-)
-from transformers.models.gemma.modeling_gemma import GemmaDecoderLayer, GemmaRMSNorm
-from transformers.models.granite.modeling_granite import (
-    GraniteDecoderLayer,
-    GraniteRMSNorm,
-)
 from transformers.models.llama.configuration_llama import LlamaConfig
-from transformers.models.llama.modeling_llama import LlamaDecoderLayer, LlamaRMSNorm
-from transformers.models.mistral.modeling_mistral import (
-    MistralDecoderLayer,
-    MistralRMSNorm,
-)
-from transformers.models.mixtral.modeling_mixtral import (
-    MixtralDecoderLayer,
-    MixtralRMSNorm,
-)
-from transformers.models.qwen3.modeling_qwen3 import Qwen3DecoderLayer, Qwen3RMSNorm
 from typing_extensions import Self
 
 from speculators import SpeculatorModel, SpeculatorModelConfig
@@ -47,16 +29,6 @@ __all__ = [
     "EagleSpeculator",
     "EagleSpeculatorConfig",
 ]
-
-ARCHITECTURE_CLASSES = {
-    "LlamaDecoderLayer": (LlamaDecoderLayer, LlamaRMSNorm),
-    "MistralDecoderLayer": (MistralDecoderLayer, MistralRMSNorm),
-    "Qwen3DecoderLayer": (Qwen3DecoderLayer, Qwen3RMSNorm),
-    "GemmaDecoderLayer": (GemmaDecoderLayer, GemmaRMSNorm),
-    "MixtralDecoderLayer": (MixtralDecoderLayer, MixtralRMSNorm),
-    "DeepseekV3DecoderLayer": (DeepseekV3DecoderLayer, DeepseekV3RMSNorm),
-    "GraniteDecoderLayer": (GraniteDecoderLayer, GraniteRMSNorm),
-}
 
 
 @SpeculatorModelConfig.register("eagle")
@@ -296,13 +268,6 @@ class EagleSpeculator(SpeculatorModel):
                 f"got {type(config)} instead."
             )
 
-        if config.transformer_layer_architecture not in ARCHITECTURE_CLASSES:
-            raise ValueError(
-                f"Invalid transformer layer architecture: {config.transformer_layer_architecture}. "
-                f"Must be one of: {list(ARCHITECTURE_CLASSES.keys())}"
-            )
-        self._architecture = config.transformer_layer_architecture
-
         # Initialize model parameters from config
         self.vocab_size = config.transformer_layer_config.vocab_size
         self.hidden_size = config.transformer_layer_config.hidden_size
@@ -320,6 +285,7 @@ class EagleSpeculator(SpeculatorModel):
             verifier_attachment_mode=verifier_attachment_mode,
         )
 
+        self._architecture_classes = self._import_model_classes()
         # Initialize layers based on the configuration
         self.embedding_layernorm: Optional[nn.Module] = self._create_layernorm()
         self.fusion_fc: nn.Linear = nn.Linear(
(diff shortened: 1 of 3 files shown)


src/speculators/models/eagle.py
```diff
@@ -289,6 +285,7 @@ def __init__(
             verifier_attachment_mode=verifier_attachment_mode,
         )
 
+        self._architecture_classes = self._import_model_classes()
```
markurtz: NIT: can we break these out into separate classes set at the instance level such as layernorm_class and transformer_layer class? 

If not the above, then we should implement a quick named tuple for this rather than a generic tuple to make the later references and indexing easier to read and follow

src/speculators/models/eagle.py
```diff
@@ -532,7 +529,45 @@ def _create_transformer_layer(self) -> nn.Module:
         return layer
 
     def _layernorm_class(self) -> type[nn.Module]:
-        return LlamaRMSNorm
+        return self._architecture_classes[1]
```
markurtz: Since these are just returning from the instance variable now and combined with the earlier comment on either instance variables for each class or a named tuple, I think removing this method would be better to reduce code complexity

src/speculators/models/eagle.py
```diff
@@ -532,7 +529,45 @@ def _create_transformer_layer(self) -> nn.Module:
...
 
     def _transformer_layer_class(self) -> type[nn.Module]:
-        return LlamaDecoderLayer
+        return self._architecture_classes[0]
```
markurtz: Since these are just returning from the instance variable now and combined with the earlier comment on either instance variables for each class or a named tuple, I think removing this method would be better to reduce code complexity

src/speculators/models/eagle.py
```diff
@@ -532,7 +529,45 @@ def _create_transformer_layer(self) -> nn.Module:
...
+        Get the decoder layer class and layer normalization class for a given config
+        and decoder layer name. Uses the config qualified name to find the corresponding
+        modeling module to import the decoder layer and normalization classes from.
+        """
+        # e.g. LlamaConfig + LlamaDecoderLayer
+        # -> get qualname: transformers.models.llama.configuration_llama.LlamaConfig
+        # -> expected path: transformers.models.llama.modeling_llama.LlamaDecoderLayer
+        # -> expected path: transformers.models.llama.modeling_llama.LlamaRMSNorm
+
+        module_path = self.config.transformer_layer_config.__class__.__module__
```
markurtz: I'm worried about how delicate this is since it's an assumption that is not enforced by the Transformers repo. A pathway that better matches Transformers' internal logic is determining the module that contains the model class by utilizing the MODEL_FOR_CAUSAL_LM_MAPPING from transformers and looking up by either the config class or the architecture contained in the architectures property.

src/speculators/models/eagle.py
```diff
@@ -532,7 +529,45 @@ def _create_transformer_layer(self) -> nn.Module:
...
+                "model specific config. i.e. LlamaConfig"
+            )
+
+        # Import the modeling module
+        modeling_module = importlib.import_module(
+            f"transformers.models.{model_name}.modeling_{model_name}"
+        )
+
+        transformer_arch_name = self.config.transformer_layer_architecture
+        decoder_class = getattr(modeling_module, transformer_arch_name)
```
markurtz: NIT: can we raise a more informative error other than a generic AttributeError

src/speculators/models/eagle.py
```diff
@@ -532,7 +529,45 @@ def _create_transformer_layer(self) -> nn.Module:
...
+                "Could not import model classes. Please ensure the config is a valid "
+                "model specific config. i.e. LlamaConfig"
+            )
+
+        # Import the modeling module
+        modeling_module = importlib.import_module(
+            f"transformers.models.{model_name}.modeling_{model_name}"
+        )
+
+        transformer_arch_name = self.config.transformer_layer_architecture
```
markurtz, line 574: A nice to have here would be supporting an "auto" or "from_layer_config" reserved name that would match a ".*DecoderLayer" regex against the classes available in the module. Once found, it would update the conig with the proper value and use that class instance. The use case for this is for easier creation of the speculator model so only the transformer config needs to be determined and passed in ahead of time and we remove potential conflicting input args in the config.

Not critical, though, so if it's too much to include here, we can open up an issue for it to put it on the backlog

src/speculators/models/eagle.py
```diff
@@ -532,7 +529,45 @@ def _create_transformer_layer(self) -> nn.Module:
...
+        decoder_class = getattr(modeling_module, transformer_arch_name)
+        try:
+            layer_norm_class = getattr(
+                modeling_module,
+                f"{transformer_arch_name.removesuffix('DecoderLayer')}RMSNorm",
```
markurtz: Rather than assuming the "{DecoderLayer}RMSNorm" name pattern that would have to exist and may not transfer across model implementations, a pathway that better ensures generality would be to check a regex or multiple regexes against the class list in the module. Something like the following:

```python
classes = dict(inspect.getmembers(module, inspect.isclass))
for pat in [r".*RMSNorm$", r".*Norm$"]:
    for name, cls in classes.items():
        if re.match(pat, name): return cls
```

src/speculators/models/eagle.py
```diff
@@ -532,7 +529,45 @@ def _create_transformer_layer(self) -> nn.Module:
...
+            )
+        except AttributeError:
+            layer_norm_class = torch.nn.LayerNorm
```
markurtz: I worry about silently setting this without the user knowing. If we do this, we should raise a loud warning that this happened.

Another pathway that might work better would be to enable the user to pass into the constructor or the config an optional LayerNorm package+class name that can be used / signal that falling back on torch.nn.LayerNorm is fine.

markurtz (changes requested):
Left a few comments, take a look through and let me know what you think @fynnsu

src/speculators/models/eagle.py
```diff
@@ -532,7 +529,45 @@ def _create_transformer_layer(self) -> nn.Module:
...
 
     def _layernorm_class(self) -> type[nn.Module]:
-        return LlamaRMSNorm
+        return self._architecture_classes[1]
 
     def _transformer_layer_class(self) -> type[nn.Module]:
-        return LlamaDecoderLayer
+        return self._architecture_classes[0]
+
+    def _import_model_classes(self) -> tuple[type, type]:
```
markurtz:
```suggestion
    def _import_model_classes(self) -> tuple[type[nn.Module], type[nn.Module]]:
```

src/speculators/models/eagle.py
fynnsu: Yeah, this was not my favorite implementation but in practice it does seem to hold. I ran a test script that found every instance of ".*DecoderLayer" in the transformers model directory and the matching config class, and then checked if the classes could load correctly and it worked every time. Also a lot of the `modeling_*` modules are auto-generated from a `modular_*` module in the same folder. So in practice this pattern is pretty consistent. 

That being said, using `MODEL_FOR_CAUSAL_LM_MAPPING` to get to modeling module should be a bit more robust, so I will try using that instead.

src/speculators/models/eagle.py
fynnsu: Yeah, I used this for the fallback because it seems to be the most common alternative to a custom class, but I agree on adding a warning. 

All the more popular models I looked at seem to use a custom `.*RMSNorm` implementation, so I would probably lean towards just setting the warning for now and only adding the config option if it becomes a problem. Mainly, because I think more config options confuse the user when 99% of the time, this should just be set to the corresponding `.*RMSNorm` class.

src/speculators/models/eagle.py
fynnsu: This probably wouldn't be too difficult to add given we're already doing similar for the `RMSNorm` class. On this one, I probably would recommend we just check for `model_name + "DecoderLayer"` because 1. this pattern is very consistent within the code base and 2. there are a few modules that have multiple DecoderLayer's (sometimes for different modalities), and we are more likely to get the wrong one if we're pattern matching.

markurtz (approved):
Looks great!

PR #56, opened 2025-07-16
[Examples] Add Eagle and Eagle3 Conversion Scripts

dsikka:


rahul-tuli (approved):

shanjiaz (approved):

dsikka:
@MeganEFlynn For Eagle3 - I can remove the examples to not show both with and without `norm_before_residual` - what would we want for Llama and Qwen?

PR #62, opened 2025-07-17
Update speculative token default

dsikka:
- Changes it from 5 to 3


markurtz (changes requested):
Putting a request changes on this to talk through further. Why are we changing the default to 3? The best default here is going to be extremely dependent on the algorithm, the deployment scenario, the verifier, and the speculator architecture. 5 is the default across most of the research and research benchmark frameworks, so I'm heavily in favor of keeping that and not overindexing for our one use case we are looking at right now.

For the converter refactor, I do have an easy way so we can pass in args like this and we can potentially hack that into the current state of conversion. Additionally, the user should always have an option to override that param on deployment with vLLM

PR #63, opened 2025-07-17
Standardization and documentation for speculators conversion pathways

markurtz:


dsikka (changes requested):
Since this is now integrated with vLLM, any changes we do to the conversion pathway should be tested against this branch: https://github.com/vllm-project/vllm/pull/20436

PR #76, opened 2025-08-04
Enhanced Eagle3 Conversion Testing

shanjiaz:
## Overview
Expanded Eagle3 test coverage to address conversion workflow validation before the 0.1.0 release

## Changes
Enhanced Test Coverage
- Adds integration test to validate eagle3 conversion 
- Updates the workflow to test all test types for nightly 
- Move conversion tests to integration folder, not e2e

## Testing
Set transformer to newest version and these tests caught the same error. 
Tests passed with the older transformers version.


dsikka (approved):
Do you mind adding this to the `development.yml` as well?

markurtz:
> > Still need to go through in detail, but immediate reaction is that we should not remove the flags for test levels to run in the nightly.yml. What were the issues we were hitting that led to needing that?
> 
> We just dont get full coverage for anything that isn't marked. IMO nightly should be capturing everything beyond smoke/sanity as well

We can remove the coverage report for now, since that's on unit tests it is running everything, but we don't have enough coverage yet. But I do want to keep the levels in place. It allows us to designate things that should only be run at a weekly or release basis due to the requirements on resources. unit-tests were intended to run in full on nightly, integration would run everything but regression to push those off to a weekly / release cadence, and e2e was only set to run smoke due to those likely being longer running tests, but we could expand e2e to smoke + sanity and reserve regression for weekly / release

shanjiaz:
> This looks alright for now to get some active tests running on main immediately. Some points, though, that I think we should address in a follow up PR:
> 
> Potential issues:
> 
> * Calling out to the separate non test functions and asserting in there doesn't log the error messages/failures correctly generally. Raising exceptions there would be better for logs and debugging
> * Cleanup only happens in the setup method and there's no teardown cleanup logic currently, so failed tests can persist state
> * Currently not testing any potential failure modes to ensure those error correctly
> * Test case reporting will show as only a single function, so lacks granularity for failures/successes
> 
> Recommendation: A parameterized fixture to handle running the conversion and then passing into individual, focused test_* functions would help out a ton towards granularity, reporting, and potential expansion. The parameterized fixture can be setup as a yield style one to handle full cleanup after. Also, if any errors within the fixture happen, then it automatically fails every test dependent on it.

I can add these in a follow-up PR 🫡

PR #79, opened 2025-08-05
Polish Eagle3 config and model definitions

markurtz:
## Description

- Share code between Eagle and Eagle3 implementations as base class to minimize redundancy
- Remove redundant params and source as much info from the transformers config as possible
- Remove current implementation for token mappings/head pruning, more needs to be worked in to solidify the direction and implementation there
- Minor cleanup for eagle3 module

### DIFF
--- a/src/speculators/models/eagle3.py
+++ b/src/speculators/models/eagle3.py
@@ -13,15 +13,14 @@ Classes:
 """
 
 import os
-from typing import Any, ClassVar, Literal, Optional, Union
+from typing import ClassVar, Literal, Optional, Union
 
 import torch
-from pydantic import Field, field_serializer, field_validator
+from pydantic import Field
 from torch import nn
 from transformers import PretrainedConfig, PreTrainedModel
 from transformers.modeling_attn_mask_utils import _prepare_4d_causal_attention_mask
 from transformers.modeling_outputs import CausalLMOutputWithPast
-from transformers.models.llama.configuration_llama import LlamaConfig
 from transformers.models.llama.modeling_llama import (
     LlamaMLP,
     LlamaRMSNorm,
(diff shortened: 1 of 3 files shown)


src/speculators/models/eagle3.py
```diff
@@ -345,42 +310,36 @@ def __init__(
...
+            padding_idx=(
+                config.transformer_layer_config.pad_token_id
+                if hasattr(config.transformer_layer_config, "pad_token_id")
+                else None
+            ),
         )
-
         self.fc = nn.Linear(
-            3 * self.target_hidden_size,  # Use target model's hidden size
+            3 * self.hidden_size,  # Use target model's hidden size
```
rahul-tuli, line 339: @markurtz since the hidden states are coming from the verifier, shouldn't this be `3 * self.target_hidden_size`?

src/speculators/models/eagle3.py
```diff
@@ -390,32 +349,35 @@ def __init__(
...
         )
 
-        self.register_buffer(  # type: ignore[attr-defined]
-            "d2t",
-            torch.zeros(self.draft_vocab_size, dtype=torch.long),
-        )
-        self.register_buffer(  # type: ignore[attr-defined]
-            "t2d",
-            torch.zeros(self.target_vocab_size, dtype=torch.bool),
-        )
```
rahul-tuli, line 412: Why remove these? wouldn't this break the conversion of existing checkpoints? I think it would be nice to have their presence reflected in the config, and initialize these based on that arg; something like: `reduced_vocab: bool` I would also argue if this arg is True we should keep `target_vocab_size` along with `vocab_size`

dsikka:
Closing as no longer needed

PR #90, opened 2025-08-19
Model architectures

anmarques:
This PR lists the model architectures currently supported and planned to be supported with the speculators format in vLLM. It specifies model architectures that are currently supported and others for which there are trained checkpoints, from Red Hat AI, EAGLE or HASS.


README.md
```diff
@@ -105,11 +105,28 @@ Served models can then be benchmarked using [GuideLLM](https://github.com/vllm-p
...
+### Supported Models Architectures
+
+The following model architectures are currently supported or are planned to be supported in the short term.
+
+**Currently supported:**
+
+- Llama-3
+- Qwen3
+
+**Trained checkpoints:**
```
dsikka: Does this mean they will be supported?

anmarques: What is they?

dsikka: The models you have listed.
Your headings are currently unclear in terms of what is and isn’t supported

markurtz (changes requested):
@anmarques I'd like to adapt this further to ideally balance discoverability, clarity, and depth. Can we change this into a table format and either add it as a new sub section in either the Overview (after key features) or in Resources (### Models and then ### Research Implementations). 

I'm thinking something like the following for the table format so it is compact and informative as a really quick example:

| Model / Verifier | Training & Creation | Deployment (vLLM) | Pretrained Checkpoints |
| ---------------- | ------------------- | ----------------- | ---------------------- |
| Qwen3 (8B)       | Eagle 3 ✔️, HASS ⏳  | Eagle 3 ✔️        | [Eagle 3 ✔️](link)     |
| Llama-3.1-70B    | Eagle 3 ⏳           | Eagle 3 ⏳         | ✖️                     |

✔️ = Supported,⏳ = In Progress,✖️ = Not Available

rahul-tuli:
I like having this central place for everything supported, LGTM once the existing comments are resolved!

dsikka:
I agree with the suggested changes. Rahul and I are putting together a similar table for in-depth vLLM overview that we'll add in the next week or so.

PR #91, opened 2025-08-20
[Testing][vLLM] Add vLLM Eagle3 Test Cases

dsikka:
- Test entrypoints with speculators models

### DIFF
--- a/tests/e2e/test_vllm.py
+++ b/tests/e2e/test_vllm.py
@@ -3,7 +3,7 @@ import torch
 from loguru import logger
 
 try:
-    from vllm import LLM, SamplingParams
+    from vllm import LLM, SamplingParams  # type: ignore
 
     vllm_installed = True
 except ImportError:
@@ -53,6 +53,3 @@ class TestEagle3vLLM:
         output = self._run_vllm_engine(model_path=model_path)
         print(output)  # noqa: T201
         assert output
-
-    def test_vllm_server_eagle3(self):
-        pass
--- a/tests/e2e/test_vllm.py
+++ b/tests/e2e/test_vllm.py
@@ -13,6 +13,7 @@ except ImportError:
 
 @pytest.mark.skipif(not vllm_installed, reason="vLLM is not installed, skipping test")
 class TestEagle3vLLM:
+    # move boilerplate code to some base testing class
     def setup_method(self):
         import gc
 
@@ -26,6 +27,15 @@ class TestEagle3vLLM:
             "My name is",
         ]
 
+    @pytest.fixture
+    def temp_cache_dir(self, tmp_path, monkeypatch):
+        cache_dir = tmp_path / "hf_cache"
+        cache_dir.mkdir(exist_ok=True)
+        monkeypatch.setenv("HF_HOME", str(cache_dir))
+        monkeypatch.setenv("TRANSFORMERS_CACHE", str(cache_dir))
+        monkeypatch.setenv("HUGGINGFACE_HUB_CACHE", str(cache_dir))
+        return cache_dir
+
     def _run_vllm_engine(self, model_path):
         sampling_params = SamplingParams(temperature=0.80, top_p=0.95)
         llm = LLM(model=model_path, gpu_memory_utilization=0.9)
@@ -35,14 +45,14 @@ class TestEagle3vLLM:
     @pytest.mark.parametrize(
         "model_path",
         [
-            (
-                "nm-testing/SpeculatorLlama3-1-8B-Eagle3-converted-0717-quantized",
-                "nm-testing/Speculator-Qwen3-8B-Eagle3-converted-071-quantized",
-            )
+            "nm-testing/SpeculatorLlama3-1-8B-Eagle3-converted-0717-quantized",
+            "nm-testing/Speculator-Qwen3-8B-Eagle3-converted-071-quantized",
         ],
     )
     def test_vllm_engine_eagle3(self, model_path):
-        assert self._run_vllm_engine(model_path=model_path)
+        output = self._run_vllm_engine(model_path=model_path)
+        print(output)  # noqa: T201
+        assert output
 
     def test_vllm_server_eagle3(self):
         pass
--- a/tests/e2e/test_vllm.py
+++ b/tests/e2e/test_vllm.py
@@ -13,7 +13,6 @@ except ImportError:
 
 @pytest.mark.skipif(not vllm_installed, reason="vLLM is not installed, skipping test")
 class TestEagle3vLLM:
-    # move boilerplate code to some base testing class
     def setup_method(self):
         import gc
 
@@ -51,5 +50,5 @@ class TestEagle3vLLM:
     )
     def test_vllm_engine_eagle3(self, model_path):
         output = self._run_vllm_engine(model_path=model_path)
-        print(output)  # noqa: T201
+        logger.info(output)
         assert output
--- a/tests/e2e/test_vllm.py
+++ b/tests/e2e/test_vllm.py
@@ -1,0 +1,48 @@
+import pytest
+import torch
+from loguru import logger
+
+try:
+    from vllm import LLM, SamplingParams
+
+    vllm_installed = True
+except ImportError:
+    vllm_installed = False
+    logger.warning("vllm is not installed. This test will be skipped")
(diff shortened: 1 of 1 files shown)


tests/e2e/test_vllm.py
```diff
@@ -0,0 +1,55 @@
...
+    @pytest.mark.parametrize(
+        "model_path",
+        [
+            "nm-testing/SpeculatorLlama3-1-8B-Eagle3-converted-0717-quantized",
+            "nm-testing/Speculator-Qwen3-8B-Eagle3-converted-071-quantized",
+        ],
+    )
+    def test_vllm_engine_eagle3(self, model_path):
+        output = self._run_vllm_engine(model_path=model_path)
+        print(output)  # noqa: T201
```
rahul-tuli: Let's remove print and log the outputs instead?

tests/e2e/test_vllm.py
```diff
@@ -0,0 +1,55 @@
...
+
+    vllm_installed = True
+except ImportError:
+    vllm_installed = False
+    logger.warning("vllm is not installed. This test will be skipped")
+
+
+@pytest.mark.skipif(not vllm_installed, reason="vLLM is not installed, skipping test")
+class TestEagle3vLLM:
+    # move boilerplate code to some base testing class
```
rahul-tuli: nit: we can remove this comment

rahul-tuli (commented):
Diff looks good, what are your thoughts on adding vllm installable, like `pip install "speculators[dev, vllm]"`?

rahul-tuli (approved):

tests/e2e/test_vllm.py
```diff
@@ -0,0 +1,54 @@
+import pytest
+import torch
+from loguru import logger
+
+try:
```
markurtz, line 5: Replace this try catch with the built in pytest method for this:
`pytest.importorskip("vllm", reason="vLLM is not installed")`


If vllm isn't imported, it will automatically skip all tests within the module. After that, you can use `from vllm import LLM, SamplingParams` normally without having to indent and without having to mark the later tests

tests/e2e/test_vllm.py
```diff
@@ -0,0 +1,54 @@
+import pytest
```
markurtz, line 1: Let's restructure this to either `tests/e2e/vllm/test_eagle3.py` or `tests/e2e/eagle3/test_vllm.py` to make it clearer exactly what this is testing and not over complicate the scope of the each individual test module in the future as more are added.

tests/e2e/test_vllm.py
```diff
@@ -0,0 +1,54 @@
...
+    @pytest.mark.smoke
+    @pytest.mark.parametrize(
+        "model_path",
+        [
+            "nm-testing/SpeculatorLlama3-1-8B-Eagle3-converted-0717-quantized",
+            "nm-testing/Speculator-Qwen3-8B-Eagle3-converted-071-quantized",
+        ],
+    )
+    def test_vllm_engine_eagle3(self, model_path):
+        output = self._run_vllm_engine(model_path=model_path)
```
markurtz, line 52: Calling into the helper function is going to hide any specific exceptions, asserts, etc within due to issues with pytest being able to get context from locals within another function. Given the code is minimal, at a minimum the generate should be moved to the test function. For construction, see the comment on the fixture

tests/e2e/test_vllm.py
```diff
@@ -0,0 +1,54 @@
...
+        logger.info(output)
+        assert output
```
markurtz, line 54: At a minimum we should at least assert the shape is correct and simple checks on the output values that they are not 0's, NANs, etc

tests/e2e/test_vllm.py
```diff
@@ -0,0 +1,54 @@
...
+    def temp_cache_dir(self, tmp_path, monkeypatch):
+        cache_dir = tmp_path / "hf_cache"
+        cache_dir.mkdir(exist_ok=True)
+        monkeypatch.setenv("HF_HOME", str(cache_dir))
+        monkeypatch.setenv("TRANSFORMERS_CACHE", str(cache_dir))
+        monkeypatch.setenv("HUGGINGFACE_HUB_CACHE", str(cache_dir))
+        return cache_dir
+
+    def _run_vllm_engine(self, model_path):
+        sampling_params = SamplingParams(temperature=0.80, top_p=0.95)
```
markurtz, line 39: This should have a seed set to better fix reproducibility and stability of the test

tests/e2e/test_vllm.py
```diff
@@ -0,0 +1,54 @@
...
+        llm = LLM(model=model_path, gpu_memory_utilization=0.9)
+        return llm.generate(self.prompts, sampling_params)
```
markurtz, line 41: Generation should be capped with max_tokens to limit the length of the tests, add predictability to them, and make them more readable/debuggable

tests/e2e/test_vllm.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+@pytest.mark.skipif(not vllm_installed, reason="vLLM is not installed, skipping test")
+class TestEagle3vLLM:
+    def setup_method(self):
+        import gc
+
+        gc.collect()
+        if torch.cuda.is_available():
+            torch.cuda.empty_cache()
```
markurtz, line 21: synchronize should be called before to better ensure cache is reliably emptied

tests/e2e/test_vllm.py
```diff
@@ -0,0 +1,54 @@
...
+    @pytest.mark.parametrize(
+        "model_path",
+        [
+            "nm-testing/SpeculatorLlama3-1-8B-Eagle3-converted-0717-quantized",
+            "nm-testing/Speculator-Qwen3-8B-Eagle3-converted-071-quantized",
+        ],
+    )
+    def test_vllm_engine_eagle3(self, model_path):
+        output = self._run_vllm_engine(model_path=model_path)
+        logger.info(output)
```
markurtz, line 53: This can be a very long and noisy output, we should log out a more controlled set here

tests/e2e/test_vllm.py
```diff
@@ -0,0 +1,54 @@
...
+class TestEagle3vLLM:
+    def setup_method(self):
+        import gc
+
+        gc.collect()
+        if torch.cuda.is_available():
+            torch.cuda.empty_cache()
+
+        self.prompts = [
+            "The capital of France is",
```
markurtz, line 24: Utilize more deterministic prompts that don't rely on specific knowledge that may change across models. For example, "Output the following sentence: vLLM is the best inference engine for speculative decoding"

tests/e2e/test_vllm.py
```diff
@@ -0,0 +1,54 @@
...
+            "The president of the US is",
+            "My name is",
+        ]
+
+    @pytest.fixture
+    def temp_cache_dir(self, tmp_path, monkeypatch):
```
markurtz, line 30: This fixture is a no op in the current code since it isn't used anywhere. It needs to either be included in the specific test case later or it needs to be decorated with @pytest.fixture(autouse=True).

But, to future-proof these tests / make them more maintainable as well as removing the later mentioned issue around helper functions, this would be much better if we replace it with a parameterized yield style fixture which handles setup, vLLM instantiation, and teardown like the following:

@pytest.fixture(params=[...model_ids])
def llm_engine(request, tmp_path, monkeypatch):
    model_id = request.param

    # Ensure clean and temporary state
    _clean_cuda()
    cache_dir = tmp_path / "hf_cache"
    cache_dir.mkdir(exist_ok=True)
    for env_key in ("HF_HOME", "TRANSFORMERS_CACHE", "HUGGINGFACE_HUB_CACHE"):
        monkeypatch.setenv(env_key, str(cache_dir))
    monkeypatch.setenv("HF_HUB_DISABLE_TELEMETRY", "1")
    monkeypatch.setenv("HF_HUB_ENABLE_HF_TRANSFER", "1")

    # Construct LLM and yield
    try:
        llm = LLM(model=model_id, gpu_memory_utilization=0.9)

        yield llm, model_id
    finally:
        # Ensure clean ending state
        if llm is not None:
            try:
                del llm
            except Exception as exc:
                logger.warning(f"del llm raised: {exc}")

        _clean_cuda()

PR #93, opened 2025-08-25
[Tests] Extend vLLM test cases with conversion step

dsikka:
- Extends the vLLM test flow to add an additional test with conversion
- Converts a speculators model and a model from the Eagle3 repo
- Runs the converted models in vllm


tests/e2e/vllm/test_eagle3.py
```diff
@@ -0,0 +1,78 @@
...
+        converter = Eagle3Converter()
+
+        converter.convert(
+            input_path=unconverted_model,
+            output_path=converted_path,
+            base_model=base_model,
+            cache_dir=temp_cache_dir,
+            norm_before_residual=norm_before_residual,
+        )
+        self._run_vllm_engine(model_path=str(converted_path))
```
markurtz, line 66: If vLLM isn't installed, this is going to fail. Ideally we'd have a base case of comparing to a known converted model / expected structure and then optionally run in vLLM. At a min need to skip this section if not available

markurtz (changes requested):
One major issue that I left as a comment on the code, otherwise there are several minor cleanups that we could get in for this or for a later PR:
- The sampling params use temperature and top_p which causes extra, unneeded variability in the tests
- The test is asserting that the number of returned tokens has to be 20. Should be updated to >1 and <=20 or we should null out the stop tokens so it's forced to go to 20
- There's no cleanup of the vLLM resources (del llm, gc.collect, empty_cache, etc). A parameterized fixture with a yield would ensure both setup and teardown always run
- Currently only checking for vLLM imports are available, we likely would want to put an extra case on there for torch.cuda / GPU availability so we don't waist resources
- We could require in the dev environments hf_transfer and set HF_HUB_ENABLE_HF_TRANSFER for faster downloads
- _run_vllm_engine runs as a helper, so any failures within there are not going to surface properly with pytest
- Enabling a test config/settings longer term would be good so we can pull down things like gpu_memory_utilization and have a central place to update them for workers running tests
- Base comparisons / interface comparisons for what the output from conversion should look like before running in vLLM

tests/e2e/vllm/test_eagle3.py
markurtz: Missed line 51, resolving

PR #101, opened 2025-08-29
Converters refactor [utils]: Base utility improvements and updates to for refactor enablement

markurtz:
## **Summary**

Refactors the registry pattern from `ClassRegistryMixin` to a generic `RegistryMixin`, introduces comprehensive type annotations, and improves API consistency. Lays the foundation for a more extensible class-based converter architecture by providing better type safety, cleaner and expanded APIs, and enhanced auto-discovery capabilities.

## **Details**

- general:
    - Added `eval_type_backport` dependency for better type evaluation support
    - Updated and added type annotations using `from __future__ import annotations`
- registry.py:
    - Renamed `ClassRegistryMixin` to `RegistryMixin` with generic type support `RegistryMixin[RegistryObjT]`
    - Introduced new generic type variables: `RegistryObjT`, `RegisterT`, `BaseModelT`, `RegisterClassT`
    - Enhanced registry methods with `registered_objects()`, `is_registered()`, and `get_registered_object()`
    - Updated so that typing systems will evaluate wrapped classes correctly
- pydantic_utils.py:
    - Updated `PydanticClassRegistryMixin` to use generic types: `PydanticClassRegistryMixin[BaseModelT]`
    - Improved polymorphic validation with automatic schema reloading
    - Enhanced error handling and type validation for registered classes
    - Added `registered_classes()` method for better Pydantic-specific functionality
    - Updated so that typing systems will evaluate wrapped classes correctly
- auto_importer.py:
    - Streamlined documentation and improved type annotations
    - Better error handling and validation for package discovery
    - Enhanced module import tracking and duplicate prevention
- model.py and config.py:
    - Updated to use the changes from registry.py and pydantic_utils.py

## **Test Plan**

- Expanded and added unit tests to for the full functionality

## **Related Issues**

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,13 +47,13 @@ The following models are currently supported or are planned to be supported in t
 <tr>
 <td rowspan="3">Llama</td>
 <td>8B-Instruct</td>
-<td>EAGLE-3 ✅ | HASS ✅</td>
+<td><a href="https://huggingface.co/RedHatAI/Llama-3.1-8B-Instruct-speculator.eagle3">EAGLE-3</a> ✅ | HASS ✅</td>
 <td>✅</td>
 <td><a href="https://huggingface.co/yuhuili/EAGLE3-LLaMA3.1-Instruct-8B">EAGLE-3</a> ✅</td>
 </tr>
 <tr>
 <td>70B-Instruct</td>
-<td>EAGLE-3 ⏳</td>
+<td><a href="https://huggingface.co/RedHatAI/Llama-3.3-70B-Instruct-speculator.eagle3">EAGLE-3</a> ✅</td>
 <td>✅</td>
 <td><a href="https://huggingface.co/yuhuili/EAGLE3-LLaMA3.3-Instruct-70B">EAGLE-3</a> ✅</td>
 </tr>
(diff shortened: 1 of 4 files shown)


src/speculators/utils/pydantic_utils.py
```diff
@@ -1,138 +1,214 @@
...
+                ):
+                    potential_parents.add(subclass)
+                    stack.append(subclass)
+
+        for check in cls.__mro__:
+            if isinstance(check, type) and issubclass(check, BaseModel):
+                cls._reload_schemas_depending_on(check, potential_parents)
+
+    @classmethod
+    def _reload_schemas_depending_on(cls, target: type[BaseModel], types: set[type]):
```
shanjiaz: Thanks for updating the utils. I see we want to keep schemas consistent when models change. My concern is that this feels a little heavier than necessary, walking all BaseModel subclasses and diffing JSON schemas could be expensive and harder to maintain. Do we always want a full recursive reload? Maybe a lighter approach (targeted model_rebuild(force=True) on affected models, or tracking dependents directly) would cover the same needs with less complexity.

markurtz: Generally we need to have a full recursive reload to enable automation to ensure the pydantic polymorphism is updated when anything is registered. The alternative is relying on implementers to know about that system and to constantly update external code paths from the class they are implementing, which is very likely to become out of date and lead to hard to find issues.

In terms of performance, these updates only traverse the pydantic models and their base classes that we explicitly included, is defined as part of a pydantic registry, and includes a field that references one of the types that changed. Currently it takes a few hundreths of a second to run the update. This is in comparison with anywhere from 1-2 seconds just to import transformers. Ultimately it would take hundreds of classes to start having a noticable impact, and if we reach that point, we can definitely revisit

src/speculators/utils/pydantic_utils.py
```diff
@@ -1,138 +1,214 @@
...
-        ClassRegistryMixin to support polymorphic model instantiation based on
-        a discriminator field
+Provides integration between Pydantic and the registry system, enabling
+polymorphic serialization and deserialization of Pydantic models using
+a discriminator field and dynamic class registry. Includes base model classes
+with standardized configurations and generic status breakdown models for
+structured result organization.
 """
 
+from __future__ import annotations
```
rahul-tuli, line 11: nice!

pyproject.toml
```diff
@@ -40,6 +40,7 @@ dependencies = [
     "click",
     "datasets",
     "deepspeed",
+    "eval_type_backport",
```
rahul-tuli, line 43: This is for pydantic working with `|` with older python versions right?

rahul-tuli (approved):

pyproject.toml
markurtz: Yep, correct, without 3.9 crashes in pydantic class creation

PR #108, opened 2025-09-01
Update README.md with badges

dsikka:
- Add badges for our docs, test status (nightly and main for now), and docs


markurtz (commented):
Some minor NITs that we could push off, but sending over that revolve around ordering as well as including some more badges:
- There doesn't seem to be any type of ordering for the badges. Can we order a bit more around grouping like items together and most frequently helpful items? Something along the lines of the following Docs, Version (GitHub release), PyPi release, Python supported versions, License, CI/CD reports
- On the above, added in GitHub release so we have a link that goes to the latest source code package and the release docs
- Also, we should standardize on capitalizing everything

PR #115, opened 2025-09-03
Update mkdocs

aireilly:
Updates mkdocs


aireilly:
Still not right. will have another look tomorrow.

PR #117, opened 2025-09-04
Converters refactor [base config/model]: Updates to the base config and model for better loading support and standardization

markurtz:
## **Summary**

Refactors base configuration and model classes to improve loading capabilities, standardize APIs, and enhance type safety. This PR modernizes the codebase with better type annotations, streamlined loading methods, and improved documentation while maintaining backward compatibility.

## **Details**

- **Enhanced VerifierConfig loading**:
    - Renamed `from_config()` to `from_pretrained()` for consistency with transformers
    - Added support for loading from multiple sources (PreTrainedModel, PretrainedConfig, paths)
    - Improved parameter resolution and error handling
- **Modernized type annotations**:
    - Migrated from `Union` types to Python 3.10+ union syntax (`str | int`)
    - Added `from __future__ import annotations` for Python 3.9 support
    - Improved type hints throughout config and model classes
- **Standardized documentation**:
    - Enhanced docstrings with consistent formatting and comprehensive parameter descriptions
    - Improved class-level documentation with better examples and usage patterns
    - Added clear inheritance and compatibility information
- **Improved model loading and attachment**:
    - Enhanced verifier attachment with `add_to_config` parameter for better config management
    - Improved error messages and validation in model initialization
    - Streamlined state dictionary handling to exclude verifier parameters
- **Code quality improvements**:
    - Added TC003 to ruff ignore list to allow typing imports outside type checking blocks
    - Consistent code formatting and structure across modules
    - Better separation of concerns in class methods
- **Updated tests**:
    - Reorganized test structure with comprehensive class-based organization
    - Enhanced test coverage with parameterized fixtures and edge case testing
    - Improved test naming and documentation for better maintainability

## **Test Plan**

- Expanded and added unit tests to for the full functionality

## **Related Issues**

- N/A


src/speculators/config.py
```diff
@@ -62,139 +58,157 @@ def __pydantic_schema_base_type__(cls) -> type["TokenProposalConfig"]:
...
-        :param name_or_path: The name or path for the verifier model.
-            Set to None to not add a specific name_or_path.
-            If not provided, the name_or_path from the config will be used.
-        :return: A VerifierConfig object with the extracted parameters.
+
+        Extracts required parameters from the original verifier config and creates
+        a VerifierConfig instance for compatibility validation.
+
+        :param config: The configuration source to extract parameters from
+        :param name_or_path: The name or path for the verifier model
```
rahul-tuli: Could we describe here what `UNSET` means

src/speculators/model.py
```diff
@@ -375,37 +315,22 @@ def resolve_verifier(
 
     def attach_verifier(
         self,
-        verifier: Union[str, os.PathLike, PreTrainedModel],
-        mode: Optional[Literal["full", "train_only"]] = None,
+        verifier: str | os.PathLike | PreTrainedModel,
+        mode: Literal["full", "train_only"] | None = None,
```
rahul-tuli, line 316: nit: Would be nice to convert mode into an enum as well, (for autocomplete instead of having to type out the string)

src/speculators/model.py
```diff
@@ -502,55 +433,44 @@ def forward(self, *args, **kwargs):
...
+        prefix_allowed_tokens_fn: (  # noqa: ARG002
+            Callable[[int, torch.Tensor], list[int]] | None
+        ) = None,
+        synced_gpus: bool | None = None,  # noqa: ARG002
+        assistant_model: PreTrainedModel | None = None,  # type: ignore[override]  # noqa: ARG002
+        streamer: BaseStreamer | None = None,  # noqa: ARG002
+        negative_prompt_ids: torch.Tensor | None = None,  # noqa: ARG002
+        negative_prompt_attention_mask: torch.Tensor | None = None,  # noqa: ARG002
+        use_model_defaults: bool | None = None,  # noqa: ARG002
+        custom_generate: str | None = None,  # noqa: ARG002
```
rahul-tuli: nit: do we still need all these `noqa`

rahul-tuli (approved):
A few nits but looks good overall!

shanjiaz (approved):
Looks good! Can we try to resolve some of the failed tests as well?

src/speculators/model.py
```diff
@@ -82,91 +81,51 @@ class SpeculatorModel(  # type: ignore[misc]
...
+        cls: type[SpeculatorModel],
+        pretrained_model_name_or_path: str | os.PathLike | None,
         *model_args,
-        verifier: Optional[Union[str, os.PathLike, PreTrainedModel]] = None,
-        verifier_attachment_mode: Optional[
-            Literal["detached", "full", "train_only"]
-        ] = None,
-        config: Optional[Union[PretrainedConfig, str, os.PathLike]] = None,
-        cache_dir: Optional[Union[str, os.PathLike]] = None,
+        verifier: str | os.PathLike | PreTrainedModel | None = None,
```
dsikka, line 84: Wont | fail for Python 3.9? We have said in our ReadMe that we support Python 3.9 for the time being

pyproject.toml
```diff
@@ -156,6 +156,7 @@ ignore = [
     "COM812",
     "ISC001",
     "TC002",
+    "TC003", # allow imports for typing outside of type checking blocks
```
fynnsu, line 159: Is there a reason not to just use type checking blocks here?

src/speculators/model.py
fynnsu: This should be okay with the `from __future__ import annotations` header at the top. That basically parses the file and replaces stuff like this with a string (in this case `"str | os.PathLike | PreTrainedModel | None"`) which is valid syntax for py 3.9.

src/speculators/config.py
```diff
@@ -62,139 +58,157 @@ def __pydantic_schema_base_type__(cls) -> type["TokenProposalConfig"]:
...
     @classmethod
-    def from_config(
-        cls, config: PretrainedConfig, name_or_path: Optional[str] = "UNSET"
-    ) -> "VerifierConfig":
+    def from_pretrained(
+        cls,
+        config: str | os.PathLike | PreTrainedModel | PretrainedConfig | None,
+        name_or_path: str | None = "UNSET",
+        **kwargs,
+    ) -> VerifierConfig:
```
fynnsu: Should this signature match `SpeculatorModelConfig.from_pretrained`? If not in this pr, then at some point in the future?

pyproject.toml
markurtz: Mainly around a few points that come from personal preference from my side and wanting to limit the overhead of putting up a PR:
- The if type_checking blocks can make the imports at the top more complicated, especially for auto sorting
- There's not great support for automatically handling injecting the type checking blocks, especially when not using unsafe fixes which we generally shouldn't run
- There's arguably minimal perf gains for using the TYPE_CHECKING blocks since everything we're typing with should already be loaded in the namespace due to usage

src/speculators/model.py
markurtz: This is a much more involved change and going to leave that off for a potential future PR. There are pros and cons here that we'd need to weight, particularly around requiring another Enum/class import to be able to use the function. Additionally, modern typing libraries and autocomplete have started supporting Literals properly

pyproject.toml
fynnsu: > - The if type_checking blocks can make the imports at the top more complicated, especially for auto sorting
> - There's not great support for automatically handling injecting the type checking blocks, especially when not using unsafe fixes which we generally shouldn't run

I think `ruff` supports these use cases. 

> - There's arguably minimal perf gains for using the TYPE_CHECKING blocks since everything we're typing with should already be loaded in the namespace due to usage

Yeah that's fair, I think the advantage is less of a performance one and more that it:
1. Improves readability, makes it clear what's actually being used in the code
2. Reduces risk of circular imports

src/speculators/config.py
markurtz: Yes, good catch, updated

src/speculators/model.py
markurtz: fixed

pyproject.toml
markurtz: Ruff wouldn't autofix locally for me without running unsafe fixes, could be the version we're on

src/speculators/model.py
```diff
@@ -478,80 +410,36 @@ def state_dict(
...
+        Generate text using speculative decoding with attached verifier model.
+
+        Follows the standard transformers generation interface, making it compatible
+        with existing generation workflows while adding speculative decoding
+        capabilities for faster generation.
+
+        :param kwargs: Generation keyword arguments
+        :return: Generated token sequences as GenerateOutput object or LongTensor
         """
+        _ = (kwargs,)  # To avoid unused variable linting error
```
fynnsu: This is okay but just fyi you can also start the kwargs name with an underscore to indicate it isn't used.

i.e. this passes the ruff check:
```python
def generate(self, **_kwargs) -> GenerateOutput | torch.LongTensor: # type: ignore[override]
```

pyproject.toml
fynnsu: Yeah it seems like it can't always do it safely. This change is fine, just want to avoid the slow creep of loosening lint checks, so wanted to make sure there was a reason for it.

src/speculators/model.py
markurtz: yeah, fair on kwargs, this is a left over remnant from when I had the full list of args in there and we needed to preserve the names

markurtz: Updated to _kwargs

PR #125, opened 2025-09-09
Remove the auto-issue link check CI job

aireilly:


markurtz (changes requested):
I'd like to keep the link checking CI on main/nightly. This is primarily because links on README.md or other places that need absolute references will not work until they are landed on main. Secondarily, adds one less job to wait on for PRs.

Towards the above, I think a nightly cadence would work best so we're not overloading main

fynnsu:
Offering [this](https://github.com/vllm-project/speculators/pull/129) as an alternative to removing the nightly link checker. 

Instead of opening a new issue each time, we just have a single issue which we add comments to.

PR #129, opened 2025-09-10
Update link checker so that it comments on existing issue

fynnsu:


.github/workflows/nightly.yml
```diff
@@ -16,13 +16,12 @@ jobs:
...
-      - name: Create Issues
+      - name: Create Comment
         if: steps.lychee.outputs.exit_code != '0'
-        uses: peter-evans/create-issue-from-file@v5
+        uses: peter-evans/create-or-update-comment@v4
         with:
-          title: Link Checker Report
-          content-filepath: ./lychee/out.md
-          labels: report, automated issue
+          issue-number: 128
```
markurtz, line 23: Is there any easy way we could check to see if an issue already exists of a given title matching a set format we define, if so use that issue-number, and if not, then it would create a new one?

I'm primarily worried about us missing updates to a static issue that is updated in place and not knowing when it is passing vs when it isn't without opening that issue and checking the history. 

The ideal case in my mind would be that we reference the issue in the fix PR so it can be closed out. Once a new issue happens, then we open up a new issue and continually update it until it's closed out by a fix.

fynnsu: Hi Mark, I considered this but couldn't find any well established github actions on the marketplace that search for an issue. We could make our own but that's probably too much effort spent on this problem. 

> I'm primarily worried about us missing updates to a static issue that is updated in place and not knowing when it is passing vs when it isn't without opening that issue and checking the history.

Yeah, my thinking was that whoever is interested can subscribe to the issue, and then they would be notified when there are new link checker errors. The comments are only added when a problem is detected so any new comment on that issue would correspond to a failing run. 

> The ideal case in my mind would be that we reference the issue in the fix PR so it can be closed out. Once a new issue happens, then we open up a new issue and continually update it until it's closed out by a fix.

Yeah that'd be ideal but I think this a reasonable compromise that solves the current issue, while still uploading the link checker output on failures and notifying stakeholders.

markurtz: The https://github.com/marketplace/actions/find-last-issue action is made for this and pairs wells with the create-issue pathways. Here's an end to end example (not tested):

```
  link-checks:
    runs-on: ubuntu-latest
    permissions:
      issues: write
    steps:
      - uses: actions/checkout@v4

      - name: Run Link Checker
        id: lychee
        uses: lycheeverse/lychee-action@v2
        with:
          fail: false
          args: --verbose --no-progress --output ./lychee/out.md

      - name: Find the last open link-check issue
        id: last-issue
        if: steps.lychee.outputs.exit_code != '0'
        uses: micalevisk/last-issue-action@v2
        with:
          state: open
          labels: |
            link-check
            automated issue

      - name: Create or Update link-check issue
        if: steps.lychee.outputs.exit_code != '0'
        uses: peter-evans/create-issue-from-file@v4
        with:
          title: "Link Checker Issue"
          content-filepath: ./lychee/out.md
          issue-number: ${{ steps.last-issue.outputs.issue-number }}
          labels: |
            link-check
            automated issue

      - name: Fail job if issues found
        if: steps.lychee.outputs.exit_code != '0'
        run: |
          echo "Link Checker found issues. See the created/updated issue for details."
          exit 1
```

fynnsu: I looked into this and unfortunately the `find-last-issue` action is unmaintained (no commits for the last two years and only 2 github stars).

I was considering using it regardless, cause it does seem like it's still in use on some repos but when I tried I got a deprecation warning saying that it uses deprecated github actions commands (see attempt [here](https://github.com/fynnsu/speculators/actions/runs/17957041957/job/51070886591?pr=5#step:4:10)). Therefore I don't think it makes sense to use this as it's only a matter of time before github removes the deprecated feature and this action breaks.

markurtz: Gotcha, thanks for double checking on that @fynnsu

PR #134, opened 2025-09-17
Fix for Eagle attention arch when head_dim is given in config.json

eldarkurtic:
Most Llama-like models follow the standard symmetry when it comes to creating attention layers, which is that: `head_dim = hidden_size // num_attention_heads`. However, Qwen3-32B doesn't follow this convention, which is indicated by `head_dim` explicitly given in `config.json`. Our current EagleAttention assumes that all models will have `head_dim == hidden_size // num_attention_heads` which is not correct for this model.

Before this PR:
Qwen3-32B models converted/created with speculators would crash in vLLM due to incompatible shapes. Our existing codebase assumes that `head_size = hidden_size // num_attention_heads = 5120 // 64 = 80`, which is not correct. It should be `head_size = 128`, as specified in `config.json` at https://huggingface.co/Qwen/Qwen3-32B/blob/main/config.json#L9

After this PR:
Qwen3-32B models can be converted/created with speculators, and are running fine (with expected acceptance rates) in vLLM.

### DIFF
--- a/.lycheeignore
+++ b/.lycheeignore
@@ -1,0 +1,1 @@
+/reference/speculators
--- a/README.md
+++ b/README.md
@@ -6,13 +6,13 @@
     <img alt="Speculators logo" src="https://raw.githubusercontent.com/neuralmagic/speculators/main/docs/assets/branding/speculators-logo-black.svg" height="64" />
   </picture>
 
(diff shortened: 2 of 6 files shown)


eldarkurtic:
quality/type tests are green now (I don't think that the red ones are due to these diffs, but please correct me if I am wrong).
Can we re-review this please and try to merge into main so others can use it easily?
@markurtz @shanjiaz

markurtz (approved):

PR #136, opened 2025-09-18
Fix for draft models always being in fp32 datatype

eldarkurtic:
Even though they are trained and saved in bf16 format, after calling `speculators convert` all of our draft models are converted and saved with FP32 data-type. We want to preserve the original data-type of the trained draft model.

Before this PR:
All drafter weights in FP32 data-type.

After this PR:
All drafter weights in the target data-type given by the model's config.json file.

### DIFF
--- a/src/speculators/convert/eagle/eagle3_converter.py
+++ b/src/speculators/convert/eagle/eagle3_converter.py
@@ -215,6 +215,7 @@ class Eagle3Converter:
             mlp_bias=eagle_config.get("mlp_bias", False),
             tie_word_embeddings=False,
             torch_dtype=eagle_config.get("torch_dtype"),
+            head_dim=eagle_config.get("head_dim"),
         )
 
     def _save_converted_checkpoint(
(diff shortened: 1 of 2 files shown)


shanjiaz (approved):
Niiiice thanks for doing this!

markurtz (approved):
Looks good pending code formatting/styling fixes

src/speculators/convert/eagle/eagle3_converter.py
```diff
@@ -228,6 +229,11 @@ def _save_converted_checkpoint(
             verifier_attachment_mode="detached",
         )
         model.load_state_dict(weights, strict=False)  # type: ignore[attr-defined]
+        weights_dtype = getattr(
```
dsikka: Shouldn't we know this info on load and use this precision throughout as opposed to the end?

eldarkurtic: Llama layers (like LlamaMLP) don't accept dtype as an argument to initialize linear operators in the target dtype. That's why we have to do it after weight loading

PR #142, opened 2025-09-29
Update speculator config & converter to support hidden states indexing

shanjiaz:
## Changes:
- Added support for optional arguments `eagle_aux_hidden_state_layer_ids` and `inference_type`. 
- Added more robust logic for `target_vocab_size`. We default on using "t2d" length, if not available, load the config file of verifier model, recursively search the dict for `vocab_size`. (The search is needed for nested dict. e.g. target_config_dict["text_config"]["vocab_size"] )
- Removed tests for adding verifier embeddings as it's handled on the vllm side now.
- Removed forward pass tests since forward function is defined on the vllm side.

## Command used:
```
speculators convert nvidia/Llama-4-Maverick-17B-128E-Eagle3 \
  --algorithm eagle3 \
  --verifier RedHatAI/Llama-4-Maverick-17B-128E-Instruct-quantized.w4a16 \
  --output-path Llama4-Maverick-Eagle3-Speculators \
  --validate-device cuda:0 \
  --algorithm-kwargs '{"eagle_aux_hidden_state_layer_ids": [1,23,44], "inference_type": "text"}'
```

## Converted checkpoint:
[shanjiaz/Llama4-Maverick-Eagle3-Speculators-converted](https://huggingface.co/shanjiaz/Llama4-Maverick-Eagle3-Speculators-converted)

### DIFF
--- a/src/speculators/convert/eagle/eagle3_converter.py
+++ b/src/speculators/convert/eagle/eagle3_converter.py
@@ -60,7 +60,9 @@ class Eagle3Converter:
             target_config_dict, _ = PretrainedConfig.get_config_dict(base_model)
             vocab_size = find_vocab_size(target_config_dict)
             if vocab_size is None:
-                raise ValueError("Could not determine vocab_size from target model config.")
+                raise ValueError(
+                    "Could not determine vocab_size from target model config."
+                )
             eagle_config["target_vocab_size"] = vocab_size
             logger.debug(
                 f"Using target_vocab_size from config: "


src/speculators/models/eagle3.py
```diff
@@ -78,6 +76,16 @@ class Eagle3SpeculatorConfig(SpeculatorModelConfig):
         description="Hidden size of the target model (if different from draft model)",
     )
 
+    eagle_aux_hidden_state_layer_ids: Optional[list[int]] = Field(
+        default=None,
+        description="Layer IDs of the Eagle auxiliary hidden state layers",
+    )
+
+    inference_type: Optional[str] = Field(
```
rahul-tuli: This argument is not needed anymore

src/speculators/models/eagle3.py
rahul-tuli: We don't have embed_tokens in the drafter we need to make this optional as well

src/speculators/convert/eagle/eagle3_converter.py
```diff
@@ -34,6 +35,8 @@ def convert(
         base_model: str,
         validate: bool = True,
         norm_before_residual: bool = False,
+        inference_type: Optional[str] = None,
```
rahul-tuli: To be deleted

src/speculators/convert/eagle/eagle3_converter.py
```diff
@@ -157,6 +180,8 @@ def _build_eagle3_speculator_config(
         eagle_config: dict,
         base_model: str,
         norm_before_residual: bool = False,
+        inference_type: Optional[str] = None,
```
rahul-tuli: To be deleted

examples/convert/eagle3/apply_eagle3_llama4_maverick.sh
```diff
@@ -0,0 +1,6 @@
+speculators convert nvidia/Llama-4-Maverick-17B-128E-Eagle3 \
```
rahul-tuli, line 1: Should we make it explicit and set norm_before_residual to False for this checkpoint? even though it defaults to the same thing, I think it'll be nice?

rahul-tuli (changes requested):

examples/convert/eagle3/apply_eagle3_llama4_maverick.sh
shanjiaz: sure!

src/speculators/models/eagle3.py
shanjiaz: Do we not need it because vllm side is pulling the embeddings directly from verifier?

src/speculators/convert/eagle/eagle3_converter.py
```diff
@@ -92,58 +119,7 @@ def _process_checkpoint_weights(
...
-
-        # Only add verifier embeddings if not present in eagle model
-        if "embed_tokens.weight" not in processed_weights:
-            logger.info("Eagle model missing embeddings - adding verifier embeddings")
-            return self._add_verifier_embeddings(processed_weights, base_model)
-        else:
-            logger.info("Eagle model already has embeddings - keeping originals")
-            return processed_weights
-
-    def _add_verifier_embeddings(
```
rahul-tuli, line 104: We would need this functionality during attach verifier/forward passes no? maybe this would be better placed in a util file somewhere

src/speculators/convert/eagle/eagle3_converter.py
rahul-tuli: nit: we can do this in a follow up, but we don't even need this part for vllm

examples/convert/eagle3/apply_eagle3_llama4_maverick.sh
rahul-tuli, line 1: nice!

src/speculators/convert/eagle/eagle3_converter.py
shanjiaz: Clarified during standup. Will add this to a util in a following PR.

src/speculators/convert/eagle/eagle3_converter.py
shanjiaz: Oh nice! I can removed it. I guess this is the main logic for this function, we can removed the whole thing!

src/speculators/models/eagle3.py
```diff
@@ -335,6 +338,8 @@ def __init__(
         verifier_attachment_mode: Optional[
             Literal["detached", "full", "train_only"]
         ] = None,
+        reduce_vocab_size: bool = True,
+        embed_tokens_available: bool = True,
```
fynnsu: Rename to something like Separate draft embedding

src/speculators/models/eagle3.py
fynnsu, line 349: Add docstrings for new args

src/speculators/models/eagle3.py
```diff
@@ -401,158 +409,20 @@ def __init__(
...
-            torch.zeros(self.target_vocab_size, dtype=torch.bool),
-        )
-
-        # Type hints for buffers
-        self.d2t: torch.Tensor
-        self.t2d: torch.Tensor
-
-        self.post_init()  # type: ignore[attr-defined]
-
-    def forward(
```
rahul-tuli, line 420: instead of removing the forward method, could we raise a NotImplementedError?

src/speculators/convert/eagle/eagle3_converter.py
```diff
@@ -44,107 +46,48 @@ def convert(
...
+                f"Using target_vocab_size from t2d tensor: "
+                f"{eagle_config['target_vocab_size']}"
+            )
+            reduce_vocab_size = True
+        else:
+            # fall back to target model config - search for vocab_size at any level
+            target_config_dict, _ = PretrainedConfig.get_config_dict(base_model)
+            vocab_size = find_vocab_size(target_config_dict)
+            eagle_config["target_vocab_size"] = (
+                vocab_size if vocab_size is not None else 128000
```
rahul-tuli: Is It safe to  hardcode target_vocab_size to 128000 here?

src/speculators/convert/eagle/eagle3_converter.py
rahul-tuli, line 145: This reads a lil weird that the vocab_size for the drafter is the target_vocab_size no? shouldn't it be renamed to draft vocab size?

But we can do that as a follow up

rahul-tuli (approved):
few questions/nits which can be addressed in a follow up, good work on this, LGTM once we raise the NotImplementedError for forward passes

src/speculators/convert/eagle/eagle3_converter.py
shanjiaz: yeah I'll raise an error here instead

src/speculators/models/eagle3.py
shanjiaz: yes!

rahul-tuli (dismissed):
LGTM!

dsikka (changes requested):
Do we have test cases for multiple decoder layers?

src/speculators/convert/eagle/eagle3_converter.py
```diff
@@ -59,9 +59,9 @@ def convert(
             # fall back to target model config - search for vocab_size at any level
             target_config_dict, _ = PretrainedConfig.get_config_dict(base_model)
             vocab_size = find_vocab_size(target_config_dict)
-            eagle_config["target_vocab_size"] = (
-                vocab_size if vocab_size is not None else 128000
-            )
+            if vocab_size is None:
+                logger.error("Could not determine vocab_size from target model config.")
```
fynnsu: This just logs an error, it doesn't raise one. Is that what we want? Or should the converter script fail if this isn't set?

shanjiaz: Huh yeah converter should probably fail for this. Will update.

PR #143, opened 2025-10-03
[WIP] Eagle3 Training Implementation

fynnsu:
WORK IN PROGRESS

Todos/Missing:
- [ ] Loss Masking
- [ ] Training loop
- [ ] Data batching
- [ ] Distributed support
- [ ] Better Verifier `lm_head`, `embed_tokens` and config values loading
- [ ] `Eagle3DraftModel.__init__` signature cleanup/better configuration
- [ ] Code relocation / merging with existing definitions (Currently just have everything under `speculators/train` but this will need to change)
- [ ] Metric collection/reporting
- [ ] Verify correctness of key components (attention masking, data token alignment, etc).
- [ ] General testing
- [ ] All of the stuff we don't know we need yet but will find as we go
- [ ] + all the stuff I forgot when writing this list

### DIFF
--- a/src/speculators/train/data.py
+++ b/src/speculators/train/data.py
@@ -1,0 +1,82 @@
+import os
+
+import torch
+from torch.utils.data import Dataset
+import torch.nn.functional as F
+
+
+class AddGaussianNoise:
+    def __init__(self, mean=0.0, std=0.0):
+        self.mean = mean
+        self.std = std
+
+    def __call__(self, data):
+        tensor = data["hidden_state_big"]
+        noise = torch.randn(tensor.size()) * self.std + self.mean
+        noisy_tensor = tensor + noise
+        data["hidden_state_big"] = noisy_tensor
+        return data
+
+
+class AddUniformNoise:
+    def __init__(self, std=0.0):
+        self.std = std
+
+    def __call__(self, data):
+        tensor = data["hidden_state_big"]
+        noise = (torch.rand_like(tensor) - 0.5) * self.std * 512 / tensor.shape[1]
+        noisy_tensor = tensor + noise
+        data["hidden_state_big"] = noisy_tensor
+        return data
+
+
+def list_files(path):
+    datapath = []
+    for root, _directories, files in os.walk(path):
+        for file in files:
+            file_path = os.path.join(root, file)
+            datapath.append(file_path)
+
+    return datapath
+
+
+def pad_to_length(tensor, dim, length):
+    padding = [0, 0] * tensor.dim()
+    padding[-dim * 2 - 1] = length - tensor.shape[dim]
+    return F.pad(tensor, padding)
+
+
+class Eagle3SampleFileDataset(Dataset):
+    def __init__(self, datapath: str, max_len: int, transform=None):
+        self.data = list_files(datapath)
+        self.max_len = max_len
+        self.transform = transform
+
+    def __len__(self):
+        return len(self.data)
+
+    def __getitem__(self, index):
+        data = torch.load(self.data[index])
+
+        new_data = {}
+        hidden_state = pad_to_length(
+            data["hidden_state"][: self.max_len], 0, self.max_len
+        )
+        # shape: [max_len, 3 * hidden_dim]
+
+        input_ids = pad_to_length(data["input_ids"][: self.max_len], 0, self.max_len)
+        # shape: [max_len]
+        loss_mask = pad_to_length(data["loss_mask"][: self.max_len], 0, self.max_len)
+        # shape: [max_len]
+        target = pad_to_length(data["target"][: self.max_len], 0, self.max_len)
+        # shape: [max_len, hidden_dim]
+
+        new_data["loss_mask"] = loss_mask
+        new_data["target"] = target
+        new_data["hidden_state_big"] = hidden_state
+        new_data["input_ids"] = input_ids
+
+        if self.transform:
+            new_data = self.transform(new_data)
+
+        return new_data
--- a/src/speculators/train/eagle3/attention.py
+++ b/src/speculators/train/eagle3/attention.py
@@ -1,0 +1,166 @@
+import torch
+from transformers.modeling_utils import AttentionInterface
+from transformers.integrations.flex_attention import repeat_kv
+from torch.nn.attention.flex_attention import flex_attention
+from torch.nn.attention.flex_attention import or_masks, and_masks, BlockMask
+from typing import Callable
+
+
+def create_combined_mask_mod(lengths: torch.Tensor):
+    total_seq_len = lengths.sum().item()
+    document_ids = torch.repeat_interleave(
+        torch.arange(lengths.shape[0], device=lengths.device, dtype=torch.long), lengths
+    ).contiguous()
+    N = document_ids.shape[0]
+
+    def causal_mask_mod(b, h, q_idx, kv_idx):
+        return q_idx >= kv_idx
+
+    def document_mask_mod(b, h, q_idx, kv_idx):
+        return document_ids[q_idx] == document_ids[kv_idx % N]
+
+    def diagonal_draft_mask_mod(b, h, q_idx, kv_idx):
+        return kv_idx % total_seq_len == q_idx
+
+    return or_masks(
+        and_masks(causal_mask_mod, document_mask_mod), diagonal_draft_mask_mod
+    )
+
+
+def extend_mask_for_draft_tokens(block_mask, new_mask_mod: Callable | None = None):
+    """
+    Extend the block mask to include new draft tokens. Concatenates a diagonal mask for the new draft tokens.
+
+    Assumptions:
+    - block_mask BLOCK_SIZE := KV_BLOCK_SIZE == Q_BLOCK_SIZE
+    - The number of query values is the original total_seq_len (or equivalently the number of query blocks is the original total_seq_len // BLOCK_SIZE)
+
+    i.e. if block_mask is:
+    [
+        [
+            [1, 0, 0],
+            [1, 1, 0],
+            [0, 0, 1],
+        ]
+    ]
+    the result will be:
+    [
+        [
+            [1, 0, 0, 1, 0, 0],
+            [1, 1, 0, 0, 1, 0],
+            [0, 0, 1, 0, 0, 1],
+        ]
+    ]
+    and then callinga again will give:
+    [
+        [
+            [1, 0, 0, 1, 0, 0, 1, 0, 0],
+            [1, 1, 0, 0, 1, 0, 0, 1, 0],
+            [0, 0, 1, 0, 0, 1, 0, 0, 1],
+        ]
+    ]
+
+    """
+    kv_num_blocks = block_mask.kv_num_blocks
+    # shape: [B, H, Q_LEN // BLOCK_SIZE]
+
+    kv_indices = block_mask.kv_indices
+    # shape: [B, H, Q_LEN // BLOCK_SIZE, KV_LEN // BLOCK_SIZE]
+    b, h, q_blocks, kv_blocks = kv_indices.shape
+
+    # extend kv indices if needed
+    kv_indices = torch.cat(
+        [kv_indices, kv_indices.new_zeros((b, h, q_blocks, q_blocks))], dim=-1
+    )
+    new_block_indices = torch.arange(
+        kv_blocks,
+        kv_blocks + q_blocks,
+        dtype=kv_indices.dtype,
+        device=kv_indices.device,
+    ).reshape(1, 1, q_blocks, 1)
+    kv_indices.scatter_(
+        dim=-1, index=kv_num_blocks.unsqueeze(-1), src=new_block_indices
+    )
+
+    kv_num_blocks = kv_num_blocks + 1
+
+    return BlockMask.from_kv_blocks(
+        kv_num_blocks,
+        kv_indices,
+        block_mask.full_kv_num_blocks,
+        block_mask.full_kv_indices,
+        mask_mod=new_mask_mod or block_mask.mask_mod,
+    )
+
+
+def block_mask_to_dense_attention_mask(
+    block_mask: BlockMask, device: torch.device, dtype: torch.dtype
+):
+    attention_mask = torch.ones(block_mask.shape, device=device, dtype=dtype)
+
+    for q_idx in range(attention_mask.shape[2]):
+        attention_mask[0, 0, q_idx, :] = block_mask.mask_mod(
+            torch.zeros(1, device=device, dtype=torch.long),
+            torch.zeros(1, device=device, dtype=torch.long),
+            torch.ones(1, device=device, dtype=torch.long) * q_idx,
+            torch.arange(attention_mask.shape[3], device=device, dtype=torch.long),
+        )
+    return attention_mask
+
+
+
+
+def flex_attention_forward(
+    module: torch.nn.Module,
+    query: torch.Tensor,
+    key: torch.Tensor,
+    value: torch.Tensor,
+    attention_mask,
+    scaling: float | None = None,
+    softcap: float | None = None,
+    head_mask: torch.Tensor | None = None,
+    s_aux: torch.Tensor | None = None,
+    **kwargs,
+) -> tuple[torch.Tensor, torch.Tensor | None]:
+    block_mask = attention_mask
+    enable_gqa = False
+
+    num_local_query_heads = query.shape[1]
+    # When running TP this helps:
+    if (num_local_query_heads & (num_local_query_heads - 1)) != 0:
+        key = repeat_kv(key, query.shape[1] // key.shape[1])
+        value = repeat_kv(value, query.shape[1] // value.shape[1])
+
+    return_lse = query.device.type != "cpu"
+
+    query = query.contiguous()
+    key = key.contiguous()
+    value = value.contiguous()
+
+    flex_attention_output = flex_attention(
+        query,
+        key,
+        value,
+        score_mod=None,
+        block_mask=block_mask,
+        enable_gqa=enable_gqa,
+        scale=scaling,
+        kernel_options=None,
+        # Last time checked on PyTorch == 2.5.1: Flex Attention always computes the lse regardless.
+        # For simplification, we thus always return it as no additional computations are introduced.
+        return_lse=return_lse,
+    )
+    # lse is returned in float32
+    if return_lse:
+        attention_output, lse = flex_attention_output  # type: ignore[misc]
+        lse = lse.to(value.dtype)
+    else:
+        attention_output = flex_attention_output  # type: ignore[assignment]
+        lse = None
+
+    attention_output = attention_output.transpose(1, 2).contiguous()
+    return attention_output, lse
+
+
+ALL_ATTENTION_FUNCTIONS = AttentionInterface() # Singleton class used for registry
+ALL_ATTENTION_FUNCTIONS.register("simple_flex_attention", flex_attention_forward)
--- a/src/speculators/train/eagle3/core.py
+++ b/src/speculators/train/eagle3/core.py
@@ -1,0 +1,187 @@
+import torch
+from transformers.configuration_utils import PretrainedConfig
+from transformers import AutoModelForCausalLM, DynamicCache
+
+from torch.nn.attention.flex_attention import create_block_mask
+
+from speculators.train.eagle3.attention import (
+    create_combined_mask_mod,
+    extend_mask_for_draft_tokens,
+)
+from speculators.train.eagle3.model_definitions import model_classes
+
+
+class Eagle3DraftModel(torch.nn.Module):
+    def __init__(
+        self,
+        hidden_size: int,  # Must be same for verifier and draft
+        # Vocab mappings
+        t2d_vocab: torch.Tensor,
+        d2t_vocab: torch.Tensor,
+        decoder_layer_config: PretrainedConfig,
+        # Verifier
+        verifier_vocab_size: int,
+        verifier_pad_token_id: int,
+        # Draft config
+        num_layers: int = 1,
+        ttt_steps: int = 3,
+    ):
+        super().__init__()
+        self.hidden_size = hidden_size
+        self.num_layers = num_layers
+        self.decoder_layer_config = decoder_layer_config
+        self.ttt_steps = ttt_steps
+        self.t2d_vocab = t2d_vocab  # shape: [verifier_vocab_size], bool
+        self.d2t_vocab = d2t_vocab  # shape: [draft_vocab_size], int offsets
+        self.draft_vocab_size = t2d_vocab.sum(dtype=torch.long).item()
+        model_definitions = model_classes[decoder_layer_config.model_type]
+
+        self.fc_layer = torch.nn.Linear(3 * hidden_size, hidden_size)
+        self.layers = torch.nn.ModuleList(
+            [
+                model_definitions.decoder_layer_class(decoder_layer_config, layer_idx)
+                for layer_idx in range(num_layers)
+            ]
+        )
+        self.norm = model_definitions.norm_class(
+            hidden_size, eps=decoder_layer_config.rms_norm_eps
+        )
+        self.rotary_emb = model_definitions.rotary_emb_class(decoder_layer_config)
+        self.embed_tokens = torch.nn.Embedding(
+            verifier_vocab_size, hidden_size, padding_idx=verifier_pad_token_id
+        )
+        # shape: [verifier_vocab_size, hidden_size]
+
+        self.verifier_lm_head = torch.nn.Linear(
+            hidden_size, self.draft_vocab_size, bias=False
+        )
+        self.lm_head = torch.nn.Linear(hidden_size, self.draft_vocab_size, bias=False)
+        # shape: [hidden_size, draft_vocab_size]
+
+    def load_verifier_lm_head(self, verifier_model_name_or_path: str):
+        verifier_model = AutoModelForCausalLM.from_pretrained(
+            verifier_model_name_or_path
+        )
+        verifier_lm_head_data = verifier_model.lm_head.weight.data.to(
+            self.t2d_vocab.device
+        )
+        self.verifier_lm_head.weight.data = verifier_lm_head_data[self.t2d_vocab, :]
+
+    def loss_function(self, logits: torch.Tensor, targets: torch.Tensor):
+        logits = torch.nn.functional.log_softmax(logits, dim=-1)
+        targets = torch.nn.functional.log_softmax(targets, dim=-1)
+        return torch.nn.functional.kl_div(
+            logits, targets, reduction="sum", log_target=True
+        )
+
+    def forward(
+        self,
+        hidden_states: torch.Tensor,  # shape: [1, total_seq_len, 3 * hidden_size]
+        input_ids: torch.Tensor,  # shape: [1, total_seq_len]
+        lengths: torch.Tensor | None = None,  # shape: [batch_size]
+        verifier_last_hidden_state: torch.Tensor
+        | None = None,  # shape: [1, total_seq_len, hidden_size]
+        ttt_steps: int | None = None,
+        use_off_policy_tokens: bool = False,
+        **kwargs,
+    ):
+        device = hidden_states.device
+        total_seq_len = hidden_states.shape[1]
+        return_loss = verifier_last_hidden_state is not None
+
+        if ttt_steps is None:
+            ttt_steps = self.ttt_steps
+        if lengths is None:
+            lengths = torch.tensor([total_seq_len], dtype=torch.long, device=device)
+
+        past_key_values = DynamicCache(config=self.decoder_layer_config)
+        position_ids = (
+            torch.cat(
+                [
+                    torch.arange(length, dtype=torch.long, device=device)
+                    for length in lengths
+                ],
+                dim=0,
+            )
+            .unsqueeze(0)
+            .contiguous()
+        )
+        # shape: [1, total_seq_len]
+
+        combined_mask_mod = create_combined_mask_mod(lengths.to(device))
+        block_mask = create_block_mask(
+            combined_mask_mod,
+            B=None,
+            H=None,
+            Q_LEN=total_seq_len,
+            KV_LEN=total_seq_len,
+            device=device,
+        )
+
+        hidden_states = self.fc_layer(hidden_states)
+        # shape: [1, total_seq_len, hidden_size]
+
+        if return_loss:
+            verifier_logits = self.verifier_lm_head(verifier_last_hidden_state)
+            loss = torch.tensor(0.0, device=device)
+
+        draft_tokens = []
+        for ttt_step in range(ttt_steps):
+            input_embeds = self.embed_tokens(input_ids)
+            # shape: [1, total_seq_len, hidden_size]
+            cache_position = torch.arange(
+                ttt_step * total_seq_len,
+                (ttt_step + 1) * total_seq_len,
+                dtype=torch.long,
+                device=device,
+            )
+            # shape: [total_seq_len]
+
+            hidden_states = torch.cat([input_embeds, hidden_states], dim=-1)
+            # shape: [1, total_seq_len, 2 * hidden_size]
+
+            position_embeddings = self.rotary_emb(hidden_states, position_ids)
+
+            for decoder_layer in self.layers:
+                hidden_states = decoder_layer(
+                    hidden_states,
+                    attention_mask=block_mask,  # block_mask_to_dense_attention_mask(block_mask, device, torch.bool),
+                    position_ids=position_ids,
+                    past_key_values=past_key_values,
+                    cache_position=cache_position,
+                    position_embeddings=position_embeddings,
+                    **kwargs,
+                )
+
+            hidden_states = self.norm(hidden_states)
+
+            logits = self.lm_head(hidden_states)
+            # shape: [1, total_seq_len, draft_vocab_size]
+
+            if return_loss:
+                loss += self.loss_function(
+                    logits[:, : -(ttt_step + 1)], verifier_logits[:, (ttt_step + 1) :]
+                )
+
+            input_ids = torch.argmax(logits, dim=-1)
+            # shape: [1, total_seq_len]
+            # Use d2t to map draft tokens to verifier tokens.
+            # Must be in verifier vocabulary space because we use full verifier vocabulary in embedding
+            input_ids = input_ids + self.d2t_vocab[input_ids]
+            draft_tokens.append(input_ids)
+
+            if use_off_policy_tokens:
+                # Overwrite input_ids with ground truth tokens
+                # shift input_ids by 1 to the left and pad with 0
+                # note: inputs_ids will no longer line up with verifier_last_hidden_state
+                # the draft logits generated from the padded tokens are ignored sliced out for loss calculation
+                input_ids = torch.cat(
+                    [input_ids[:, 1:], input_ids.new_zeros(1, 1)], dim=-1
+                )
+                # shape: [1, total_seq_len]
+
+            block_mask = extend_mask_for_draft_tokens(block_mask)
+            position_ids = position_ids + 1
+            # shape: [1, total_seq_len]
+
+        return draft_tokens, loss if return_loss else None
--- a/src/speculators/train/eagle3/model_definitions.py
+++ b/src/speculators/train/eagle3/model_definitions.py
@@ -1,0 +1,100 @@
+from typing import NamedTuple, Optional
+import torch
+from transformers.configuration_utils import PretrainedConfig
+import copy
+from transformers import Cache, LlamaConfig
+from transformers.models.llama.modeling_llama import (
+    LlamaDecoderLayer,
+    LlamaRMSNorm,
+    LlamaRotaryEmbedding,
+)
+
+from transformers.processing_utils import Unpack
+from transformers.utils.generic import TransformersKwargs
+
+
+class LlamaConcatInputDecoderLayer(LlamaDecoderLayer):
+    def __init__(self, config: PretrainedConfig, layer_idx: int):
+        super().__init__(config, layer_idx)
+
+        ##### CHANGES START #####
+        if layer_idx == 0:
+            self.self_attn.q_proj = torch.nn.Linear(
+                2 * config.hidden_size,  # previous: config.hidden_size
+                config.num_attention_heads * config.head_dim,
+                bias=config.attention_bias,
+            )
+            self.self_attn.k_proj = torch.nn.Linear(
+                2 * config.hidden_size,  # previous: config.hidden_size
+                config.num_key_value_heads * config.head_dim,
+                bias=config.attention_bias,
+            )
+            self.self_attn.v_proj = torch.nn.Linear(
+                2 * config.hidden_size,  # previous: config.hidden_size
+                config.num_key_value_heads * config.head_dim,
+                bias=config.attention_bias,
+            )
+            self.input_layernorm = LlamaRMSNorm(
+                2 * config.hidden_size,  # previous: config.hidden_size
+                eps=config.rms_norm_eps,
+            )
+        ##### CHANGES END #####
+
+    def forward(
+        self,
+        hidden_states: torch.Tensor,
+        attention_mask: Optional[torch.Tensor] = None,
+        position_ids: Optional[torch.LongTensor] = None,
+        past_key_values: Optional[Cache] = None,
+        use_cache: Optional[bool] = False,
+        cache_position: Optional[torch.LongTensor] = None,
+        position_embeddings: Optional[
+            tuple[torch.Tensor, torch.Tensor]
+        ] = None,  # necessary, but kept here for BC
+        **kwargs: Unpack[TransformersKwargs],
+    ) -> torch.Tensor:
+        ##### CHANGES START #####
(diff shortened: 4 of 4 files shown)


src/speculators/train/eagle3/attention.py
```diff
@@ -0,0 +1,177 @@
...
+    ]
+    the result will be:
+    [
+        [
+            [1, 0, 0, 1, 0, 0],
+            [1, 1, 0, 0, 1, 0],
+            [0, 0, 1, 0, 0, 1],
+        ]
+    ]
+    and then callinga again will give:
```
brian-dellabetta:
```suggestion
    and then calling again will give:
```

src/speculators/train/data.py
brian-dellabetta, line 1: consider splitting transforms-specific stuff into a separate file

src/speculators/train/data.py
```diff
@@ -0,0 +1,218 @@
...
+from typing import Any
+
+import torch
+import torch.nn.functional as F  # noqa: N812
+from torch.utils.data import Dataset
+
+BatchType = dict[str, Any]
+
+
+class TransformTensors:
```
brian-dellabetta: It's unfortunately a part of `torchvision` and not native torch, but it has an abstraction to handle this. Might be of use -- https://docs.pytorch.org/vision/main/auto_examples/transforms/plot_custom_transforms.html

src/speculators/train/utils.py
```diff
@@ -0,0 +1,62 @@
...
+
+local_rank = int(os.environ.get("LOCAL_RANK", "0"))
+world_size = int(os.environ.get("WORLD_SIZE", "1"))
+is_distributed = "LOCAL_RANK" in os.environ
+
+logger = logging.getLogger("speculators")
+
+
+def maybe_setup_distributed():
+    # Based off of https://docs.pytorch.org/tutorials/intermediate/ddp_tutorial.html#initialize-ddp-with-torch-distributed-run-torchrun
```
brian-dellabetta: might be better off as a docstring, with a quick summary for what this does

src/speculators/train/logger.py
brian-dellabetta, line 1: seems like a lot of code for logging, there's no dependency that can handle this without so much code to maintain?

brian-dellabetta (commented):
Excellent work! In future, consider splitting a PR like this up into separate smaller PRs that can be merged over time. This looks like it could be split up into a few -- logging, trainer, dataset class, llama-specific code.

A few comments from an outsider's perspective. Since this is all entirely new, i'm sure there will be some validation after this lands, but consider an e2e test here or in a follow-up

src/speculators/train/data.py
```diff
@@ -0,0 +1,218 @@
...
+            math.ceil(Path(fname).stat().st_size / size_0 * lengths_0)
+            for fname in self.data
+        ]
+
+    def __getitem__(self, index) -> BatchType:
+        data = torch.load(self.data[index])
+
+        # todo: standardize names during data generation and then remove this
+        data["hidden_states"] = data["hidden_state"]
+        data["verifier_last_hidden_states"] = data["target"]
```
shanjiaz: Can we remove this? I think we agreed on saving the last hidden state with rest of the hidden states.  Thanks!

src/speculators/train/distributed_batch_sampler.py
```diff
@@ -0,0 +1,221 @@
...
+        if heap[0].rank == rank:
+            # minimum bucket corresponds to the local rank -> add idx to local batch
+            local_batch.append(start_index + idx)
+
+        _ = heapreplace(heap, _Bin(new_fill, heap[0].rank))
+
+    return local_batch
+
+
+def _assign_to_packed_batches(
```
kylesayrs, line 85: Can you give more detail about this function? Seems like a pretty costly operation?

src/speculators/train/data.py
```diff
@@ -0,0 +1,272 @@
...
+
+    return {
+        "hidden_states": data["hidden_state"],
+        "input_ids": data["input_ids"],
+        "verifier_last_hidden_states": data["target"],
+        "loss_mask": data["loss_mask"],
+    }
+
+
+def standardize_data_v1(data: dict[str, Any]) -> dict[str, Any]:
```
kylesayrs, line 111: Do you want to create a dataclass for this format?

src/speculators/train/eagle3/attention.py
```diff
@@ -0,0 +1,177 @@
...
+import torch
+from torch.nn.attention.flex_attention import (
+    BlockMask,
+    and_masks,
+    flex_attention,
+    or_masks,
+)
+from transformers.modeling_utils import AttentionInterface
+
+flex_attention = torch.compile(flex_attention)
```
kylesayrs: Does torch compiling the entire forward function not work?

src/speculators/train/eagle3/attention.py
```diff
@@ -0,0 +1,177 @@
...
+        block_mask=block_mask,
+        enable_gqa=enable_gqa,
+        scale=scaling,
+    )
+    attention_output: torch.Tensor = cast("torch.Tensor", flex_attention_output)
+    attention_output = attention_output.transpose(1, 2).contiguous()
+    return attention_output, None
+
+
+ALL_ATTENTION_FUNCTIONS = AttentionInterface()  # Singleton class used for registry
```
kylesayrs, line 172: Beware that `config.attn_implementation` is used to key both AttentionInterface and AttentionMaskInterface. Make sure that AttentionMaskInterface is not being used, or alternatively register your mask implementation with AttentionMaskInterface.


https://github.com/huggingface/transformers/blob/4d0b6758b90aa54e4077171e6d42c55e0c01c622/src/transformers/masking_utils.py#L677

src/speculators/train/data.py
fynnsu: As discussed offline, updated the code to support both formats.

src/speculators/train/data.py
fynnsu: Yeah, I think in this case it doesn't make sense to add a `torchvision` dep. This `TransformTensors` is just meant to be a very lightweight abstraction for the two different noise functions.

src/speculators/train/logger.py
fynnsu: Yeah it is a bit of code of but
- 234/582 of the lines in this file are docstrings
- This code allows us to interface with multiple different experiment tracking backends. Simplifying the code would require selecting a single experiment tracking tool and enforcing it for our users.

There are potentially improvements that could be made to this logic, but unfortunately there's no third-party library that achieves this.

src/speculators/train/distributed_batch_sampler.py
fynnsu: Yeah, there's a bit of info in the "Batch sampling" section in the pr description but the idea is:
1. We want to batch our data so that training is faster
2. We can't just combine our data samples because the sequence lengths differ. There are two solutions to this: pad/slice the samples to match or combine the samples along the sequence dimension and use attention masking to treat them as separate samples. We're doing the later as it is more efficient (we don't waste compute on padding tokens).
3. If we naively fix our batch size to some value (e.g. 5) and then select the first 5 samples for batch 1, next 5 for batch 2, etc. we will have batches with drastically different sizes. e.g. batch_1_lengths = [3, 7, 2, 4, 2], batch_2_lengths = [20, 32, 5, 18, 17] which wastes compute on small batches and risks OOMs on larger batches
4. The next idea might be to have a dynamic batch size where we greedily grab the next sample repeatedly until our batch reaches some `max_seq_len`. This is a bit better but can still result in underutilized batches. For example, if we have `max_seq_len=20` and `lengths=[10, 11, 3, 5, 8]` we will get 3 batches if greedily batching (`[10], [11], [3, 5, 8]`) whereas a more efficient packing could give you 2 batches (`[10, 3, 5], [11, 8]`).
5. This code is designed to find a reasonable packing, that takes the multiple gpus into account.

As an example, with a small dataset of 5000 real samples and a max_seq_len of 4096:
- The greedy algorithm produces 1973 batches
- The sampler produces 1,696 (212/gpu) batches for 8 gpus (14% less)
- The sampler produces 1,740 (435/gpu) batches for 4 gpus (12% less) 

It takes about `0.017` seconds to run. Here are some additional times for larger datasets:
```
num_lengths: 10002, elapsed: 0.032926 secs
num_lengths: 20004, elapsed: 0.066594 secs
num_lengths: 40008, elapsed: 0.131806 secs
num_lengths: 80016, elapsed: 0.268784 secs
num_lengths: 160032, elapsed: 0.534865 secs
num_lengths: 320064, elapsed: 1.080697 secs
num_lengths: 640128, elapsed: 2.149603 secs
num_lengths: 1280256, elapsed: 4.425388 secs
```
which suggest it grows roughly linearly with dataset size.

On the other hand, assuming each batch takes roughly same amount of time to compute, we save significantly more time per epoch by having less batches for the same set of data.

As for the algorithm itself, it works roughly like so:
1. Shuffle the data
2. Create `N` buckets (where `N` is the number of gpus)
3. Perform a binary search over `S`:
  a. Take the next `S` samples
  b. Attempt to pack them into `N` buckets by repeatedly adding the smallest remaining sample into the emptiest bucket
  c. If success, try with a larger `S` (more samples), else try a smaller `S`
4. Assign the `S` samples to their buckets and mark those samples as accounted for. Repeat with the next `S` samples.
5. return the batches assigned to the local rank's bucket

src/speculators/train/data.py
fynnsu: I don't think it makes sense here because this is just a temporary format for the data, which then gets modified further in the next step. We would essentially be instantiating the dataclass with the values just to immediately extract those values back out into a new dict.

src/speculators/train/eagle3/attention.py
fynnsu: Hmm, it didn't at some point when I was first testing this but I just tried now and it seems like it works! Will make the change.

src/speculators/train/eagle3/attention.py
fynnsu: Yeah, the llama model's seem to use `AttentionInterface`, and I'd imagine other CasualLMs would be the same but might be something to watch out for in the future as we add other drafter model architectures.

fynnsu:
> The only part that makes me slightly nervous is that reimplementation of model definitions, and the fixed ModelComponents structure, as this pattern makes supporting new models harder/ more rigid. If there's any way to make this more general/ provide a good programming model, that would be nice. Otherwise, this is good for now.

Yeah that's fair. This is also my least favorite part :(

Unfortunately, we do need to make a few changes to the DecoderLayer but I did my best to minimize these and clearly mark them with comments. I'm happy to look further into methods for removing these modifications but right now it isn't super clear to me how we could do that.

As for `ModelComponents`, this is really just meant as a super simple temporary way to bundle up all the classes associated with a model architecture. It's possible we will need to extend or modify this in the future but it feels premature to do so before we even have a second drafter architecture that we would want to support.

Supporting other drafter architectures for spec decoding is less critical than supporting other model architectures is because we don't need to match the drafter architecture with the verifier's. e.g. there's nothing stopping you from training a Llama drafter model on a Qwen verifier.

src/speculators/train/eagle3/core.py
```diff
@@ -0,0 +1,339 @@
...
+            if use_off_policy_tokens:
+                # Overwrite input_ids with ground truth tokens
+                # shift input_ids by 1 to the left and pad with 0
+                # note: inputs_ids no longer line up with verifier_last_hidden_states
+                # the draft logits generated from the padded tokens are ignored
+                # and sliced out for loss calculation
+                input_ids = torch.cat(
+                    [
+                        original_input_ids[:, 1 + ttt_step :],
+                        original_input_ids.new_zeros(1, 1 + ttt_step),
```
shanjiaz, line 402: should we pad with a padding token?

fynnsu: It doesn't matter because we don't calculate loss on these tokens.
e.g.
for tokens 
`[A, B, C, D]`
we generate
`[B*, C*, D*, E*]`
and compute loss on those. Then using we would repeat this process and use these tokens in the next step of generation.
However, this `use_off_policy_tokens` section replaces the tokens with
`[B, C, D, 0]`
then we do another draft forward pass to get
`[C*, D*, E*, 0*]`
however, we slice out the `0*` prediction (in `align_for_step`) before computing the loss. So whatever output we get is ignored.

src/speculators/train/eagle3/core.py
```diff
@@ -0,0 +1,339 @@
...
+            Q_LEN=total_seq_len,
+            KV_LEN=total_seq_len,
+            device=device,
+        )
+
+        hidden_states = self.fc(hidden_states)
+        # shape: [1, total_seq_len, hidden_size]
+
+        original_input_ids = input_ids.detach().clone()
+        return_loss = verifier_last_hidden_states is not None
```
HDCharles, line 322: feels like return_loss should be set via module setting along the lines of module.train_mode(), rather than triggered via passing in the verifier's hidden states

src/speculators/train/eagle3/core.py
```diff
@@ -0,0 +1,339 @@
...
+                    past_key_values=past_key_values,
+                    cache_position=cache_position,
+                    position_embeddings=position_embeddings,
+                    **kwargs,
+                )
+
+            logits = self.lm_head(self.norm(hidden_states))
+            # shape: [1, total_seq_len, draft_vocab_size]
+
+            if return_loss:
```
HDCharles, line 373: nit: i prefer to keep main loop clean and move extraneous code like this to a helper function

src/speculators/train/noise_transforms.py
```diff
@@ -0,0 +1,33 @@
...
+        return tensor + torch.randn_like(tensor) * self.std + self.mean
+
+
+class AddUniformNoise(TransformTensors):
+    def __init__(self, std=0.2, tensors=("hidden_states",)):
+        super().__init__(tensors)
+        self.std = std
+
+    def transform(self, tensor: torch.Tensor) -> torch.Tensor:
+        return tensor + (torch.rand_like(tensor) - 0.5) * self.std
```
HDCharles: bit weird for this one to be forced to be centered around 0 while gaussian noise is not.

src/speculators/train/eagle3/core.py
fynnsu: This is modeled after the way transformers handles this ([link](https://github.com/huggingface/transformers/blob/571352d3785469bbfa8cb124ab08ead1d10c4c5b/src/transformers/models/llama/modeling_llama.py#L502-L504)):
```python
loss = None
if labels is not None:
    loss = self.loss_function(logits=logits, labels=labels, vocab_size=self.config.vocab_size, **kwargs)
```

I think it's a reasonable way to handle this because if you don't pass target values (or in this case the verifier hidden states) then you can't do the computation. Otherwise we need to have stuff like
```python
if return_loss and verifier_last_hidden_states is None:
    raise ValueError(...)
```

src/speculators/train/noise_transforms.py
fynnsu: Yeah, although honestly I may "fix" this by removing the `mean` option from Gaussian because I'm not sure non-zero-centered noise makes sense here.

src/speculators/train/eagle3/core.py
fynnsu: That's fair. I will see if anything can be moved out but it is a bit of a challenge because a lot of this is very tightly coupled with the training logic.

scripts/train.py
```diff
@@ -0,0 +1,196 @@
...
+    llama_config._attn_implementation = "simple_flex_attention"  # noqa: SLF001
+
+    # Load t2d and d2t tensors
+    d2t = torch.from_numpy(np.load(args.d2t_path)).to(device)
+    t2d = torch.from_numpy(np.load(args.t2d_path)).to(device)
+    draft_vocab_size = d2t.shape[0]
+
+    speculator_config = Eagle3SpeculatorConfig(
+        transformer_layer_config=llama_config,
+        draft_vocab_size=draft_vocab_size,
```
HDCharles, line 111: This nested Config design seems like it could be a bit more streamlined. The combination of configs + extraneous data that could probably be in a config makes it hard to have an intuitive mental model of where all the info needs to be.

- transform_layer_config: 
  - seems like it should have all the data from the draft model which I would expect to include the draft vocab if we're going to overwrite the num_hidden_layers, seems like that would be the place to pass that in.
- Eagle3DraftModel: 
  - takes in speculator_config which has ttt_steps and takes ttt_steps as a separate arg. 
  - Takes in t2d and d2t as tensors, whereas all the other tensor data for the verifier+speculator is passed in as paths that are then loaded later. Feels like t2d and d2t may make more sense to be normally handled as paths in the Eagle3SpeculatorConfig (that can be overridden if you use optional arguments in the Eagle3DraftModel)

src/speculators/train/eagle3/core.py
```diff
@@ -0,0 +1,339 @@
...
+            verifier_attachment_mode="train_only",
+        )
+        self.hidden_size = config.transformer_layer_config.hidden_size
+        self.num_layers = config.transformer_layer_config.num_hidden_layers
+        self.decoder_layer_config = config.transformer_layer_config
+        self.ttt_steps = ttt_steps
+        self.ttt_step_loss_decay = ttt_step_loss_decay
+        self.register_buffer("t2d", t2d)  # shape: [verifier_vocab_size], bool
+        self.register_buffer("d2t", d2t)  # shape: [draft_vocab_size], int offsets
+        self.draft_vocab_size = int(t2d.sum(dtype=torch.long).item())
```
HDCharles: probably want to assert shape  of t2d/d2t but take the draft vocab size from the config

src/speculators/train/eagle3/core.py
```diff
@@ -0,0 +1,339 @@
...
+    def forward(
+        self,
+        hidden_states: torch.Tensor,  # shape: [1, total_seq_len, 3 * hidden_size]
+        input_ids: torch.Tensor,  # shape: [1, total_seq_len]
+        lengths: torch.Tensor | None = None,  # shape: [batch_size]
+        loss_mask: torch.Tensor | None = None,  # shape: [1, total_seq_len]
+        position_ids: torch.Tensor | None = None,  # shape: [1, total_seq_len]
+        verifier_last_hidden_states: torch.Tensor
+        | None = None,  # shape: [1, total_seq_len, hidden_size]
+        ttt_steps: int | None = None,
```
HDCharles: why is ttt_steps an input to forward rather than an attribute of the model?

src/speculators/train/eagle3/core.py
```diff
@@ -0,0 +1,339 @@
...
+        ttt_step_loss_decay: float = 1.0,
+    ):
+        super().__init__(
+            config=config,
+            verifier=None,
+            verifier_attachment_mode="train_only",
+        )
+        self.hidden_size = config.transformer_layer_config.hidden_size
+        self.num_layers = config.transformer_layer_config.num_hidden_layers
+        self.decoder_layer_config = config.transformer_layer_config
```
HDCharles: is there a way to streamline these names so they don't swap around?

src/speculators/train/eagle3/core.py
```diff
@@ -0,0 +1,339 @@
...
+
+        original_input_ids = input_ids.detach().clone()
+        return_loss = verifier_last_hidden_states is not None
+        if return_loss:
+            with torch.no_grad():
+                target_logits = self.verifier_lm_head(verifier_last_hidden_states)
+                # shape: [1, total_seq_len, draft_vocab_size]
+
+        loss = torch.tensor(0.0, device=device)
+        prev_correct = (
```
HDCharles: i'm not sure i understand what prev_correct means in this context also, it seems like its only used in teh return loss branch so should this be moved with line 250?

src/speculators/train/eagle3/model_definitions.py
```diff
@@ -0,0 +1,116 @@
...
+
+        # Fully Connected
+        residual = hidden_states
+        hidden_states = self.post_attention_layernorm(hidden_states)
+        hidden_states = self.mlp(hidden_states)
+        hidden_states = residual + hidden_states
+        return hidden_states  # noqa: RET504
+
+
+class LlamaConcatInputRotaryEmbedding(LlamaRotaryEmbedding):
```
HDCharles: feels like you could have Eagle3Draft do the config modifications since you're not modifying the class, just how it preprocesses the config.

src/speculators/train/eagle3/model_definitions.py
```diff
@@ -0,0 +1,116 @@
...
+class LlamaConcatInputDecoderLayer(LlamaDecoderLayer):
+    def __init__(
+        self,
+        config: LlamaConfig,
+        layer_idx: int,
+        norm_before_residual: bool = False,
+    ):
+        super().__init__(config, layer_idx)
+
+        ##### CHANGES START #####
```
HDCharles: Feels more accurate to say its 

1) initialize normally

2) updates for speculation/eagle3

rather than

1) code that is the same

2) code that is different

src/speculators/train/eagle3/model_definitions.py
```diff
@@ -0,0 +1,116 @@
...
+        hidden_states: torch.Tensor,
+        attention_mask: torch.Tensor | None = None,
+        position_ids: torch.LongTensor | None = None,
+        past_key_values: Cache | None = None,
+        use_cache: bool | None = False,
+        cache_position: torch.LongTensor | None = None,
+        position_embeddings: tuple[torch.Tensor, torch.Tensor] | None = None,
+        **kwargs: Unpack[TransformersKwargs],  # type: ignore[valid-type]
+    ) -> torch.Tensor:
+        ##### CHANGES START #####
```
HDCharles: the changes start block makes a lot more sense here.

src/speculators/train/eagle3/model_definitions.py
```diff
@@ -0,0 +1,116 @@
...
+            # Apply norms
+            embeds = self.input_layernorm(embeds)
+            hidden = self.hidden_norm(hidden)
+            if self.norm_before_residual:
+                residual = hidden  # set residual to normalized hidden
+            hidden_states = torch.cat([embeds, hidden], dim=-1)
+        else:
+            residual = hidden_states
+            hidden_states = self.input_layernorm(hidden_states)
+        ##### CHANGES END #####
```
HDCharles: may be clearer to just do

if not layer_idx == 0:
    return super().forward(...)

src/speculators/train/trainer.py
```diff
@@ -0,0 +1,208 @@
...
+    def setup_model(self):
+        if self.is_distributed:
+            apply_fully_sharded(self.model)
+
+            if self.resume_from_checkpoint and self.checkpointer.previous_epoch != -1:
+                self.checkpointer.load_model_state_dict(self.model)
+            else:
+                # Currently we make assumptions based on the Eagle3DraftModel
+                # architecture, including the existence of a layers attribute.
+                # todo: generalize to non-Eagle3DraftModel
```
HDCharles: given this todo, add assert for model type?

src/speculators/train/eagle3/attention.py
```diff
@@ -0,0 +1,175 @@
...
+        )
+    return attention_mask
+
+
+def flex_attention_forward(
+    module: torch.nn.Module,  # noqa: ARG001
+    query: torch.Tensor,
+    key: torch.Tensor,
+    value: torch.Tensor,
+    attention_mask,
```
HDCharles, line 146: feels like attention_mask vs block_mask swaps back and forth multiple times

Eagle3DraftModel -> block_mask
decoder layer -> attention mask
flex_attention_forward -> attention_mask
flex_attention -> block_mask

src/speculators/train/eagle3/core.py
```diff
@@ -0,0 +1,339 @@
...
+    # logits_indices_ttt_step_2 = [3, 4, 5, 6, 7, 8, 9, 10, 11]
+    # The indices for the loss_mask need to be kept in line with the targets indices
+    logits = logits[:, :-ttt_step] if ttt_step > 0 else logits
+    # shape: [1, total_seq_len - ttt_step, draft_vocab_size]
+    targets = targets[:, ttt_step:]
+    # shape: [1, total_seq_len - ttt_step, draft_vocab_size]
+    if loss_mask is not None:
+        loss_mask = loss_mask[:, ttt_step:]
+        # shape: [1, total_seq_len - ttt_step]
+    if prev_correct is not None:
```
HDCharles, line 48: i thought prev correct was always not None?

src/speculators/train/eagle3/core.py
HDCharles: if that's the standard thats fair

src/speculators/train/noise_transforms.py
HDCharles: yeah that makes sense

src/speculators/train/eagle3/core.py
HDCharles: honestly the logic with prev_correct vs loss_mask could probably be fleshed out to be either clearer or explained somewhere because i'm not sure whats going on there.

scripts/train.py
fynnsu: As discussed offline, although I agree that this config could use some cleanup, I don't think that's within the scope of the pr, because changes to this config also require changes to `vllm` to support loading the new config.

Part of the weirdness around `transform_layer_config` is that we are primarily using this just to instantiate the decoder blocks, but the config we pass in is a full `LlamaConfig` which would usually be used to configure the entire llama model. That means we have extra fields in this config that aren't being used. This is obviously not ideal, but it is the way it's been set up so far. 

I agree with removing `ttt_steps` from here but as discussed, don't think the d2t and t2d paths should end up in the Eagle3 config.

src/speculators/train/eagle3/attention.py
fynnsu: Yeah that's fair. The object we create is a `block_mask` but it serves the same role as an `attention_mask` and so we use the existing `attention_mask` arguments to pipe the `block_mask` through the different levels of the stack. This is the same way `transformers` operates if you use their flex attention backend.

Unfortunately, we are a little bit constrained in the renaming because:
- Eagle3DraftModel: uses the pytorch's `create_block_mask` utility to instantiate the mask.
- decoder layer: this is meant to be a minimal changes modification of the regular llama decoder block, so we shouldn't do renames. We especially want to make sure the function signature matches.
- flex_attention_forward: This slots into the transformers "attention registry" and so we want to keep the function signature consistent with other functions in the registry.
- flex_attention: this is a pytorch function, so we can't rename the arg.

src/speculators/train/eagle3/core.py
fynnsu: Yeah that's fair. I will try to clarify this a bit in comment.

> i thought prev correct was always not None?

Yeah I think that may be True now. Initially I think I was trying out having it be `None` on the first iteration and the set on subsequent ones. I will remove this option.

src/speculators/train/eagle3/core.py
fynnsu: This was a requested feature, as it is possible you might choose to change the number of ttt_steps during training.

src/speculators/train/eagle3/core.py
fynnsu: Yeah I can move this in there.

src/speculators/train/eagle3/model_definitions.py
fynnsu: Sure, can update

src/speculators/train/eagle3/model_definitions.py
fynnsu: Yep, will update

src/speculators/train/eagle3/model_definitions.py
fynnsu: Yeah I can try that.

src/speculators/train/eagle3/core.py
HDCharles: you could just use a shallow function to declutter the logit calculation vs loss calculation

scripts/train.py
```diff
@@ -0,0 +1,203 @@
...
+PREFETCH_FACTOR = 4
+NOISE_STD = 0.05
+
+
+def setup_dataloader(
+    file_list: list[str],
+    world_size: int,
+    local_rank: int,
+    add_noise: bool = True,
+    data_format_version: int = 1,
```
dsikka, line 41: docstring

scripts/train.py
```diff
@@ -0,0 +1,203 @@
...
+    MultipackDistributedBatchSamplerV2,
+)
+from speculators.train.eagle3.core import Eagle3DraftModel
+from speculators.train.logger import setup_metric_logger, setup_root_logger
+from speculators.train.noise_transforms import AddUniformNoise
+from speculators.train.trainer import Trainer, TrainerConfig
+from speculators.train.utils import maybe_destroy_distributed, maybe_setup_distributed
+
+# DRAFTER MODEL HYPARAMETERS
+NORM_BEFORE_RESIDUAL = True
```
dsikka, line 28: A readme explaining the script, how it can be used, would be helpful

src/speculators/train/utils.py
```diff
@@ -0,0 +1,75 @@
...
+from speculators.train.eagle3.core import Eagle3DraftModel
+
+local_rank = int(os.environ.get("LOCAL_RANK", "0"))
+world_size = int(os.environ.get("WORLD_SIZE", "1"))
+is_distributed = "LOCAL_RANK" in os.environ
+
+logger = logging.getLogger("speculators")
+
+
+def maybe_setup_distributed():
```
dsikka: typing, docstring

src/speculators/train/utils.py
```diff
@@ -0,0 +1,75 @@
...
+        return
+
+    dist.destroy_process_group()
+    logger.info(
+        f"Destroyed distributed with local_rank={local_rank}, world_size={world_size}",
+        extra={"override_rank0_filter": True},
+    )
+
+
+def apply_fully_sharded(model: torch.nn.Module):
```
dsikka, line 59: docstring

src/speculators/train/noise_transforms.py
```diff
@@ -0,0 +1,25 @@
+import torch
+
+
+class TransformTensors:
```
dsikka, line 4: What is the purpose of this?
Data augmentation?

src/speculators/train/eagle3/model_definitions.py
```diff
@@ -0,0 +1,119 @@
...
+from transformers.models.llama.modeling_llama import (
+    LlamaDecoderLayer,
+    LlamaRMSNorm,
+    LlamaRotaryEmbedding,
+)
+from transformers.processing_utils import Unpack
+from transformers.utils.generic import TransformersKwargs
+
+
+class LlamaDecoderEagle3FirstLayer(LlamaDecoderLayer):
```
dsikka, line 15: Why do we call it FirstLayer?
This is the class for a single layer in the draft model, correct?

src/speculators/train/eagle3/core.py
```diff
@@ -0,0 +1,360 @@
...
+        default_dtype = self.embed_tokens.weight.dtype
+
+        embed_tokens_sd = {
+            "weight": verifier_weights["model.embed_tokens.weight"].to(default_dtype)
+        }
+        self.embed_tokens.load_state_dict(embed_tokens_sd)
+        self.embed_tokens.weight.requires_grad = False
+
+        # LM HEADS
+        self.lm_head = torch.nn.Linear(
```
dsikka, line 258: This is the LM head for the draft model? Why isn't it part of model_classes?

src/speculators/train/eagle3/core.py
```diff
@@ -0,0 +1,360 @@
...
+                f"t2d.shape[0] ({t2d.shape[0]}) must match"
+                f" verifier_vocab_size ({verifier_model_config.vocab_size})."
+            )
+
+        verifier_weights = load_model_layers(
+            ["model.embed_tokens.weight", "lm_head.weight"], config.name_or_path
+        )
+
+        # EMBEDDINGS
+        self.embed_tokens = torch.nn.Embedding(
```
dsikka, line 243: Seems like MLR often trains the embedding weight (maybe only in the case of gpt-oss but I can be wrong). Do we have a way to train the embedding layer?

Similar case for the vocab?

src/speculators/train/eagle3/core.py
```diff
@@ -0,0 +1,360 @@
...
+        self, config: Eagle3SpeculatorConfig, t2d: torch.Tensor, d2t: torch.Tensor
+    ):
+        super().__init__(
+            config=config,
+            verifier=None,
+            verifier_attachment_mode="train_only",
+        )
+        self.hidden_size = config.transformer_layer_config.hidden_size
+        self.register_buffer("t2d", t2d)  # shape: [verifier_vocab_size], bool
+        self.register_buffer("d2t", d2t)  # shape: [draft_vocab_size], int offsets
```
dsikka, line 167: Fairly naive question but do you mind explaining how dt2 and t2d are being used during training?

tests/unit/train/test_eagle3_attention.py
```diff
@@ -0,0 +1,139 @@
+# ruff: noqa: ERA001
+import pytest
```
dsikka, line 2: Do we have tests for all the other training, checkpointing, etc utilities that you've written?

src/speculators/train/noise_transforms.py
fynnsu: This is just to add a little bit of noise to the hidden states we train on to reduce overfitting.

src/speculators/train/eagle3/model_definitions.py
fynnsu: Only the first layer (in a multi-layer draft model) is modified because it needs to accept both hidden states and input embeddings. Subsequent layers use `LlamaDecoderLayer` directly.

Previously, I used the same decoder block for all layers and had if statements in the init + forward checking the layer id. Now, I've updated it so that we only use the modified block for the first layer.

src/speculators/train/eagle3/core.py
fynnsu: `model_classes` is intended as a future extension point for different draft model architectures (i.e. not always using `LlamaDecoderLayer` blocks). However, all the draft models will use the `torch.nn.Linear` class (at least as far as I know right now), so there is no reason to add it to the dictionary.

src/speculators/train/eagle3/core.py
fynnsu: Yeah, I think the research code wasn't actually training the embedding because it was being called within a `torch.no_grad` context. I don't currently have support for toggling on training of the embedding layer but it wouldn't be difficult to have if we think this is an important usecase.

Conversely for the `lm_head`, we currently always train a separate `lm_head` for the drafter model.

src/speculators/train/eagle3/core.py
fynnsu: Because our draft model has a reduced vocab size, the token ids predicted by the drafter are relative to the drafter vocab. We need to map these to the verifier vocab space before embedding them for the next ttt_step.

e.g. if verifier vocab size is 10k and drafter vocab size is 5k, then the drafter predicted tokens will be in range [0, 5k], and need to be mapped to their corresponding verifier indices before we call the verifier embeddings on them.

That's what `d2t` is used for. 

Note: it is stored as int offsets such that
```python
verifier_ids = draft_ids + self.d2t[draft_ids]
```

As for `t2d`, the verifier's "target" distribution contains all 10k possible verifier tokens, but we only want to compare the tokens that also exist in our drafter vocab. So we use `t2d` to mask out the tokens that are in the verifier but not in the drafter.  The way we actually accomplish this is by using `t2d` to mask the `verifier_lm_head` to only include the drafter tokens.

tests/unit/train/test_eagle3_attention.py
fynnsu: No, that's something that I still need to work on.

src/speculators/train/trainer.py
```diff
@@ -0,0 +1,264 @@
...
+                            sub_module.reset_parameters()  # type: ignore[operator]
+                # todo: Ensure lm_head and embed_tokens are loaded after reset
+        else:
+            self.model.to(self.local_rank)  # type: ignore[arg-type]
+            if self.resume_from_checkpoint and self.checkpointer.previous_epoch != -1:
+                self.checkpointer.load_model_state_dict(self.model)
+
+    def setup_optimizer(self):
+        # Setup optimizer
+        self.opt = torch.optim.AdamW(self.model.named_parameters(), lr=self.config.lr)
```
rahul-tuli, line 116: Should this be `parameters()` instead of `named_parameters()`?

```suggestion
        self.opt = torch.optim.AdamW(self.model.parameters(), lr=self.config.lr)
```

scripts/train.py
```diff
@@ -0,0 +1,213 @@
...
+NOISE_STD = 0.05
+
+
+def setup_dataloader(
+    file_list: list[str],
+    world_size: int,
+    local_rank: int,
+    add_noise: bool = True,
+    data_format_version: int = 1,
+):
```
rahul-tuli:
```suggestion
): -> DataLoader
```

src/speculators/train/checkpointer.py
```diff
@@ -0,0 +1,237 @@
...
+    def optimizer_path(self, epoch: int):
+        optimizer_fname = "optimizer_state_dict.pt"
+        return self.path / str(epoch) / optimizer_fname
+
+    def scheduler_path(self, epoch: int):
+        scheduler_fname = "scheduler_state_dict.pt"
+        return self.path / str(epoch) / scheduler_fname
+
+
+def convert_float_dtype(sd: dict[str, Any], dtype: torch.dtype):
```
rahul-tuli: nit:
```suggestion
def convert_float_dtype(sd: dict[str, torch.Tensor | Any], dtype: torch.dtype):
```

src/speculators/train/eagle3/core.py
```diff
@@ -0,0 +1,360 @@
...
+    # before computing the loss/accuracy. Likewise, there are no logits for the first
+    # ttt_step tokens, so we mask them out.
+    # This is equivalent to shifting the target values by ttt_step + 1 to the left
+    # which puts them in the correct position for the generated tokens.
+    # e.g.
+    # targets_indices = [1, 2, 3, 4, 5, 6, 7, 8, 9]
+    # logits_indices_ttt_step_0 = [1, 2, 3, 4, 5, 6, 7, 8, 9]
+    # logits_indices_ttt_step_1 = [2, 3, 4, 5, 6, 7, 8, 9, 10]
+    # logits_indices_ttt_step_2 = [3, 4, 5, 6, 7, 8, 9, 10, 11]
+    # The indices for the loss_mask need to be kept in line with the targets indices
```
rahul-tuli: nit: should we convert this to a docstring?

src/speculators/train/data.py
```diff
@@ -0,0 +1,246 @@
...
+        self,
+        max_len: int,
+        datapath: str | None = None,
+        file_list: list[str] | None = None,
+        transform: TransformTensors | None = None,
+        hidden_states_dtype=torch.float,
+        standardize_fn: StandardizeFnSig = standardize_data_v1,
+    ):
+        if datapath is not None and file_list is not None:
+            raise ValueError("Only one of datapath or file_list may be provided")
```
rahul-tuli: nit: I think there is some benefit here in calling out that users should `"Use 'datapath' to auto-discover files or 'file_list' for explicit paths."` in the error message?

src/speculators/train/checkpointer.py
```diff
@@ -0,0 +1,237 @@
...
+            full_state_dict[key] = f.get_tensor(key)
+    return full_state_dict
+
+
+class SingleGPUCheckpointer(BaseCheckpointer):
+    def load_model_state_dict(
+        self, model: PreTrainedModel, float_dtype: torch.dtype | None = None
+    ):
+        full_state_dict = load_safetensors_state_dict(
+            self.model_path(self.previous_epoch), "cuda:0"
```
rahul-tuli: should we be hardcoding to "cuda:0"?

src/speculators/train/trainer.py
fynnsu: Both work and in my opinion `named_parameters()` is slightly preferable as the parameters will be named within the optimizer state dict. From the AdamW docs:

> params (iterable) – iterable of parameters or named_parameters to optimize or iterable of dicts defining parameter groups. When using named_parameters, all parameters in all groups should be named

src/speculators/train/checkpointer.py
fynnsu: I actually decided to change the type of `sd` to `pytree.PyTree` as it better represents the arguments this function supports. Added `pytree.PyTree` as the return type too.

src/speculators/train/checkpointer.py
fynnsu: No, probably not in the long term. For now this is only used in the single gpu code where the script is likely being launched with `CUDA_VISIBLE_DEVICES=M` anyways, which will result on `cuda:0` mapping to `M`. In the long term we may also want to generalize to non-cuda accelerators.

For now, I added a todo to generalize this.

src/speculators/train/eagle3/core.py
HDCharles: Wonder how annoying it would be to permute the verifier input/output embeddings so they just match the draft vocab until the cutoff. Seems like it should be possible

fynnsu: As discussed offline, I don't think this would be possible without also modifying the verifier model in vllm.

PR #144, opened 2025-10-06
added loading util for specific layers

shanjiaz:
We were loading the full verifier model only to get the embeddings, specific layers. This util helps saving memory by loading only the shards we needed and returns a dict of layer name -> layers. 

Tested locally:

``` python
model_path = "shanjiaz/Meta-Llama-3-8B-Instruct-FP8-BLOCK"
layer_names = ["lm_head.weight", "model.embed_tokens.weight"]
layer = load_model_layers(layer_names, model_path)
for k, v in layer.items():
    print(k, v.shape)
    print(f"sample data: {v.flatten()[:5]}")
```

``` python
model_path = "/home/hzhao/.cache/huggingface/hub/models--shanjiaz--Meta-Llama-3-8B-Instruct-FP8-BLOCK/snapshots/ea6d7c1a6a0874d9db6511ce93da2b777f24376f"
layer_names = ["lm_head.weight", "model.embed_tokens.weight"]
layer = load_model_layers(layer_names, model_path)
for k, v in layer.items():
    print(k, v.shape)
    print(f"sample data: {v.flatten()[:5]}")
```

```
2025-10-06 14:02:21.042 | INFO     | __main__:_resolve_file:64 - Loading from local directory:
2025-10-06 14:02:21.043 | INFO     | __main__:_resolve_file:64 - Loading from local directory:
2025-10-06 14:02:21.044 | INFO     | __main__:_resolve_file:64 - Loading from local directory:
lm_head.weight torch.Size([128256, 4096])
sample data: tensor([ 0.0098,  0.0175,  0.0037,  0.0222, -0.0194], dtype=torch.bfloat16)
model.embed_tokens.weight torch.Size([128256, 4096])
sample data: tensor([ 0.0013,  0.0054, -0.0022,  0.0003, -0.0024], dtype=torch.bfloat16)
2025-10-06 14:02:52.385 | INFO     | __main__:_resolve_file:70 - Loading from huggingface directory:
2025-10-06 14:02:52.468 | INFO     | __main__:_resolve_file:70 - Loading from huggingface directory:
2025-10-06 14:02:52.508 | INFO     | __main__:_resolve_file:70 - Loading from huggingface directory:
lm_head.weight torch.Size([128256, 4096])
sample data: tensor([ 0.0098,  0.0175,  0.0037,  0.0222, -0.0194], dtype=torch.bfloat16)
model.embed_tokens.weight torch.Size([128256, 4096])
sample data: tensor([ 0.0013,  0.0054, -0.0022,  0.0003, -0.0024], dtype=torch.bfloat16)
```

### DIFF
--- a/src/speculators/utils/loading.py
+++ b/src/speculators/utils/loading.py
@@ -1,6 +1,6 @@
 import json
 from pathlib import Path
-from typing import Any, Optional, Union
+from typing import Any
 
 import torch
 from huggingface_hub import hf_hub_download
@@ -72,5 +72,5 @@ def _resolve_file(model_path: str, file_name: str) -> Path:
             raise FileNotFoundError(f"Expected local file missing: {p}")
         return p
     # Treat as repo_id on the Hub
-    logger.info("Loading from huggingface directory: {}", model_path)
+    logger.info(f"Loading from huggingface directory: {model_path}: {file_name}")
     return Path(hf_hub_download(repo_id=model_path, filename=file_name))


src/speculators/utils/loading.py
```diff
@@ -0,0 +1,79 @@
...
+        index = json.load(f)
+
+    weight_map: dict[str, str] = index["weight_map"]
+
+    # group requested names by shard filename
+    shard_to_names: dict[str, list[str]] = {}
+    for name in layer_names:
+        shard = weight_map.get(name)
+        if shard is None:
+            logger.warning(f"Tensor '{name}' not found in index weight_map.")
```
fynnsu: It might be better to make this an error.

src/speculators/utils/loading.py
```diff
@@ -0,0 +1,79 @@
...
+
+    # fetch each required shard and extract only the requested tensors
+    out: dict[str, Any] = {}
+    for shard_file, names in shard_to_names.items():
+        shard_path = _resolve_file(model_path, shard_file)
+        with safe_open(shard_path, framework="pt", device="cpu") as f:
+            available = set(f.keys())
+            for name in names:
+                if name not in available:
+                    logger.warning(
```
fynnsu: Same here

src/speculators/utils/loading.py
```diff
@@ -0,0 +1,79 @@
...
+    """
+    model_path_obj = Path(model_path)
+    if model_path_obj.is_dir():
+        logger.info("Loading from local directory: {}", model_path)
+        p = model_path_obj / file_name
+        if not p.exists():
+            raise FileNotFoundError(f"Expected local file missing: {p}")
+        return p
+    # Treat as repo_id on the Hub
+    logger.info("Loading from huggingface directory: {}", model_path)
```
fynnsu: This (or the other logger.info call) is going to print out each time this function is called. If we load multiple values (i.e. `embed_tokens.weight` and `lm_head.weight`) it would be nice to also include the `file_name` in these log calls.

fynnsu (dismissed):
Looks good!

Added a few suggestions

src/speculators/utils/loading.py
shanjiaz: Will do!

src/speculators/utils/loading.py
shanjiaz: sure!

src/speculators/utils/loading.py
```diff
@@ -0,0 +1,79 @@
...
+    Load one or more named tensors from a HF repo using safetensors shards.
+    Returns a single tensor if len(layer_names)==1, else a dict[name] -> tensor.
+
+    :param layer_names: list of tensor names to load, e.g.
+    ["model.embed_tokens.weight", "lm_head.weight"]
+    :param model_path: either a local directory containing model.safetensors.index
+    :return: a single tensor or a dict of tensors
+    """
+    # download the index file
+    index_file = _resolve_file(model_path, "model.safetensors.index.json")
```
kylesayrs, line 24: This file doesn't always exist

src/speculators/utils/loading.py
```diff
@@ -0,0 +1,79 @@
...
+
+import torch
+from huggingface_hub import hf_hub_download
+from loguru import logger
+from safetensors import safe_open
+
+
+def load_model_layers(
+    layer_names: list[str], model_path: str
+) -> Optional[Union[torch.Tensor, dict[str, torch.Tensor]]]:
```
brian-dellabetta: Better to make your output types as simple as possible so that users don't need their own if/elif checks to parse the return type. Can you return a single dict that is empty if there aren't any, and a dict with a single key if `len(layer_names)==1`?
```suggestion
) -> dict[str, torch.Tensor]:
```

src/speculators/utils/loading.py
shanjiaz: Yeah I think in that case this function will raise FileNotFoundError, and we'll have to load the entire model.

src/speculators/utils/loading.py
shanjiaz: sure!

brian-dellabetta (approved):
👏

src/speculators/utils/loading.py
fynnsu: Sorry, I actually meant
```python
raise ValueError(...)
```
That way we actually crash when this case occurs, instead of just logging the issue.

PR #147, opened 2025-10-09
add num_hidden_layers

shanjiaz:
Instead of assuming 1 hidden layer, pull num_hidden_layers from checkpoint config.

### DIFF
--- a/tests/unit/convert/test_eagle3_converter.py
+++ b/tests/unit/convert/test_eagle3_converter.py
@@ -242,3 +242,24 @@ class TestEagle3ConverterFixes:
             config_dict = config.to_dict()
             assert "eagle_aux_hidden_state_layer_ids" in config_dict
             assert config_dict["eagle_aux_hidden_state_layer_ids"] == layer_ids
+
+    @pytest.mark.sanity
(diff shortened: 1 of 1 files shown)


tests/unit/convert/test_eagle3_converter.py
```diff
@@ -93,6 +93,44 @@ def test_config_max_position_embeddings_logic(
...
+        llama_config = converter._create_transformer_config_from_eagle(
+            sample_eagle3_config, "meta-llama/Llama-3.1-8B"
+        )
+        assert llama_config.num_hidden_layers == 3
+
+    @pytest.mark.sanity
+    @patch(
+        "speculators.convert.eagle.eagle3_converter.PretrainedConfig.get_config_dict"
+    )
+    def test_config_num_hidden_layers_default(
```
rahul-tuli, line 122: Could we also add a test for converting an actual multilayer checkpoint? We could use https://huggingface.co/nm-testing/random-weights-llama3.1.8b-2layer-eagle3/tree/main

shanjiaz: yep!

PR #151, opened 2025-10-21
Remove PyPI publishing steps from nightly workflow

dsikka:
- Removed steps to push wheel and tar.gz artifacts to PyPI.
- DIsabling until enabled by QA


fynnsu (approved):
Mostly looks good but I'm wondering if we want to keep parts of the `.github/workflows/release-candidate.yml` workflow. 

This runs our test / quality suite on any push to a `release/*` branch which is behavior we will want to keep. Alternatively and possibly better, would be to add `- 'release/*'` to our `main.yml` test so that any test we run on pushes to `main` we also run on pushes to release branches.

PR #152, opened 2025-10-23
Remove remaining python 3.9 usages

fynnsu:
I tried to just update the 3.9 usages and it resulted in a bunch of new ruff lint errors. 

Setting the target version for `ruff` to `3.10` resulted in a bunch of upgrade lints failing. Most of these were automatically handled by `ruff check --fix` and `ruff format`. A few required manual intervention.

Lastly, I updated the version of ruff we use because there was an upgrade rule on the old version that has since been deprecated. I could've added an ignore rule instead but I don't think it hurts here to update to a newer ruff version.

### DIFF
--- a/tests/e2e/vllm/run_vllm.py
+++ b/tests/e2e/vllm/run_vllm.py
@@ -1,0 +1,50 @@
+import argparse
+import json
+from pathlib import Path
+
+from vllm import LLM, SamplingParams  # type: ignore[import-not-found]
+
+
+def parse_args():
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--sampling-params-args",
+        type=str,
+        required=True,
+        help="JSON-serialized kwargs for SamplingParams instantiation",
+    )
+    parser.add_argument(
+        "--llm-args",
+        type=str,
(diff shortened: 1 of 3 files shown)


fynnsu:
> Can you resolve the conflict? LGTM otherwise

Rebased

shanjiaz (dismissed):

pyproject.toml
```diff
@@ -245,6 +245,7 @@ select = [
     "PGH004",  # allow general style ignores
     "ERA001", # allow commented-out code
     "N80",  # allow upper case names
+    "PLC0415", # ignore import not at top of file errors
```
dsikka, line 248: Should we add something to ignore TODOs?

dsikka (approved):

pyproject.toml
fynnsu: Actually all we need to do is remove this line:
https://github.com/vllm-project/speculators/blob/b46b376d029b384a1bb2f22b871b06dc3e396d7a/pyproject.toml#L214

Currently, I've already done that in my training pr, so I think it's okay to leave until that is merged.

PR #153, opened 2025-10-23
Refactor e2e tests to support external vLLM

dbarbuzzi:
## Summary

Refactor the vLLM-based end-to-end tests (`tests/e2e/vllm/test_eagle3.py`) to support using vLLM from a separate virtual environment to keep the `speculators` test virtual environment clean and the strategy more similar to how the tools would actually be used.

## Details

The changes are similar to the work done to enable this in `llm-compressor`:

* The running of vLLM with the tests’ parameters is separated to a different file
* The test finds the Python from the correct virtual environment from an env var (`VLLM_PYTHON`), defaulting to the same env if not set
* The test serializes and passes the various inputs needed, calls the new `run_vllm.py` file with them, and then reads a file with serialized results to be able to perform the same level of validation (just the presence/count/type of the token IDs)
* Similarly, the existing logging was preserved as well as possible, with logging around the use of `run_vllm.py` added for debugging purposes during test failures

## Testing

An abbreviated run with only one test, and the model being passed to `run_vllm.py` hard-coded to one that works in vLLM, is here: https://github.com/neuralmagic/llm-compressor-testing/actions/runs/18754731123

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -129,7 +129,7 @@ warn_redundant_casts = true
 warn_unused_ignores = false
 show_error_codes = true
 namespace_packages = true
-exclude = ["venv", ".tox", "build", "dist", "research", "tests/e2e/vllm/run_vllm.py"]
+exclude = ["venv", ".tox", "build", "dist", "research"]
 
 # Silence "type import errors" as our 3rd-party libs does not have types
 # Check: https://mypy.readthedocs.io/en/latest/config_file.html#import-discovery


tests/e2e/vllm/run_vllm.py
```diff
@@ -0,0 +1,50 @@
+import argparse
+import json
+from pathlib import Path
+
+from vllm import LLM, SamplingParams
```
fynnsu:
```suggestion
from vllm import LLM, SamplingParams  # type: ignore[import-not-found]
```

This might be better than the full file exclude in `pyproject.toml`. This way we will still be warned if there are other type issues in the file.

pyproject.toml
```diff
@@ -129,7 +129,7 @@ warn_redundant_casts = true
 warn_unused_ignores = false
 show_error_codes = true
 namespace_packages = true
-exclude = ["venv", ".tox", "build", "dist", "research"]
+exclude = ["venv", ".tox", "build", "dist", "research", "tests/e2e/vllm/run_vllm.py"]
```
fynnsu:
```suggestion
exclude = ["venv", ".tox", "build", "dist", "research"]
```

See above.

tests/e2e/vllm/run_vllm.py
dbarbuzzi: I like this a lot better, thanks!

pyproject.toml
dbarbuzzi: Thanks!

fynnsu (approved):

PR #155, opened 2025-10-24
Added loading util tests

shanjiaz:
Added unit testing for loading utils. 

Since the util relies on index file existing, created a random weights fake llama model that has two weights file for testing.

### DIFF
--- a/tests/unit/utils/test_loading.py
+++ b/tests/unit/utils/test_loading.py
@@ -14,7 +14,7 @@ TEST_MODEL_REPO = "nm-testing/tiny-testing-random-weights"
 # _resolve_file Tests
 
 
-@pytest.mark.smoke
+@pytest.mark.sanity
 def test_resolve_file_hub_download():
     """Test resolving a file from HuggingFace Hub using real model."""
     result = _resolve_file(TEST_MODEL_REPO, "config.json")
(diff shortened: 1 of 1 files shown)


tests/unit/utils/test_loading.py
```diff
@@ -0,0 +1,103 @@
...
+        )
+
+        # Check values are identical
+        assert torch.equal(util_tensor, model_tensor), (
+            f"Tensor values don't match for {layer_name}"
+        )
+
+    # Clean up
+    del full_model
+    del state_dict
```
brian-dellabetta: won't these be garbage collected after the test completes?

tests/unit/utils/test_loading.py
```diff
@@ -0,0 +1,103 @@
...
+
+from speculators.utils.loading import _resolve_file, load_model_layers
+
+# Test model from HuggingFace
+TEST_MODEL_REPO = "nm-testing/tiny-testing-random-weights"
+
+# _resolve_file Tests
+
+
+@pytest.mark.regression
```
brian-dellabetta: aren't these unit tests? maybe speculators uses a different pattern for marking

shanjiaz: They are! I'm just not too sure how to categorize these tests, they don't seem like smoke tests for me either.

tests/unit/utils/test_loading.py
shanjiaz: right!

tests/unit/utils/test_loading.py
dsikka: https://github.com/vllm-project/speculators/blob/fc59c6ae830b1ffd12fb7bdb1c6c080d4fe11e8b/pyproject.toml#L258

Sanity

shanjiaz: Will do!

PR #157, opened 2025-10-27
Offline data generation for eagle3 training

shanjiaz:
This PR includes the entire pipeline of data generation, the PR kinda grew organically and become gigantic. To help mitigating the stress of reviewing, here's a guide to facilitate the review process. (Hope this helps!)
🔴 means most important and contains key logic
🟡 means less important, usually user-facing entry points
🟢 means least important, supporting infrastructure 

Here are the details on each part organized based on the dependencies:
## preprocessing 
🔴`src/speculators/data_generation/preprocessing.py`
This file contains the main logic of preprocessing, including raw dataset loading and formatting, loss mask calculation, sample tokenization, cache key generation and the main function `load_and_preprocess_dataset`.  
🟡`scripts/preprocess_data.py`
This is a standalone simple script that preprocess raw conversational data and saves them in a cache file specified by users. It's optional, users have the option to run this script if they want to separate preprocessing from hidden states generation. The default is running `data_generation_offline.py` directly, which handles preprocessing, hidden states generation, caching and savings. 
🟢 `src/speculators/data_generation/configs.py`
Configuration registries, defines chat templates for different model formats and dataset loading configurations.
## Hidden States Generation
🔴 `src/speculators/data_generation/vllm_hidden_states_generator.py`
This file contains the core logic for extracting hidden states from intermediate layers during prefill using vLLM. Key features include:
    - Auto-selection of layers following EAGLE3 pattern: [layer_2, layer_mid, layer_-3, layer_-1] (first 3 for fusion, last for target logits)
    - Multi-GPU tensor parallelism support via MultiprocExecutor
    - Batch processing with variable-length sequences
    - Returns dict with input_ids, hidden_state (concatenated layers), and loss_mask

🔴 `src/speculators/data_generation/custom_worker.py`
Custom vLLM worker extension that hooks into the model's forward pass to capture hidden states from specified layers during prefill. This integrates with vLLM's v1 API architecture.

🟡 `scripts/data_generation_offline.py`
Main end-to-end pipeline script that users should use. This script:
  - Automatically handles preprocessing (runs it if cache not found, loads from cache otherwise)
  - Uses vLLM to extract hidden states in batches
  - Saves each sample as a separate .pt file (to support variable-length sequences)
  - Generates data_config.json with metadata for reproducibility
  - Supports auto-resume by detecting existing files
  - Optimized for throughput with configurable batch_size and max_num_batched_tokens

## Vocabulary Mapping
🟡 `src/speculators/data_generation/vocab_mapping.py`
Contains logic for building vocabulary mappings (d2t and t2d) between draft and target models when vocabularies differ. Maps draft tokens to target tokens based on frequency distribution.

🟡 `scripts/build_vocab_mapping.py`
Standalone script to generate d2t.npy and t2d.npy files from token frequency distributions collected during preprocessing.

## Infrastructure & Utilities
 🟢 `src/speculators/data_generation/logging_utils.pyCustom` 
logger with colored sections, subsections, and config display for better pipeline visibility.
 🟢 `src/speculators/data_generation/__init__.py`
Package exports.

## Tests
  🟡 `tests/integration/data_generation/test_vllm_hidden_states.py`
Focused tests for the vLLM hidden states generator component.

## Key Design Decisions
  1. Single-file storage: Each training sample is saved as a separate .pt file to support variable-length sequences efficiently. Training code now handles batching nicely, but we're planning on moving batching to the data generation side to avoid the batch -> unbatch -> rebatch overhead.
  2. 4-layer extraction, auto-selects 4 layers from the target model if layer ids are not specified
  3. Automatic preprocessing: data_generation_offline.py handles preprocessing transparently via cache, eliminating manual steps and preventing parameter mismatches.
  4. Metadata tracking: data_config.json saves all generation parameters for reproducibility.

## Usage Example
```
 python scripts/data_generation_offline.py \
      --target-model-path meta-llama/Llama-3.1-8B \
      --train-data-path sharegpt \
      --chat-template llama3 \
      --output-dir ./training_data \
      --max-samples 5000 \
      --batch-size 16
```
```
  # Generate vocabulary mappings (if needed)
  python scripts/build_vocab_mapping.py \
      --token-freq-path ./cache/token_frequencies/xxx_token_freq.pt \
      --draft-vocab-size 32000 \
      --target-vocab-size 128256 \
      --output-path ./training_data/
```

More details can be found in `src/speculators/data_generation/README.md`. Please feel free to reach out with any questions.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -131,7 +131,7 @@ warn_redundant_casts = true
 warn_unused_ignores = false
 show_error_codes = true
 namespace_packages = true
-exclude = ["venv", ".tox", "build", "dist", "research"]
+exclude = ["venv", ".tox", "build", "dist"]
 
 # Silence "type import errors" as our 3rd-party libs does not have types
 # Check: https://mypy.readthedocs.io/en/latest/config_file.html#import-discovery
@@ -236,6 +236,7 @@ select = [
 "src/speculators/convert/**/*.py" = [
     "BLE001", # allow catching Exception for conversion errors
 ]
+
 "src/speculators/data_generation/**/*.py" = [
     "S106", # false positives for chat template tokens
     "S324", # MD5 is used for cache keys, not security
@@ -251,18 +252,6 @@ select = [
 ]
 "scripts/**/*.py" = [
     "PTH", # os.path is acceptable in scripts
-]
-"research/**/*.py" = [
-    "PTH",  # ignore using os.* instead of Path errors
-    "T20",  # ignore usage of print statement errors
-    "PLR",  # ignore code complexity errors
-    "NPY",  # ignore numpy formatting/usage errors
-    "E722",  # allow bare exception catching
-    "C90",  # ignore code complexity errors
-    "PGH004",  # allow general style ignores
-    "ERA001", # allow commented-out code
-    "N80",  # allow upper case names
-    "PLC0415", # ignore import not at top of file errors
 ]
 
 [tool.ruff.lint.isort]
--- a/README.md
+++ b/README.md
@@ -193,9 +193,9 @@ print(speculators.__version__)
 
 Here you can find links to our research implementations. These provide prototype code for immediate enablement and experimentation, with plans for productization into the main package soon.
 
-- [eagle3](https://github.com/neuralmagic/speculators/tree/main/research/eagle3): This implementation trains models similar to the EAGLE 3 architecture, specifically utilizing the Train Time Test method.
+- eagle3: This implementation trains models similar to the EAGLE 3 architecture, specifically utilizing the Train Time Test method.
 
-- [hass](https://github.com/neuralmagic/speculators/tree/main/research/hass): This implementation trains models that are a variation on the EAGLE 1 architecture using the [HASS](https://github.com/HArmonizedSS/HASS) method.
+- hass: This implementation trains models that are a variation on the EAGLE 1 architecture using the [HASS](https://github.com/HArmonizedSS/HASS) method.
 
 ## License
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -101,5 +101,5 @@ VLLM_USE_V1=1 vllm serve RedHatAI/Qwen3-8B-speculator.eagle3
 
 Speculators includes prototype implementations of cutting-edge speculative decoding research:
 
-- **[EAGLE 3](https://github.com/neuralmagic/speculators/blob/main/research/eagle3/README.md)**: Train Time Test method implementation for advanced token speculation
-- **[HASS](https://github.com/neuralmagic/speculators/blob/main/research/hass/README.md)**: EAGLE 1 architecture variation using the HASS training method
+- **EAGLE 3**: Train Time Test method implementation for advanced token speculation
+- **HASS**: EAGLE 1 architecture variation using the HASS training method
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,6 +1,6 @@
 # Training
 
-Speculators currently supports training of Eagle, Eagle 3 and HASS-style speculative decoders. This functionality is still under development and includes the necessary data generation and training steps required to produce the models. Examples of the training process can be found under the [`research`](https://github.com/neuralmagic/speculators/tree/main/research) directory for [Eagle 3](https://github.com/neuralmagic/speculators/blob/main/research/eagle3/README.md) and [HASS](https://github.com/neuralmagic/speculators/blob/main/research/hass/README.md).
+Speculators currently supports training of Eagle, Eagle 3 and HASS-style speculative decoders. This functionality is still under development and includes the necessary data generation and training steps required to produce the models. Examples of the training process can be found under the research directory for Eagle 3 and HASS.
 
 # Conversion
 


src/speculators/data_generation/custom_worker.py
```diff
@@ -0,0 +1,115 @@
...
+    _captured_states: list[Any] | None
+    _should_capture: bool
+    model_runner: Any
+
+    def _store_captured_states(self, aux_hidden_states):
+        if self._captured_states is None:
+            self._captured_states = aux_hidden_states
+        else:
+            for i, h in enumerate(aux_hidden_states):
+                self._captured_states[i] = torch.cat(
```
kylesayrs: Doing many cats might be costly. Consider keeping these at as a list, and concatting only when you need to read them (_get_captured_states)?

src/speculators/data_generation/custom_worker.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def _enable_capture(self):
+        """Enable hidden states capture"""
+        self._should_capture = True
+        self._captured_states = None
+
+    def _disable_capture(self):
+        """Disable hidden states capture and clear captured data"""
+        self._should_capture = False
+        self._captured_states = None
```
kylesayrs: Consider deleting intermediates `self._captured_states`

src/speculators/data_generation/configs.py
```diff
@@ -0,0 +1,140 @@
...
+        user_header="[INST] ",
+        assistant_header="",
+        end_of_turn_token=" [/INST]",
+        bos_token="<s>",
+        eos_token="</s>",
+    ),
+}
+
+
+def format_conversation(
```
kylesayrs: [apply_chat_template](https://huggingface.co/docs/transformers/en/chat_templating)?

src/speculators/data_generation/preprocessing.py
```diff
@@ -0,0 +1,262 @@
...
+    chat_template: str,
+    seq_length: int,
+    train_data_path: str,
+) -> str:
+    """Generate MD5 cache key from preprocessing parameters."""
+    key_string = f"{target_model_path}_{chat_template}_{seq_length}_{train_data_path}"
+    return hashlib.md5(key_string.encode()).hexdigest()
+
+
+def load_and_preprocess_dataset(
```
kylesayrs, line 231: FYI this is very similar to [TextGenerationDataset](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/transformers/finetune/data/base.py). I don't think these two things have to be the same, but something to keep in mind when expanding to new datasets / processing techniques.

src/speculators/data_generation/preprocessing.py
```diff
@@ -0,0 +1,262 @@
...
+
+    raw_dataset = load_raw_dataset(train_data_path, num_proc=build_dataset_num_proc)
+    raw_dataset = raw_dataset.shuffle(seed=seed)
+
+    if max_samples is not None and len(raw_dataset) > max_samples:
+        raw_dataset = raw_dataset.select(range(max_samples))
+
+    log.info(f"Loaded {len(raw_dataset)} samples")
+
+    cache_key = generate_cache_key(
```
kylesayrs: FYI some of this functionality is already built into HF Datasets
https://huggingface.co/docs/datasets/en/about_cache

src/speculators/data_generation/vllm_hidden_states_generator.py
```diff
@@ -0,0 +1,275 @@
...
+        # Increment to ensure unique request IDs across calls
+        # (prevents KV cache corruption with delayed block freeing)
+        self._request_counter += 1
+
+        self.executor.collective_rpc("_enable_capture")
+
+        schedule_iterations = 0
+        while True:
+            scheduler_output = self.scheduler.schedule()
+            if scheduler_output.total_num_scheduled_tokens == 0:
```
kylesayrs: nit: consider moving this into the while condition

src/speculators/data_generation/vllm_hidden_states_generator.py
```diff
@@ -0,0 +1,275 @@
...
+            results.append(
+                {
+                    "input_ids": input_ids_tensor,
+                    "hidden_states": layer_states,
+                    "loss_mask": None,
+                }
+            )
+            offset += seq_len
+
+        del aux_hidden_states
```
kylesayrs: Doesn't actually do anything

src/speculators/data_generation/vllm_hidden_states_generator.py
```diff
@@ -0,0 +1,275 @@
...
+        torch.cuda.empty_cache()
```
kylesayrs, line 293: This is quite costly. Is it needed?

scripts/data_generation_offline.py
```diff
@@ -0,0 +1,362 @@
...
+            batch_input_ids = batch["input_ids"]
+            batch_loss_mask = batch["loss_mask"]
+
+            results = generator.generate(batch_input_ids)
+
+            # Submit save operations to thread pool (async I/O)
+            for j, result in enumerate(results):
+                result_cleaned = {
+                    "input_ids": result["input_ids"].clone(),
+                    "hidden_states": [h.clone() for h in result["hidden_states"]],
```
kylesayrs:
```suggestion
                    "hidden_states": [h.contiguous() for h in result["hidden_states"]],
```

scripts/data_generation_offline.py
```diff
@@ -0,0 +1,362 @@
...
+                    "loss_mask": batch_loss_mask[j].clone(),
```
kylesayrs: Needs clone?

scripts/data_generation_offline.py
```diff
@@ -0,0 +1,362 @@
...
+            batch = dataset[i:batch_end]
+            batch_input_ids = batch["input_ids"]
+            batch_loss_mask = batch["loss_mask"]
+
+            results = generator.generate(batch_input_ids)
+
+            # Submit save operations to thread pool (async I/O)
+            for j, result in enumerate(results):
+                result_cleaned = {
+                    "input_ids": result["input_ids"].clone(),
```
kylesayrs: Needs clone?

scripts/data_generation_offline.py
```diff
@@ -0,0 +1,362 @@
...
+                    "hidden_states": [h.clone() for h in result["hidden_states"]],
+                    "loss_mask": batch_loss_mask[j].clone(),
+                }
+                output_path = os.path.join(args.output_dir, f"data_{file_idx}.pt")
+                future = thread_executor.submit(
+                    save_sample_to_disk, result_cleaned, output_path
+                )
```
kylesayrs, line 292: Why not just call save_sample_to_disk in this thread? Otherwise you risk thread starvation and oom

kylesayrs: Also reduced overhead from thread launching.

kylesayrs: If generation is the bottleneck, you need disk writing to be coupled or else you'll oom
If disk writing is the bottleneck, decoupling the two step won't make total throughput any better.

scripts/build_vocab_mapping.py
```diff
@@ -0,0 +1,100 @@
...
+
+def main():
+    args = parse_args()
+
+    if not os.path.exists(args.token_freq_path):
+        raise FileNotFoundError(
+            f"Token frequency file not found: {args.token_freq_path}"
+        )
+
+    token_freq_dict = torch.load(args.token_freq_path, weights_only=False)
```
rahul-tuli: Why are we setting `weights_only` to `False`, this can be a potential security concern, we should default this argument to `True`

src/speculators/data_generation/vllm_hidden_states_generator.py
```diff
@@ -0,0 +1,275 @@
...
+
+        return results
+
+    def __del__(self):
+        if hasattr(self, "executor"):
+            try:
+                self.executor.shutdown()
+            except Exception:
```
rahul-tuli, line 301: Silently swallowing all exceptions can be dangerous here, I would ideally like to make VllmHiddenStatesGenerator a context manager, if not could we atleast log a warning here?

src/speculators/data_generation/custom_worker.py
rahul-tuli: +1, Concatenate once when retrieving

src/speculators/data_generation/vllm_hidden_states_generator.py
```diff
@@ -0,0 +1,275 @@
...
+        cache_config = CacheConfig(
+            block_size=16,
+            gpu_memory_utilization=gpu_memory_utilization,
+        )
+
+        # For prefill-only workloads, use conservative scheduler limits
+        # to reduce warmup memory allocation. max_num_seqs controls the
+        # warmup allocation size (see gpu_worker.py:441-444).
+        # We set it to a small value since we only do prefill in batches.
+        max_num_seqs = 32
```
rahul-tuli: This file has several magic numbers spread out, could we extract these into CONSTANTS with comments please?

scripts/build_vocab_mapping.py
```diff
@@ -0,0 +1,100 @@
...
+        type=int,
+        required=True,
+        help="Vocabulary size for the target model",
+    )
+    parser.add_argument(
+        "--output-path",
+        type=str,
+        required=True,
+        help="Path to save the vocabulary mapping file (.pt)",
+    )
```
fynnsu, line 62: This is confusing/maybe outdated. 

The current code logic is that we take the `args.output_path`, then go to the parent directory and then create `t2d.npy` and `d2t.npy` files in that parent directory. There are also no `.pt` files being saved.

We should change this so that we just create the `t2d.npy` and `d2t.npy` files in the provided directory, and update this help (as well as the example usage at the start of the file) to reflect that.

fynnsu: As a side-note, I wonder if we should be saving these differently. We would need to update the loading on the training side too of course, but I just used this format because that's what was being used by the research code. We could instead save a single `.pt` file containing a dict with both tensors? We can leave this for a future pr, but just wanted to mention it.

scripts/build_vocab_mapping.py
```diff
@@ -0,0 +1,100 @@
...
+            f"Token frequency file not found: {args.token_freq_path}"
+        )
+
+    token_freq_dict = torch.load(args.token_freq_path, weights_only=False)
+
+    d2t, t2d = build_vocab_mappings_from_distribution(
+        token_freq_dict=token_freq_dict,
+        draft_vocab_size=args.draft_vocab_size,
+        target_vocab_size=args.target_vocab_size,
+    )
```
fynnsu, line 80: (Potentially future pr) I think we might need to have explicit handling for the "special tokens" to make sure they get included in the reduced vocab. Rather than adding the special tokens as inputs to the cli (which would be very messy), we could just pass in the target_model name/path and load the tokenizer config from it, which would also give us the target_vocab_size.

Again, not a change that needs to be made now, but something we might want/need in the future.

src/speculators/data_generation/custom_worker.py
```diff
@@ -0,0 +1,115 @@
...
+
+        if not supports_eagle3(model):
+            raise ValueError(
+                f"Model {type(model).__name__} does not support hidden state extraction"
+            )
+
+        base_model = model.model  # type: ignore[attr-defined]
+        base_model.aux_hidden_state_layers = tuple(layer_ids)
+
+        base_model.forward = self._create_patched_forward(base_model)
```
rahul-tuli: This is not idempotent, if _setup_hidden_states_capture is called twice it will double wrap; could we add an internal flag to prevent this? something like  this `_overridden` flag in modify_save_pretrained in llmcompressor https://github.com/vllm-project/llm-compressor/blob/15b29be3cbabdd4b7ccbbc66d2d8d88ec87d031e/src/llmcompressor/transformers/compression/compressed_tensors_utils.py#L113

scripts/build_vocab_mapping.py
```diff
@@ -0,0 +1,100 @@
...
+        help="Path to save the vocabulary mapping file (.pt)",
+    )
+
+    return parser.parse_args()
+
+
+def main():
+    args = parse_args()
+
+    if not os.path.exists(args.token_freq_path):
```
fynnsu: nit: I think it might be nicer to use `pathlib.Path` objects in this function. Both here and instead of `os.path.dirname`, `os.makedirs`, `os.path.join`, etc.

src/speculators/data_generation/custom_worker.py
```diff
@@ -0,0 +1,115 @@
...
+from vllm.model_executor.models.interfaces import supports_eagle3
+from vllm.sequence import IntermediateTensors
+
+__all__ = ["HiddenStatesWorkerExtension"]
+
+logger = logging.getLogger(__name__)
+
+
+class HiddenStatesWorkerExtension:
+    """Worker extension that adds hidden states capture functionality."""
```
rahul-tuli: Could we add a more extensive docstring here? Something calling out that this class hooks into VLLM's Worker initialization and only captures on TP rank 0 to avoid duplicates and also that it stores states in GPU memory during batch processing

rahul-tuli (commented):
This is a very good implementation, however we could beef up the testing before merge:  
- At least 2-3 integration tests for main functionality
 - And unit tests for preprocessing logic

Great Job!

scripts/preprocess_data.py
```diff
@@ -0,0 +1,129 @@
...
+        help="Path to training data (JSON/JSONL, sharegpt, or ultrachat)",
+    )
+
+    # Processing parameters
+    parser.add_argument(
+        "--chat-template",
+        type=str,
+        required=True,
+        help="Chat template name (e.g., qwen2, llama3, etc.)",
+    )
```
fynnsu: I think this should eventually be derived directly from the `--target-model-path` (i.e. directly loading the chat template for the model, instead of a having a preset list), but could be a future pr.

scripts/preprocess_data.py
```diff
@@ -0,0 +1,129 @@
...
+    parser.add_argument(
+        "--seq-length",
+        type=int,
+        default=2048,
+        help="Maximum sequence length (default: 2048)",
```
fynnsu, line 52: Should confirm with research on this default. I think we might want higher (4096).

src/speculators/data_generation/configs.py
```diff
@@ -0,0 +1,140 @@
...
+    name: str
+    system_prompt: str
+    user_header: str
+    assistant_header: str
+    end_of_turn_token: str
+    bos_token: str | None = None
+    eos_token: str | None = None
+
+
+CHAT_TEMPLATES: dict[str, ChatTemplate] = {
```
fynnsu: As mentioned in previous comment, I think we will want to eventually use the model's chat template directly, instead of this dataclass + fixed format.

src/speculators/data_generation/custom_worker.py
```diff
@@ -0,0 +1,115 @@
...
+
+    _layer_ids: list[int]
+    _captured_states: list[Any] | None
+    _should_capture: bool
+    model_runner: Any
```
fynnsu: These are not really being used.

This sets up a bunch of class variables, which then get shadowed by instance variables in `_setup_hidden_states_capture`

In the example below, you can see that `T.a` and `T.b` (class vars) are completely separate from `t.a` and `t.b` (instance vars).
```python

>>> class T:
...     a: int
...     b: list[int]
...     def __init__(self):
...         self.a = 3
...         self.b = []
...         self.b.append(2)
...
>>> T.a = 1
>>> T.b = [4, 5, 6]
>>> t = T()
>>> t.a
3
>>> t.b
[2]
>>> T.a
1
>>> T.b
[4, 5, 6]
```

If you want to keep the type hints, I would just add them in `_setup_hidden_states_capture` when you instantiate the attributes.

src/speculators/data_generation/custom_worker.py
shanjiaz: Will do!

fynnsu: Agreed, and then you can also simplify the logic so that `self._captured_states` is always a list, which you just append to. 

Also as a side note, why are combining multiple `aux_hidden_states` on the sequence dimension? Is this because the prefill is being chunked somehow?

scripts/data_generation_offline.py
shanjiaz: Nice!

scripts/data_generation_offline.py
shanjiaz: Removed!

src/speculators/data_generation/custom_worker.py
```diff
@@ -0,0 +1,115 @@
...
+                residual = None
+            else:
+                assert intermediate_tensors is not None
+                hidden_states = intermediate_tensors["hidden_states"]
+                residual = intermediate_tensors["residual"]
+
+            aux_hidden_states = []
+            should_capture = self._should_capture and get_tp_group().rank_in_group == 0
+            target_layers = (
+                base_model.aux_hidden_state_layers if should_capture else frozenset()
```
fynnsu: Why are we setting `base_model.aux_hidden_state_layers` in `_setup_hidden_states_capture` and then retrieving the value here instead of just using `self._layer_ids` (which seems to store the same value)?

I think we should just have one or the other. Currently it looks like `self._layer_ids` isn't really used, although maybe that's the better one to use so that we're not setting attributes of `base_model` from outside the class.

src/speculators/data_generation/custom_worker.py
```diff
@@ -0,0 +1,115 @@
...
+
+        base_model = model.model  # type: ignore[attr-defined]
+        base_model.aux_hidden_state_layers = tuple(layer_ids)
+
+        base_model.forward = self._create_patched_forward(base_model)
+        logger.info(f"Hidden states capture setup complete for layers {layer_ids}")
+
+    def _enable_capture(self):
+        """Enable hidden states capture"""
+        self._should_capture = True
```
fynnsu: What is the scenario where we would want to run this model with `_should_capture = False` ? If we're modifying the model, aren't we using it for capturing? Or is there another use case?

src/speculators/data_generation/logging_utils.py
fynnsu, line 1: I'm not entirely convinced we need this file. This could be a future pr, but I think we're introducing a few too many different logging setups and we should try to consolidate them.

If the goal is mostly to just have pretty logs (w/ colors) there are packages (like rich) that can do that for you by just specifying a `RichHandler` for the logs. Happy to discuss this further offline.

src/speculators/data_generation/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""Data generation utilities for EAGLE-style speculative decoding training."""
+
+from .vllm_hidden_states_generator import VllmHiddenStatesGenerator
```
fynnsu: We should use fully qualified paths

e.g.
```python
from speculators.data_generation.vllm_hidden_states_generator import VllmHiddenStatesGenerator
```

src/speculators/data_generation/preprocessing.py
```diff
@@ -0,0 +1,262 @@
...
+import re
+
+import torch
+from datasets import Dataset as HFDataset
+from datasets import load_dataset, load_from_disk
+from transformers import AutoTokenizer, PreTrainedTokenizer
+
+from .configs import CHAT_TEMPLATES, DATASET_CONFIGS, ChatTemplate, format_conversation
+from .logging_utils import PipelineLogger
+from .vocab_mapping import save_token_frequency_distribution
```
fynnsu: Fully qualified paths here too

scripts/DATAGEN.md
fynnsu, line 1: This file should probably be elsewhere. Maybe in the scripts folder or in some kind of docs or something.

src/speculators/data_generation/vllm_hidden_states_generator.py
```diff
@@ -0,0 +1,275 @@
...
+log = PipelineLogger(__name__)
+
+
+class VllmHiddenStatesGenerator:
+    """Extracts hidden states from intermediate layers during prefill only."""
+
+    layer_ids: list[int]
+    model_path: str
+    tensor_parallel_size: int
+    _request_counter: int
```
fynnsu: Same comment as for `HiddenStatesWorkerExtension`. These are class vars, so aren't actually used.

scripts/build_vocab_mapping.py
shanjiaz: Good catch! Will do.

src/speculators/data_generation/preprocessing.py
brian-dellabetta: I think we also started moving away from handling the data processing in `oneshot` in llm-compressor, instead preferring the user to set up the dataset on their own and pass that in. This can be hard to maintain, but ultimately depends on the needs of users.

brian-dellabetta (commented):
I did a quick high level review, structure and entrypoints look good to me. Biggest question is on how much we build logic around dataset preprocessing, as Kyle mentioned with the TextGenerationDataset that we've stopped encouraging people to use in our llm-compressor examples.

nice work!

src/speculators/data_generation/vllm_hidden_states_generator.py
```diff
@@ -0,0 +1,275 @@
...
+            schedule_iterations += 1
+            log.debug(
+                f"Scheduler iteration {schedule_iterations} - tokens: "
+                f"{scheduler_output.total_num_scheduled_tokens}"
+            )
+
+            self.executor.execute_model(scheduler_output)
+
+            for req_id in scheduler_output.num_scheduled_tokens:
+                self.scheduler.finish_requests([req_id], RequestStatus.FINISHED_ABORTED)
```
fynnsu, line 260: I'm not sure I understand this bit here

src/speculators/data_generation/vocab_mapping.py
```diff
@@ -0,0 +1,76 @@
...
+    if os.path.exists(freq_dist_path):
+        return freq_dist_path
+
+    token_freq: Counter[int] = Counter()
+    for item in tqdm(dataset, desc="Counting token frequencies"):
+        input_ids = item["input_ids"]
+        loss_mask = item["loss_mask"]
+        masked_ids = input_ids[loss_mask == 1]
+        unique_ids, counts = masked_ids.unique(return_counts=True)
+        batch_token_freq = dict(zip(unique_ids.tolist(), counts.tolist(), strict=False))
```
fynnsu: Should use `strict=True`. If these don't line up then it means we have a different number of unique_ids than counts for those ids which would be a problem.

src/speculators/data_generation/vocab_mapping.py
```diff
@@ -0,0 +1,76 @@
...
+    freq_dist_path = os.path.join(cache_dir, f"{cache_key}_token_freq.pt")
+
+    if os.path.exists(freq_dist_path):
+        return freq_dist_path
+
+    token_freq: Counter[int] = Counter()
+    for item in tqdm(dataset, desc="Counting token frequencies"):
+        input_ids = item["input_ids"]
+        loss_mask = item["loss_mask"]
+        masked_ids = input_ids[loss_mask == 1]
```
fynnsu: nit: It might be better (more explicit) to use `loss_mask.to(torch.bool)`.

src/speculators/data_generation/vocab_mapping.py
```diff
@@ -0,0 +1,76 @@
...
+    return freq_dist_path
+
+
+def build_vocab_mappings_from_distribution(
+    token_freq_dict: dict[int, int],
+    draft_vocab_size: int,
+    target_vocab_size: int,
+) -> tuple[torch.Tensor, torch.Tensor]:
+    """Build vocabulary mappings for draft model from token frequency distribution."""
+    sorted_tokens = sorted(token_freq_dict.items(), key=lambda x: (-x[1], x[0]))
```
fynnsu: nit: Would be nice to have a comment here on the sort order.

e.g.
`# Sort by decreasing frequency, then increasing token id`

fynnsu: Also you could make this line:

```python
sorted_tokens = sorted(token_freq_dict, key=lambda tid: (-token_freq_dict[tid], tid))
```
which is maybe slightly more readable and also means you don't need to remove the frequency values in lines 52-54 since the list already just contains the token ids.

src/speculators/data_generation/vocab_mapping.py
```diff
@@ -0,0 +1,76 @@
...
+
+    num_tokens_to_select = min(draft_vocab_size, len(sorted_tokens))
```
fynnsu, line 60: nit: Technically you don't need this line, because taking a slice that is longer than the list just returns the full list, so you can use `draft_vocab_size` directly. 

```python
>>> x = [1, 2, 3]
>>> x[:10]
[1, 2, 3]
```

src/speculators/data_generation/vocab_mapping.py
```diff
@@ -0,0 +1,76 @@
...
+        token_id for token_id, _ in sorted_tokens[:num_tokens_to_select]
+    ]
+
+    if len(selected_token_ids) < draft_vocab_size:
+        current_ids = set(selected_token_ids)
+        for tid in range(draft_vocab_size):
+            if tid not in current_ids:
+                selected_token_ids.append(tid)
+            if len(selected_token_ids) >= draft_vocab_size:
+                break
```
fynnsu, line 69: ~This seems to be for handling the situation where the target vocab is smaller than the draft_vocab_size. 
It seems like in this scenario we should probably just raise an error and exit.~

Okay I thought about this a bit further, and I guess there's a scenario where we have tokens in our vocab that don't exist in the calibration data we use for computing the frequencies.

src/speculators/data_generation/vocab_mapping.py
```diff
@@ -0,0 +1,76 @@
...
+
+    selected_token_ids.sort()
+
+    # Store offset: target_token_id = draft_idx + draft_to_target[draft_idx]
+    draft_to_target = torch.zeros(draft_vocab_size, dtype=torch.long)
+    for draft_idx, target_token_id in enumerate(selected_token_ids):
+        draft_to_target[draft_idx] = target_token_id - draft_idx
```
fynnsu: Probably can do something like:

```python
d2t = torch.tensor(selected_token_ids, dtype=torch.long) - torch.arange(draft_vocab_size, dtype=torch.long)
```

src/speculators/data_generation/vocab_mapping.py
```diff
@@ -0,0 +1,76 @@
...
+
+    target_to_draft = torch.zeros(target_vocab_size, dtype=torch.bool)
+    for target_token_id in selected_token_ids:
+        if target_token_id < target_vocab_size:
+            target_to_draft[target_token_id] = True
```
fynnsu: Similarly, I think this can be vectorized to
```python
target_to_draft[selected_token_ids] = True
```

I also don't think `target_token_id < target_vocab_size:` should ever be `False` no? If it is then the vectorized version doesn't work but shouldn't all the token_ids be in range `[0, target_vocab_size)` ?

pyproject.toml
```diff
@@ -107,6 +107,8 @@ dev = [
     "mkdocs-linkcheck~=1.0.6",
 ]
 
+datagen = ["vllm>=0.11.0"]
```
fynnsu: We may want to pin this and update it regularly, otherwise I think the integration will break too often.

src/speculators/data_generation/custom_worker.py
fynnsu: I think it's actually technically okay because the original forward isn't called by the patched forward. So call this multiple times will repeatedly replace `base_model.forward` with the same function but won't actually break anything.

fynnsu: Although one improvement here would be to use `MethodType`

e.g.
```python
import types # not typing (different stdlib package)

class A:
    def __init__(self, ...):
         ...

a = A()

# don't need to define patched forward dynamically (can get rid of _create_patched_forward)
def patched_forward(self, x): 
    ...

# Patch the forward function like so
a.forward = types.MethodType(patched_forward, a)

a.forward(x) # patched_forward receives (self=a, x=x)
```

scripts/build_vocab_mapping.py
shanjiaz: Yeah I originally saved them in `.pt` format and switched to `.npy` when testing with the training code. If the training code can consume it, I don't mind making the changes at all.

scripts/preprocess_data.py
shanjiaz: Yeah, making the change now.

src/speculators/data_generation/vllm_hidden_states_generator.py
shanjiaz: Will do!

src/speculators/data_generation/custom_worker.py
shanjiaz: Done.

src/speculators/data_generation/vocab_mapping.py
shanjiaz: Nice!

scripts/build_vocab_mapping.py
shanjiaz: sure!

src/speculators/data_generation/vllm_hidden_states_generator.py
```diff
@@ -1,4 +1,28 @@
-"""Extract hidden states from intermediate layers during prefill using vLLM."""
+"""Extract hidden states from intermediate layers during prefill using vLLM.
+
+This module provides a generator for extracting hidden states from transformer models
```
brian-dellabetta: nit -- consider moving this to the docstring of `class VllmHiddenStatesGenerator`

shanjiaz: Will do!

scripts/build_vocab_mapping.py
fynnsu: Could you resolve the first part of this (regarding output_path)?
To clarify the issue:

Currently if I want to save `t2d.npy` and `d2t.npy` in a `./vocab_mapping_outputs` folder I need to actually use `--output-path ./vocab_mapping_outputs/something_else_that_gets_ignored`. Then the script extracts the parent dir of the `args.output_path` to get `./vocab_mapping_outputs` before producing the final file paths `./vocab_mapping_outputs/d2t.npy` and `./vocab_mapping_outputs/t2d.npy`.

This is super unintuitive and is also not reflected properly in argparse descriptions, README instructions, etc. Instead the user should just pass `--output-path ./vocab_mapping_outputs` directly.

scripts/build_vocab_mapping.py
```diff
@@ -67,27 +67,25 @@ def parse_args():
...
         token_freq_dict=token_freq_dict,
         draft_vocab_size=args.draft_vocab_size,
         target_vocab_size=args.target_vocab_size,
     )
 
-    output_dir = os.path.dirname(args.output_path)
-    if output_dir:
-        os.makedirs(output_dir, exist_ok=True)
+    output_path = Path(args.output_path)
+    output_dir = output_path.parent
```
fynnsu: This line is the cause of the [issue mentioned above ](https://github.com/vllm-project/speculators/pull/157/files/bfa577015859eeb3d9463a545028e892ca438d70..6a48b9f73140ea03fbd79a1cc260456dc2dc07fd#r2519543737). We should just use `output_path` directly.

scripts/data_generation_offline.py
```diff
@@ -78,30 +80,34 @@ def parse_args():
...
         "--max-samples",
         type=int,
         default=None,
         help="Maximum number of samples to process (default: None, process all)",
     )
+    parser.add_argument(
+        "--token-freq-path",
+        type=str,
+        default="./token_freq.pt",
+        help="Path to save token frequency distribution (default: ./token_freq.pt)",
```
fynnsu, line 99: Is this actually the path for the already saved distribution? We aren't creating this file in this script but reading it, right?

shanjiaz: Actually depends, user has the option to run preprocessing separately first, which might save the token_freq.pt in a user-defined path way. Or, this script would automatically run preprocessing and save token frequencies in `./token_freq.pt`.

scripts/build_vocab_mapping.py
shanjiaz: Good catch will fix!

scripts/build_vocab_mapping.py
shanjiaz: yep!

scripts/data_generation_offline.py
fynnsu: How is this determined? Is it just if a file exists at the `--token-freq-path` location?

brian-dellabetta (approved):
to get the approval ball rolling here 😅

HDCharles (approved):
Looks good now that concerns have been addressed

rahul-tuli (approved):

scripts/data_generation_offline.py
shanjiaz: yeah

PR #168, opened 2025-10-31
Update config for EAGLE data generation

rahul-tuli:
# Add Type-Safe Config Generator for Data Generation (v2.0)

Replaces manual dict-based config (v1.0) with typed dataclass architecture (v2.0).

## Changes

**New file:** `src/speculators/data_generation/config_generator.py` (+340 lines)
- Dataclass-based configuration with full type safety
- Reproducibility tracking: package versions, GPU info, cache keys, command line
- Schema documentation with dtypes, shapes, and descriptions

**Updated:** `scripts/data_generation_offline.py` (-40, +65 lines)
- Uses `DataGenerationConfig.from_generator()` instead of manual dict construction
- Removed unused datetime import

## Type Safety

**Before:**
```python
@dataclass
class DataGenerationConfig:
    reproducibility: dict[str, Any]
    model: dict[str, Any]
    # ... lost type information
```

**After:**
```python
@dataclass
class DataGenerationConfig:
    reproducibility: ReproducibilityInfo
    model: ModelConfig
    # ... fully typed throughout
```

### v1.0
```json
{
  "version": "1.0",
  "model": {
    "target_model_path": "meta-llama/Llama-3.1-8B-Instruct",
    "tensor_parallel_size": 1,
    "max_model_len": 2048,
    "gpu_memory_utilization": 0.8
  },
  "data": {
    "train_data_path": "sharegpt",
    "seq_length": 2048,
    "num_samples": 1000,
    "seed": 0
  },
  "hidden_states": {
    "layer_ids": [2, 16, 29, 31],
    "num_layers": 4
  },
  "format": {
    "fields": ["input_ids", "hidden_states", "loss_mask"],
    "hidden_states_type": "List[torch.Tensor]"
  }
}
```

### v2.0
```json
{
  "version": "2.0",
  "generated_at": "2025-11-20T09:53:22.885881",
  "speculators_version": "0.3.0.dev14",
  "reproducibility": {
    "command": "data_generation_offline.py --target-model-path meta-llama/Llama-3.1-8B-Instruct ...",
    "cache_key": "9b43ea75a2370fef19c5c580e549846d",
    "gpu": "NVIDIA B200",
    "package_versions": {
      "torch": "2.8.0+cu128",
      "vllm": "0.11.0",
      "transformers": "4.57.1",
      "speculators": "0.3.0.dev14"
    }
  },
  "model": {
    "target_model_path": "meta-llama/Llama-3.1-8B-Instruct",
    "tensor_parallel_size": 1,
    "max_model_len": 2048,
    "gpu_memory_utilization": 0.8,
    "hidden_size": 4096
  },
  "data": {
    "train_data_path": "sharegpt",
    "seq_length": 2048,
    "max_samples": 10,
    "num_samples": 10,
    "seed": 0,
    "chat_template_note": "Uses tokenizer's built-in chat template"
  },
  "hidden_states": {
    "layer_ids": [2, 16, 29, 31],
    "description": "Layers selected for EAGLE3 fusion and target logits"
  },
  "generation": {
    "batch_size": 4,
    "cache_dir": "~/.cache/huggingface"
  },
  "format": {
    "file_pattern": "data_{idx}.pt",
    "schema": {
      "input_ids": {
        "dtype": "torch.long",
        "shape": "[seq_len]",
        "description": "Tokenized input sequence"
      },
      "hidden_states": {
        "dtype": "list[torch.bfloat16]",
        "shape": "list of [seq_len, 4096]",
        "num_tensors": 4,
        "description": "Hidden states from 4 layers"
      },
      "loss_mask": {
        "dtype": "torch.long",
        "shape": "[seq_len]",
        "description": "1 for assistant tokens to train on, 0 elsewhere"
      }
    }
  }
}
```

## What's Added

| Feature | v1.0 | v2.0 |
|---------|------|------|
| Package versions | No | Yes (torch, vLLM, transformers, speculators) |
| GPU information | No | Yes (auto-detected) |
| Cache key | No | Yes (MD5 hash) |
| Command line | No | Yes (full command) |
| Hidden size | No | Yes (from model config) |
| Schema docs | Field list | dtype, shape, description per field |
| Type safety | `dict[str, Any]` | Typed dataclasses |

## Testing

Ran e2e test with 10 samples on GPU 5:
```bash
CUDA_VISIBLE_DEVICES=5 python scripts/data_generation_offline.py \
    --target-model-path meta-llama/Llama-3.1-8B-Instruct \
    --train-data-path sharegpt \
    --seq-length 2048 \
    --batch-size 4 \
    --max-samples 10 \
    --output-dir local/test_config_v2
```

Config generated successfully with all fields populated.

### DIFF
--- a/scripts/DATAGEN.md
+++ b/scripts/DATAGEN.md
@@ -74,26 +74,146 @@ results = generator.generate(token_ids=token_ids)
 
 Tokenizes conversations and creates loss masks to identify trainable tokens.
 
+The preprocessing pipeline:
+1. Loads raw conversation data (JSON/JSONL or from datasets like ShareGPT, UltraChat)
+2. Normalizes conversation format
+3. Applies the model's chat template
+4. Creates loss masks to train only on assistant responses
+5. Tokenizes and caches the result
+6. Computes token frequency statistics and stores to file
+
+##### Basic Usage
+
+```bash
+python scripts/preprocess_data.py \
+    --target-model-path meta-llama/Llama-3.1-8B-Instruct \
+    --train-data-path sharegpt \
+    --seq-length 2048 \
+    --build-dataset-num-proc 8
+```
+
+##### Python API
+
 ```python
 from speculators.data_generation.preprocessing import load_and_preprocess_dataset
 
+# Basic usage
 dataset, tokenizer = load_and_preprocess_dataset(
     target_model_path="meta-llama/Llama-3.1-8B-Instruct",
     train_data_path="sharegpt",
     seq_length=2048,
-    max_samples=1000,
-    token_freq_path="./token_freq.pt",
-    cache_dir="/path/to/cache",  # Optional
+)
+
+# With turn dropout
+dataset, tokenizer = load_and_preprocess_dataset(
+    target_model_path="meta-llama/Llama-3.1-8B-Instruct",
+    train_data_path="sharegpt",
+    seq_length=2048,
+    turn_dropout=True,
+)
+
+# With custom pattern
+dataset, tokenizer = load_and_preprocess_dataset(
+    target_model_path="openai/gpt-oss-20b",
+    train_data_path="sharegpt",
+    seq_length=2048,
+    assistant_pattern=r'<\|start\|>assistant<\|channel\|>final<\|message\|>(.*?)<\|return\|>',
 )
 ```
 
-**Features:**
+##### Advanced Features
 
-- Uses tokenizer's built-in chat template via `apply_chat_template`
-- Automatically creates loss masks for assistant responses
-- ShareGPT format datasets
-- HuggingFace datasets (sharegpt, ultrachat)
-- Local JSON/JSONL files
+**Turn Dropout**
+
+Turn dropout is a data augmentation technique that randomly truncates conversations to different lengths during preprocessing. This helps the model learn to generate continuations from various conversation contexts.
+
+When enabled, each conversation is randomly truncated to keep only the first N consecutive turns (where N is randomly chosen between 1 and the total number of turns).
+
+```bash
+python scripts/preprocess_data.py \
+    --target-model-path meta-llama/Llama-3.1-8B-Instruct \
+    --train-data-path sharegpt \
+    --seq-length 2048 \
+    --turn-dropout
+```
+
+Benefits:
+- Increases training data diversity
+- Helps model generalize across different conversation lengths
+- Prevents overfitting to full-length conversations
+
+**Custom Assistant Pattern**
+
+By default, the assistant response pattern is auto-detected from the model's chat template. However, you can override this with a custom regex pattern if needed.
+
+```bash
+python scripts/preprocess_data.py \
+    --target-model-path openai/gpt-oss-20b \
+    --train-data-path sharegpt \
+    --seq-length 2048 \
+    --assistant-pattern '<\|start\|>assistant<\|channel\|>final<\|message\|>(.*?)<\|return\|>'
+```
+
+When to use:
+- Custom chat template formats not recognized automatically
+- Fine-grained control over what content is marked as trainable
+- Models with non-standard message boundaries
+
+Pattern requirements:
+- Must be a valid Python regex pattern
+- Should contain exactly one capture group `(.*?)` for the assistant's response content
+- Must not match across role boundaries (use negative lookahead if needed)
+
+##### Preprocessing Arguments
+
+| Argument | Type | Required | Default | Description |
+|----------|------|----------|---------|-------------|
+| `--target-model-path` | str | Yes | - | HuggingFace model ID or local path |
+| `--train-data-path` | str | Yes | - | Dataset name (sharegpt, ultrachat) or path to JSON/JSONL file |
+| `--seq-length` | int | No | 2048 | Maximum sequence length |
+| `--seed` | int | No | 0 | Random seed for shuffling |
+| `--build-dataset-num-proc` | int | No | 8 | Number of parallel processes for preprocessing |
+| `--max-samples` | int | No | None | Maximum number of samples to process |
+| `--token-freq-path` | str | No | ./token_freq.pt | Path to save token frequency distribution |
+| `--hf-cache-dir` | str | No | None | Directory for HuggingFace datasets cache |
+| `--assistant-pattern` | str | No | None | Custom regex pattern for assistant responses |
+| `--turn-dropout` | flag | No | False | Enable turn dropout for data augmentation |
+
+##### Supported Datasets
+
+- **sharegpt**: ShareGPT conversation dataset
+- **ultrachat**: UltraChat conversation dataset
+- **Custom JSON/JSONL**: Local files with conversation format
+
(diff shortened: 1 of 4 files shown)


scripts/data_generation_offline.py
```diff
@@ -181,47 +181,32 @@ def save_sample_to_disk(data_dict, output_path):
...
-    }
+    """Save metadata config file for reproducibility.
 
+    Uses the v2.0 config generator with enhanced metadata including
+    package versions, GPU info, example prompts, and detailed schema.
+    """
+    log.subsection("Saving configuration metadata")
+
+    # Determine cache_dir - use hf_cache_dir if provided, otherwise use default
+    cache_dir = args.hf_cache_dir if args.hf_cache_dir else "~/.cache/huggingface"
```
fynnsu: Discussed with @shanjiaz offline, and ideally we drop `args.hf_cache_dir` and use hf env variables like `HF_HUB_CACHE` / `HF_HOME` instead (not directly but by finding a util in the `huggingface_hub` package that resolves the cache dir).

Note: this will be done by @shanjiaz in a future pr, so no changes needed here.

src/speculators/data_generation/config_generator.py
```diff
@@ -0,0 +1,328 @@
...
+    layer_ids: list[int],
+    tensor_parallel_size: int,
+    max_model_len: int,
+    gpu_memory_utilization: float,
+    batch_size: int,
+    cache_dir: str,
+    num_samples: int,
+    max_samples: int | None = None,
+    seed: int = 0,
+) -> DataGenerationConfig:
```
fynnsu: Maybe this should be a `classmethod` of `DataGenerationConfig`? Or even the `__init__` implementation? 

Otherwise seems strange that `DataGenerationConfig.from_generator` calls this function internally.

src/speculators/data_generation/config_generator.py
```diff
@@ -0,0 +1,328 @@
...
+        """
+        return asdict(self)
+
+    @classmethod
+    def from_generator(
+        cls,
+        generator: "VllmHiddenStatesGenerator",
+        train_data_path: str,
+        seq_length: int,
+        batch_size: int,
```
fynnsu: Do we need `batch_size`? This is just the `batch_size` used during generation but doesn't actually describe how the data is saved.

src/speculators/data_generation/config_generator.py
```diff
@@ -0,0 +1,320 @@
...
+        from importlib.metadata import version  # noqa: PLC0415
+
+        import transformers  # noqa: PLC0415
+
+        try:
+            import vllm  # noqa: PLC0415
+
+            vllm_version = vllm.__version__
+        except (ImportError, AttributeError):
+            vllm_version = "unknown"
```
rahul-tuli: I don't actually need this since vllm is a requirement for datagen

src/speculators/data_generation/config_generator.py
```diff
@@ -0,0 +1,328 @@
...
+from speculators.data_generation.logging_utils import PipelineLogger
+
+if TYPE_CHECKING:
+    from speculators.data_generation.vllm_hidden_states_generator import (
+        VllmHiddenStatesGenerator,
+    )
+
+log = PipelineLogger(__name__)
+
+CONFIG_VERSION = "2.0"
```
fynnsu: Why is this `"2.0"`?

For the current data format (the data files themselves not the config), we are referring to them as version `1`. I'm not sure if that's the same version as the `CONFIG_VERSION`? Otherwise we should add the data format version to `FormatConfig`.

Also currently for data format version we're just using an int (0 or 1) but I'm okay with updating that to a `"x.y"` string, but then we need to make this change in a few other places in the repo.

fynnsu: Ahh okay, I see the `"1.0"` vs `"2.0"` example above. I guess I would still like clarity on if the config versioning should be the same as the data format version. Or if not, we should include the data format version in the config.

src/speculators/data_generation/config_generator.py
shanjiaz: True, but this is just for reproduction here I think

scripts/data_generation_offline.py
rahul-tuli: ACK!

src/speculators/data_generation/config_generator.py
rahul-tuli: Integrated `generate_config()` logic directly into `DataGenerationConfig.from_generator()`, Good suggestion!

src/speculators/data_generation/config_generator.py
rahul-tuli: I think Fynn's comment makes sense, so removed it for now, can add back later if requested

src/speculators/data_generation/config_generator.py
rahul-tuli: Moved `CONFIG_VERSION` from module-level constant to `DataGenerationConfig.VERSION` class variable

src/speculators/data_generation/config_generator.py
```diff
@@ -0,0 +1,281 @@
...
+
+    cache_dir: str
+
+
+@dataclass
+class FormatConfig:
+    """Output format specification for generated data files."""
+
+    file_pattern: str
+    schema: dict[str, dict[str, Any]]
```
fynnsu, line 124: What I meant by this https://github.com/vllm-project/speculators/pull/168#discussion_r2546906977 was that we currently use `--data-format-version` ([link](https://github.com/vllm-project/speculators/blob/b20d626b719d5dbfd91fe254c91ec2ae2589898d/scripts/train.py#L213)) in `train.py` and have the data format versions shown here https://github.com/vllm-project/speculators/blob/b20d626b719d5dbfd91fe254c91ec2ae2589898d/src/speculators/train/data.py#L94-L129

which we support. Version `1` is the schema described here, whereas version `0` is what the old data generation code produces. I think it would probably be useful to include this version number here in the `FormatConfig`. i.e. `data_format_version: int`. This would be an easy way for code loading generated data to check the format without parsing the schema. 

Alternatively, if we want the `DataGenerationConfig.version` to also correspond with a specific data format schema, that's also a reasonable approach. But then we should open an issue to update the `--data-format-version` parameter and then function names in `data.py` to reflect the new version `"2.0"` instead of the existing data format version `1`. 

However, it is probably more straightforward to do the first option and add a data format version field to this class.

rahul-tuli: @fynnsu Great suggestion! Added `data_format_version: int` to FormatConfig in commit 9db70a0.

  I went with Option 1 as you recommended - keeping the config schema version (2.0) separate
  from the data format version (1). This makes it easy for training code to check which
  standardization function to use without parsing the full schema.

  The generated config now includes:
  - `DataGenerationConfig.version = "2.0"` (config file schema)
  - `FormatConfig.data_format_version = 1` (data format - list of hidden_states)

  This aligns with the existing `--data-format-version` parameter in train.py. Let me know 
  if you'd like any adjustments!

PR #183, opened 2025-11-14
Add truncate_long_samples option to distributed batch sampler

fynnsu:
When truncate_long_samples is True, the distributed batch sampler will truncate long samples to the batch_max_length.

When truncate_long_samples is False, the distributed batch sampler will drop long samples.

The default is True.


brian-dellabetta (commented):
👍

PR #184, opened 2025-11-14
Fix small model weight loading

fynnsu:
This pr fixes two issues:
- Small models that only have a single weight shard don't have a `model.safetensors.index.json` file. Now we will fallback to checking for `model.safetensors` if the index file doesn't exist
- Some models share the `lm_head` and `embed_tokens` weights and don't have separate entries for both in the checkpoints. We now fallback to using the `embed_tokens` value for the `lm_head` if we can't find an `lm_head` entry.

### DIFF
--- a/src/speculators/train/distributed_batch_sampler.py
+++ b/src/speculators/train/distributed_batch_sampler.py
@@ -154,6 +154,7 @@ class MultipackDistributedBatchSamplerV2(Sampler):
         lengths: ArrayLike,
         num_replicas: int,
         rank: int,
+        truncate_long_samples: bool = True,
         seed: int = 0,
     ):
         """Efficient distributed packing sampler for linear attention style models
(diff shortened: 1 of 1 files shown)


shanjiaz (approved):
Can you add a small test case for this?

src/speculators/train/eagle3/core.py
```diff
@@ -235,9 +235,12 @@ def _setup_embeddings_and_lm_heads(self, config: VerifierConfig, t2d: torch.Tens
                 f" verifier_vocab_size ({verifier_model_config.vocab_size})."
             )
 
+        lm_head_key = "lm_head.weight"
         verifier_weights = load_model_layers(
-            ["model.embed_tokens.weight", "lm_head.weight"], config.name_or_path
+            ["model.embed_tokens.weight", lm_head_key], config.name_or_path
         )
+        if lm_head_key not in verifier_weights:
```
brian-dellabetta, line 242: just fyi, in case these might help, there's usually a `tie_word_embeddings` key in the config, example [here](https://huggingface.co/Qwen/Qwen3-VL-8B-Instruct/blob/main/config.json#L37). And a `get_input_embeddings` helper on the model itself that we use [here](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/transformers/compression/compressed_tensors_utils.py#L131)

fynnsu: Yeah I consider this a little bit but wasn't sure if `tie_word_embeddings` always exists. Unfortunately, `get_input_embeddings` isn't an option because we don't actually load the verifier model. If you think `tie_word_embeddings` makes sense, I can switch to that but otherwise this seems like a fairly safe fallback.

brian-dellabetta: It will error out either way, so if it works for now i think it's fine

PR #185, opened 2025-11-17
Added data generation testing when merging to main

shanjiaz:
Added separate data generation(vllm dependent) tests when merging to main.


.github/workflows/development.yml
```diff
@@ -110,6 +110,22 @@ jobs:
         run: pip install tox tox-uv
       - name: Run integration tests
         run: tox -e test-integration
+  
+  integration-tests:
```
dsikka: Should be called `datagen-tests`?

shanjiaz: oh yes!

PR #191, opened 2025-11-20
Add script for E2E Datagen + training

fynnsu:
1. Add `scripts/gen_and_train.py` which contains a `run_e2e()` function that calls the other data gen + training scripts in order and coordinates the passing the outputs between them.
2. Combine `scripts/DATAGEN.md` and `scripts/TRAINING.md` into a single `scripts/README.md` file and add E2E instructions. 
3. Make fixes to data gen to get the e2e Qwen3 example script running:
a. Change default tensor parallel size to `torch.accelerator.device_count()`
b. Add assistant pattern and turn dropout to argparse for `scripts/data_generation_offline.py`
c. Update ultrachat dataset split from `"train"` -> `"train_sft"`
d. Update turn dropout logic to only drop after after an assistant response
e. Create folder if it doesn't exist when saving token frequency dist

# E2E implementation

Separated this into two files: `scripts/gen_and_train.py` which contains all the logic and `examples/e2e/qwen3_8b_sharegpt_ultrachat.py` which demonstrates creating an example config and calling the e2e script with it. 

## `examples/e2e/qwen3_8b_sharegpt_ultrachat.py`
This is maybe not the best implementation long term. We may want to look into some kind of config file format (e.g. yaml/toml) which would replace this file, and be passed into `scripts/gen_and_train.py`. As it is, we currently need to do some `sys.path` hackery to access the `run_e2e` fn + the argument NamedTuples. That being said, this solution works well for the time being and is doesn't lock us into maintaining some kind of recipe/config format prematurely.

## `scripts/gen_and_train.py`
- Calls the `data_generation_offline.py`, `build_vocab_mapping.py`, and `train.py` as subprocesses to delegate all work to these existing scripts. Acts as glue to connect these different steps together. 
- Enforces a strict output folder structure (see `Output structure` comment at top of file), which simplifies user arguments.
- Uses a sentinel class for default unset values and only passes values that are explicitly set by the user on to the subprocesses. This allows us to avoid redefining default values for optional arguments. 
- Has logic to forcefully shut down subprocesses if a `KeyboardInterrupt` is received. This is important as otherwise it is easy to create zombie processes that need to be manually killed.
- Supports running datagen on multiple datasets and combines the token frequency data from each dataset before creating the vocab mapping file.

### DIFF
--- a/examples/e2e/llama3_8b_sharegpt_5k.py
+++ b/examples/e2e/llama3_8b_sharegpt_5k.py
@@ -35,20 +35,20 @@ from gen_and_train import (  # noqa: E402
 if __name__ == "__main__":
     VERIFIER_NAME_OR_PATH = "meta-llama/Llama-3.1-8B-Instruct"
     OUTPUT_PATH = "./output/llama3_8b_sharegpt_5k"
-    TOTAL_SEQ_LEN = 8_192
+    TOTAL_SEQ_LEN = 8192
 
     # Data Generation
     data_gen_args_sharegpt = DataGenArgs(
         train_data_path="sharegpt",
         max_model_len=TOTAL_SEQ_LEN,
         seq_length=TOTAL_SEQ_LEN,
-        max_samples=5_000,  # Only use 5000 samples from ShareGPT
+        max_samples=5000,  # Only use 5000 samples from ShareGPT
     )
 
     # Vocab Mapping
     vocab_mapping_args = VocabMappingArgs(
-        draft_vocab_size=8_192,  # Use a very small draft vocabulary for this example
-        target_vocab_size=128_256,  # From https://huggingface.co/meta-llama/Llama-3.1-8B-Instruct/blob/main/config.json#L37
+        draft_vocab_size=8192,  # Use a very small draft vocabulary for this example
+        target_vocab_size=128256,  # From https://huggingface.co/meta-llama/Llama-3.1-8B-Instruct/blob/main/config.json#L37
     )
 
     # Training
--- a/examples/e2e/qwen3_8b_sharegpt_ultrachat.py
+++ b/examples/e2e/qwen3_8b_sharegpt_ultrachat.py
@@ -38,7 +38,7 @@ from gen_and_train import (  # noqa: E402
 if __name__ == "__main__":
     VERIFIER_NAME_OR_PATH = "Qwen/Qwen3-8B"
     OUTPUT_PATH = "./output/qwen3_8b_sharegpt_ultrachat"
-    TOTAL_SEQ_LEN = 8_192
+    TOTAL_SEQ_LEN = 8192
 
     # Data Generation
     data_gen_args_sharegpt = DataGenArgs(
@@ -57,8 +57,8 @@ if __name__ == "__main__":
 
     # Vocab Mapping
     vocab_mapping_args = VocabMappingArgs(
-        draft_vocab_size=32_000,  # Use an 32k draft vocabulary
-        target_vocab_size=151_936,  # From https://huggingface.co/Qwen/Qwen3-8B/blob/main/config.json#L29
+        draft_vocab_size=32000,  # Use an 32k draft vocabulary
+        target_vocab_size=151936,  # From https://huggingface.co/Qwen/Qwen3-8B/blob/main/config.json#L29
     )
 
     # Training


pyproject.toml
```diff
@@ -252,6 +253,14 @@ select = [
     "PTH", # os.path is acceptable in scripts
 ]
 
+"examples/**/*.py" = [
+    "INP001", # allow implicit namespace packages in examples
+]
+
+"scripts/gen_and_train.py" = [
```
dsikka, line 260: Just use log?

scripts/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+```bash
+python scripts/data_generation_offline.py \
+    --target-model-path meta-llama/Llama-3.1-8B-Instruct \
+    --train-data-path sharegpt \
+    --output-dir ./training_data \
+    --max-samples 5000
+```
+
+The script automatically uses the tokenizer's built-in chat template via `apply_chat_template`.
```
dsikka: It may be useful to provide a quick overview of what this is file + args are doing using vLLM.

scripts/README.md
```diff
@@ -0,0 +1,202 @@
...
+    --cache-dir ./cache \
+    --output-dir ./training_data \
+    --layer-ids 2 28 54 \
+    --tensor-parallel-size 4 \
+    --batch-size 16 \
+    --max-samples 10000
+```
+
+
+#### Dataset Configs
```
dsikka, line 135: We're saying we support the automatic processing of these two datasets only?

scripts/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+#### Caching
+
+Preprocessing is automatically cached by HuggingFace datasets using fingerprint-based cache invalidation. The cache automatically updates when:
+
+- Tokenizer changes
+- Preprocessing parameters change (seq_length, etc.)
+- Dataset changes
+
+**Cache Location:**
```
dsikka, line 151: Are we saying here that if we set our cache, data gen will use that location automatically, without having to point to any dataset?

I guess the ... here `python scripts/data_generation_offline.py ...` has left me a bit confused.

examples/e2e/qwen3_8b_sharegpt_ultrachat.py
```diff
@@ -0,0 +1,54 @@
+import sys
+from pathlib import Path
+
+# Add scripts directory to path so we can import the run_e2e function.
```
dsikka, line 4: This is great and exactly what we'd want for well-lit paths.

We should add a script for llama3

dsikka (approved):
This looks great. Some small comments for clarity. I think we should be more verbose not less to start in our README.md docs. 


We should add in the llama3 e2e in the example files as well.

scripts/README.md
shanjiaz: These are the two datasets that users can just pass in by name. For other datasets users have to either pass in a huggingface dataset or a path.

scripts/README.md
shanjiaz: This is just to specify the huggingface cache to save the preprocessed dataset, userd would still have to pass in a dataset. `python scripts/data_generation_offline.py ...` would be a normal command we use for data generation, perhaps we should remove it.

shanjiaz (approved):
Looks great! Thanks!

pyproject.toml
fynnsu: Because we're using the `gen_and_train.py` to call a bunch of scripts as subprocesses, it can be a little bit difficult to distinguish between where scripts start and end. The way I'm using print here is to print big blocks like this
```
############################ RUNNING data_generation_offline.py ############################

uv run --no-sync --no-dev --no-default-groups --isolated --with-editable .[datagen] /home/fynnsu/repos/vllm-project/speculators/scripts/data_generation_offline.py --train-data-path sharegpt --max-model-len 8192 --seq-length 8192 --target-model-path Qwen/Qwen3-8B --token-freq-path /mnt/data/machine/fynnsu/qwen3_8b_sharegpt_ultrachat/vocab_mapping/token_freq_sharegpt.pt --output-dir /mnt/data/machine/fynnsu/qwen3_8b_sharegpt_ultrachat/gen/sharegpt --turn-dropout

############################################################################################
```
and this
```
############################ COMPLETED train.py ############################

Time taken: 30041.80 seconds. Exit code: 0

############################################################################
```

That's kind of challenging to do with log because it adds a bunch of (usually helpful) metadata about where the print comes for.

scripts/README.md
```diff
@@ -0,0 +1,202 @@
...
+    <!-- duplicate subsection name, requires -1 suffix to avoid conflict -->
+    - **[Quick Start](#quick-start-1)**<br>
+    - **[Arguments](#arguments)**<br>
+    - **[Example Command](#example-command)**<br>
+- **[E2E Pipeline](#e2e-pipeline)**<br>
+    - **[Overview](#overview-1)**<br>
+    - **[Prerequisites](#prerequisites)**<br>
+    - **[Usage](#usage)**<br>
+
+## Data Generation
```
dsikka, line 27: Can we add a note that we generate a config with the information about the data generation as well as a sample?

fynnsu: Added

scripts/README.md
fynnsu: Adding

scripts/README.md
fynnsu: Will clarify this a bit in the README

examples/e2e/qwen3_8b_sharegpt_ultrachat.py
fynnsu: Added an example script for llama3. I decided to make this one just use 5k samples from Sharegpt. It's not enough for a full training run but enough to see the drafter learning something. I also added some comments to this script + the qwen one describing their uses.

scripts/README.md
```diff
@@ -0,0 +1,277 @@
...
+
+
+
+#### Dataset Configs
+
+Built-in datasets (can be used directly by name in the `--train-data-path` argument):
+- `sharegpt` - ShareGPT Vicuna unfiltered
+- `ultrachat` - HuggingFace UltraChat 200k
+
+Alternatively, you can use a different dataset by passing the HuggingFace dataset path or local JSON/JSONL file path in the `--train-data-path` argument.
```
shanjiaz, line 141: Thanks for adding this!

scripts/README.md
```diff
@@ -0,0 +1,277 @@
...
+- Dataset changes
+
+**Cache Location:**
+
+- Default: `~/.cache/huggingface/datasets`
+- Custom: Set `HF_DATASETS_CACHE` environment variable
+
+```bash
+# Example: Use custom cache directory
+export HF_DATASETS_CACHE=/path/to/your/cache
```
shanjiaz: Should we update this to `export HF_HUB_CACHE = /path/to/your/cache` and maybe remove the rest based on our discussion? Maybe also add a sentence clarifying that this is optional, or move this section to troubleshooting?

shanjiaz (changes requested):
Left one comment about the huggingface environment variables, otherwise looks good to me!

scripts/README.md
fynnsu: Updated!

shanjiaz (dismissed):
Thanks for updating!

examples/e2e/llama3_8b_sharegpt_5k.py
```diff
@@ -0,0 +1,69 @@
...
+# first token accuracy: 0.40
+# second token accuracy: 0.13
+# third token accuracy: 0.04
+# average acceptance length: 1.57
+
+
+if __name__ == "__main__":
+    VERIFIER_NAME_OR_PATH = "meta-llama/Llama-3.1-8B-Instruct"
+    OUTPUT_PATH = "./output/llama3_8b_sharegpt_5k"
+    TOTAL_SEQ_LEN = 8_192
```
dsikka: typo?

dsikka (dismissed):
LGTM. I think one typo?

examples/e2e/llama3_8b_sharegpt_5k.py
fynnsu: It actually was intentional, python allows you to add arbitrary underscores in integers which can be helpful for breaking up long numbers. But after discussing with @shanjiaz as well, it's clear that these are a little confusing and not super necessary because the longest numbers we have are 6 digits.

PR #197, opened 2025-12-03
Update ReadME with datagen install

dsikka:


shanjiaz (dismissed):

README.md
```diff
@@ -184,6 +184,12 @@ For development with additional tools:
 pip install -e ".[dev]"
 ```
 
+To enable the generation of data (i.e hidden states) from vLLM for speculator training:
+
+```bash
+pip install -e ".[dev,datagen]"
```
fynnsu:
```suggestion
pip install -e ".[datagen]"
```

We don't actually need `dev` for datagen. `dev` should only be required for people developing speculators, and not for end users.

shanjiaz (approved):

PR #199, opened 2025-12-03
Update `src/speculators/models/eagle3` to new training implementation

fynnsu:
Background:
We previously had an Eagle3 model implementation in `src/speculators/models/eagle3.py`. This model definition to date has only been used to support the convert pathway and doesn't have a functional forward implementation. When developing the new training implementation #143, a new `Eagle3DraftModel` class was produced to eventually replace the `Eagle3Speculator` model. Although the new class is not an exact drop in replacement, it represents the new structure we are going for with speculator models (i.e. focusing on converting + training, dropping support for inference/loading verifier models).

This pr:
- moves the training implementing `src/speculators/train/eagle3` -> `src/speculators/models/eagle3`
- splits the old implementation `src/speculators/models/eagle3.py` into 2 files:
  - `config.py` which goes into the new eagle3 folder and contains the `Eagle3SpeculatorConfig`
  - `eagle3_legacy_model.py` which goes under `src/speculators/convert/eagle` and contains the old `Eagle3Speculator` class
- Updates all import links

Note: eventually we will want to drop the old `Eagle3Speculator` entirely, but for now we hold onto it to use during conversion. This is because the `Eagle3DraftModel` has a slightly different init interface + doesn't have fully tested `from_pretrained` support.


shanjiaz (approved):
Nice!

PR #206, opened 2025-12-04
[Testing] Add e2e train <> vLLM tests + clean-up test utils

dsikka:
# Summary
- Add e2e tests which run the following steps:
1. Fetch pre-computed hidden states and token freq data from hf hub
2. Compute the vocab mapping
3. Running single gpu training for one epoch
4. Launch the trained speculator in vLLM and ensure that it runs + generates tokens

- Clean-up and unify shared utils between the different vLLM tests 

# Testing:
Sample Output:
[test_output.txt](https://github.com/user-attachments/files/23941795/test_output.txt)


tests/e2e/vllm/test_train.py
```diff
@@ -0,0 +1,88 @@
...
+            draft_vocab_size=8192,
+            target_vocab_size=128256,
+        )
+
+        np.save(d2t_path, d2t.cpu().numpy())
+        np.save(t2d_path, t2d.cpu().numpy())
+
+    def _run_training(self, script_path: str, args_dict: dict):
+        cmd = [
+            "python",
```
fynnsu, line 35: Maybe use `sys.executable` here (gives the absolute path to currently running python interpreter)?

tests/e2e/vllm/test_train.py
```diff
@@ -0,0 +1,88 @@
...
+            "data-path": local_dir,
+            "save-path": SAVE_PATH,
+            "data-format-version": 1,
+            "log-dir": str(tmp_path / "logs"),
+            "d2t-path": str(tmp_path / "d2t.npy"),
+            "t2d-path": str(tmp_path / "t2d.npy"),
+        }
+        # 3. Train draft model for one epoch
+        p = self._run_training("scripts/train.py", training_args)
+        p.wait()
```
fynnsu, line 87: To be explicit, you could check `p.returncode` and raise error if it's nonzero.

fynnsu (approved):

tests/e2e/vllm/test_train.py
```diff
@@ -19,7 +19,7 @@ class TestTrainvLLM:
     and runs the trained model in vLLM.
     """
 
-    def _generate_t2d_d2t(self, token_freq: str, d2t_path: str, t2d_path: str):
+    def _generate_t2d_d2t(self, token_freq: str, d2t_path: Path, t2d_path: Path):
```
fynnsu: If we're making `d2t_path` and `t2d_path` `Path` objects, we should probably update `token_freq` too.

PR #217, opened 2025-12-09
pin datasets version

shanjiaz:
python 3.10 was failing because pyarrow 22.0.0 and datasets 2.14.4 were not compatible. Pinning datasets version to >= 4.4.1 to avoid this failure.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ keywords = [
 dependencies = [
     "accelerate",
     "click",
-    "datasets>=4.4.1",
+    "datasets>=4.0.0",
     "deepspeed",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -38,7 +38,7 @@ keywords = [
 dependencies = [
     "accelerate",
     "click",
-    "datasets",
+    "datasets>=4.4.1",
```
fynnsu: Does it need to be this high? What about `4.0.0`? Does that work?

It's just generally good practice to have these requirements be as permissive as possible.

shanjiaz: Sure!

PR #223, opened 2025-12-15
support preprocessing of regenerated gpt-oss data

shanjiaz:
We're currently using regenerated gpt-oss data, and preprocess needed some fixes to support that. Also added some minor changes/bug fixes.

### DIFF
--- a/src/speculators/data_generation/vllm_hidden_states_generator.py
+++ b/src/speculators/data_generation/vllm_hidden_states_generator.py
@@ -12,9 +12,12 @@ from vllm.config import (
     VllmConfig,
 )
 from vllm.sampling_params import SamplingParams
+from vllm.utils.hashing import get_hash_fn_by_name
 from vllm.v1.core.kv_cache_utils import (
     _get_kv_cache_groups_uniform_spec,
     get_kv_cache_config_from_groups,
+    get_request_block_hasher,
+    init_none_hash,
     unify_hybrid_kv_cache_specs,
 )
 from vllm.v1.core.sched.scheduler import Scheduler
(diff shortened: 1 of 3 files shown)


scripts/train.py
```diff
@@ -175,7 +175,7 @@ def main(args: argparse.Namespace):
         is_distributed=is_distributed,
         local_rank=local_rank,
         train_call_kwargs={
-            "use_off_policy_tokens": False,
+            "use_off_policy_tokens": True,
```
fynnsu: We should make this an actual option

src/speculators/data_generation/vllm_hidden_states_generator.py
```diff
@@ -2,7 +2,14 @@
 
 import torch
 from transformers import AutoConfig, AutoTokenizer
-from vllm.config import (
+
+# Fix CUDA multiprocessing: Set vLLM to use 'spawn' instead of 'fork'
+# This must be done before importing vLLM modules
+from vllm import envs
+
+envs.VLLM_WORKER_MULTIPROC_METHOD = "spawn"
```
fynnsu: I feel like we shouldn't have this both in the library and the scripts. Maybe we can remove here and only use in scripts? Probably better if the library avoids changing env settings.

fynnsu: Then we can also remove the `# noqa: E402` lines below

scripts/train.py
fynnsu: I can also do this separately, but we should avoid changing back and forth in prs for specific models.

src/speculators/data_generation/vllm_hidden_states_generator.py
shanjiaz: Sure!

scripts/train.py
shanjiaz: Oh Megan says the default for this should be true. I can make this into an argument!

fynnsu (approved):

PR #226, opened 2025-12-18
End to end acceptance rate regression test

fynnsu:
- Add acceptance rate metric collection to `tests/e2e/vllm/run_vllm.py`
- Add acceptance rate optional asserts to `tests/e2e/vllm/utils.py`
- Add `tests/e2e/vllm/test_gen_train_acceptance.py` based on `examples/data_generation_and_training/llama3_8b_sharegpt_5k.py`, which trains a llama 3.1 8B model on 5k samples from sharegpt and then checks the acceptance rate on several test prompts. This test uses the functionality added to the above files.

### DIFF
--- a/examples/evaluate/eval-guidellm/README.md
+++ b/examples/evaluate/eval-guidellm/README.md
@@ -30,7 +30,8 @@ bash setup.sh  # or: bash setup.sh --use-uv for faster installation
 **Or run with custom parameters:**
 ```bash
 ./run_evaluation.sh \
-  -m "RedHatAI/Llama-3.1-8B-Instruct-speculator.eagle3" \
+  -b "meta-llama/Llama-3.1-8B-Instruct" \
+  -s "RedHatAI/Llama-3.1-8B-Instruct-speculator.eagle3" \
   -d "emulated"
 ```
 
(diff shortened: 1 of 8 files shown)


shanjiaz (approved):
Looks awesome! Do we want to test regression for different models? Maybe the models in examples?

tests/e2e/vllm/run_vllm.py
```diff
@@ -31,20 +32,66 @@ def parse_args():
     return parser.parse_args()
 
 
+def extract_metrics(
```
dsikka, line 35: Is this different than how Rahul set-up extracting metrics from the logs through is benchmarking work? Is there anyway we could share that functionality?

dsikka (approved):
Just a question otherwise lgtm

tests/e2e/vllm/run_vllm.py
fynnsu: Yeah it's a different approach. I got this approach from the vllm [examples/offline_inference/spec_decode.py](https://github.com/vllm-project/vllm/blob/main/examples/offline_inference/spec_decode.py) script.

The challenge is, this system uses vllm metrics and works because we're running vllm through the python api. Rahul's testing instead uses the cli to spin up a vllm instances which guidellm then interacts with. The advantage of the guidellm approach is that it allows us to simulate slightly more "real world" workloads and measure actual server response times. I can look into if there's a way to use the metrics system to get acceptance rates instead of the current log scrapping method, but either way I don't think we can easily combine the implementations.

PR #228, opened 2025-12-18
Added data generation & training support for qwen3 vl model

shanjiaz:
vl model support:
- Extract `text_config` from multimodal model configs in vocab mapping, training, and model initialization
- Added `kwargs` to support vision language parameters like `deepstack_input_embeds` in custom worker for Qwen3-VL's multimodal processing
- Use vLLM's native `get_language_model()` method to access the text model in vision-language architectures
- Detect base model structure for vision-language models

Util changes:
- Add suffix pattern matching for weight keys (e.g., "embed_tokens.weight" matches "model.embed_tokens.weight")
- Support single-file safetensors models (no index.json) by building virtual weight maps

- [ ] Run e2e test for a qwen 3 vl model

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -113,7 +113,7 @@ dev = [
     "mkdocs-linkcheck~=1.0.6",
 ]
 
-datagen = ["vllm==0.12.0"]
+datagen = ["vllm>=0.12.0,<=0.13.0"]
 
 [project.entry-points.console_scripts]
 speculators = "speculators.__main__:app"


src/speculators/data_generation/custom_worker.py
```diff
@@ -14,13 +14,21 @@
 
 
 def _patched_forward(
-    self, input_ids, positions, intermediate_tensors=None, inputs_embeds=None
+    self,
+    input_ids,
+    positions,
+    intermediate_tensors=None,
+    inputs_embeds=None,
+    deepstack_input_embeds=None,  # noqa: ARG001
```
fynnsu: If this is just to allow the `forward` method being called with `deepstack_input_embeds` then maybe we should use `**_kwargs` instead?

shanjiaz: Sure!

PR #229, opened 2025-12-19
Only add comments to non-fork prs

fynnsu:
Due to permissions issues, only non-fork prs are able to add comments to prs.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -81,6 +81,7 @@ dev = [
     "pre-commit~=3.5.0",
     "sphinx~=7.1.2",
     "tox~=4.16.0",
+    "tox-uv", # version is tied to tox version
 
     # testing
     "pytest~=8.2.2",


dsikka (commented):
I dont think I follow - what is this PR enabling?

fynnsu:
> I dont think I follow - what is this PR enabling?

Currently, every pr from a fork will have a failing build job because the last step of the build job is to add a comment to the pr (like [this one](https://github.com/vllm-project/speculators/pull/229#issuecomment-3675818797)). Unfortunately, prs from forks don't have the permissions required to actually add the comment, causing that step to fail. There is also no way to give them that permission.

This pr modifies the build and link-check jobs so that if they're running on a community pr, they won't attempt to add a comment. Their outputs will still be visible by going to the job outputs but there won't be a comment for them.

PR #232, opened 2025-12-30
support ascend npu platform.

momo609:
support ascend npu platform.

### DIFF
--- a/scripts/gen_and_train.py
+++ b/scripts/gen_and_train.py
@@ -30,10 +30,12 @@ from pathlib import Path
 from typing import Any, NamedTuple
 
 import psutil
+import torch
 
 from speculators.train.vocab_mapping import (
     combine_token_frequency_distributions,
 )
+from speculators.utils.util import is_npu_available
 
 
 class _NS(enum.Enum):
@@ -167,6 +169,16 @@ def print_block(title: str, content: str):
     print("\n", "#" * term_width, "\n", sep="")
 
 
+def npu_command(python_alt, command):
+    if is_npu_available():
+        if python_alt and python_alt.strip():
+            command = []
+            command.extend(python_alt.split())
+        else:
+            command = ["python"]
+    return command
+
+
 def run_script(
     script_name: str,
     script_args: list[str],
@@ -190,6 +202,8 @@ def run_script(
 
     if python_alt:
         command.extend(python_alt.split())
+
+    command = npu_command(python_alt, command)
 
     script_path = (Path(__file__).parent / script_name).absolute()
     command.append(str(script_path))
@@ -329,10 +343,10 @@ def run_e2e(
             loggers = loggers.split(",")
         loggers = [logger.strip() for logger in loggers]
         packages.extend(loggers)
-
+    device_count = torch.accelerator.device_count()
     run_script(
         "train.py",
         ta_list,
         packages,
-        python_alt="torchrun --standalone --nproc_per_node=gpu",
+        python_alt=f"torchrun --standalone --nproc_per_node={device_count}",
     )
--- a/src/speculators/data_generation/config_generator.py
+++ b/src/speculators/data_generation/config_generator.py
@@ -16,6 +16,7 @@ import torch
 from transformers import AutoConfig
 
 from speculators.data_generation.logging_utils import PipelineLogger
+from speculators.utils.util import get_device_name
 
 if TYPE_CHECKING:
     from speculators.data_generation.vllm_hidden_states_generator import (
@@ -30,14 +31,15 @@ log = PipelineLogger(__name__)
 def _get_gpu_info() -> str:
     """Get GPU information string.
 
-    :return: GPU model and count, or "CPU only" if no GPU available
+    :return: GPU model and count or NPU model and count,
+             or "CPU only" if no GPU/NPU available
     """
-    if not torch.cuda.is_available():
-        return "CPU only"
-
-    gpu_count = torch.cuda.device_count()
-    gpu_name = torch.cuda.get_device_name(0)
-    return gpu_name if gpu_count == 1 else f"{gpu_count}x {gpu_name}"
+    device_count = torch.accelerator.device_count()
+    device_name = get_device_name(0)
+    if device_name == "NO ACCELERATOR":
+        return "CPU ONLY"
+    else:
+        return device_name if device_count == 1 else f"{device_count}x {device_name}"
 
 
 @dataclass
--- a/src/speculators/data_generation/vllm_hidden_states_generator.py
+++ b/src/speculators/data_generation/vllm_hidden_states_generator.py
@@ -25,13 +25,15 @@ from vllm.v1.executor.multiproc_executor import MultiprocExecutor
 from vllm.v1.request import Request, RequestStatus
 from vllm.v1.structured_output import StructuredOutputManager
 
+from speculators.utils.util import empty_cache, is_npu_available, mem_get_info
+
 from .logging_utils import PipelineLogger
 
 __all__ = ["VllmHiddenStatesGenerator"]
 
 # Constants
 CACHE_MEMORY_FRACTION = 0.2  # Fraction of GPU memory for KV cache
-VLLM_BLOCK_SIZE = 16  # Block size for KV cache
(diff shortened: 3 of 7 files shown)


src/speculators/train/checkpointer.py
```diff
@@ -13,6 +13,7 @@
     set_optimizer_state_dict,
 )
 from transformers.modeling_utils import PreTrainedModel
+from speculators.utils.util import is_torch_npu_available
```
dsikka: For many of these checks, I think it would be easier to do something like this:

```python 

from transformers.utils import is_torch_npu_available

def fetch_device():
    if is_torch_npu_available():
       return "npu:0"
   else:
       return "cuda:0"
```

And then just update the calls as opposed to repeating lines:

``` python
full_state_dict = load_safetensors_state_dict(
            self.model_path(self.previous_epoch),
            fetch_device()  
 )
```

fynnsu: You could also use the [torch.accelerator](https://docs.pytorch.org/docs/stable/accelerator.html) api which generalizes across different accelerator backends.

e.g. `torch.accelerator.is_available()`

fynnsu: Or to get the device label (type:idx)
```python
device = f"{torch.accelerator.current_accelerator().type}:{torch.accelerator.current_device_index()}"
```
returns `cuda:0` on an nvidia box and should generalize to other hardware types.

src/speculators/data_generation/vllm_hidden_states_generator.py
```diff
@@ -24,6 +24,7 @@
 from vllm.v1.executor.multiproc_executor import MultiprocExecutor
 from vllm.v1.request import Request, RequestStatus
 from vllm.v1.structured_output import StructuredOutputManager
+from speculators.utils.util import is_torch_npu_available
```
shanjiaz: The imports are failing because the util is called `is_npu_available`

src/speculators/data_generation/config_generator.py
```diff
@@ -30,15 +31,19 @@
...
+    :return: GPU model and count or NPU model and count, or "CPU only" if no GPU/NPU available
     """
-    if not torch.cuda.is_available():
+    if torch.cuda.is_available():
+        gpu_count = torch.cuda.device_count()
+        gpu_name = torch.cuda.get_device_name(0)
+        return gpu_name if gpu_count == 1 else f"{gpu_count}x {gpu_name}"
+    elif is_torch_npu_available():
+        npu_count = torch.npu.device_count()
+        npu_name = torch.cnpuuda.get_device_name(0)
```
shanjiaz: Typo. This should probably be `torch.npu.get_device_name`

shanjiaz (changes requested):
Thanks so much for contributing! Would you mind taking another look at the failing tests? It would be great if you could try adding some NPU tests and share the results with us as well. Feel free to reach out with any questions/concerns. You can also find us in vllm slack speculators channel!

momo609:
> Thanks so much for contributing! Would you mind taking another look at the failing tests? It would be great if you could try adding some NPU tests and share the results with us as well. Feel free to reach out with any questions/concerns. You can also find us in vllm slack speculators channel!

I'll adding some test results on NPU. We also have some of our own trained model weights, would it be okay to include a link to these in the repository as well?

momo609:
> > > Thanks so much for contributing! Would you mind taking another look at the failing tests? It would be great if you could try adding some NPU tests and share the results with us as well. Feel free to reach out with any questions/concerns. You can also find us in vllm slack speculators channel!
> > 
> > 
> > I'll adding some test results on NPU. We also have some of our own trained model weights, would it be okay to include a link to these in the repository as well?
> 
> @momo609 Could you take another look at the failed quality tests? You can run `tox -e quality` and `pre-commit run --all-files` locally to check. Thanks!

fixed

scripts/gen_and_train.py
```diff
@@ -163,6 +165,16 @@ def print_block(title: str, content: str):
...
+def npu_command(python_alt, command):
+    if is_npu_available():
+        if python_alt and python_alt.strip():
+            command = []
+            command.extend(python_alt.split())
+        else:
+            command = ["python"]
+    return command
+
+
```
fynnsu, line 181: It seems like the purpose of this is to remove the use of `uv` to setup the environment and instead just call python directly (using whatever environment is enabled).

The reason we're using `uv` for this is because the environment requirements for the data generation phase are different than for training. The logic in `run_script` allows us to create an ephemeral environment with all the required packages for each command that is run.

If the problem is that you require additional packages (such as to support torch npu), then they can be added to the `requires` list of `run_script` when calling `run_script`.

src/speculators/data_generation/vllm_hidden_states_generator.py
```diff
@@ -310,8 +312,7 @@ def generate(self, token_ids: list[list[int]] | torch.Tensor) -> list[dict]:
                 }
             )
             offset += seq_len
-
-        torch.cuda.empty_cache()
+        empty_cache()
```
fynnsu, line 315: There is actually a `torch.accelerator` fn for this one [torch.accelerator.empty_cache()](https://docs.pytorch.org/docs/stable/generated/torch.accelerator.memory.empty_cache.html).

src/speculators/train/trainer.py
```diff
@@ -101,7 +102,10 @@ def setup_model(self):
                 for m in self.model.layers.children():  # type: ignore[union-attr]
                     if not isinstance(m, FSDPModule):
                         continue
-                    m.to_empty(device="cuda")  # type: ignore[attr-defined]
+                    if is_npu_available():
+                        m.to_empty(device="npu")  # type: ignore[attr-defined]
+                    else:
+                        m.to_empty(device="cuda")  # type: ignore[attr-defined]
```
fynnsu:
```suggestion
                    acc_type = torch.accelerator.current_accelerator().type
                    m.to_empty(device=acc_type)  # type: ignore[attr-defined]
```

scripts/gen_and_train.py
momo609: > It seems like the purpose of this is to remove the use of `uv` to setup the environment and instead just call python directly (using whatever environment is enabled).
> 
> The reason we're using `uv` for this is because the environment requirements for the data generation phase are different than for training. The logic in `run_script` allows us to create an ephemeral environment with all the required packages for each command that is run.
> 
> If the problem is that you require additional packages (such as to support torch npu), then they can be added to the `requires` list of `run_script` when calling `run_script`.

Yes, but vllm_ascend currently depends on PyTorch 2.8, which conflicts with the latest vLLM dependencies, so we directly launched it with Python and manually installed the required dependencies.

momo609:
I have run 'tox -e quality' and successs，but ci seems don't have this fuc?
image

dsikka:
@momo609 I believe you are seeing type failures, not style. Do you mind running `tox -e quality`?
This should show the failures I think

src/speculators/train/trainer.py
```diff
@@ -21,6 +21,7 @@
     SingleGPUCheckpointer,
 )
 from speculators.train.utils import apply_fully_sharded
+from speculators.utils.util import is_npu_available
```
rahul-tuli: This function is not used in current file, we need to remove it to make quality pass, kindly refer to the patch I pasted

momo609: @rahul-tuli  fixed.

rahul-tuli: Nice, Let's go!

scripts/gen_and_train.py
```diff
@@ -167,6 +169,16 @@ def print_block(title: str, content: str):
     print("\n", "#" * term_width, "\n", sep="")
 
 
+def npu_command(python_alt, command):
```
dsikka, line 172: nit: this should also go in utils, not in the generic train script

scripts/gen_and_train.py
```diff
@@ -167,6 +169,16 @@ def print_block(title: str, content: str):
...
+    if is_npu_available():
+        if python_alt and python_alt.strip():
+            command = []
+            command.extend(python_alt.split())
+        else:
+            command = ["python"]
+    return command
+
```
dsikka, line 180: same as above

PR #236, opened 2026-01-06
updated vllm version

shanjiaz:
Updated vllm requirement to the latest release version.


fynnsu (commented):
Should we do a range since no code changes were needed? Like >= 0.12.0 <= 0.13.0

rahul-tuli (approved):

PR #247, opened 2026-01-16
[CI] Update stale PR / Issue conditions 

dsikka:
SUMMARY:

- Add a message for stale issues
- Bump min days from 60 to 90 to be consistent with vLLM
- Add keep-open condition to ignore PRs / Issues with this label

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -17,6 +17,8 @@ jobs:
         with:
           operations-per-run: 1000
           exempt-draft-pr: true
+          exempt-issue-labels: 'keep-open'
+          exempt-pr-labels: 'keep-open'
 
           days-before-issue-stale: 90
           days-before-issue-close: 30


shanjiaz (dismissed):

.github/workflows/stale.yml
```diff
@@ -17,8 +17,22 @@ jobs:
         with:
           operations-per-run: 1000
           exempt-draft-pr: true
+          exempt-issue-labels: 'keep-open'
+          exempt-pr-labels: 'keep-open'
 
-          days-before-pr-stale: 60
+          days-before-issue-stale: 90
```
fynnsu, line 23: I think 60 days (or even less) makes more sense for marking as stale. 3 months is a long time for no progress on a pr. Maybe for issues it's okay since the progress could be happening on the linked pr.

fynnsu (approved):

PR #250, opened 2026-01-20
Add lr scheduler options to train.py and gen_and_train.py

fynnsu:
We already have support in `trainer.py` for different learning rate schedulers, however the options to configure them aren't currently exposed in the user scripts (`train.py` and `gen_and_train.py`). This pr exposes those options.


MeganEFlynn (approved):
lgtm!

PR #261, opened 2026-01-24
use the full verifier vocab

### LINKED ISSUE #200: Create `Eagle3DraftModel`s which use the full verifier vocab
Currently `Eagle3DraftModel` has `t2d` and `dt2` (vocab mapping) tensors as required inputs to it's init function. This means if a model uses the full verifier vocab, users must create "no-op" mapping tensors and pass them to the drafter. Furthermore, these mapping tensors will still be saved in the model checkpoints during training and then subsequently used by vllm when running the model.

It would be better if we made these optional, and used the full verifier vocab when they aren't provided.

Todo:
- Make `t2d` and `dt2` tensors optional inputs to `Eagle3DraftModel`
- Verify that if one is provided, the other is as well
- Make sure all existing uses of the tensors in `Eagle3DraftModel` can handle non-existent/`None` values
- Update `scripts/train.py` so that `t2d` and `d2t` are not required inputs
- Update `scripts/gen_and_train.py` if needed.

gDINESH13:
closes #200 
1. Make t2d and dt2 tensors optional inputs to Eagle3DraftModel
2. Verify that if one is provided, the other is as well
3. Update scripts/train.py so that t2d and d2t are not required inputs
4. Update scripts/gen_and_train.py

### DIFF
--- a/scripts/data_generation_offline.py
+++ b/scripts/data_generation_offline.py
@@ -196,7 +196,7 @@ def find_last_checkpoint(output_dir: str) -> int:
 
 
 def save_sample_to_disk(data_dict, output_path):
-    """Helper function to save a single sample to disk (for async execution)"""
+    """Save a single sample to disk for async execution."""
     torch.save(data_dict, output_path)
     return output_path
 
@@ -229,6 +229,8 @@ def generate_and_save_hidden_states(args, dataset):
     Path(args.output_dir).mkdir(parents=True, exist_ok=True)
 
     start_file_idx = find_last_checkpoint(args.output_dir)
+    sample_lengths = {}
+
     if start_file_idx > 0:
         log.subsection(f"Resuming: {start_file_idx} files already exist")
 
@@ -279,6 +281,7 @@ def generate_and_save_hidden_states(args, dataset):
             for j, result in enumerate(results):
                 # Truncate loss_mask to match input_ids length (generator may truncate)
                 input_len = len(result["input_ids"])
+                sample_lengths[file_idx] = input_len
                 loss_mask = batch_loss_mask[j][:input_len]
 
                 result_cleaned = {
@@ -300,6 +303,11 @@ def generate_and_save_hidden_states(args, dataset):
             future.result()
 
     samples_saved = file_idx - start_file_idx
+
+    sample_lengths_output_path = Path(args.output_dir) / "sample_lengths.json"
+    with open(sample_lengths_output_path, "w") as f:
+        json.dump(sample_lengths, f, indent=2)
+
     log.info(f"Saved {samples_saved} new data points to {args.output_dir}")
 
     save_config(args, generator, num_samples, args.output_dir)
--- a/src/speculators/train/data.py
+++ b/src/speculators/train/data.py
@@ -1,4 +1,5 @@
 # ruff: noqa: ERA001
+import json
 import math
 import os
 import random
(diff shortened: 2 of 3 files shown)


scripts/gen_and_train.py
```diff
@@ -314,24 +314,35 @@ def run_e2e(
...
+    else:
+        # Training without vocab mapping (use full verifier vocab)
+        ta_dict = {
+            **train_args._asdict(),
+            "verifier-name-or-path": verifier_name_or_path,
+            "data-path": str(output_path / "gen"),
+            "save-path": str(output_path / "checkpoints"),
+            "data-format-version": 1,
+            "log-dir": str(output_path / "logs"),
+        }
```
fynnsu: Could we avoid the duplication here?

e.g.
```python
ta_dict = {
    **train_args._asdict(),
    "verifier-name-or-path": verifier_name_or_path,
    "data-path": str(output_path / "gen"),
    "save-path": str(output_path / "checkpoints"),
    "data-format-version": 1,
    "log-dir": str(output_path / "logs"),
}
if vocab_mapping_args is not None:
    ta_dict["d2t-path"] = ...
    ta_dict["t2d-path"] = ...
```

scripts/train.py
```diff
@@ -121,10 +121,19 @@ def main(args: argparse.Namespace):
...
-    t2d = torch.from_numpy(np.load(args.t2d_path)).to(device)
-    draft_vocab_size = d2t.shape[0]
+    # Load t2d and d2t tensors if provided
+    if args.d2t_path and args.t2d_path:
+        d2t = torch.from_numpy(np.load(args.d2t_path)).to(device)
+        t2d = torch.from_numpy(np.load(args.t2d_path)).to(device)
+        draft_vocab_size = d2t.shape[0]
+    else:
+        d2t = None
+        t2d = None
```
fynnsu, line 137: If the user passes in only one path, we will set `d2t` and `t2d` to `None` here and then the logic for this error in `EagleDraftModel.__init__` won't be hit and we'll just silently use the full verifier vocab.
```
"Both t2d and d2t must be provided together, or both must be None. "
                f"Got t2d={'provided' if t2d is not None else 'None'}, "
                f"d2t={'provided' if d2t is not None else 'None'}"
```

We may want to instead have a check here:
```python
elif args.d2t_path or args.t2d_path:
    raise ...
else:
    ...
```

src/speculators/models/eagle3/core.py
```diff
@@ -285,17 +306,31 @@ def _setup_embeddings_and_lm_heads(self, config: VerifierConfig, t2d: torch.Tens
...
+            # Use full verifier vocab (no masking)
+            full_lm_head_weight = lm_head_weight.to(dtype=default_dtype)
+            if full_lm_head_weight.shape != self.lm_head.weight.shape:
+                raise ValueError(
+                    f"Verifier lm head data shape "
+                    f"{full_lm_head_weight.shape} does not match draft "
+                    f"lm head shape {self.lm_head.weight.shape}"
+                )
+            self.lm_head.weight.data = full_lm_head_weight.detach().clone()
+            self.verifier_lm_head.weight.data = full_lm_head_weight.detach().clone()
```
fynnsu: Can we also reduce the duplication here?

Something like:
```python
# lm_head_weight already defined above

if t2d is not None:
    # Reduce to limited vocab
    lm_head_weight = lm_head_weight.to(
        device=t2d.device, dtype=default_dtype
    )[t2d.to(torch.bool), :]

# shape check and attribute setting generalized for both cases
if lm_head_weight.shape != self.lm_head.weight.shape:
    raise ValueError(
        f"Masked verifier lm head data shape "
        f"{lm_head_weight.shape} does not match draft "
        f"lm head shape {self.lm_head.weight.shape}"
    )
self.lm_head.weight.data = lm_head_weight.detach().clone()
self.verifier_lm_head.weight.data = lm_head_weight.detach().clone()
```

scripts/train.py
gDINESH13: Ah thank you for pointing this. I have handled it now

scripts/gen_and_train.py
gDINESH13: done.

src/speculators/models/eagle3/core.py
gDINESH13: Done.

PR #262, opened 2026-01-24
fix: Save and load exact sample lengths for training

### LINKED ISSUE #221: Incorrect length estimates for data sample files
During data generation we write to disk a separate file for each dataset sample we process, which stores the verifier hidden states, loss mask, and token ids for the sample. During training, we then use these samples as input to the training process.

In order to efficiently pack the samples into batches, we need to know the (approximate) length of each data sample, so that we know which samples can be combined into the same batch. 

The way this is handled currently is ([code](https://github.com/vllm-project/speculators/blob/4d41d302e769f62c67ca41ffb0b76f94521cc096/src/speculators/train/data.py#L185)):
- Load the first sample and measure the sequence length (`lengths_0`) + file size (`size_0`)
- Then approximate the length of every other sample `i` by computing `math.ceil(size_i / size_0 * lengths_0)`

There are two problems with this approach:
1. Although the file size should theoretically scale linearly with sequence length, the approximation is sometimes off, particularly for samples with very different sizes/lengths than the first reference sample.
2. It seems like our method for extracting file size (`Path(fname).stat().st_size`) sometimes incorrectly returns a file size of `1` byte when being applied to large datasets (i.e. 300,000 files). This can happen as frequently as 10% of the time

The training code is designed to handle these issues by truncating off extra tokens if we have more than `max_seq_len` in a batch, however it would be better to avoid these issues entirely.

It's possible that `2` could be patched by simply retrying the file size check on files with "size 1", however the more robust approach would be to just save the sequence lengths at while generating. These could be written to a separate file (e.g. `data_sample_lengths.pt`) and loaded by the training process.

Steps:
1. Record the sequence lengths of generated samples during the data generation process and write to file. These can either be stored as a mapping (from filename/index to length) or as a simple list where the indices correspond to the file indices (the data files are stored in format `data_{i}.pt`).
2. During training, when initializing the [Eagle3SampleFileDataset](https://github.com/vllm-project/speculators/blob/4d41d302e769f62c67ca41ffb0b76f94521cc096/src/speculators/train/data.py#L132) check if this file exists. If it does, use it to get the sequence lengths, otherwise fallback to the existing approximation method.

Vishnu-sai-teja:
## Summary
This PR addresses issue #221 by saving exact sequence lengths during data generation and loading them during training.

## Problem
During training, sequence lengths were estimated by comparing file sizes, which:
1. Sometimes produced inaccurate approximations
2. Could return incorrect file sizes (1 byte) ~10% of the time for large datasets

## Solution
1. **Data Generation** (`data_generation_offline.py`): Save sample lengths to `sample_lengths.json` during generation
2. **Training** (`train/data.py`): Load exact lengths from `sample_lengths.json` if available, otherwise fall back to the original approximation method

## Changes
- `scripts/data_generation_offline.py`: Added logic to collect and save sample lengths to JSON
- `src/speculators/train/data.py`: Added logic to load lengths from JSON with graceful fallback

## Testing
- All 6 existing unit tests pass
- Manual testing verified:
  - ✅ Exact lengths loaded from `sample_lengths.json`
  - ✅ Fallback to approximation when file doesn't exist

## Backward Compatibility
- Fully backward compatible - existing datasets without `sample_lengths.json` will continue to work with the fallback method

Fixes #221

### DIFF
--- a/scripts/data_generation_offline.py
+++ b/scripts/data_generation_offline.py
@@ -29,10 +29,20 @@ import torch
 from datasets import config as datasets_config
 from tqdm import tqdm  # type: ignore[import-untyped]
 
-from speculators.data_generation.config_generator import DataGenerationConfig
-from speculators.data_generation.logging_utils import PipelineLogger
-from speculators.data_generation.preprocessing import load_and_preprocess_dataset
-from speculators.data_generation.vllm_hidden_states_generator import (
+# Set vLLM to use 'spawn' instead of 'fork'
+# to prevent "Cannot re-initialize CUDA in forked subprocess" errors
+from vllm import envs
+
+envs.VLLM_WORKER_MULTIPROC_METHOD = "spawn"
+
+from speculators.data_generation.config_generator import (  # noqa: E402
+    DataGenerationConfig,
+)
+from speculators.data_generation.logging_utils import PipelineLogger  # noqa: E402
+from speculators.data_generation.preprocessing import (  # noqa: E402
+    load_and_preprocess_dataset,
+)
+from speculators.data_generation.vllm_hidden_states_generator import (  # noqa: E402
     VllmHiddenStatesGenerator,
 )
 
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -1,0 +1,39 @@
+pull_request_rules:
+  - name: label-documentation
+    description: Automatically apply documentation label
+    conditions:
+      - label != stale
+      - -closed
+      - or:
+          - files~=^[^/]+\.md$
+          - files~=^docs/
+          - files~=^examples/
+    actions:
+      label:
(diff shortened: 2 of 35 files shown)


scripts/data_generation_offline.py
fynnsu, line 292: Could save the file_name directly here instead of index. This would potentially make the matching more stable since we wouldn't need to extract the index and we wouldn't be tied to this filename scheme. That being said, this is also fine.

e.g.
```python
file_name = f"data_{file_idx}.pt"
sample_lengths[file_name] = input_len
output_path = Path(args.output_dir) / file_name
```

Vishnu-sai-teja: Noted, thanks for the feedback !

PR #268, opened 2026-01-31
[FEAT][TRAINING] Add fine-tuning support for EAGLE3 from HF Hub

VincentG1234:
## Summary

Adds support for fine-tuning EAGLE3 models from pretrained checkpoints, enabling users to initialize training from existing models stored locally or on HuggingFace Hub.

## Changes

- Added `--pretrained-model-path` CLI argument to `scripts/train.py`
- Implemented `load_safetensors_state_dict()` function supporting:
  - Local single/sharded safetensors files
  - Automatic download from HuggingFace Hub
- Automatic extraction of `d2t`/`t2d` vocab mappings from pretrained models
- Automatic derivation of `draft_vocab_size` from loaded mappings

### Fine-tune from HuggingFace Hub:
python scripts/train.py \
  --pretrained-model-path RedHatAI/Llama-3.1-8B-Instruct-speculator.eagle3 \
  --data-path ./new_data \
  --epochs 3 \
  --lr 5e-5

## Related issue

issue #269

### DIFF
--- a/tests/integration/test_pretrained_loading.py
+++ b/tests/integration/test_pretrained_loading.py
@@ -73,10 +73,8 @@ def test_end_to_end_pretrained_loading():
 
     assert config.draft_vocab_size == d2t.shape[0]
     model_class = SpeculatorModel.registered_model_class_from_config(config)
-    model = model_class(
+    model = model_class(  # type: ignore[call-arg]
         config=config,
-        verifier=None,
-        verifier_attachment_mode=None,
         t2d=t2d,
         d2t=d2t,
     )


VincentG1234:
Hello guys @dsikka @fynnsu, just to keep you in touch. I come with good news: 

1) I investigated the acceptance drop after a light fine-tuning of an EAGLE-3 1B draft model (I test with this one: RedHatAI/Llama-3.1-8B-Instruct-speculator.eagle3). Although the weights were nearly identical to the HF checkpoint (only fp-level differences), the issue was caused by a config.json mismatch between my local setup and the HF model. In particular, rope_theta (and related RoPE scaling parameters) differed significantly, which changed the verifier’s hidden states and broke draft–verifier alignment, leading to a large drop in mean acceptance length (2.6 to 1.9).
 After aligning the RoPE configuration with the HF config, acceptance metrics returned to expected values. I will fix that in the code and I think we will be good ! 

2) I exported the loading_weights funtions in the right script as suggested 

3) I’m now validating the full set of changes in my PR that enable fine-tuning EAGLE(-3) models end-to-end (the earlier config.json/RoPE mismatch should be the last issue). I could use guidance on which tests you’d consider sufficient/most relevant for this PR. btw, I don't know why, but running tox locally is currently difficult because downloading model weights takes hours on my setup... 

have a nice week end

tests/integration/test_pretrained_loading.py
```diff
@@ -0,0 +1,89 @@
...
+def test_end_to_end_pretrained_loading():
+    """Test full workflow: load state dict → extract mappings → create model → load."""
+    state_dict = load_full_state_dict(TEST_MODEL_ID)
+    device = torch.device("cpu")
+    d2t, t2d = extract_vocab_mappings(state_dict, device)
+    config = SpeculatorModelConfig.from_pretrained(TEST_MODEL_ID)
+
+    assert config.draft_vocab_size == d2t.shape[0]
+    model_class = SpeculatorModel.registered_model_class_from_config(config)
+    model = model_class(  # type: ignore[call-arg]
```
VincentG1234, line 76: That line caused the fail of the last CI. 

Context:
SpeculatorModel.registered_model_class_from_config(config) is typed as returning type[SpeculatorModel]. The base class SpeculatorModel.__init__ (in src/speculators/model.py) requires verifier and verifier_attachment_mode. For EAGLE3 configs, the actual class returned is Eagle3DraftModel (in src/speculators/models/eagle3/core.py), whose __init__ only takes config, t2d, and d2t and passes fixed values to super().__init__(config=..., verifier=None, verifier_attachment_mode="train_only"). 
So: calling without verifier/verifier_attachment_mode triggers mypy [call-arg]; calling with them triggers a runtime TypeError from Eagle3DraftModel.__init__.

fix:
To keep it simple, and do not modify another file, I fixed it by adding # type: ignore[call-arg] for this call so mypy skips the check.

Follow-up:
It would be good to keep this mismatch in mind and consider aligning constructors  in a later change, so that code using registered_model_class_from_config() can call the constructor in a type-safe way without ignores.

Hoping we are finally good 😄

VincentG1234:
I ran a small experiment to better understand the French performance gap and wanted to share some preliminary results.

I used **MT-Bench French** as a benchmark:
[https://huggingface.co/datasets/bofenghuang/mt-bench-french](https://huggingface.co/datasets/bofenghuang/mt-bench-french)

To isolate the language effect, I ran the exact same prompts in English and in French.

## Baseline

**EN weighted acceptance**

```
[0.768 0.557 0.376]
```

**FR weighted acceptance**

```
[0.586 0.349 0.188]
```

This reveals a large gap between English and French speculative acceptance.

## Small targeted fine-tuning

I fine-tuned Eagle3 on ~2k MT-Bench-French samples, using verifier-generated assistant responses (angeluriot/french_instruct).

* verifier: `meta-llama/Llama-3.1-8B-Instruct`
* base speculator: `RedHatAI/Llama-3.1-8B-Instruct-speculator.eagle3`
* 3 epochs, lr 5e-5, seq len 8192
* 1 layer, TTT=3, cosine scheduler

Training took ~40 minutes on an A100 40GB.

## After fine-tuning

**EN (no English samples seen)**

```
[0.698 0.463 0.303]
```

**FR**

```
[0.609 0.368 0.206]
```

French acceptance improves across all positions, especially deeper ones, while English performance remains strong despite seeing no English data.

This is still a very small and somewhat superficial experiment, but the results are encouraging and suggest that light language-specific fine-tuning can noticeably improve speculative decoding efficiency and reduce the language gap.

I can share a Colab notebook if that would be useful.

---

If helpful, I can also iterate on the PR, for example by reducing diffs, simplifying parts of the logic, or improving tests, whatever makes review easier and the code more robust.

VincentG1234:
@fynnsu 

Thanks for the explanation and for taking the time to refactor this.

It makes sense to rely on the `transformers` `from_pretrained()` pathway rather than re-implementing model loading logic from scratch
I'll take a look at #333 and test the workflow on my side.

If everything looks good there, I can close this PR once #333 is merged.

VincentG1234:
Feature implemented in the PR #333

PR #271, opened 2026-02-02
Generalize training framework to support algorithms beyond Eagle 3

shanjiaz:
## Background
We're anticipating multiple training algorithms to be added in the next few weeks. This PR refactors the training infrastructure to support multiple speculative decoding algorithms (e.g., Eagle3,    
  DFlash, FastMTP) without code duplication.                                                                                                                                                             
                                                                                                                                                                                                         
 ## Changes                                                                                                                                                                                                
  - Added a readme/guideline on how to add a new algorithm to the speculators codebase                                                                                                                   
  - Each algorithm owns its training logic through classmethods              
  - Added `base_components.py` for shared transformer definitions (Llama and Qwen3)                                                                                                                      
  - Added `--speculator-type` as the main argument in `train.py` to determine the algorithm used                                                                                                         
  - Removed algorithm-specific code from `trainer.py` and `utils.py`  

## Test
- Tested with the existing examples
- Tested with the dflash training in development

### DIFF
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -1,4 +1,5 @@
 import argparse
+import random
 
 import numpy as np
 import torch
@@ -21,6 +22,19 @@ from speculators.train.logger import setup_metric_logger, setup_root_logger
 from speculators.train.noise_transforms import AddUniformNoise
 from speculators.train.trainer import Trainer, TrainerConfig
 from speculators.train.utils import maybe_destroy_distributed, maybe_setup_distributed
+
+
+def set_seed(seed: int, deterministic: bool = False):
+    """Set random seeds for reproducibility."""
+    random.seed(seed)
+    np.random.seed(seed)  # noqa: NPY002
+    torch.manual_seed(seed)
+    torch.cuda.manual_seed_all(seed)
+
+    if deterministic:
+        # For deterministic behavior (may impact performance)
+        torch.backends.cudnn.deterministic = True
+        torch.backends.cudnn.benchmark = False
 
 
 def setup_dataloader(
@@ -107,6 +121,9 @@ def create_transformer_layer_config(
 
 
 def main(args: argparse.Namespace):
+    # Set random seed for reproducibility
+    set_seed(args.seed, args.deterministic_cuda)
+
     # Setup logging
     setup_root_logger()
     setup_metric_logger(
@@ -239,6 +256,15 @@ def parse_args():
     parser.add_argument("--t2d-path", type=str, default=None)
     parser.add_argument("--ttt-steps", type=int, default=3)
     parser.add_argument("--ttt-step-loss-decay", type=float, default=1.0)
+    parser.add_argument(
+        "--seed", type=int, default=42, help="Random seed for reproducibility"
+    )
+    parser.add_argument(
+        "--deterministic-cuda",
+        action="store_true",
+        default=False,
+        help="Sets cuda to deterministic mode. This may impact performance.",
+    )
     parser.add_argument(
         "--use-off-policy-tokens",
         action="store_true",


src/speculators/models/eagle3/core.py
```diff
@@ -465,3 +465,73 @@ def forward(
             return draft_tokens, loss, metrics
         else:
             return draft_tokens
+
+    @classmethod
+    def from_training_args(
```
rahul-tuli, line 471: Should we make this a contract? By maybe adding this as an abstract method in the SpeculatorModel?

docs/adding_new_algorithms.md
```diff
@@ -0,0 +1,103 @@
...
+- Add algorithm-specific parameters as needed
+
+### 3. Implement Model Class
+
+In `core.py`, create a model class with the `@register` decorator and training factory methods.
+
+**Reference:** See `src/speculators/models/eagle3/core.py` for a complete example:
+
+- Use `@SpeculatorModel.register("myalgo")` decorator
+- Set `config_class` to link config and model
```
rahul-tuli: Could we clarify this a bit more, maybe something that makes it explicit that the config being set here refers to the once created in step 2?

src/speculators/models/eagle3/core.py
shanjiaz: Sure!

docs/adding_new_algorithms.md
shanjiaz: Yes! I can specify that.

docs/adding_new_algorithms.md
```diff
@@ -0,0 +1,103 @@
...
+3. **Training factory methods** as classmethods on the model
+4. **CLI arguments** in `train.py` (if needed)
+
+## Step-by-Step Guide
+
+### 1. Create Algorithm Module
+
+Create a directory under `src/speculators/models/<algorithm_name>/`:
+
+```bash
```
dsikka: I dont think we need to put bash commands on creating a new directory :)

docs/adding_new_algorithms.md
```diff
@@ -0,0 +1,103 @@
+# Adding New Speculative Decoding Algorithms
+
+This guide explains how to add a new speculative decoding algorithm to the Speculators library.
+
+## Quick Start
+
+Adding a new algorithm requires:
+
+1. **Configuration class** with `@register` decorator
+2. **Model class** with `@register` decorator
```
dsikka: Would be helpful to indicate what register does

docs/developer/add_new_algorithms.md
```diff
@@ -0,0 +1,103 @@
+# Adding New Speculative Decoding Algorithms
+
+This guide explains how to add a new speculative decoding algorithm to the Speculators library.
```
dsikka, line 3: I think my general feedback is that for a dev guide, explaining what / why you're doing something is a bit more useful than just what to do as we want to enable others to have the tools to debug as well.

src/speculators/train/utils.py
```diff
@@ -57,15 +55,14 @@ def maybe_destroy_distributed():
...
-    """Applies torch FSDP fully_shard to the model, wrapping layers in FSDPModule."""
+    """Applies torch FSDP fully_shard to the model, wrapping layers in FSDPModule.
+
+    Assumes the model has a `layers` attribute containing the decoder layers.
+    """
     mp_policy = MixedPrecisionPolicy(
         param_dtype=torch.bfloat16,
         reduce_dtype=torch.float32,
     )
-    if not isinstance(model, Eagle3DraftModel):
```
dsikka, line 65: Couldn't we just check against the list of registered models and make sure that the current model falls in this list? The registry should give you this

src/speculators/train/trainer.py
```diff
@@ -92,12 +91,6 @@ def setup_model(self):
             if self.resume_from_checkpoint and self.checkpointer.previous_epoch != -1:
                 self.checkpointer.load_model_state_dict(self.model)
             else:
-                if not isinstance(self.model, Eagle3DraftModel):
```
dsikka, line 95: same here

src/speculators/models/base_components.py
```diff
@@ -0,0 +1,55 @@
...
+)
+from transformers.models.qwen3.modeling_qwen3 import (
+    Qwen3Attention,
+    Qwen3DecoderLayer,
+    Qwen3RMSNorm,
+    Qwen3RotaryEmbedding,
+)
+
+
+class ModelComponents(NamedTuple):
```
dsikka, line 17: What is this class doing? Why do we need this? 
Dcostrings explaining the params and what this is doing would be very useful

src/speculators/models/eagle3/core.py
dsikka, line 471: is this only supposed to be eagle3 training args?

dsikka (dismissed):
I think as we make things more generic, we should add some form of lifecycle flow in the docstring, as we do in compressed-tensors: https://github.com/vllm-project/compressed-tensors/blob/46dd0c74534ace91cc360464c66dfdf0bb64f0ce/src/compressed_tensors/compressors/model_compressors/model_compressor.py#L93

docs/adding_new_algorithms.md
shanjiaz: Haha fair

src/speculators/train/utils.py
shanjiaz: Yes!

src/speculators/models/base_components.py
shanjiaz: This used to be in eagle3/model_definitions.py, eagle3 uses llama and dflash uses qwen, I figured they could share the same logic and make the types flexible.

docs/developer/add_new_algorithms.md
fynnsu, line 1: Please review the generated docs to ensure that this page renders correctly.

I would also suggest moving this under `docs/developer` and then adding a link to the page in `docs/developer/index.md`

Since the name is a bit long it doesn't show up nicely in the sidebar. Maybe we could simplify the title to "Add an algorithm" or similar.

docs/adding_new_algorithms.md
```diff
@@ -0,0 +1,160 @@
...
+
+
+## Quick Start
+
+Adding a new algorithm requires:
+
+1. **Configuration class** with `@register` decorator.
+
+Note: When Python imports your module, the `@register("myalgo")` decorator adds your class to a global registry dictionary. The training script looks up `"myalgo"` in the registry to find your class. This is helpful because the training script doesn't need to know about every algorithm and adding a new algorithm doesn't require modifying the training script.
+2. **Model class** with `@register` decorator
```
fynnsu: This is currently not rendering well (the list is getting broken). I think you can fix this with more lines between things.

docs/adding_new_algorithms.md
```diff
@@ -0,0 +1,160 @@
...
+3. **Training factory methods** as classmethods on the model
+4. **CLI arguments** in `train.py`
+
+
+## Step-by-Step Guide
+
+### 1. Create Algorithm Module
+
+**What:** Create a self-contained directory for your algorithm.
```
fynnsu: Where?

fynnsu: What should it contain?

I would add an example directory structure:

Something like this:
```
src/speculators/models
|-> eagle3
|-> ...
|-> new_model
    |-> __init__.py
    |-> core.py
    |-> config.py

File: docs/adding_new_algorithms.md [hunk 11]
```diff
@@ -0,0 +1,160 @@
...
+
+**Why:** Keeps algorithm logic isolated and maintainable. Each algorithm owns its configuration, model definition, and any custom components.
```
fynnsu: I'm also not entirely sold on this `What:` + `Why:` format, it feels overly verbose. They "why" information can be included naturally when it makes sense but I think the explicit format results in unnecessary fluff like
```
### 2. Implement Configuration Class

**What:** Define how your algorithm is configured.
```

I would instead just do something like:
```
1. Create Algorithm Module.

Create a self-contained directory for your algorithm under (`src/speculators/models`). All of the algorithm specific logic should be implemented in this directory, including configuration logic, model definitions, and custom components.
```

docs/developer/add_new_algorithms.md
```diff
@@ -0,0 +1,160 @@
...
+**Reference:** See `src/speculators/models/eagle3/core.py` for a complete example.
+
+**Required for the training infrastructure:**
+
+Model attributes:
+- `layers`: ModuleList of decoder layers (each layer is individually wrapped by FSDP for distributed training)
+- `lm_head`: Language model head for token prediction
+- `embed_tokens`: Embedding layer for input tokens
+
+Methods:
```
fynnsu, line 72: This repeats what we have above.

I would simplify this and focus on the required public api:
- Users need to register their model to ensure so that it's accessible to the library (But we don't need 5 lines on this)
- Models must have these attributes (layers, lm_head, embed_tokens). Please confirm that this is correct. Unless we access these directly from outside the class, they aren't actually part of the public api. I'm pretty sure we currently access layers in the fsdp wrapping but not sure about the others.
- Required/optional methods. I would make sure the base methods have docstrings with their expected inputs and outputs. I would also mention here that when adding a new alg it's important to verify the method signature / behaviour matches what is expected. But we don't need to describe it all here (since it will be duplicated anyways in the docstring).

Also is `__init__` used directly or is the public api only these three methods?

docs/adding_new_algorithms.md
```diff
@@ -0,0 +1,160 @@
...
+- `embed_tokens`: Embedding layer for input tokens
+
+Methods:
+- `from_training_args(cls, verifier_config, **kwargs)`: Factory method to build from CLI args (receives all args as kwargs)
+- `get_trainer_kwargs(args)`: Returns `(train_kwargs, val_kwargs)` dicts passed to `forward()`
+- `forward(...)`: Must return `(output, loss, metrics)` where metrics includes a `"loss"` key
+
+### 4. Export Classes
+
+**What:** Make your classes importable from the package.
```
fynnsu: Re-export public classes in the new algorithm's `__init__.py`

e.g.
```
from speculators.models.eagle3.config import Eagle3SpeculatorConfig
from speculators.models.eagle3.core import Eagle3DraftModel

__all__ = [
    "Eagle3DraftModel",
    "Eagle3SpeculatorConfig",
]
```

docs/adding_new_algorithms.md
```diff
@@ -0,0 +1,160 @@
...
+1. User runs: `python train.py --speculator-type myalgo`
+2. Training script calls: `model_class = SpeculatorModel.get_class("myalgo")`
+3. Registry returns: `MyAlgoDraftModel` class
+4. Script converts args to dict: `vars(args)` and calls: `model_class.from_training_args(verifier_config, **vars(args))`
+5. Your factory method extracts the kwargs it needs and builds the model instance
+6. Trainer validates the model is registered (via checks in `setup_model()` and `apply_fully_sharded()`)
+
+This pattern is similar to how `transformers` uses `.from_pretrained()` - each model owns its own instantiation logic.
+
+**Reference:** See `scripts/train.py` lines 150-178 for how the registry and factory methods are used.
```
fynnsu: I wouldn't reference line numbers, they will get outdated too quickly. Just mention the file

scripts/train.py
```diff
@@ -146,28 +142,26 @@ def main(args: argparse.Namespace):
...
-            verifier=VerifierConfig(
-                name_or_path=args.verifier_name_or_path,
-                architectures=["LlamaForCausalLM"],
-            ),
-        ),
-    )
+    # Attach computed values to args for model factory method
+    args.t2d = t2d
+    args.d2t = d2t
+    args.draft_vocab_size = draft_vocab_size
```
fynnsu: Can we just pass these into `from_training_args`?

e.g. 
```python
draft_model = model_class.from_training_args(
        verifier_config=transformer_layer_config,
        t2d=t2d,
        d2t=d2t,
        draft_vocab_size = draft_vocab_size,
        **vars(args),
    )

```

scripts/train.py
```diff
@@ -177,33 +171,33 @@ def main(args: argparse.Namespace):
...
         add_noise=False,
         data_format_version=args.data_format_version,
+        noise_std=args.noise_std,
+        num_workers=args.num_workers,
+        prefetch_factor=args.prefetch_factor,
     )
 
-    # Setup trainer
+    # Get trainer kwargs from model class
+    train_call_kwargs, val_call_kwargs = model_class.get_trainer_kwargs(args)
```
fynnsu: Should this be `model_class.get_trainer_kwargs(**vars(args))` ?

src/speculators/train/utils.py
```diff
@@ -57,15 +57,27 @@ def maybe_destroy_distributed():
...
+        raise TypeError(f"Model must be a SpeculatorModel, got {type(model).__name__}")
+
+    model_class = type(model)
+    registry = SpeculatorModel.registry
+    if registry is None or model_class not in registry.values():
+        raise ValueError(
+            f"Model {model_class.__name__} is not registered in "
+            f"SpeculatorModel.registry. "
+            f"Available models: {list(registry.keys()) if registry else []}"
+        )
```
fynnsu: Combined with the above, we now have two different implementations of this check which are slightly different (the one above assumes the registry exists). Can we make this a helper?
e.g.
```
SpeculatorModel.check_registered(model)

# raises error if unregistered, otherwise returns none
```

src/speculators/train/utils.py
```diff
@@ -57,15 +57,27 @@ def maybe_destroy_distributed():
...
     mp_policy = MixedPrecisionPolicy(
         param_dtype=torch.bfloat16,
         reduce_dtype=torch.float32,
     )
-    if not isinstance(model, Eagle3DraftModel):
-        # todo: generalize to non-Eagle3DraftModel
-        msg = "Only Eagle3DraftModel is supported for sharded training"
-        raise ValueError(msg)
 
     for layer in model.layers:  # type: ignore[union-attr]
```
fynnsu, line 69: We may also want an explicit check (here and above) that the model has a `self.layers` attribute. Or we can include this in `SpeculatorModel.check_registered` (maybe change name to `check_compatible` or something).

docs/developer/add_new_algorithms.md
shanjiaz: the public api is only these three methods.

fynnsu: This still has 2 method sections. Can we just combine into one list?

docs/developer/add_new_algorithms.md
```diff
@@ -0,0 +1,145 @@
+# Adding New Speculative Decoding Algorithms
```
fynnsu, line 1: Update title to match what's on index.md: "Add new algorithms" ?

shanjiaz: I think this title gives a little more information? I feel like we shortened it for the sidebar UI.

fynnsu: Screenshot 2026-02-04 at 4 45 19 PM

I think it still shows the full thing in the sidebar UI. But at least now it's line-wrapping better, so this is also fine

shanjiaz: Ok!

PR #272, opened 2026-02-03
feat: add Qwen3-VL preprocessing and multimodal data generation path

shx2005:
# Qwen3-VL (single image-text) training support

- [x] Add multimodal parsing in preprocessing (image + text) and use processor instead of tokenizer
- [x] Pass multimodal inputs into vLLM request path during hidden-state extraction
- [x] Add a new data format version for multimodal samples in training (vl-v1)
- [x] Update training pipeline to consume multimodal data format (blocked on a dedicated VL schema)
- [x] Document dataset format and E2E usage for VL training

## Background
This PR adds vision‑language (image‑text) data generation and training support, aligning preprocessing with vLLM’s multimodal tokenization.

## Test

No impact to existing branches.
Successfully ran end‑to‑end training on llamafactory/pokemon-gpt4o-captions.

### DIFF
--- a/scripts/convert_pokemon_gpt4o_to_jsonl.py
+++ b/scripts/convert_pokemon_gpt4o_to_jsonl.py
@@ -1,0 +1,258 @@
+#!/usr/bin/env python3
+"""Convert a HF dataset with images + conversations into speculators JSONL.
+
+This script loads a HuggingFace dataset split, attaches each sample's images to the
+first user turn, and writes a JSONL file with speculators-style multimodal messages.
+
+Example:
+    python scripts/convert_pokemon_gpt4o_to_jsonl.py \
+    --dataset llamafactory/pokemon-gpt4o-captions \
+    --split train \
+    --output ./data/pokemon_vl.jsonl \
+    --image-output-dir ./pokemon_images
+
+"""
+
+import argparse
+import json
+import logging
+import os
+import re
+from io import BytesIO
+from typing import Any
+
+from datasets import Image, Sequence, load_dataset
+from PIL import Image as PILImage
+from tqdm import tqdm
+
+logger = logging.getLogger(__name__)
+
+
+def parse_args() -> argparse.Namespace:
+    # Parse CLI arguments for dataset conversion to JSONL.
+    parser = argparse.ArgumentParser(
+        description="Convert pokemon-gpt4o-captions to JSONL with multimodal content."
+    )
+    parser.add_argument(
+        "--dataset",
+        type=str,
+        default=None,
+        help="HuggingFace dataset name (e.g., llamafactory/pokemon-gpt4o-captions).",
+    )
+    parser.add_argument(
+        "--split",
+        type=str,
+        default="train",
+        help="Dataset split to load (default: train).",
+    )
+    parser.add_argument(
+        "--output",
+        type=str,
+        required=True,
+        help="Output JSONL path.",
+    )
+    parser.add_argument(
+        "--image-root",
+        type=str,
+        default=None,
+        help=(
+            "Optional root directory for resolving image paths. If set, image"
+            " paths are written as basenames and resolved later using --image-root"
+            " in data generation."
+        ),
+    )
+    parser.add_argument(
+        "--image-output-dir",
+        type=str,
+        default=None,
+        help=(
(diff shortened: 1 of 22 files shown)


src/speculators/data_generation/preprocessing.py
```diff
@@ -343,13 +583,108 @@ def _preprocess_batch(
     return results
 
 
+def _preprocess_batch_multimodal(
```
shanjiaz, line 586: It seems like lots of these helper function are extremely similar to their text-only counterparts, can we merge them so the code is simpler and easier to maintain in the future?

scripts/convert_pokemon_gpt4o_to_jsonl.py
```diff
@@ -0,0 +1,258 @@
+#!/usr/bin/env python3
+"""Convert a HF dataset with images + conversations into speculators JSONL.
+
+This script loads a HuggingFace dataset split, attaches each sample's images to the
+first user turn, and writes a JSONL file with speculators-style multimodal messages.
+
+Example:
```
shanjiaz, line 7: This is great! Can we move this file under `examples/data_egeneration_and_training` since it's technically an example of converting a specific dataset?

scripts/data_generation_offline.py
```diff
@@ -224,7 +275,117 @@ def save_config(args, generator, num_samples, output_dir):
     log.info(f"Saved config v{config.version} to {config_path}")
 
 
-def generate_and_save_hidden_states(args, dataset):
+def _resolve_image_ref(
```
shanjiaz, line 316: Can we move these helper function to `speculators/data_generation under` a different file? maybe `multimodal_utils.py`. Could you also add a docstring of what data type we support for image_ref here for users to reference?

src/speculators/data_generation/vllm_hidden_states_generator.py
```diff
@@ -240,26 +266,106 @@ def generate(self, token_ids: list[list[int]] | torch.Tensor) -> list[dict]:
                 raise ValueError("token_ids cannot be empty")
             input_ids_list = token_ids
 
+        # Guard against multimodal misalignment: prompt text + mm inputs must be
+        # present and batch-aligned with token IDs, or vLLM will mix samples.
+        if multimodal_inputs is not None:
```
shanjiaz: Can we add a `is_multimodal` bool variable on the top level so we don't have to check for multimodal inputs every time?

src/speculators/data_generation/vllm_hidden_states_generator.py
```diff
@@ -240,26 +266,106 @@ def generate(self, token_ids: list[list[int]] | torch.Tensor) -> list[dict]:
...
+                    arrival_time=INITIAL_ARRIVAL_TIME,
+                )
+                # vLLM should return multimodal features; if missing, the prompt
+                # likely lacks image placeholders or preprocessing failed.
+                if engine_req.mm_features is None:
+                    raise ValueError(
+                        "Multimodal request missing mm_features; "
+                        "prompt may lack multimodal placeholders."
+                    )
+                for mm_index, mm_feature in enumerate(engine_req.mm_features):
```
shanjiaz: This is just a preference, feel free to ignore. could we maybe break this into its own function? Anticipation more multimodal function support we want to support in the future, might be cleaner that way. : )

shanjiaz (commented):
@shx2005 Thanks again for adding this feature, left some comments. We'd love to discuss this more with you if possible. We're available on vllm slack in #speculators channel.

shx2005:
> @shx2005 Thanks again for adding this feature, left some comments. We'd love to discuss this more with you if possible. We're available on vllm slack in #speculators channel.

@shanjiaz By the way, I have a quick question about the RFC process.

Should I file it first in:

[vllm-project/vllm (main repo)](https://github.com/vllm-project/vllm) or
[vllm-project/speculators (this repo](https://github.com/vllm-project/speculators/pull/272))?
My understanding is that the training/data-generation workflow is in speculators, and the issue I’m seeing appears to be an integration/calling-path question (how speculators calls vLLM internals for multimodal speculative decoding), rather than a confirmed bug in the default vllm serve path.

Could you confirm your preferred repo for the initial RFC?
I can file it there first and cross-link the other repo.

Thanks again for your guidance and support !

PR #285, opened 2026-02-10
[Docs] Move algorithm guide + exclude branding doc

dsikka:
- Move adding a new algorithm guide to a new folder
- Exclude branding doc from readthedocs build

Sample build: https://vllm--285.org.readthedocs.build/projects/speculators/en/285/


mkdocs.yml
```diff
@@ -48,6 +48,9 @@ theme:
         icon: material/brightness-4
         name: Switch to system preference
 
+exclude_docs: |
+  developer/branding.md
+
 validation:
```
fynnsu, line 54: Should we just remove the doc instead of excluding it like this?

fynnsu (changes requested):

mkdocs.yml
dsikka: It's good to keep around for marketing / when working with Addie / Sasa as a reference and we may want to show it again in the future when the project grows.

Could potentially live somewhere internally but we dont maintain anything like that atm

PR #288, opened 2026-02-11
Remove torch upper bound to support 2.10

dsikka:
- We need to follow-up with upper and lower bounds


fynnsu (dismissed):
Looks like the e2e test failed. We should get that working first

PR #293, opened 2026-02-18
Response regeneration

MeganEFlynn:
Added scripts that automatically start vllm servers, send requests through them to regenerate responses for the ultrachat 200k and magpie datasets, and then shutdown the servers.  It saves the results as a json file compatible with the datagen script.


shanjiaz (approved):
Looks good!

PR #296, opened 2026-02-23
Add local environment option to gen_and_train.py script

fynnsu:
This pr adds an option to use the local environment instead of a `uv run` ephemeral environment for the training step in the `gen_and_train.py` script.
Setting `LOCAL_TRAIN_ENV` to anything other than an empty string will enable this option.

I also updated the npu logic in `gen_and_train.py` to use the same pathway.

Lastly, we enable this option in
`tests/e2e/vllm/test_gen_train_acceptance.py::test_gen_train_acceptance` and ensure the e2e test runs with the correct uv torch backend.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,6 +2,34 @@
 
 Thank you for considering contributing to Speculators! We welcome contributions from the community to help improve and grow this project. This document outlines the process and guidelines for contributing.
 
+## Claiming Work
+
+**We value your time and want to ensure your contributions have the greatest impact.**
+
+To make the contribution process as smooth as possible, we ask that you coordinate with maintainers before diving into significant changes. This helps us:
+
+- **Avoid duplicate effort** if someone else is already working on a similar solution.
+- **Align on architecture** early on so your PR can be merged quickly.
+- **Protect your time** by ensuring the proposed change fits the project’s long-term roadmap.
+
+### How to Get Started
+
+1. **Find or create an issue:** Check if your idea is already being discussed. If not, open a new issue to propose the change.
+2. **Start a conversation:** Comment on the issue to let us know you’re interested! A brief outline of your planned approach is always helpful.
+3. **Wait for the "Green Light":** A maintainer will assign the issue to you. This is our signal that the approach looks good and the "floor is yours."
+4. **Build away:** Once assigned, you’re all set to begin implementation.
+
+### When should you check in?
+
+| **Let's chat first if...**                      | **Feel free to dive right in for...**   |
+| :---------------------------------------------- | :-------------------------------------- |
+| Adding new training algorithms or model support | Fixing typos or improving documentation |
+| Modifying the data pipeline or CLI/API          | Small bug fixes (generally < 20 lines)  |
+| Large refactors or performance optimizations    | Adding missing type annotations         |
+| Changes affecting 3+ files                      | Minor dependency updates                |
+
+*Maintainers typically respond in 2-3 business days. If you haven’t heard from us after a week, please feel free to give the thread a nudge!*
+
 ## How Can You Contribute?
 
 There are many ways to contribute to Speculators:
@@ -142,7 +170,7 @@ We require a DCO signoff for all commits. For more information, see the [DCO Sig
 
 ```bash
 # Sign off a commit as you're making it
-git commit --signoff -m"my commit"
+git commit --signoff -m "my commit"
 
 # Add a signoff to the last commit you made
 git commit --amend --signoff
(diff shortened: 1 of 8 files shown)


tests/e2e/vllm/test_gen_train_acceptance.py
```diff
@@ -43,6 +44,9 @@ def test_gen_train_acceptance(tmp_path: Path):
         epochs=NUM_EPOCHS,
     )
 
+    # Use local environment for training run
+    os.environ["LOCAL_TRAIN_ENV"] = "1"
```
rahul-tuli: nit: Should we use `monkeypatch.setenv("LOCAL_TRAIN_ENV", "1")` from pytest?

scripts/gen_and_train.py
```diff
@@ -169,41 +170,31 @@ def print_block(title: str, content: str):
...
+            "--no-dev",
+            "--no-default-groups",
+            "--isolated",
+        ]
+        for package in requires[:1]:
+            command.append("--with-editable")
+            command.append(package)
+        for package in requires[1:]:
+            command.append("--with")
+            command.append(package)
```
rahul-tuli: nit: Use enumerate (to avoid slicing)?

```python
for i, pkg in enumerate(requires):
            prefix = "--with-editable" if i == 0 else "--with"
            command.extend([prefix, pkg])
```

scripts/gen_and_train.py
```diff
@@ -169,41 +170,31 @@ def print_block(title: str, content: str):
...
+
+    command.extend(python_alt.split())
 
     script_path = (Path(__file__).parent / script_name).absolute()
```
rahul-tuli, line 196: nit: I think we should use .resolve() here instead of .absolute() but we can ignore this w.r.t current diff

tox.ini
```diff
@@ -43,6 +43,7 @@ passenv =
     HF_HOME
     HF_TOKEN
     VLLM_PYTHON
+    UV_TORCH_BACKEND
```
rahul-tuli, line 46: Should we also pass in `LOCAL_TRAIN_ENV` here?

fynnsu:
> @fynnsu Thank you for fixing the issue. Just trying to understand the changes better, can you please be a bit more specific how the envs are separated for the datagen, training, and run inference in vLLM steps? When you set LOCAL_TRAIN_ENV to local env, are you referring to the native python env from the test machine, or we still use some virtual env?
> 
> Also I see the `run_e2e()` called run_script() multiple times with `use_uv` set to `not is_npu_available()` except for the last step where it's using `local_train_env = is_npu_available() or bool(os.environ.get("LOCAL_TRAIN_ENV", ""))`, can you please explain why the differences here for each steps?

@dhuangnm, yeah for sure! So the problems arise from the fact that the datagen process has a dependency on vllm, which then in turn fixes almost all of the other dependencies, including some that may interfere with training. To get around this we only include vllm in the dependencies when running datagen, and use a separate virtual environment when running the training code.

This was part of the purpose of the `gen_and_train.py` script, which is meant to be an easy to use script for running datagen (with vllm env), processing the outputs (also using env with vllm), and then running training (using an env without vllm). This script also manages making sure the inputs/outputs are passed correctly through the pipeline.

Now, on top of that we have the CI runner environment, and for historical reasons we also use `tox` to run our tests (which has it's own environment). 

So to summarize we have:
-> CI ENV (w/ tox and tox-uv) runs `tox -e test-e2e`
-> TOX ENV (installs '.[dev]') run `pytest ...`
-> Calls `run_e2e()` in `gen_and_train.py` which creates (via `uv run`)
1. DATAGEN ENV (installs '.[datagen]') runs `data_generation_offline.py`
2. DATAGEN ENV 2 (installs '.[datagen]') runs `build_vocab_mapping.py`
3. TRAIN ENV (installs '.' + 'trackio/tensorboard/wandb' if required) runs 'train.py'

Now, what this pr does is adds the option to not create a new env for 3 (TRAIN ENV) and instead use the existing local env. Which in the case of the CI, means that `train.py` runs in the TOX ENV. 

We don't enable this option for data generation and build vocab mapping because those two steps need a separate env with the vllm dependency. It also makes sense to treat the last step specially here, because we want to have more control over the test environment it runs it, whereas the datagen will hopefully be deprecated soon and therefore just needs to work in specific environments for now.

tox.ini
fynnsu: Not needed, because this is set inside the test itself. PASSENV is only required when passing variables from outside the testing process.

scripts/gen_and_train.py
fynnsu: Leaving for now

tox.ini
rahul-tuli: Thanks! Learnt something new

PR #297, opened 2026-02-23
Fix --norm-before-residual flag being impossible to disable

guan404ming:
## Why                                                                                                                  
  
Using action="store_true" with default=True makes the flag always True, preventing users from training with norm_before_residual=False.                                                                                             

## How

- Replace `store_true` with `BooleanOptionalAction` so `--no-norm-before-residual` sets it to False

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,6 +2,34 @@
 
 Thank you for considering contributing to Speculators! We welcome contributions from the community to help improve and grow this project. This document outlines the process and guidelines for contributing.
 
+## Claiming Work
+
+**We value your time and want to ensure your contributions have the greatest impact.**
+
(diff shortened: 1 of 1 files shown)


scripts/train.py
```diff
@@ -304,7 +304,7 @@ def parse_args():
     # Model hyperparameters
     parser.add_argument(
         "--norm-before-residual",
-        action="store_true",
+        action=argparse.BooleanOptionalAction,
         default=True,
         help="Whether to normalize before residual connection",
```
rahul-tuli: Could you also update the help message to something like:
```
Toggle normalization before residual connections (default: True)
```

guan404ming: Thanks, just updated.

PR #298, opened 2026-02-24
support sp ulysses

momo609:
## Core Functionality
This commit primarily implements support for Sequence Parallelism Ulysses (SP Ulysses) , used for parallel processing of long sequences in distributed training.

### 1. Distributed Training Framework Refactoring (distributed.py)
- New maybe_setup_distributed() function : Initializes distributed training environment
  
  - Supports both SP Ulysses and SP Ring parallel strategies
  - Creates device_mesh for managing draft_dp and sp dimensions
  - Sets up Ulysses process groups (size=4) and Ring process groups (size=1)
- New all-gather operation functions : all_gather_tensor() and Gather custom autograd functions for distributed tensor aggregation
- New gather_outputs_and_unpad() function : Collects outputs along sequence dimension and removes padding
### 2. Eagle3 Model Core Modifications (core.py)
- New prepare_usp_input() method : Prepares input for Ulysses sequence parallelism, chunking full input along SP dimension
- Modified forward() method :
  
  - Calls prepare_usp_input() on hidden_states and input_ids for sharding
  - Supports conversion from global input to local shards
  - Processes sharded input in each TTT (Test-Time Training) step
- New sequence parallelism related attributes :
  
  ```
  self.sp_ring_degree = torch.distributed.get_world_size(get_sp_ring_group())
  self.sp_ulysses_degree = torch.distributed.get_world_size(get_sp_ulysses_group
  ())
  self.sp_world_size = self.sp_ring_degree * self.sp_ulysses_degree
  self.sp_rank = torch.distributed.get_rank() % self.sp_world_size
  ```
### 3. Eagle3 Attention Layer USP Implementation (model_definitions.py)
Core class LlamaUSPFlashAttention : Implements Flash Attention supporting both Ulysses SP and Ring SP

- Initialization parameters :
  
  - Gets Ring and Ulysses process groups
  - Sets number of heads, head dimensions, position encoding, etc.
- Forward propagation forward() :
  
  1. Projects Q/K/V and reshapes
  2. Uses SeqAllToAll4D to exchange sequence and head dimensions within Ulysses group
  3. Applies RoPE position encoding (supports sharded sequences)
  4. Calls ring_attention_hybrid_masked() to execute Ring Attention
  5. Uses SeqAllToAll4D again to restore tensor shapes
  6. Output projection
- Ring Attention implementation ring_attention_hybrid_masked() :
  
  - Loops within Ring process group to compute attention
  - Supports KV Cache accumulation
  - Uses Online Softmax algorithm for numerically stable computation
  - Supports attention masks
- Auxiliary function all_to_all_4D() : Implements all-to-all communication for 4D tensors, used for Ulysses SP
### 4. Utility Functions Update (utils.py)
- Removes import and definition of maybe_setup_distributed and maybe_destroy_distributed functions
- Keeps apply_fully_sharded() function for FSDP sharding
### 5. Training Script Update (train.py)
- Modifies import statement to import distributed functions from speculators.train.distributed (instead of speculators.train.utils )
## Technical Highlights
1. Ulysses SP : Distributes long sequences across multiple GPUs along sequence dimension, with each GPU processing part of the sequence
2. Ring SP : Exchanges KV Cache while computing attention through ring communication pattern, supporting even longer sequences
3. All-to-All Communication : Exchanges tensor sequence and head dimensions within Ulysses group to achieve load balancing
4. Online Softmax : Uses numerically stable online Softmax algorithm in Ring Attention
This commit significantly enhances the Eagle3 model's capability for long sequence training, supporting larger batch sizes and longer context windows through sequence parallelism techniques.

### DIFF
--- a/src/speculators/models/eagle3/core.py
+++ b/src/speculators/models/eagle3/core.py
@@ -2,7 +2,6 @@
 import copy
 import warnings
 from typing import ClassVar
-from typing import ClassVar, Optional
 
 import torch
 from torch.nn.attention.flex_attention import create_block_mask
@@ -17,8 +16,12 @@ from speculators.models.eagle3.attention import (
 )
 from speculators.models.eagle3.model_definitions import model_classes
 from speculators.proposals.greedy import GreedyTokenProposalConfig
+from speculators.train.distributed import (
+    gather_outputs_and_unpad,
+    get_sp_ring_group,
+    get_sp_ulysses_group,
+)
 from speculators.utils.loading import load_model_layers
-from speculators.train.distributed import get_sp_ring_group, get_sp_ulysses_group, gather_outputs_and_unpad
 
 
 def align_for_step(
@@ -153,6 +156,7 @@ def conditional_torch_compile(func):
     else:
         return func
 
+
 # Copied from transformers.models.bart.modeling_bart._make_causal_mask
 def _make_causal_mask(
     input_ids_shape: torch.Size,
@@ -184,9 +188,7 @@ def _make_causal_mask(
     )
 
 
-def _expand_mask(
-    mask: torch.Tensor, dtype: torch.dtype, tgt_len: Optional[int] = None
-):
+def _expand_mask(mask: torch.Tensor, dtype: torch.dtype, tgt_len: int | None = None):
     """
     Expands attention_mask from `[bsz, seq_len]` to
     `[bsz, 1, tgt_seq_len, src_seq_len]`.
@@ -201,6 +203,7 @@ def _expand_mask(
     return inverted_mask.masked_fill(
         inverted_mask.to(torch.bool), torch.finfo(dtype).min
     )
+
 
 @SpeculatorModel.register("eagle3")
 class Eagle3DraftModel(SpeculatorModel):
@@ -277,7 +280,6 @@ class Eagle3DraftModel(SpeculatorModel):
         )
         self.sp_world_size = self.sp_ring_degree * self.sp_ulysses_degree
         self.sp_rank = torch.distributed.get_rank() % self.sp_world_size
-
 
     def _setup_decoder_layers(
         self, transformer_layer_config: PretrainedConfig, norm_before_residual: bool
@@ -432,37 +434,35 @@ class Eagle3DraftModel(SpeculatorModel):
     def prepare_usp_input(self, full_input):
         if self.enable_sp_ulysses:
             return self.basic_extract_local(
-                        full_input,
-                        rank=self.sp_rank,
-                        world_size=self.sp_world_size,
-                    ).clone()
+                full_input,
+                rank=self.sp_rank,
+                world_size=self.sp_world_size,
+            ).clone()
         else:
             return full_input
 
     def prepare_decoder_attention_mask(
-            self, attention_mask, input_shape, inputs_embeds, past_key_values_length
-        ):
-            # create causal mask
-            # [bsz, seq_len] -> [bsz, 1, tgt_seq_len, src_seq_len]
-            combined_attention_mask = None
-            if input_shape[-1] > 1:
-                combined_attention_mask = _make_causal_mask(
-                    input_shape,
-                    inputs_embeds.dtype,
-                    device=inputs_embeds.device,
-                    past_key_values_length=past_key_values_length,
-                )
-
-            expanded_attn_mask = _expand_mask(
-                attention_mask, inputs_embeds.dtype, tgt_len=input_shape[-1]
-            ).to(inputs_embeds.device)
-            combined_attention_mask = (
-                expanded_attn_mask
-                if combined_attention_mask is None
-                else expanded_attn_mask + combined_attention_mask
+        self, attention_mask, input_shape, inputs_embeds, past_key_values_length
+    ):
+        # create causal mask
+        # [bsz, seq_len] -> [bsz, 1, tgt_seq_len, src_seq_len]
+        combined_attention_mask = None
+        if input_shape[-1] > 1:
+            combined_attention_mask = _make_causal_mask(
+                input_shape,
+                inputs_embeds.dtype,
+                device=inputs_embeds.device,
+                past_key_values_length=past_key_values_length,
             )
 
-            return combined_attention_mask
+        expanded_attn_mask = _expand_mask(
+            attention_mask, inputs_embeds.dtype, tgt_len=input_shape[-1]
+        ).to(inputs_embeds.device)
+        return (
+            expanded_attn_mask
+            if combined_attention_mask is None
+            else expanded_attn_mask + combined_attention_mask
+        )
 
     def sp_ulysses_forward(
         self,
@@ -521,7 +521,7 @@ class Eagle3DraftModel(SpeculatorModel):
                 else torch.ones(1, total_seq_len, device=device, dtype=torch.bool)
             )
             metrics = {}
-        
+
         hidden_states = self.prepare_usp_input(hidden_states)
         global_input_ids = input_ids
         draft_tokens = []
@@ -607,16 +607,18 @@ class Eagle3DraftModel(SpeculatorModel):
         **kwargs,
     ):
         if self.enable_sp_ulysses:
-            self.sp_ulysses_forward(hidden_states,
-                                    input_ids,
-                                    lengths,
-                                    loss_mask,
-                                    position_ids,
-                                    verifier_last_hidden_states,
-                                    ttt_steps,
-                                    ttt_step_loss_decay,
-                                    use_off_policy_tokens,
-                                    **kwargs,)
+            self.sp_ulysses_forward(
+                hidden_states,
+                input_ids,
+                lengths,
+                loss_mask,
+                position_ids,
+                verifier_last_hidden_states,
+                ttt_steps,
+                ttt_step_loss_decay,
+                use_off_policy_tokens,
+                **kwargs,
+            )
         device = hidden_states.device
         total_seq_len = hidden_states.shape[1]
 
--- a/src/speculators/models/eagle3/core.py
+++ b/src/speculators/models/eagle3/core.py
@@ -2,6 +2,7 @@
 import copy
 import warnings
 from typing import ClassVar
+from typing import ClassVar, Optional
 
 import torch
 from torch.nn.attention.flex_attention import create_block_mask
@@ -17,6 +18,7 @@ from speculators.models.eagle3.attention import (
 from speculators.models.eagle3.model_definitions import model_classes
 from speculators.proposals.greedy import GreedyTokenProposalConfig
 from speculators.utils.loading import load_model_layers
+from speculators.train.distributed import get_sp_ring_group, get_sp_ulysses_group, gather_outputs_and_unpad
 
 
 def align_for_step(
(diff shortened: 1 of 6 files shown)


momo609:
> Hi @momo609, thanks for the contribution! I have a few comments to help us review this PR:
> 
> 1. Could you please rebase on main? It looks like this branch was created directly on top of your earlier NPU support branch.
> 2. Please add a PR description briefly outlining the motivation, the main changes introduced, and the tests you’ve run.
>    Let us know if you have any questions — happy to help!

before use sp_ulysses，8k input will oom
image
after use sp_ulysses with ring_size=1,ulysses_size=4, 8k input can training.
image

src/speculators/models/eagle3/core.py
```diff
@@ -348,6 +425,171 @@ def _setup_embeddings_and_lm_heads(
...
+                    hidden_states,
+                    attention_mask=attention_mask,
+                    position_ids=position_ids,
+                    past_key_values=past_key_values,
+                    **kwargs,
+                )
+
+            logits = self.lm_head(self.norm(hidden_states))
+            # shape: [1, total_seq_len, draft_vocab_size]
+            logits = gather_outputs_and_unpad(logits, gather_dim=1)
```
uygnef, line 548: You can use distributed computation for the logits and compute the local loss, then perform an all-reduce at the end. Otherwise, gathering will consume a lot of GPU memory.

src/speculators/models/eagle3/core.py
```diff
@@ -348,6 +425,171 @@ def _setup_embeddings_and_lm_heads(
...
+        past_key_values = DynamicCache(config=self.config.transformer_layer_config)
+
+        attention_mask = torch.ones(
+            (batch_size, seq_length), dtype=torch.bool, device=hidden_states.device
+        )
+        attention_mask = self.prepare_decoder_attention_mask(
+            attention_mask, (batch_size, seq_length), hidden_states, 0
+        )
+
+        hidden_states = self.fc(hidden_states)
```
uygnef, line 502: It may be better to split the hidden states before passing them through self.fc; otherwise, during backpropagation, the gradients may be repeatedly accumulated across the SP group, which could negatively affect the results.

src/speculators/models/eagle3/core.py
```diff
@@ -364,6 +606,19 @@ def forward(
...
+        if self.enable_sp_ulysses:
+            self.sp_ulysses_forward(
+                hidden_states,
+                input_ids,
+                lengths,
+                loss_mask,
+                position_ids,
+                verifier_last_hidden_states,
+                ttt_steps,
+                ttt_step_loss_decay,
```
speculatorsbot, line 618: **Bug (critical): `sp_ulysses_forward` is called but its return value is discarded, then the original forward code runs unconditionally.**

The `if self.enable_sp_ulysses:` block calls `self.sp_ulysses_forward(...)` without `return`, so execution falls through to the original `forward` body below. This means:
- The SP path's results (draft tokens, loss, metrics) are thrown away
- The regular forward path runs on un-sharded inputs
- If training with SP enabled, this produces silently wrong results

This should be:
```python
if self.enable_sp_ulysses:
    return self.sp_ulysses_forward(
        hidden_states,
        ...
    )
```

speculatorsbot (commented):
This PR adds Ulysses Sequence Parallelism (USP) support to the Eagle3 training pipeline. The feature addresses a real need (OOM on long sequences), but there are several critical correctness bugs that would prevent this from working, and the approach needs significant rethinking given how the codebase has evolved since the PR was opened.

**Key concerns:**

1. **Severe merge conflicts with main**: The codebase has been substantially refactored since this branch was created. `distributed.py` now exists on main with a completely different architecture (module-level state with getters like `get_sp_group`, `get_dp_group`, `maybe_setup_distributed(sp_size=1)`). This PR creates a new `distributed.py` from scratch, which would conflict entirely. The `model_definitions.py` file has also been restructured (uses a mixin + `base_components.override_components` pattern now). A rebase is required before this can be reviewed meaningfully.

2. **The PR removes Qwen3 support**: The rewrite of `model_definitions.py` drops `Qwen3DecoderEagle3FirstLayer` and the entire `base_components` integration, reducing the `model_classes` dict to only `"llama"`. This is a regression.

3. **The `model_definitions.py` rewrite is disproportionate**: As @shanjiaz noted, this rewrites the entire file to add USP support for a single model. The existing architecture uses a mixin pattern specifically to keep model-specific code minimal. USP attention should be designed as a composable component, not a monolithic replacement.

4. **No tests are included**: A feature of this complexity (distributed attention, ring communication, gradient-correct autograd functions) requires unit tests, especially for the numerical correctness of the ring attention and all-to-all communication patterns.

Inline comments below cover the critical bugs found in the implementation.

---

src/speculators/models/eagle3/core.py
```diff
@@ -221,27 +274,51 @@ def __init__(
             config.speculators_config.verifier, t2d, config.embed_requires_grad
         )
 
+        self.sp_ring_degree = torch.distributed.get_world_size(get_sp_ring_group())
+        self.sp_ulysses_degree = torch.distributed.get_world_size(
+            get_sp_ulysses_group()
```
speculatorsbot, line 279: **Bug (critical): `get_sp_ring_group()` and `get_sp_ulysses_group()` are called unconditionally in `__init__`, but the globals they reference (`_SP_RING_GROUP`, `_SP_ULYSSES_GROUP`) are never initialized at module level.**

In `distributed.py`, these globals are only assigned inside `maybe_setup_distributed_sp_ulysses()`. If the user is not using SP Ulysses (the default), these functions will raise `NameError`. This means `Eagle3DraftModel` cannot be instantiated at all without SP Ulysses enabled.

These process group lookups should be guarded by `if self.enable_sp_ulysses:`, or the globals should be initialized to `None` with fallback logic.

src/speculators/models/eagle3/model_definitions.py
```diff
@@ -1,41 +1,163 @@
...
+        if seq_world_size > 1:
+            dist.all_to_all_single(output, input_t, group=group)
+            if use_sync:
+                torch.cuda.synchronize()
+        else:
+            output = input_t
+        # if scattering the seq-dim, transpose the heads back to the original dimension
+        output = output.reshape(seqlen, bs, shard_hc, hs)
+
+        # (seq_len, bs, hc/P, hs) -reshape-> (bs, seq_len, hc/P, hs)
```
speculatorsbot, line 73: **Bug: `input` (Python builtin) used instead of the function parameter `x`.**

In the `scatter_idx == 1 and gather_idx == 2` branch, line 73 reads:
```python
bs, seqlen, shard_hc, hs = input.shape
```

The function parameter is `x`, not `input`. This will either crash with `AttributeError` (the builtin `input` function has no `.shape`) or, if `input` was shadowed by an earlier assignment, reference stale data. All occurrences of `input` in this branch (`input.shape`, `input.reshape(...)`) should be `x`.

src/speculators/models/eagle3/model_definitions.py
```diff
@@ -111,33 +227,487 @@ def forward(
...
+                    reqs.append(dist.isend(curr_k, dst=send_dst, group=group))
+                    reqs.append(dist.isend(curr_v, dst=send_dst, group=group))
+                    reqs.append(dist.irecv(next_k, src=recv_src, group=group))
+                    reqs.append(dist.irecv(next_v, src=recv_src, group=group))
+                else:
+                    reqs.append(dist.irecv(next_k, src=recv_src, group=group))
+                    reqs.append(dist.irecv(next_v, src=recv_src, group=group))
+                    reqs.append(dist.isend(curr_k, dst=send_dst, group=group))
+                    reqs.append(dist.isend(curr_v, dst=send_dst, group=group))
+
```
speculatorsbot, line 579: **Bug: `NormLlamaDecoderLayer.forward` passes kwargs that `LlamaUSPFlashAttention.forward` does not accept.**

The decoder layer calls `self.self_attn(...)` with `past_key_values`, `output_attentions`, and `use_cache`, but `LlamaUSPFlashAttention.forward` only accepts `hidden_states`, `attention_mask`, and `position_ids`. This will crash at runtime with a `TypeError`.

src/speculators/models/eagle3/model_definitions.py
```diff
@@ -111,33 +227,487 @@ def forward(
...
+        ).transpose(1, 2)
+
+        value_states = self.v_proj(hidden_states)
+        value_states = value_states.view(
+            bsz, q_len, self.num_key_value_heads, self.head_dim
+        )
+        value_states = SeqAllToAll4D.apply(
+            self.ulysses_pg,
+            value_states,
+            self.scatter_idx,
```
speculatorsbot, line 451: **Bug: `ring_attention_hybrid_masked` is called with 5 positional arguments but the method signature only accepts 4 (plus `self`).**

The call site passes `q_len` as a fifth positional argument:
```python
attn_output = self.attention_impl(query_states, attention_mask, cache_k, cache_v, q_len)
```

But the method signature is:
```python
def ring_attention_hybrid_masked(self, local_q, attention_mask, cache_k, cache_v):
```

This will crash with `TypeError: ring_attention_hybrid_masked() takes 5 positional arguments but 6 were given`.

src/speculators/train/distributed.py
```diff
@@ -0,0 +1,262 @@
...
+        mesh_dim_names=("draft_dp", "sp"),
+    )
+    sp_ring_degree = sp_ring_size
+    sp_ulysses_degree = sp_ulysses_size
+    sp_degree = sp_ring_degree * sp_ulysses_degree
+    dp_degree = world_size // sp_degree
+
+    if world_size % sp_degree != 0:
+        raise ValueError(
+            f"world_size {world_size} % sp_degree {sp_ulysses_degree} == 0"
```
speculatorsbot, line 79: **Bug: hardcoded `"npu"` device type in `init_device_mesh`.**

```python
draft_device_mesh = dist.device_mesh.init_device_mesh(
    "npu",
    ...
)
```

This hardcodes Huawei NPU as the device type. On CUDA GPUs (which are the primary target for this repo), this will fail. This should use the detected accelerator type, e.g., `torch.accelerator.current_accelerator().type` or `"cuda"`.

src/speculators/train/distributed.py
```diff
@@ -0,0 +1,262 @@
...
+
+    logger.info(
+        f"Started distributed with local_rank={local_rank}, world_size={world_size}",
+        extra={"override_rank0_filter": True},
+    )
+    return local_rank, world_size, rank, True
+
+
+def maybe_destroy_distributed():
+    """Destroys distributed process group if using distributed training."""
```
speculatorsbot, line 164: **Bug: `maybe_setup_distributed` calls `init_process_group` and then conditionally calls `maybe_setup_distributed_sp_ulysses`, which calls `init_process_group` again.**

When `enable_sp_ulysses=True`, the execution path is:
1. `maybe_setup_distributed` calls `dist.init_process_group(...)` (line 155)
2. Then calls `maybe_setup_distributed_sp_ulysses(...)` (line 162)
3. Which calls `dist.init_process_group(...)` again (line 54)

Calling `init_process_group` twice raises `RuntimeError: trying to initialize the default process group twice!`

src/speculators/models/eagle3/model_definitions.py
```diff
@@ -111,33 +227,487 @@ def forward(
...
+
+            # 2.5 [FP32] Online Softmax Update
+            m_block = attn_block.max(dim=-1).values  # fp32
+            m_new = torch.maximum(m, m_block)
+
+            alpha = torch.exp(m - m_new)
+            p_block = torch.exp(attn_block - m_new.unsqueeze(-1))  # fp32
+
+            l_new = loss * alpha + p_block.sum(dim=-1)
+
```
speculatorsbot, line 607: **Regression: `norm_before_residual` behavior removed.**

The PR removes the `norm_before_residual` logic from `LlamaDecoderEagle3FirstLayer.forward`:
```python
-        if self.norm_before_residual:
-            residual = hidden  # set residual to normalized hidden
```

This is a configurable behavior that existing checkpoints may depend on. `NormLlamaDecoderLayer` also omits it. Removing this changes the model's computation for configs with `norm_before_residual=True`, silently producing different training results.

PR #300, opened 2026-02-24
Feature/tool use support

olisicky:
This PR adds tool-use support for data generations. Many applications uses tool-use and training speculators with such data might be beneficial for the model.

### DIFF
--- a/src/speculators/data_generation/preprocessing.py
+++ b/src/speculators/data_generation/preprocessing.py
@@ -489,6 +489,8 @@ def _parse_conv_tools(conv_tools: object, idx: int) -> list | None:
     on invalid JSON or unexpected types."""
     if not conv_tools:
         return None
+    if isinstance(conv_tools, list):
+        return conv_tools
     if not isinstance(conv_tools, str):
         log.warning(
             f"Non-string value in tools column for conversation {idx}: "
--- a/src/speculators/data_generation/preprocessing.py
+++ b/src/speculators/data_generation/preprocessing.py
@@ -531,13 +531,19 @@ def _preprocess_batch(
         # Parse tools JSON string if present; warn and skip tools on invalid JSON
         parsed_tools = None
         if conv_tools:
-            try:
-                parsed_tools = json.loads(conv_tools)
-            except json.JSONDecodeError as e:
-                log.warning(
-                    f"Invalid JSON in tools column for conversation {idx}: {e}, "
-                    "proceeding without tools"
-                )
+            if isinstance(conv_tools, str):
+                try:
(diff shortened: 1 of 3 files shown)


src/speculators/data_generation/preprocessing.py
```diff
@@ -292,13 +292,19 @@ def _preprocess_batch(
...
-                )
+            if isinstance(conv_tools, str):
+                try:
+                    parsed_tools = json.loads(conv_tools)
+                except json.JSONDecodeError as e:
+                    log.warning(
+                        f"Invalid JSON in tools column for conversation {idx}: {e}, "
+                        "proceeding without tools"
+                    )
+                else:
```
fynnsu: I think this needs to be un-indented.

fynnsu (commented):
Looks like the quality checks are failing. Could you run this locally and fix the type issues?

```
uv pip install -e .[dev]
make quality
```

The link checks are also failing but that maybe be unrelated. I can look into it.

olisicky:
Hi @fynnsu . I totally missed that two commits were not signed before and the branch is blocked. I fixed it and rebased, hopefully everything is fine. 

Thanks,

Ondřej

fynnsu:
@olisicky Hey we just landed a bigger change to preprocessing to support multi-modal. Would you be able to rebase one more time?

olisicky:
Hi @fynnsu . Sure, just rebased.

fynnsu (approved):
Looks good! Thanks for adding this

fynnsu:
Okay, looks like you have a type error. This seems like it should pass and mypy just doesn't accept it for some reason, so you can add `# type: ignore [arg-type]` to the problem lines to just ignore the issue. 

And then run `make quality` again just to check everything is working.

olisicky:
My bad. I rebased but didn't synced new pinned versions of ruff and it missed checks which were indicated here. Sorry.

olisicky:
Hi @fynnsu . Partial implementation of this PR was merged 2 weeks ago #561 but they didn't add the tools in the `apply_chat_template` so the model can't see the tool definitions.

fynnsu:
Hi @olisicky, could you rebase/merge main? I think there may also be some duplicated logic in `_normalize_conversation` now.

Can you also add an early return to `_parse_conv_tools` to handle already parsed datasets where the tools column is already a list? Something like:
```python
 if isinstance(conv_tools, list):
      return conv_tools
```

I think this should be good to merge then. Sorry for the confusion with the other tool calling pr.

fynnsu:
Hi @olisicky, I just merged the pr. Sorry this took so long, but glad we were able to get it in.

PR #301, opened 2026-02-24
Add FastMTP multi-token prediction speculator

rahul-tuli:
## Motivation

Multi-token prediction (MTP) heads are now standard in frontier models (DeepSeek-V3,
Qwen3, TencentBAC MiMo), but each ships in its own bespoke checkpoint format with
no standardized inference or fine-tuning interface. Loading them today requires either
the originating codebase or a custom conversion script.

This PR adds `FastMTPSpeculator` — a Hugging Face-compatible MTP module in the
speculators format. It provides:

1. A generic MTP layer definition suitable for fine-tuning from scratch or continued
   training on top of converted upstream checkpoints.
2. A registered config (`"mtp"`) and model (`FastMTPSpeculator`) that integrate with
   `SpeculatorModel.from_pretrained` and the existing vLLM deployment path.
3. Weight key normalization utilities to support the conversion pipeline that maps
   upstream checkpoint formats (MiMo, Qwen3-Next, DeepSeek) into the speculators format.

> **Note:** Upstream checkpoints require a conversion step before they can be loaded
> directly by speculators. The conversion utilities are part of a separate script;
> this PR defines the target format and module.

### `FastMTPConfig` — configuration

Pydantic config registered as `"mtp"` under `SpeculatorModelConfig`:

- `num_speculative_steps` (default: 3, range: 1–10) — tokens predicted per forward pass
- `num_nextn_predict_layers` (always 1) — validated at construction
- `mtp_loss_step_weights` (default: `[0.51, 0.31, 0.18]`) — per-step loss weighting for training
- Standard transformer dims (`hidden_size`, `intermediate_size`, `num_attention_heads`, etc.)
- `transformer_config: PretrainedConfig` — architecture config for MTP layer components

### `FastMTPLayer` — single MTP decoder layer

A single transformer layer that fuses verifier hidden states with token embeddings to
predict the next token:

```mermaid
flowchart TD
    H["hidden_states\n(from verifier)"] --> HN["hidden_layernorm"]
    E["embed_tokens(input_ids)"] --> TN["token_layernorm"]
    HN --> CAT["Concatenate\n[hidden ‖ embed]"]
    TN --> CAT
    CAT --> PROJ["input_proj\n(2H → H, no bias)"]
    PROJ --> LN1["input_layernorm"]
    LN1 --> ATT["Self-Attention\n+ RotaryEmb"]
    ATT --> RES1["+ residual"]
    RES1 --> LN2["post_attention_layernorm"]
    LN2 --> MLP["MLP / SparseMoeBlock"]
    MLP --> RES2["+ residual"]
    RES2 --> LN3["final_layernorm"]
    LN3 --> OUT["layer output"]
```

### `FastMTPSpeculator` — multi-step predictor

Performs `num_speculative_steps` auto-regressive passes, each using the previous
step's argmax output as the next input:

```
for step in range(num_speculative_steps):
    layer_out = FastMTPLayer(input_ids, hidden_states)
    logits    = lm_head(layer_out)          # shared with verifier
    loss     += weight[step] * CE(logits, labels[step+1:])
    input_ids = argmax(logits)              # feeds next step
    hidden_states = layer_out
```

`embed_tokens` and `lm_head` are shared with the verifier model and frozen during
training.

### Architecture-agnostic component registry

`FastMTPComponents` maps architecture names to their concrete module classes:

| Architecture | Attention | MLP | Norm | MoE |
|---|---|---|---|---|
| `mimo` | `Qwen2Attention` | `Qwen2MLP` | `Qwen2RMSNorm` | No |
| `qwen3_next` | `Qwen3NextAttention` | `Qwen3NextSparseMoeBlock` | `Qwen3NextRMSNorm` | Yes |

New architectures can be added by registering a new entry — no changes to the forward pass.

### Checkpoint key normalization

`normalize_weight_key` maps weight keys from upstream checkpoint formats into the
speculators key namespace, which is used by the conversion pipeline:

| Source format | Example key | Speculators key |
|---|---|---|
| `qwen3_next` | `mtp.layers.0.self_attn.q_proj.weight` | `model.mtp_layers.0.self_attn.q_proj.weight` |
| `qwen3_next` | `mtp.fc.weight` | `model.mtp_layers.0.input_proj.weight` |
| `qwen3_next` | `mtp.norm.weight` | `model.mtp_layers.0.final_layernorm.weight` |
| `deepseek` | `mtp.<component>` | `model.mtp_layers.0.<component>` |

### `MiMoConfig` — AutoConfig shim

Registers `model_type="mimo"` with `transformers.CONFIG_MAPPING` so TencentBAC
MiMo checkpoints can be loaded via `AutoConfig` without requiring their patched
transformers fork.

## Test Checkpoints

Integration tests run against two HuggingFace-hosted test checkpoints that have
already been converted to the speculators format:

- [`inference-optimization/test_tencentbac_fastmtp`](https://huggingface.co/inference-optimization/test_tencentbac_fastmtp) — MiMo/TencentBAC (Qwen2-based)
- [`inference-optimization/test_qwen3_next_mtp`](https://huggingface.co/inference-optimization/test_qwen3_next_mtp) — Qwen3-Next (MoE)

## Testing

```bash
# Integration tests (requires GPU)
CUDA_VISIBLE_DEVICES=3 pytest tests/integration/models/test_fast_mtp_model.py -v --tb=short
```

Tests validate for both checkpoint formats:
- `from_pretrained` loads `FastMTPSpeculator` with the correct config type
- Forward pass completes with mock `embed_tokens` / `lm_head` in detached mode
- `logits_list` has length == `num_speculative_steps`
- Each logits tensor has shape `(batch, seq_len, vocab_size)`
- No NaN or Inf values in any output tensor

## Files Changed

| File | Change |
|---|---|
| `src/speculators/models/fast_mtp/config.py` | `FastMTPConfig` with Pydantic validators |
| `src/speculators/models/fast_mtp/core.py` | `FastMTPLayer` + `FastMTPSpeculator` |
| `src/speculators/models/fast_mtp/model_definitions.py` | `FastMTPComponents` registry (mimo, qwen3_next) |
| `src/speculators/models/fast_mtp/mimo_config.py` | `MiMoConfig` AutoConfig shim |
| `src/speculators/models/fast_mtp/__init__.py` | Package exports |
| `src/speculators/models/__init__.py` | Auto-import for registry |
| `src/speculators/model.py` | Handle `speculators_config=None` edge case |
| `tests/integration/models/test_fast_mtp_model.py` | Parametrized integration tests |

### DIFF
--- a/src/speculators/models/fast_mtp/config.py
+++ b/src/speculators/models/fast_mtp/config.py
@@ -1,0 +1,116 @@
+"""Configuration for FastMTP speculator model."""
+
+from typing import Any, Literal
+
+from pydantic import Field, field_serializer, field_validator
+from transformers import AutoConfig, PretrainedConfig
+from transformers.models.qwen2.configuration_qwen2 import Qwen2Config
+
+from speculators import SpeculatorModelConfig
+
+__all__ = ["FastMTPConfig"]
+
+
+@SpeculatorModelConfig.register("mtp")
+class FastMTPConfig(SpeculatorModelConfig):
+    """Configuration for FastMTP (Multi-Token Prediction) speculator.
+
+    Targets two checkpoint families:
+    - **MiMo (TencentBAC/FastMTP)**: Qwen2-based, ``model_type="mimo"``, hidden=4096,
+      32 attention heads, 8 KV heads, vocab=151680, rope_theta=640000.
+    - **Qwen3-Next**: sparse MoE, ``model_type="qwen3_next"``, hidden=2048, 16 heads,
+      2 KV heads, vocab=151936, rope_theta=10000000.
+
+    Architecture: a single MTP layer with attention and MLP, combining verifier hidden
+    states with token embeddings via an explicit input projection. ``embed_tokens`` and
+    ``lm_head`` share the verifier's full vocabulary.
+
+    The forward pass is teacher-forced: at step k the model receives
+    ``input_ids[:, k:k+valid_len]`` as token embeddings and
+    ``hidden_states[:, :valid_len]`` from the verifier. Each step is independent,
+    enabling parallel computation during training.
+
+    **Stored fields:**
+
+    :param transformer_layer_config: Configuration for the underlying transformer
+        architecture (e.g., ``Qwen2Config``). All architecture dimensions
+        (``hidden_size``, ``vocab_size``, attention heads, MLP dims) are derived from
+        this config. Serialised as a ``to_diff_dict()`` snapshot and reconstructed via
+        ``AutoConfig`` on load.
+    :param num_nextn_predict_layers: Number of MTP prediction heads in the checkpoint.
+        vLLM reads this field directly to instantiate the correct number of MTP head
+        instances. Currently only ``1`` is supported.
+
+    **Derived properties (not stored):**
+
+    :property hidden_size: Hidden dimension, from
+        ``transformer_layer_config.hidden_size``.
+    :property vocab_size: Vocabulary size, from
+        ``transformer_layer_config.vocab_size``.
+    :property num_speculative_steps: Number of teacher-forced prediction steps, derived
+        from ``speculators_config.proposal_methods[0].speculative_tokens``. Not a stored
+        field — set via the proposal config.
+    """
+
+    speculators_model_type: Literal["mtp"] = "mtp"
+    architectures: list[str] = Field(
+        default_factory=lambda: ["FastMTPSpeculator"],
+        description="Model architectures that can load these weights",
+    )
+
+    transformer_layer_config: PretrainedConfig = Field(
+        default_factory=Qwen2Config,
+        description="Underlying transformer architecture config (e.g., Qwen2Config)",
+    )
+
+    num_nextn_predict_layers: int = Field(
+        default=1,
+        description=(
+            "Number of MTP prediction heads in the checkpoint. vLLM reads this "
+            "field to create the correct number of speculator head instances."
+        ),
+    )
+
+    @property
+    def hidden_size(self) -> int:
+        """Hidden dimension size, derived from transformer_layer_config."""
+        return self.transformer_layer_config.hidden_size  # type: ignore[return-value]
+
+    @property
+    def vocab_size(self) -> int:
+        """Vocabulary size, derived from transformer_layer_config."""
+        return self.transformer_layer_config.vocab_size  # type: ignore[return-value]
+
+    @property
+    def num_speculative_steps(self) -> int:
+        """Number of teacher-forced prediction steps, from the proposal config."""
+        return self.speculators_config.proposal_methods[0].speculative_tokens  # type: ignore[union-attr,attr-defined]
+
+    @field_validator("num_nextn_predict_layers")
+    @classmethod
+    def validate_num_nextn_predict_layers(cls, value: int) -> int:
+        """Reject configs that request more than one FastMTP layer."""
+        if value != 1:
+            raise ValueError(
+                f"FastMTP currently only supports 1 layer, got {value}. "
+                "Multi-layer support may be added in future versions."
+            )
+        return value
+
+    @field_serializer("transformer_layer_config")
+    def serialize_transformer_layer_config(self, value: PretrainedConfig) -> dict:
+        """Serialize transformer_layer_config to dict for JSON storage."""
+        return value.to_diff_dict()
+
+    @field_validator("transformer_layer_config", mode="before")
+    @classmethod
+    def validate_transformer_layer_config(cls, value: Any) -> PretrainedConfig:
+        """Validate and convert transformer config from dict or PretrainedConfig."""
+        if isinstance(value, dict):
+            config_class: type[PretrainedConfig] = Qwen2Config
+            if "model_type" in value:
+                config_class = AutoConfig.for_model(
+                    model_type=value["model_type"]
+                ).__class__
+            return config_class(**value)
+        return value
--- a/src/speculators/models/fast_mtp/core.py
+++ b/src/speculators/models/fast_mtp/core.py
@@ -1,0 +1,247 @@
+"""FastMTP speculator model implementation."""
+
+from typing import Any, ClassVar
+
+import torch
+from torch import nn
+from transformers import PretrainedConfig
+
+from speculators import SpeculatorModel
+from speculators.config import SpeculatorsConfig, VerifierConfig
+from speculators.models.fast_mtp.config import FastMTPConfig
+from speculators.models.fast_mtp.model_definitions import fast_mtp_model_classes
+from speculators.proposals.greedy import GreedyTokenProposalConfig
+from speculators.utils.loading import load_model_layers
+
+__all__ = ["FastMTPSpeculator"]
+
+
+@SpeculatorModel.register("mtp")
+class FastMTPSpeculator(SpeculatorModel):
+    """FastMTP speculator model for multi-token prediction.
+
+    FastMTP predicts multiple future tokens (default: 3) per forward pass using
+    a single layer with weighted multi-step loss for training.
+
+    embed_tokens and lm_head are always initialized with random weights in __init__.
+    When speculators_config.verifier.name_or_path is set, _setup_embeddings_and_lm_head
+    overwrites them with weights from the verifier checkpoint. When loading a
+    self-contained checkpoint (embed_tokens + lm_head weights present in the file),
+    from_pretrained fills them directly.
+    """
+
+    config_class: ClassVar[type[FastMTPConfig]] = FastMTPConfig  # type: ignore[misc]
+
+    def __init__(self, config: FastMTPConfig) -> None:
+        super().__init__(config=config)
+        tc = config.transformer_layer_config
+        self._model_definitions = fast_mtp_model_classes[tc.model_type]
+        self.mtp_layers = nn.ModuleList(
+            [self._model_definitions.first_layer_class(tc, layer_idx=0)]
+        )
+        self.rotary_emb = self._model_definitions.rotary_emb_class(tc)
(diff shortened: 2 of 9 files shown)


src/speculators/models/fast_mtp/core.py
```diff
@@ -0,0 +1,514 @@
...
+            config=config,
+            verifier=verifier,
+            verifier_attachment_mode=verifier_attachment_mode or "detached",
+        )
+
+        self.hidden_size = config.hidden_size
+        self.vocab_size = config.vocab_size
+        self.num_speculative_steps = config.num_speculative_steps
+
+        self.mtp_layers = nn.ModuleList([FastMTPLayer(config, layer_idx=0)])
```
brian-dellabetta: why make this a module list? it looks like you only ever access index 0. is it always length 1?

src/speculators/models/fast_mtp/core.py
```diff
@@ -0,0 +1,514 @@
...
+        device = input_ids.device
+
+        all_logits = []
+        total_loss = torch.tensor(0.0, device=device) if labels is not None else None
+        metrics = {}
+
+        current_hidden = hidden_states
+        current_input_ids = input_ids
+
+        for step in range(self.num_speculative_steps):
```
brian-dellabetta: num_speculative_steps here is the same as number of tokens to predict, right?

brian-dellabetta (commented):
interesting! following this for the most part. just a couple questions

src/speculators/models/fast_mtp/config.py
```diff
@@ -0,0 +1,169 @@
...
+    :param rms_norm_eps: Epsilon for RMS normalization
+    """
+
+    speculators_model_type: Literal["mtp"] = "mtp"
+    architectures: list[str] = Field(
+        default_factory=lambda: ["FastMTPSpeculator"],
+        description="Model architectures that can load these weights",
+    )
+
+    transformer_config: PretrainedConfig = Field(
```
fynnsu: Maybe use `transformer_layer_config` for consistency with Eagle3

src/speculators/models/fast_mtp/config.py
```diff
@@ -0,0 +1,169 @@
...
+        default_factory=Qwen2Config,
+        description=(
+            "Configuration for the transformer architecture (e.g., Qwen2Config)"
+        ),
+    )
+
+    num_speculative_steps: int = Field(
```
fynnsu: I think we're currently using the proposal_methods `GreedyTokenProposalConfig` to store this.

src/speculators/models/fast_mtp/config.py
```diff
@@ -0,0 +1,169 @@
...
+        default=3,
+        ge=1,
+        le=10,
+        description="Number of future tokens to predict per forward pass",
+    )
+
+    num_nextn_predict_layers: int = Field(
```
fynnsu, line 66: Maybe use `num_hidden_layers` on the `transformer_layer_config`?

src/speculators/models/fast_mtp/config.py
```diff
@@ -0,0 +1,169 @@
...
+        default=1,
+        description="Number of MTP layers (currently only 1 is supported)",
+    )
+
+    mtp_loss_step_weights: list[float] = Field(
```
fynnsu: This is a training time setting and doesn't effect model usage. I don't think we should include this in the model config.

src/speculators/models/fast_mtp/config.py
```diff
@@ -0,0 +1,169 @@
...
+        default=[0.51, 0.31, 0.18],
+        description="Loss weights for each prediction step",
+    )
+
+    hidden_size: int = Field(
```
fynnsu: These fields are all duplicated on the `transformer_layer_config`. We should just have a single source of truth for them.

src/speculators/models/fast_mtp/config.py
```diff
@@ -0,0 +1,169 @@
...
+        default=32,
+        description="Number of attention heads",
+    )
+
+    num_key_value_heads: int = Field(
+        default=8,
+        description="Number of key-value heads for grouped query attention",
+    )
+
+    vocab_size: int = Field(
```
fynnsu: Maybe `draft_vocab_size` if we want fastmtp to support vocab mappings

src/speculators/models/fast_mtp/core.py
```diff
@@ -0,0 +1,514 @@
...
+        self.lm_head: nn.Linear | None = None
+
+        if verifier is not None and verifier_attachment_mode != "detached":
+            self.attach_verifier(verifier, mode=verifier_attachment_mode)
+
+    def attach_verifier(
+        self,
+        verifier: str | os.PathLike | PreTrainedModel,
+        mode: Literal["full", "train_only"] | None = None,
+    ):
```
fynnsu: The `attach_verifier` system is deprecated. It doesn't scale to training workflow as it requires loading the entire verifier model, which can be significantly larger than the draft model.

src/speculators/models/fast_mtp/core.py
```diff
@@ -0,0 +1,514 @@
...
+        hidden_states: torch.Tensor,
+        attention_mask: torch.Tensor | None = None,
+        position_ids: torch.Tensor | None = None,
+        labels: torch.Tensor | None = None,
+        return_dict: bool = True,
+        **kwargs,
+    ) -> dict[str, Any] | tuple:
+        """Forward pass for FastMTP multi-token prediction.
+
+        :param input_ids: Token IDs [batch, seq_len]
```
fynnsu: Can we update this so that it works with FlexAttention / packed sequences? Or if that's too big of a lift, I think we need to add data loading utils. Currently the way our training is configured, we don't load sequences in batches. 

I suppose we could might be able to get this running with batch_size=1, using the current data loaders. Although we will still need something custom to handle the fact that there's only 1 hidden state instead of 3.

src/speculators/models/fast_mtp/core.py
fynnsu: So MTP also uses train time testing?

src/speculators/models/fast_mtp/core.py
```diff
@@ -0,0 +1,514 @@
...
+        for step in range(self.num_speculative_steps):
+            current_embeddings = self.embed_tokens(current_input_ids)
+
+            mtp_output = self.mtp_layers[0](
+                hidden_states=current_hidden,
+                token_embeddings=current_embeddings,
+                attention_mask=attention_mask,
+                position_ids=position_ids,
+                **kwargs,
+            )
```
fynnsu, line 140: I'm concerned that the attention mask isn't correct here.

We need to have some kind of kv cache so that subsequent steps can attend to kv states from the original sequence. And we need the attention mask to account for that.

src/speculators/models/fast_mtp/core.py
```diff
@@ -0,0 +1,514 @@
...
+
+    def load_state_dict(self, state_dict: dict, *args, **kwargs):  # type: ignore[override]
+        """Load checkpoint with automatic format detection and key remapping."""
+        sample_keys = list(state_dict.keys())[:20]
+        detected_format = self._detect_checkpoint_format(sample_keys)
+
+        if detected_format != "speculators":
+            state_dict = self._remap_checkpoint_keys(
+                state_dict, from_format=detected_format
+            )
```
fynnsu: I don't think we should be handling multiple state dict structures here, just the native speculators one, which ideally would align to whatever the main existing format is.

Conversion logic like this should be implemented as a preprocessing step, before calling load_state_dict.

src/speculators/models/fast_mtp/mimo_config.py
```diff
@@ -0,0 +1,41 @@
...
+
+__all__ = ["MiMoConfig"]
+
+
+class MiMoConfig(Qwen2Config):
+    """Transformer config for MiMo/FastMTP models.
+
+    MiMo is a Qwen2-based architecture that adds multi-token prediction (MTP)
+    fields. This shim preserves those fields when loading/saving configs.
+    """
```
fynnsu: Why do we need this? Can we just use the `Qwen2Config`?

src/speculators/models/fast_mtp/core.py
```diff
@@ -0,0 +1,514 @@
...
+    apply MLP with residual, and final normalization.
+
+    Attribute names must match checkpoint key structure exactly. The component
+    registry ensures correct module types for different architectures.
+
+    :param config: FastMTP configuration
+    :param layer_idx: Layer index (0-indexed)
+    """
+
+    def __init__(self, config: FastMTPConfig, layer_idx: int = 0):
```
fynnsu: Could you comment on what's different between this and the `Eagle3FirstLayerMixin` logic?
Or what's different between this and a typical `Qwen2DecoderLayer` (or whatever type this is based on)?

tests/integration/models/test_fast_mtp_model.py
```diff
@@ -0,0 +1,40 @@
...
+    model = SpeculatorModel.from_pretrained(
+        model_id, verifier_attachment_mode="detached"
+    )
+    assert isinstance(model, FastMTPSpeculator)
+    config = model.config
+    assert isinstance(config, FastMTPConfig)
+
+    model.embed_tokens = nn.Embedding(config.vocab_size, config.hidden_size)
+    model.lm_head = nn.Linear(config.hidden_size, config.vocab_size, bias=False)
+    model.eval()
```
fynnsu: Why are `embed_tokens` and `lm_head` setup here? What is the intended flow for users to set these up?

fynnsu (changes requested):
Hi @rahul-tuli, thanks for the pr. I added some questions below. Can you also clarify what is the main difference between this and eagle3? I'm trying to understand if we can reduce code repetition across the two methods, since it seems like the structure is pretty similar.

PR #303, opened 2026-02-25
Update dependency bounds to be consistent with llmcompressor

dhuangnm:
Update the shared dependencies between speculators and llmcompressor and make the bounds consistent.

A run from the PR is here: https://github.com/neuralmagic/llm-compressor-testing/actions/runs/22413995068

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -249,6 +249,7 @@ select = [
 "scripts/**/*.py" = [
     "INP001", # allow implicit namespace packages in scripts
     "PTH", # os.path is acceptable in scripts
+    "T201", # print statements are acceptable in scripts
 ]
 
 "examples/**/*.py" = [
--- a/.github/workflows/development.yml
+++ b/.github/workflows/development.yml
@@ -29,21 +29,40 @@ jobs:
         uses: astral-sh/setup-uv@v6
       - name: Generate doc files
         run: uv run --with .[dev] --no-sync --isolated docs/scripts/gen_files.py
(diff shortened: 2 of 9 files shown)


pyproject.toml
```diff
@@ -37,20 +37,20 @@ keywords = [
...
     "protobuf",
     "psutil",
     "pydantic>=2.0.0",
     "pydantic-settings>=2.0.0",
     "rich",
     "safetensors",
     "setuptools",
-    "torch",
-    "tqdm",
-    "transformers<5.0.0",
```
dsikka, line 53: Can we pin transformers and torch to match llmcompressor?

dhuangnm: Yes they're updated to be consistent with llmcompressor https://github.com/vllm-project/llm-compressor/pull/2407

pyproject.toml
```diff
@@ -37,20 +37,20 @@ keywords = [
 ]
 dependencies = [
     "click",
-    "datasets>=4.0.0",
+    "datasets>=4.0.0,<=4.6.0",
     "huggingface-hub",
```
dsikka, line 41: Isn't this downloaded as a result of downloading transformers? cc @fynnsu

fynnsu: We use it directly in data gen and response regen. We will also need it even after we switch to datagen through vllm.

fynnsu (approved):

dsikka (approved):

PR #304, opened 2026-02-25
Removed data gen related files and added training examples

shanjiaz:
1. offline_data_generation.py and related scripts
2. data_gen tests & test_gen_and_train in e2e tests
3. removed data_generation folder
4. Updated examples to download huggingface datasets instead of using datagen, and run train.py directly


dhuangnm (approved):
Thanks!

PR #305, opened 2026-02-26
Extract device resolution into get_current_device utility

guan404ming:
## Why                                                                                                 
   
Duplicated accelerator device resolution logic in SingleGPUCheckpointer needed generalization per inline TODOs.                                                   

## How

- Add `get_current_device()` utility alongside existing device helpers
- Replace device blocks in `load_model_state_dict` and `load_optimizer_state_dict` with single call


guan404ming:
Thanks you all for the review!

PR #313, opened 2026-02-26
Applying fix to load norm just before lm head.

MeganEFlynn:
In the original research code, we were using transformers to get the last hidden state, and saving it to disk before computing targets.  However, that pulled the hidden state after the very last layernorm (hidden_states[-1]).  
The new data generation code using vllm does not apply that final layernorm 
(for example, here:  
https://github.com/vllm-project/vllm/blob/5e58bdc7113a2c62a9bfb71304d0d1563b0da7f3/vllm/model_executor/models/llama.py#L435
and 
github.com/vllm-project/vllm/blob/5e58bdc7113a2c62a9bfb71304d0d1563b0da7f3/vllm/model_executor/models/gpt_oss.py#L309
) 
before saving the hidden states. 

This means there is a mismatch because we do not apply the last layernorm before the lm_head when we compute the targets in core.py for eagle 3 training.  The bug should affect all models, but seems to be most salient for gpt-oss.  

This PR fixes that bug, and after fixing the acceptance rate on gpt-oss-20b on math_reasoning when trained on a very small test dataset of 20k ultrachat samples goes from 25% -> 40%.

### DIFF
--- a/src/speculators/train/checkpointer.py
+++ b/src/speculators/train/checkpointer.py
@@ -14,6 +14,8 @@ from torch.distributed.checkpoint.state_dict import (
 )
 from transformers.modeling_utils import PreTrainedModel
 
+from speculators.utils.util import get_current_device
+
 
 class BaseCheckpointer:
     """Helper class to save and load checkpoints.
@@ -126,16 +128,10 @@ class SingleGPUCheckpointer(BaseCheckpointer):
     def load_model_state_dict(
         self, model: PreTrainedModel, float_dtype: torch.dtype | None = None
     ):
-        acc = torch.accelerator.current_accelerator()
-        if acc is None:
-            device = "cuda:0"
-        else:
-            acc_type = acc.type
-            device_idx = torch.accelerator.current_device_index()
-            device = f"{acc_type}:{device_idx}"
+        device = get_current_device()
         full_state_dict = load_safetensors_state_dict(
             self.model_path(self.previous_epoch),
-            device,  # todo: generalize
+            device,
         )
         full_state_dict = convert_float_dtype(
             full_state_dict, float_dtype or model.dtype
@@ -149,17 +145,11 @@ class SingleGPUCheckpointer(BaseCheckpointer):
         optimizer: torch.optim.Optimizer,
         float_dtype: torch.dtype | None = None,
     ):
-        acc = torch.accelerator.current_accelerator()
-        if acc is None:
-            device = "cuda:0"
-        else:
-            acc_type = acc.type
-            device_idx = torch.accelerator.current_device_index()
-            device = f"{acc_type}:{device_idx}"
+        device = get_current_device()
         full_state_dict = torch.load(
             self.optimizer_path(self.previous_epoch),
             weights_only=True,
-            map_location=device,  # todo: generalize
+            map_location=device,
         )
         full_state_dict = convert_float_dtype(
             full_state_dict, float_dtype or model.dtype
--- a/src/speculators/utils/util.py
+++ b/src/speculators/utils/util.py
@@ -10,6 +10,15 @@ def is_npu_available() -> bool:
         return False
 
 
+def get_current_device() -> str:
+    """Get the current accelerator device string (e.g. 'cuda:0', 'npu:0')."""
+    acc = torch.accelerator.current_accelerator()
+    if acc is None:
+        return "cuda:0"
+    device_idx = torch.accelerator.current_device_index()
+    return f"{acc.type}:{device_idx}"
+
+
 def get_device_name(idx: int) -> str:
     acc = torch.accelerator.current_accelerator()
     if acc is None:


src/speculators/models/eagle3/core.py
```diff
@@ -156,11 +156,12 @@ class Eagle3DraftModel(SpeculatorModel):
     config_class: ClassVar[type[Eagle3SpeculatorConfig]] = Eagle3SpeculatorConfig  # type: ignore[misc]
     _keys_to_ignore_on_load_missing: ClassVar[list[str]] = [  # type: ignore[misc]
         "embed_tokens.weight",
+        "verifier_norm.weight",
         "verifier_lm_head.weight",
         "d2t",
         "t2d",
     ]
-    _keys_to_ignore_on_save: ClassVar[list[str]] = ["verifier_lm_head.weight"]  # type: ignore[misc,assignment]
+    _keys_to_ignore_on_save: ClassVar[list[str]] = ["verifier_lm_head.weight","verifier_norm.weight" ]  # type: ignore[misc,assignment]
```
fynnsu: My one concern with this, is if `verifier_norm.weight` is trainable then we could end up using affine/bias parameters which are different from those on the layernom used on the target model / different from what will be used by vllm. That could mean we again have a subtlety different target than the "true" model target. 

Maybe we need to be loading this weight from vllm?

fynnsu: Although on the flip side, I guess we might just have to accept that our target is going to be different from the real target because we reduce the verifier head to the drafter vocab size (for performance reasons) and `softmax(head[draft_vocab] * hidden) != softmax(head * hidden)[draft_vocab]`.

So there's an argument to be made that we will not have perfect targets regardless, but this change will at least mean that the range of the targets is easier for the model to learn, even if it doesn't perfectly align with the target model.

src/speculators/models/eagle3/core.py
```diff
@@ -271,7 +272,7 @@ def _setup_embeddings_and_lm_heads(
 
         # Load embedding and lm_head weights using suffix patterns (model-agnostic)
         verifier_weights = load_model_layers(
-            ["embed_tokens.weight", "lm_head.weight"],
+            ["embed_tokens.weight", "lm_head.weight", "model.norm.weight"],
```
fynnsu, line 284: Ahh okay, I see you are actually loading it here. I think we need to verify if this is the right layernorm and if this is consistently how it is stored or if it differs on various architectures.

src/speculators/models/eagle3/core.py
```diff
@@ -296,7 +300,7 @@ def _setup_embeddings_and_lm_heads(
 
         embed_tokens_sd = {"weight": embed_tokens_weight.to(default_dtype)}
         self.embed_tokens.load_state_dict(embed_tokens_sd)
-        self.embed_tokens.weight.requires_grad = False
+        # self.embed_tokens.weight.requires_grad = False
```
fynnsu: Let's make this an actual option.

src/speculators/models/eagle3/core.py
```diff
@@ -315,7 +319,13 @@ def _setup_embeddings_and_lm_heads(
         else:
             # Use full verifier vocab (no masking)
             lm_head_weight = lm_head_weight.to(dtype=default_dtype)
+        if "model.norm.weight" not in verifier_weights:
+            raise KeyError(
+                f"Could not find final norm weights in {config.name_or_path}. "
+                "Expected a key ending with 'model.norm.weight'."
+            )
 
+        verifier_norm_weight = verifier_weights["model.norm.weight"]
```
fynnsu: Maybe this should be a warning and fallback to just using newly initialized layernorm?

src/speculators/models/eagle3/core.py
```diff
@@ -373,9 +386,7 @@ def forward(
         return_loss = verifier_last_hidden_states is not None
         if return_loss:
             with torch.no_grad():
-                targets = self.verifier_lm_head(verifier_last_hidden_states)
-                # shape: [1, total_seq_len, draft_vocab_size]
-
```
fynnsu, line 378: I'd like to keep the shape comment.

src/speculators/models/eagle3/core.py
MeganEFlynn: The weights from the verifier model are being loaded below and I set the requires_grad to false.

src/speculators/models/eagle3/core.py
MeganEFlynn: I checked the llama, qwen and gpt-oss language models for the name and they're all the same. We make similar assumptions for lm_head.weight and embed_tokens.weight.  I believe this is likely to break for multi-modal models, but that is beyond the scope of this fix.

scripts/train.py
```diff
@@ -308,6 +300,12 @@ def parse_args():
         default=True,
         help="Toggle normalization before residual connections (default: True)",
     )
+    parser.add_argument(
+        "--embed-requires-grad",
+        action=argparse.BooleanOptionalAction,
+        default=True,
```
fynnsu: I think we want the default to be False here to match the current behavior.

src/speculators/models/eagle3/config.py
```diff
@@ -55,6 +55,11 @@ class Eagle3SpeculatorConfig(SpeculatorModelConfig):
         description="Layer IDs of the Eagle auxiliary hidden state layers",
     )
 
+    embed_requires_grad: bool = Field(
+        default=True,
```
fynnsu: Same here.

src/speculators/models/eagle3/core.py
```diff
@@ -492,6 +519,7 @@ def from_training_args(
             transformer_layer_config=verifier_config,
             draft_vocab_size=kwargs["draft_vocab_size"],
             norm_before_residual=kwargs["norm_before_residual"],
+            embed_requires_grad=kwargs.get("embed_requires_grad", True),
```
fynnsu: Here as well.

scripts/train.py
shanjiaz: right!

PR #316, opened 2026-02-27
[Docs] Small Updates

dsikka:
Summary
- Include VL Models
- Include links to utility scripts
- Update where response regeneration shows up

### DIFF
--- a/docs/scripts/gen_files.py
+++ b/docs/scripts/gen_files.py
@@ -130,13 +130,13 @@ def migrate_developer_docs():
             root_path=Path("CODE_OF_CONDUCT.md"),
             docs_path=Path("developer/code-of-conduct.md"),
             title="Code of Conduct",
-            weight=5,
+            weight=-10,
         ),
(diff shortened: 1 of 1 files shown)


docs/scripts/gen_files.py
```diff
@@ -145,6 +145,12 @@ def migrate_developer_docs():
             title="Train",
             weight=1,
         ),
+        ProcessFile(
+            root_path=Path("scripts/response_regeneration/README.md"),
+            docs_path=Path("response_regeneration.md"),
+            title="Response Regeneration",
+            weight=2,
+        ),
         ProcessFile(
```
fynnsu: What is the purpose for this change?

If you're trying to move the script's positioning in the sidebar you need to update the weight argument (Should be an integer, smaller/negative means earlier in the list, larger number means later).

dsikka: Oh I totally didn't see the negative numbers. Let me fix this

PR #321, opened 2026-03-02
Fix AdamW optimizer init to use parameters()

guan404ming:
## Why                                                                                                                                                          
  
named_parameters() yields (name, param) tuples, but AdamW expects bare parameter tensors.                                                                    
                                                                
## How

- Replace named_parameters() with parameters() in trainer optimizer setup


guan404ming:
Make sense to me. Let me close this one.

PR #332, opened 2026-03-05
Cleanup old implementations/features

fynnsu:
This pr aims to clean up some tech debt in the repo.

In particular:
- Independent and MLP speculators are not functional implementations and are unlikely to be implemented anytime soon due to low priority / lack of vllm support. They also don't conform with the structure of other implementations (Eagle3), which is confusing for new adopters.
- Move the base TokenProposalConfig into the proposals folder
- Remove auto importer. This logic was unnecessarily complicated, since the required modules can be imported trivially. Since this was a superclass of all models, that complexity remained in the concrete model classes. 
- Remove attach verifier related logic. Since speculators doesn't target/support generation, we remove any logic related to attaching the full verifier model. Supporting generation would require significantly better support for distributed inference to handle the full size of the verifier models and is out of scope for this library. vLLM should be used for model inference instead.
- Move the legacy `EagleSpeculator` model (which only supports conversion, not training) to a folder under `convert`. Since this model doesn't support training and uses legacy structures (like using attach_verifier logic), its inclusion in the `models/` folder is confusing.

### DIFF
--- a/.github/ISSUE_TEMPLATE/100-bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/100-bug-report.yml
@@ -1,0 +1,46 @@
+name: 🐛 Bug report
+description: Raise an issue here if you find a bug.
+title: "[Bug]: "
+labels: ["bug"]
+
+body:
+- type: markdown
+  attributes:
+    value: >
+      #### Before submitting an issue, please make sure the issue hasn't been already addressed by searching through [the existing and past issues](https://github.com/vllm-project/speculators/issues?q=is%3Aissue%20sort%3Acreated-desc).
+- type: textarea
+  attributes:
+    label: Your current environment
+    value: |
+      Please provide the following information about your environment if applicable:
+
+      - vLLM
+      - Speculators
+      - CUDA
+      - PyTorch
+      - Transformers
+      - Hardware
+      - Model
+  validations:
+    required: true
+- type: textarea
+  attributes:
+    label: 🐛 Describe the bug
+    description: |
+      Please provide a clear and concise description of what the bug is.
+
+      If relevant, add a minimal example so that we can reproduce the error by running the code. It is very important for the snippet to be as succinct (minimal) as possible, so please take time to trim down any irrelevant code to help us debug efficiently. We are going to copy-paste your code and we expect to get the same result as you did: avoid any external data, and include the relevant imports, etc.
+
+      Please also paste or describe the results you observe instead of the expected results. If you observe an error, please paste the error message including the **full** traceback of the exception. It may be relevant to wrap error messages in ```` ```triple quotes blocks``` ````.
+    placeholder: |
+      A clear and concise description of what the bug is.
+
+      ```python
+      # Sample code to reproduce the problem
+      ```
+
+      ```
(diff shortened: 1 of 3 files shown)


src/speculators/utils/auto_importer.py
```diff
@@ -1,100 +0,0 @@
-"""
```
rahul-tuli, line 1: 💯

src/speculators/convert/eagle/eagle_legacy_model.py
rahul-tuli: nit: missed change (we can also remove this docstring example entirely)
```suggestion
        from speculators.convert.eagle.eagle_legacy_model import EagleSpeculatorConfig
```

rahul-tuli: nit: Update import in docstring or remove example entirely!
```suggestion
        from speculators.convert.eagle.eagle_legacy_model import EagleSpeculator, EagleSpeculatorConfig
```

tests/unit/test_config.py
```diff
@@ -275,8 +275,6 @@ def test_speculator_model_config_auto_registry():
     class_names = [cls.__name__ for cls in classes]
     assert len(class_names) > 0
     assert "EagleSpeculatorConfig" in class_names
```
rahul-tuli: Should be `Eagle3SpeculatorConfig` since `EagleSpeculatorConfig` lives in
`src/speculators/convert/eagle/eagle_legacy_model.py` now and is not piped up via `__init__.py`?


```suggestion
    assert "Eagle3SpeculatorConfig" in class_names
```

src/speculators/convert/eagle/eagle_converter.py
```diff
@@ -9,13 +9,16 @@
 from transformers import LlamaConfig, PretrainedConfig
```
rahul-tuli, line 9: I found another bug, commenting here:
(importing `EagleConverter` from `speculators.convert.eagle` works but `Eagle3Converter` doesn't)
```python
In [7]: from speculators.convert.eagle import Eagle3Converter
------------------------------------------------------------------------
ImportError                            Traceback (most recent call last)
Cell In[7], line 1
----> 1 from speculators.convert.eagle import Eagle3Converter

ImportError: cannot import name 'Eagle3Converter' from 'speculators.convert.eagle' (/home/rahul-tuli/speculators/src/speculators/convert/eagle/__init__.py)

In [8]: from speculators.convert.eagle import EagleConverter
```

I think we need to pipe up the `Eagle3Converter` in `src/speculators/convert/eagle/__init__.py` could you include that change in this diff?

src/speculators/convert/eagle/eagle_legacy_model.py
fynnsu: I'm just going to remove the example usage from the docstring.

tests/unit/test_config.py
fynnsu: Fixing. A little bit surprising that this doesn't fail in CI.

It does fail locally when I just run that one test. I suspect that what's happening here is pytest isn't isolating the different tests sufficiently, and some of the other legacy eagle tests are adding this to the registry before this test runs. Not sure if we have a great solution for preventing that in general.

rahul-tuli (approved):

PR #333, opened 2026-03-05
Support `EagleDraftModel.from_pretrained` for finetuning

fynnsu:
## Prereq

Depends on #332, which should be merged first. 

## Purpose
We'd like to support finetuning existing Eagle3 models. This pr adds the `--from-pretrained` option, which loads a pretrained model from HF / local path. 

#### Fresh model setup, i.e. `from_training_args` pathway
1. model = Eagle3DraftModel.__init__
    - Initialize all modules/parameters/buffers
    - Use `torch.zeros` (w/ correct shape + dtype) as placeholders for t2d/d2t
    - Use `torch.nan` (w/ correct shape + dtype) as placeholders for lm_head, embed_tokens, and verifier_lm_head
2. model.load_vocab_mappings(t2d, d2t)
    - t2d/d2t can be `None`, which will cause an early return
    - Verify shapes match and then load t2d/d2t
3. model.load_verifier_weights()
    - Loads embed_tokens, lm_head (into both lm_head and verifier_lm_head), verifier_norm from verifier model
    - Checks if embed_tokens and lm_head are still `NaN` (from init) before loading. Otherwise they are skipped
    - verifier_lm_head is always loaded
    - verifier_norm is skipped if the weight isn't found (with a warning). Note we don't use NaN init for this module
 
Continues below

#### Finetuning setup, i.e. `from_pretrained` pathway
1. model = Eagle3DraftModel.__init__ call internally by `PreTrainedModel.from_pretrained` under meta device context
2. `PreTrainedModel.from_pretrained` also loads model weights for us
3. model.load_vocab_mappings(t2d, d2t) called, same as above
4. model.load_verifier_weights() called
    - lm_head and embed_token loading should be skipped because the values have already been set by `PreTrainedModel.from_pretrained`
    - verifier_lm_head and verifier_norm loaded (overriding existing values but should be the same)
 
Continues below

#### Joint (Fresh + Finetuning) next steps (in `Trainer.setup_model`)
```
If distributed:
    cache state dict on rank0
    Apply `fully_shard` to model layers and model (don't move to meta device, don't re-init weights)
    
    if checkpoint exists:
        load checkpoint (distributed)
    else:
        broadcast cached state dict from rank0 to all ranks
else:
    move model to local device
    if checkpoint exists:
        load checkpoint (single device)
```

## Implementation
This uses the `.from_pretrained` method from the base `PreTrainedModel` class to resolve local vs remote model and handle downloading checkpoints.

As part of this pr, I also clean up the model initialize process, which previously also including loading verifier weights and t2d/d2t tensors. These will now be loaded after the init function. This was required to support `.from_pretrained` as the model init is run under a meta device context.

This means the init instead sets up placeholder buffers for verifier parameters/vocab mapping. These are being set intentionally in a way that makes it easy to confirm they are overwritten when training starts (e.g. by initializing some values to `NaN` so that failing to overwrite them will result in immediate NaN outputs from the model).

I also updated the fully shard handling to ensure values are set on all ranks correctly when using distributed training. 

## Testing

Added comprehensive test coverage for loading pathway combinations (e.g. loading from checkpoint, pretrained, fresh init, w/ and w/o vocab mappings, single gpu and distributed, etc.). 

Note some tests require single or multi-gpu. These are skipped if requirements are not met. It would be good to ensure these are being run correctly at regular intervals.

### DIFF
--- a/tests/e2e/vllm/test_finetuning_sanity.py
+++ b/tests/e2e/vllm/test_finetuning_sanity.py
@@ -1,0 +1,155 @@
+"""E2E test: verify finetuning modifies weights but keeps them close (bounded rel L1).
+
+Uses relative L1 distance: ||a-b||_1 / (||b||_1 + eps) per tensor.
+All trainable tensors must have rel_l1 <= REL_L1_MAX;
+at least a few must have rel_l1 > REL_L1_MIN to ensure they changed.
+
+To see all logs (per-tensor distances, frozen checks, etc.), run:
+  pytest tests/e2e/vllm/test_finetuning_weight_sanity.py -s --log-cli-level=INFO
+Without -s, pytest captures output and you won't see logger output until failure.
+"""
+
+import logging
+import subprocess
+import sys
+from pathlib import Path
+
+import pytest
+import torch
+from huggingface_hub import snapshot_download
+from safetensors.torch import load_file
+
+from speculators import Eagle3DraftModel
+
+logger = logging.getLogger(__name__)
+
+
+@pytest.mark.e2e
+@pytest.mark.slow
+def test_finetuning_weight_sanity(tmp_path: Path):
+    """Verify finetuning changes weights but keeps rel L1 distance bounded (low LR)."""
+    # Ensure logs are visible when running with -s (no capture)
+    logging.basicConfig(
+        level=logging.INFO,
+        format="%(levelname)s %(name)s %(message)s",
+        stream=sys.stderr,
+        force=True,
+    )
+
+    # Learning rate used for training (shared for log and CLI).
+    FROZEN_KEY_PATTERNS = ("d2t", "embed_tokens.weight", "t2d")
+    # Relative distance thresholds: all trainable tensors must have rel_l1 <= REL_L1_MAX
+    # at least MIN_CHANGED tensors must have rel_l1 > REL_L1_MIN
+    # to ensure weights actually changed.
+    LR = "1e-5"
+    REL_L1_MAX = 0.05
+    REL_L1_MIN = 1e-4
+    MIN_CHANGED = 3
+    EPS = 1e-12
+    PRETRAINED = "RedHatAI/Llama-3.1-8B-Instruct-speculator.eagle3"
+    DATASET = "nm-testing/sharegpt_llama3_8b_hidden_states"
+
+    # Get initial state dict
+    model = Eagle3DraftModel.from_pretrained(PRETRAINED)
+    initial_sd = model.state_dict()
+    # Remove verifier weights which aren't saved in checkpoints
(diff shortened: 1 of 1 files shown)


src/speculators/models/eagle3/core.py
```diff
@@ -167,98 +167,132 @@ class Eagle3DraftModel(SpeculatorModel):
...
         super().__init__(config=config)
+
         self.hidden_size = config.transformer_layer_config.hidden_size
         self.draft_vocab_size = config.draft_vocab_size
+        self.verifier_vocab_size = config.transformer_layer_config.vocab_size
+
+        tl_config = self.config.transformer_layer_config
+        self._model_definitions = model_classes[tl_config.model_type]
+
+        # VOCAB MAPPINGS
```
shanjiaz, line 187: Thanks for cleaning up the vocab mapping option logic lol I didn't like the old implementation.

src/speculators/models/eagle3/core.py
```diff
@@ -167,98 +167,132 @@ class Eagle3DraftModel(SpeculatorModel):
...
-        # Add first layer
-        layers = [
-            self._model_definitions.first_layer_class(
-                transformer_layer_config,
-                layer_idx=0,
-                norm_before_residual=norm_before_residual,
+        # Initialize weights to nan
+        # This ensures it will be easy to detect if the weights are never
+        # loaded from the verifier model
+        torch.nn.init.constant_(self.lm_head.weight, torch.nan)
```
shanjiaz, line 259: I understand the intent. Maybe we could consider using meta tensors here?

scripts/train.py
```diff
@@ -298,12 +306,6 @@ def parse_args():
         default=True,
         help="Toggle normalization before residual connections (default: True)",
     )
-    parser.add_argument(
```
shanjiaz, line 301: Does this mean we now only derive `embed_requires_grad` from config?

src/speculators/models/eagle3/core.py
fynnsu: I think this is better than meta tensors, and I don't think the cost (runtime) is particularly high.

Part of the reason I think this is better is I think having a model with some concrete weights and others still set to meta could lead to unexpected behavior / is confusing.

scripts/train.py
fynnsu: Oh that's my bad. Issue with how I rebased, will add it back.

src/speculators/models/eagle3/core.py
shanjiaz: Sounds good!

tmp.pt
brian-dellabetta, line 1: intended?

fynnsu: Nope, thanks for catching that!

Issue #307, opened 2026-02-26
"Save best" and other improvements to model checkpointing


fynnsu:
Our current system for saving checkpoints is pretty basic and while it works it could be improved. Currently, we just save (in the checkpoint folder) a new folder for each epoch. 

E.g.
```
../path/
    0/          # epoch number
        config.json
        model.safetensors
        optimizer_state_dict.pt
        scheduler_state_dict.pt (optional)
    1/
        config.json
        model.safetensors
        optimizer_state_dict.pt
        scheduler_state_dict.pt (optional)
    ...
```

Then if continuing training we just look for folders with numbered names in the save dir and load the highest numbered one.

It would be good if we could improve this system. One thing that comes to mind is having a "save_best" option where we only save an epoch if it has the lowest validation loss. 

Additionally, the optimizer state dict and scheduler state dict are useful if resuming training, but not everyone is interested in doing so and they may need to be manually removed before transferring the checkpoint (e.g. uploading to huggingface). It might be better to only optionally save them when the user requests it.

fynnsu:
Hi @surojitiitg, thanks for taking up the issue! And yes, I intended this issue to be a little bit open ended, so if you have suggestions I'm happy to hear them!

fynnsu:
@NJX-njx, this is already being worked on by @surojitiitg. Please don't work on issues that have already been assigned.

PR #334, opened 2026-03-05
"Save best" and other improvements to model checkpointing, Issue #307

surojitiitg:
## Summary

This PR improves checkpoint saving behavior during training by adding configurable checkpoint frequency and stable best-checkpoint tracking.

## Changes

- Files modified: `src/speculators/train/trainer.py`, `src/speculators/train/checkpointer.py`, `scripts/train.py`
- Files added: `tests/unit/train/test_checkpoint.py`
- Added `--checkpoint-freq` to control how often numbered checkpoints are saved
- Added `--save-best` to maintain a stable `checkpoint_best/` path 
- Implemented symlink support so `checkpoint_best/` points to the best saved checkpoint directory
- Kept default behavior backward compatible by using `checkpoint_freq=1`
- Updated validation flow in `src/speculators/train/trainer.py` so that `val_epoch(...)` returns the computed `val_metrics` dictionary to the training loop
- Best-checkpoint selection uses the key `loss_epoch` from the dictionary returned by `self.val_epoch(epoch)` as the monitored metric. The implementation is carried out in `src/speculators/train/trainer.py`
- Added `tests/unit/train/test_checkpoint.py`, which contains unit tests for:
  - ignoring `checkpoint_best` during resume checkpoint discovery
  - creating and updating the best-checkpoint symlink
  - checkpoint selection behavior during training

## Behavior

- If `--checkpoint-freq` and `--save-best` are not provided, the current checkpointing behavior is preserved (backward compatible defaults).
- `checkpoint_freq=1` preserves the current behavior of saving checkpoints every epoch
- When `save_best=True`, `checkpoint_best/` points to the checkpoint with the lowest validation loss among the saved checkpoints
- Best-checkpoint tracking is based on `val_metrics["loss_epoch"]` returned from the function `val_epoch(...)`
- Resume behavior still loads the latest numbered checkpoint and ignores `checkpoint_best/`

## Notes

- The first epoch checkpoint is intentionally saved as part of the current behavior. This provides a a guaranteed recovery point and early verification artifact.
- `checkpoint_best/` is implemented as a symlink to avoid duplicating checkpoint files

## Testing

Tested with:
- Unit tests for checkpoint selection and symlink behavior. The test file is located in `tests/unit/train/test_checkpoint.py` 
- Did testing to observe the file structure of the checkpointing via `scripts/train.py` using:

```bash
CUDA_VISIBLE_DEVICES=1 python scripts/train.py \
  --data-path output \
  --save-path checkpoints_smoke \
  --log-dir logs_smoke \
  --epochs 2 \
  --lr 3e-5 \
  --total-seq-len 256 \
  --data-format-version 1 \
  --num-layers 1 \
  --d2t-path vocab_mapping/d2t.npy \
  --t2d-path vocab_mapping/t2d.npy \
  --num-workers 1 \
  --prefetch-factor 2 \
  --checkpoint-freq 2 \
  --save-best
```
- manual verification of checkpoint directory structure and checkpoint_best/ symlink target
- confirmed backward compatibility by running without `--checkpoint-freq` and `--save-best` (current behavior maintained)

## Style/Lint Notes

- On running `make style` the program suggests to make modification at `src/speculators/train/trainer.py:204:13`. Since this is unrelated to my modifications, I have not updated it to satisfy `make style`.
- `make style` also suggests to make import blocks sorted in `src/speculators/train/checkpointer.py:1:1`. However, since the suggestion is also present for the current version of the repo, I have not made any changes to the arrangement of the import block.

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -37,3 +37,29 @@ pull_request_rules:
       label:
         remove:
           - needs-rebase
+  
+  - name: add quality-failed label
+    conditions:
+      - label != stale
+      - check-failure ~= ^quality-checks
+      - -closed
+    actions:
+      label:
+        add:
+          - quality-failed
+      comment:
+        message: |
+          The quality checks have failed. Please run `make style` and `make quality` under
+          the root directory to address the lint failures. You will need to install the
+          dev optional install to get the required linting packages:
+          https://github.com/vllm-project/speculators/blob/main/CONTRIBUTING.md
+
+  - name: remove quality-failed label
+    conditions:
+      - label != stale
+      - -check-failure = quality-check
+      - -closed
+    actions:
+      label:
+        remove:
+          - quality-failed


src/speculators/train/checkpointer.py
```diff
@@ -106,6 +106,21 @@ def scheduler_path(self, epoch: int):
...
+        best_path = self.best_path()
+        target = Path(str(epoch))  # relative symlink inside checkpoint root
+
+        if best_path.is_symlink() or best_path.exists():
+            if best_path.is_dir() and not best_path.is_symlink():
+                shutil.rmtree(best_path)
+            else:
+                best_path.unlink()
+
+        best_path.symlink_to(target, target_is_directory=True)
```
rahul-tuli, line 148: Should this be wrapped in a rank-0 check for the distributed case? cc: @fynnsu

scripts/train.py
```diff
@@ -319,6 +323,20 @@ def parse_args():
         default=0.05,
         help="Standard deviation for noise augmentation",
     )
+    # Checkpoint Parameters
+    parser.add_argument(
+        "--checkpoint-freq",
+        type=int,
+        default=1,
+        help="Save a checkpoint every N epochs.",
```
rahul-tuli, line 348: Could we add validation that this is `>=1` and raise an error if not

src/speculators/train/trainer.py
```diff
@@ -83,6 +85,8 @@ def setup_trainer(self):
             root_logger.info("No previous checkpoint found. Starting from scratch.")
             self.current_epoch = 0
         self.global_step = 0
+        self.best_val_loss = float("inf")
+        self.best_epoch = None
```
rahul-tuli: When someone resumes training, the first saved checkpoint will become "best" regardless of whether prior checkpoints were better? We should read the existing `checkpoint_best` symlink and load its saved `val_metrics` to initialize `best_val_loss`

rahul-tuli: Also, If `--save-best` is passed but no `val_loader` is configured, `val_metrics` will always be `None` and `checkpoint_best` will never be updated — with no warning to the user. Could we log a warning here if save_best=True and val_loader is None.

src/speculators/train/trainer.py
```diff
@@ -246,20 +251,41 @@ def run_training(self):
...
+                )
+
+                if (
+                        self.config.save_best
+                        and val_metrics is not None
+                        and "loss_epoch" in val_metrics
+                        and val_metrics["loss_epoch"] < self.best_val_loss
+                ):
+                    self.best_val_loss = val_metrics["loss_epoch"]
+                    self.best_epoch = epoch
```
rahul-tuli: Is `best_epoch` used anywhere?

.gitignore
```diff
@@ -174,3 +174,10 @@ cython_debug/
 
 # Visual Studio Code
 .vscode/
+# IDE/editor
+.idea/
+*.iml
+
+# OS junk
+
+
```
rahul-tuli: Could we remove this from current PR?

src/speculators/train/checkpointer.py
fynnsu: Yes, either here or when calling it in the trainer.

src/speculators/train/trainer.py
fynnsu: Yeah, I do think this is an issue

> the first saved checkpoint will become "best" regardless of whether prior checkpoints were better

But we don't actually have `val_metrics` saved anywhere to load. We could add them to the checkpoints in theory, but I wonder if there's a better solution.

One thing I've considered in the past is whether we should do a val run before the first epoch to get a starting baseline, at least when resuming training. That would be a way for us to get a starting validation value. Although we would have to load the best checkpoint and not the latest for that to work. 

We are also getting into kind of obscure edge cases of the checkpoint loading code, so I think it's okay if the solution is not perfect so long as it is reasonable. e.g. load the latest checkpoint, run eval on it and call it the best checkpoint (even if it's not), would probably be okay.

src/speculators/train/trainer.py
```diff
@@ -246,20 +251,41 @@ def run_training(self):
...
-            )
-            self.save_checkpoint(epoch)
-            root_logger.info(
-                f"Finished saving checkpoint to {self.checkpointer.path / str(epoch)}"
-            )
+            should_save = False
+            if epoch == 0 or (epoch + 1) % self.config.checkpoint_freq == 0:
+                should_save = True
+
+            if should_save:
```
shanjiaz: nit: Can we remove the above logic and just say `if epoch == 0 or (epoch + 1) % self.config.checkpoint_freq == 0:` here. Reads a little redundant.

scripts/train.py
surojitiitg: Yeah, that's a good point. I can add a validation.

src/speculators/train/trainer.py
surojitiitg: I have not used `self.best_epoch` anywhere else. I had kept it for validation and doing sanity checks. 
After we conclude the discussion we are having on `save_best`, if self.best_epoch is still not used, I can remove it.

.gitignore
surojitiitg: Sure, my bad. I'll remove these.

src/speculators/train/trainer.py
surojitiitg: Good catch. Yes, I'll incorporate the suggestion.

src/speculators/train/trainer.py
surojitiitg: @rahul-tuli, thanks for catching this edge case. In slack @fynnsu and I discussed what would the checkpoint folder will be depending on the value of`--save-best`.

The following is the structure we aligned on:

If `save_best=False` (and save_freq = 2), then the structure is:
```
checkpoint_dir/
    1/
        model.safetensors
        ...
    3/
        model.safetensors
        ...
    ...
    best_checkpoint/  # Symlink to one of the folders above (e.g. folder 3/), which has the lowest validation loss
        model.safetensors
        ...
```
If `save_best=True` (and save_freq = 2), then the structure is:

```
checkpoint_dir/
    best_checkpoint/ # Symlink to only one folder, denoting the lowest validation loss. In this example, it's 3
        model.safetensors
        ...
    3/
        model.safetensors
        ...
```

Thus, irrespective of the value of `save_best`, we will have the `best_checkpoint` folder.

Hence, to address the issue raised in the thread by @rahul-tuli , I propose we initialize `self.best_val_loss` from the existing `checkpoint_best` symlink when resuming training.
The proposed behavior is:

## Proposed Behavior 

- The trainer checks whether a `checkpoint_best` symlink exists in the checkpoint directory.

- If the symlink exists, the trainer reads the epoch number that the symlink points to.

- The model weights for that epoch are temporarily loaded.

- A validation pass is run for that checkpoint to compute the validation loss.

- The resulting value is assigned to `self.best_val_loss`.

- The trainer then reloads the latest checkpoint weights and resumes training as usual.

- This ensures that the best validation loss is initialized consistently with the previously identified best checkpoint.

In the even `checkpoint_best` is missing (backward compatibility), the trainer falls back to the current behavior. Specifically:

- `self.best_val_loss` remains initialized to inf.

- The next checkpoint that improves validation loss will become the new `checkpoint_best`.

- This preserves backward compatibility and avoids requiring any changes to the checkpoint format.

@fynnsu, @rahul-tuli does this make sense or would you recommend something else?

fynnsu: Just one note, otherwise this works for me. 

If `save_best=True` we should be ignoring the save_freq entirely. Anytime the model val_loss is lower than the current best, we save the checkpoint. Maybe that's already what you intended but just wanted to be clear because you mention save_freq in the `save_best=True` case above, and its value should be irrelevant.

src/speculators/train/checkpointer.py
surojitiitg: @rahul-tuli , @fynnsu I have updates to account for distributed training. In addition, made a check with running the following script and it did execute well at my end.

``` 
CUDA_VISIBLE_DEVICES=0,1 OMP_NUM_THREADS=1 torchrun --standalone --nnodes=1 --nproc_per_node=2 scripts/train.py \
  --data-path output \
  --save-path checkpoints_smoke_multi \
  --log-dir logs_smoke_multi \
  --epochs 1 \
  --lr 3e-5 \
  --total-seq-len 128 \
  --data-format-version 1 \
  --num-layers 1 \
  --d2t-path vocab_mapping/d2t.npy \
  --t2d-path vocab_mapping/t2d.npy \
  --num-workers 1 \
  --prefetch-factor 2 \
  --no-resume-from-checkpoint

```

src/speculators/train/trainer.py
```diff
@@ -234,8 +310,9 @@ def val_epoch(self, epoch: int):
             )
 
             if self.is_distributed:
-                for v in metrics.values():
-                    dist.reduce(v, dst=0, op=dist.ReduceOp.AVG)
+                for m in metrics.values():
+                    dist.all_reduce(m, op=dist.ReduceOp.SUM)
+                    m.div_(dist.get_world_size())
```
fynnsu: Why `ReduceOp.SUM` + `div_` instead of the existing `ReduceOp.AVG` ?

src/speculators/train/trainer.py
```diff
@@ -126,6 +133,75 @@ def setup_model(self):
...
+        if self.is_distributed:
+            device = (
+                torch.device("cuda", self.local_rank)
+                if torch.cuda.is_available()
+                else torch.device("cpu")
+            )
+            t = torch.tensor(
+                [val_loss if val_loss is not None else float("nan")], device=device)
+            dist.broadcast(t, src=0)
+            val_loss = float(t.item())
```
fynnsu: We don't need this anymore, since we are using `all_reduce` in `val_epoch`. Just remove the `and ((not self.is_distributed) or dist.get_rank() == 0)` from the previous if statement.

src/speculators/train/trainer.py
```diff
@@ -126,6 +133,75 @@ def setup_model(self):
...
+
+            if math.isnan(val_loss):  # NaN is the only float where x != x
+                val_loss = None
```
fynnsu: A `nan` val_loss is likely unrecoverable. It might be better to raise an error here.

src/speculators/train/trainer.py
```diff
@@ -260,20 +338,57 @@ def run_training(self):
...
+                    root_logger.info(
+                        f"Updating new best checkpoint symlink at epoch {epoch} "
+                        f"(loss_epoch={self.best_val_loss:.6f})")
+                    self.checkpointer.update_best_symlink(epoch)
+                    root_logger.info(
+                        f"Updated checkpoint_best -> {epoch} "
+                        f"(loss_epoch={self.best_val_loss:.6f})"
+                    )
+            if self.is_distributed:
+                dist.barrier()
```
fynnsu, line 379: Could we move all of this logic into `Trainer.save_checkpoint`. You can also rename that method to `maybe_save_checkpoint` to reflect the new behavior. 

We want to keep the logic in `run_training` as simple as possible.

surojitiitg: Sure. Updated based on the conversation we had on slack.

PR #346, opened 2026-03-14
[Bugfix] Include thinking content in loss_mask for Qwen3

benchislett:
## Purpose

Thinking content in the Qwen3 8B chat template breaks the loss mask setup

## Description

By removing the `<think></think>` pattern that comes from running the example prompt into the tokenizer, we conversations with thinking to be included in the trainable assistant completion content

## Tests

Test cases included to prevent regressions for Qwen/Qwen3-8B.

I have filled in:

- [ ] ~(Optional) The necessary documentation update.~

### DIFF
--- a/.github/workflows/link-checks.yml
+++ b/.github/workflows/link-checks.yml
@@ -13,7 +13,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  UV_TORCH_BACKEND: "auto"
+  UV_TORCH_BACKEND: "cpu"
 
 jobs:
   link-checks:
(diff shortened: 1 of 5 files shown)


tests/datagen/test_preprocessing.py
```diff
@@ -612,6 +612,123 @@ def test_preprocess_batch_with_turn_dropout():
...
+    """
+    tokenizer = AutoTokenizer.from_pretrained(
+        "Qwen/Qwen3-8B", trust_remote_code=True
+    )
+    pattern = _detect_assistant_pattern(tokenizer)
+
+    # Build formatted text using the real chat template
+    conv = [
+        {"role": "user", "content": "What is the capital of France?"},
+        {"role": "assistant", "content": "Paris is the capital."},
```
fynnsu, line 690: Could you instead use `"reasoning_content"` here + `enable_thinking=True` in the tokenizer? That way we can ensure the reasoning content matches the tokenizer template.

Also same in the test above, except there we would then expect to only get the thinking tokens on the last response (which I believe is the correct behaviour).

Issue #335, opened 2026-03-05
[RFC]: Use vLLM's `extract_hidden_states` system (and enable online training)


fynnsu:
### Motivation.

vLLM recently merged https://github.com/vllm-project/vllm/pull/33736 which adds direct support in vLLM for generating model hidden states. This allows us to deprecate the `VllmHiddenStatesGenerator` and replace this system with direct calls to vLLM. This should simplify the user/test experience for speculators, as users will no longer need to use separate environments for different parts of the speculators library.

In addition, moving the hidden states generation into a separate vLLM process allows us to make use of api endpoints, which should enable on-demand hidden states generation, as required by online training. 

This RFC explores the changes required in speculators to support the `extract_hidden_states` system, and online training.

## Extract hidden states (brief overview)
The vllm pr above added three components. A dummy eagle3 style model that receives hidden states and inserts them into its KV cache, an exmaple KV Connector specifically designed for extracting hidden states, and a dummy spec decode method for facilitating the pathway. 

If the model is setup correctly (i.e. configured with `extract_hidden_states` spec decode method and `ExampleHiddenStatesConnector` kv connector), then vllm requests (to python api or openai endpoint) will also receive a filepath in the response, pointing to a file containing the hidden states and token ids for the request prompt.

There are also plans to add more performant hidden states connectors that don't write to disk, but when they are added the output is likely to contain the same two tensors.

### Dataset-level requirements
1. Token frequency statistics: 
    **Currently:** produced during preprocessing and cached to disk. 
    **Used to:** produce vocab mappings (i.e. which tokens are most frequently used and therefore should be kept). 
    **Note:** this can be collected using the full dataset or just a (reasonably large) sample, although full dataset is better.
2. Per sample tokenized sequence lengths: 
    **Currently:** produced during preprocessing and cached to disk
    **Used to:** determine efficient batch packings
    **Note:** We have a fallback that approximates the sequence lengths based on the disk size of hidden states files if this information doesn't exist. Approximate seq length is sufficient for training

### Sample-level requirements
1. Hidden states (for eagle3 input layers and last verifier layer)
    **Currently:** produced during data gen and stored in `data_{i}.pt` file
2. Token ids
    **Currently:** produced during data gen and stored in `data_{i}.pt` file
3. Assistant mask
[...]

PR #353, opened 2026-03-19
Update data generation to use vLLM extract_hidden_states

fynnsu:

## Purpose

This pr adds support for vLLM's extract_hidden_states system introduced in https://github.com/vllm-project/vllm/pull/33736. It also includes a major rework of our data generation system and the way our data is formatted on disk.

These changes enable online training (also introduced by this pr), offline training, and hybrid methods. 

### Data format
As discussed in #335, some changes to the data format are required to enable online training. The previous system generated data files containing aligned token ids, loss masks, and hidden states and saved them all to disk. The new system instead separates the hidden states from the token ids and loss masks. 

I initially proposed in the RFC that the data files could contain a file path pointing to the cached hidden states (if they exist). Although this works well, it requires updating the files as the data is generated. We also need 1 file per data sample (2 in the offline case) stored on disk, which is a substantial number of files. 

To combat this, I looked into different data formats for storing the data and realized that if we simply store the hidden states files in a file with a name linked to the data's index (e.g. `hs_{i}.safetensors`) then we don't need to explicitly add the file path. And since we're not updating the files manually, there is no need for simultaneous read/writes, and we can store the entire preprocessed data (just the token ids + loss masks) in a single arrow dataset, which is already been produced as an intermediate step during generation. That is what the new system uses.

### Training flow
Abridged version. See `examples/ONLINE_TRAINING.md` for more details.

1. Prepare data
```bash
python scripts/prepare_data.py --model Qwen/Qwen3-8B --data sharegpt --output ./output
```
2. Launch vLLM
```bash
CUDA_VISIBLE_DEVICES=0,1,2,3 python scripts/launch_vllm.py Qwen/Qwen3-8B -- --data-parallel-size 4 --port 8000
```
3. (Optional) Run offline hidden states generation
```bash
python scripts/data_generation_offline2.py --endpoint http://localhost:8000/v1 --preprocessed-data ./output --validate-outputs
```
4. Run training
```bash
CUDA_VISIBLE_DEVICES=4,5,6,7 torchrun --standalone --nproc_per_node 4 scripts/train.py --verifier-name-or-path Qwen/Qwen3-8B --data-path ./output --vllm-endpoint http://localhost:8000/v1 --save-path ./output/checkpoint --draft-model-size 32000
```

### Online training enablement
The following options were added to `scripts/train.py` to enable online/hybrid generation

- `--hidden-states-path`. Defaults to `args.data_path / "hidden_states"`. Where to look for and/or cache hidden states.
- `--vllm-endpoint`. Defaults to `http://localhost:8000/v1`. The location for the running configured vLLM instance for generating new states. Only required if `on_missing="generate"` and hidden states are missing.
- `--on-missing`. Choices `["generate", "skip", "warn", "raise"]`. Defaults to `generate`. What the data loader should do if it doesn't find a hidden state for a sample.
- `--on-generate` Choices ["cache", "delete"]. Defaults to `delete`. What should be done with newly generated hidden states after loading. 
- `--legacy-data`. Stores True. Use the old data system with individual files for each sample. Will be deprecated and removed.

The following options were added to `scripts/train.py` to enable auto vocab mapping generation
- `--token-freq-path`. Defaults to `args.data_path / "token_freq.pt"` Token frequency to load. 
- `--draft-vocab-size`, Defaults to `None`. If not provided, then we will use the full verifier vocab (same as existing behavior) but also output a warning. 

## Related Issue

#335 

## Tests

I've run online training with both `--data-parallel-size` and `tensor-parallel-size` set for 1 epoch (similar to instruction provided in example readme). We will need to add / update e2e tests for this new pipeline (but we may delay this for a future pr).

I have filled in:

### DIFF
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -250,6 +250,11 @@ def main(args: argparse.Namespace):
     # Setup dataloaders
     noise_transform = AddUniformNoise(std=args.noise_std)
     if args.legacy_data:
+        warnings.warn(
+            "Using '--legacy-data' is deprecated and will be removed soon.",
+            category=DeprecationWarning,
+            stacklevel=2,
+        )
         train_files, val_files = split_files(args.data_path, ratio=0.9)
         train_dataset: BaseEagle3Dataset = Eagle3SampleFileDataset(
             file_list=train_files, max_len=args.total_seq_len, transform=noise_transform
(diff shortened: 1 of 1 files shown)


scripts/data_generation_offline2.py
```diff
@@ -0,0 +1,321 @@
+#!/usr/bin/env python3
```
shanjiaz, line 1: Would be cool to remove the v2 since vllm release is out.

src/speculators/data_generation/preprocessing.py
```diff
@@ -225,7 +226,7 @@ def _create_loss_mask_from_offsets(
     assistant_pattern: str | Pattern[str],
 ) -> torch.Tensor:
     """Create loss mask by finding assistant response spans in formatted text."""
-    loss_mask = torch.zeros(len(offsets), dtype=torch.long)
+    loss_mask = torch.zeros(len(offsets), dtype=torch.bool)
```
shanjiaz, line 229: nice!

fynnsu: Yeah this single-handedly reduced disk space of the dataset by more than half

src/speculators/data_generation/preprocessing.py
```diff
@@ -472,40 +469,47 @@ def load_and_preprocess_dataset(
...
 
-    raw_dataset = load_raw_dataset(
-        train_data_path, num_proc=build_dataset_num_proc, cache_dir=cache_dir
-    )
-    raw_dataset = raw_dataset.shuffle(seed=seed)
+    processed_datasets = []
+    for train_data_path in train_data_paths:
+        log.subsection(f"Processing {train_data_path}")
+        raw_dataset = load_raw_dataset(train_data_path, num_proc=build_dataset_num_proc)
+        # raw_dataset = raw_dataset.shuffle(seed=seed)
```
shanjiaz: nit: remove this

src/speculators/train/data.py
```diff
@@ -113,15 +143,228 @@ def standardize_data_v1(data: dict[str, Any]) -> dict[str, Any]:
...
+            if hidden_states_path is None
+            else Path(hidden_states_path)
+        )
+        self.vllm_endpoint = vllm_endpoint
+        self.on_missing = on_missing
+        self.on_generate = on_generate
+        self.client: openai.OpenAI | None = None
+        self.model = model
+
+        # Delay super init so that `_compute_approx_lengths` has required data
```
shanjiaz, line 255: just curious, do we still need `_compute_approx_lengths` now we're saving them explicitly in preprocessing?

fynnsu: I was mostly keeping this for backwards compat with the old format.

scripts/train.py
```diff
@@ -267,6 +343,15 @@ def parse_args():
...
+    parser.add_argument("--hidden-states-path", type=str, default=None)
+    parser.add_argument("--vllm-endpoint", type=str, default="http://localhost:8000/v1")
+    parser.add_argument(
+        "--on-missing",
+        choices=["generate", "skip", "warn", "raise"],
+        default="generate",
+    )
+    parser.add_argument("--on-generate", choices=["cache", "delete"], default="delete")
+    parser.add_argument("--legacy-data", action="store_true")
     parser.add_argument("--save-path", type=str, default="./checkpoints")
```
brian-dellabetta: We covered these in deep dive call, so i know what they do, but would be good to add a help/description to them

Issue #248, opened 2026-01-19
[RFC][Training] Add dflash support in training


shanjiaz:
# Background
DFlash has recently emerged in a [blog post](https://z-lab.ai/projects/dflash/) as a potentially superior method for speculative decoding compared to Eagle-3.  

<img width="617" height="321" alt="Image" src="https://github.com/user-attachments/assets/8948a4cc-77d7-4735-a106-40a96b89f573" />

It uses a 5-layer diffusion-llm to predict a block of tokens as a draft for the base model, where the prediction is conditioned on the hidden states of the base model, similar to Eagle 3.  The architecture is shown above, and a reference implementation of the inference model is provided [here](https://github.com/z-lab/dflash/blob/main/model/dflash.py).  

The blog claims potentially up to 2-3x larger speedups than Eagle 3 for synchronous requests, and inference is actively being implemented in vLLM.

# Training algorithm

The model needs to be trained to take in an input of hidden states from the model, an embedded token, and some mask tokens (which will become the draft) and predict the correct tokens for the draft.  

Training Inputs and Attention Mask: 
For a single document/ coherent sequence:  
Given sequence $(S_1…S_L)$ of  length $L$ and mask tokens $M_1…M_L$, and block size $B$ (also the draft length) with hidden dimension $D$, 
Query: Shape:  $[B,L,D]$ composed of noise (or mask) $M_1…M_L$.  
Key/Value:  Shape: $[B, 2L, D]$ concatenated from $[S_1…S…L]$ and $[M_1…M_L]$
Attention Mask: Shape: $[L, 2L]$  
For example, for a sequence of size 128 and block size 8, it should look like this:  

*As of Thursday afternoon the LMSYS attn block was incorrectly causal in the mask block. 

- Position IDs: ${0…L-1, 0…L-1}$ concat

<img width="640" height="336" alt="Image" src="https://github.com/user-attachments/assets/3083ed1b-867d-417e-a988-3bf55d6ca7ea" />

- Target: Shape: $[L]$ The target ids for the model
- Loss mask: $[L]$ Loss mask (same as for eagle3) which only trains on tokens generated by the base model.  Based on the chat template.  - Generated by Eagle 3 data generation code already.  
- Potential:  T2D and D2T tensors:  Vocabulary matching same as we use for Eagle 3.  I don't see any reason to exclude this functionality from DFlash.  
- Difference from Eagle 3:  We only need one forward pass to train the model, we do not need TTT, so in that respect it is simpler.  

[...]

PR #354, opened 2026-03-20
[Feat]Add DFlash training support

shanjiaz:

## Purpose
DFlash has recently emerged in a [blog post](https://z-lab.ai/projects/dflash/) as a potentially superior method for speculative decoding compared to Eagle-3.  We would like to enable our users to try out this new algorithm.

## Description
As described in the issue, we have added the dflash model definiton & configuration. Instead of the TTT method of eagle3, Dflash runs a single forward step. The drafter produces a block of tokens of length $B$ for each prefix (total of $L/B$ prefixes). Updated the attention mechanism to use a non-causal $[L, 2L]$ attention pattern where queries attend to both the hidden states from the verifier (first $L$ positions) and the mask token embeddings (last $L$ positions), all in a single forward pass. 

## Related Issue
[Issue 248](https://github.com/vllm-project/speculators/issues/248)

## Tests
Validate training with qwen3-8b model and testing in vLLM for acceptance rates. 

Tested by replacing the dflash model weights with our trained weights and this is the acceptance rate.
The model used: https://huggingface.co/shanjiaz/dflash-qwen3-8b/tree/main
```
============ Serving Benchmark Result ============
Successful requests:                     80        
Failed requests:                         0         
Maximum request concurrency:             1         
Benchmark duration (s):                  103.22    
Total input tokens:                      24010     
Total generated tokens:                  20480     
Request throughput (req/s):              0.78      
Output token throughput (tok/s):         198.41    
Peak output token throughput (tok/s):    18.00     
Peak concurrent requests:                3.00      
Total token throughput (tok/s):          431.02    
---------------Time to First Token----------------
Mean TTFT (ms):                          34.35     
Median TTFT (ms):                        26.97     
P99 TTFT (ms):                           102.18    
-----Time per Output Token (excl. 1st token)------
Mean TPOT (ms):                          4.92      
Median TPOT (ms):                        4.88      
P99 TPOT (ms):                           6.73      
---------------Inter-token Latency----------------
Mean ITL (ms):                           90.17     
Median ITL (ms):                         93.50     
P99 ITL (ms):                            167.02    
---------------Speculative Decoding---------------
Acceptance rate (%):                     17.56     
Acceptance length:                       1.88      
Drafts:                                  10899     
Draft tokens:                            54495     
Accepted tokens:                         9570      
Per-position acceptance (%):
  Position 0:                            50.74     
  Position 1:                            23.11     
  Position 2:                            9.49      
  Position 3:                            3.20      
  Position 4:                            1.27      
```

I have filled in:

## Summary by CodeRabbit

* **New Features**
  * Added DFlash speculator with anchored-block drafting plus CLI options for block size and max anchors; includes DFlash metrics, utilities, and a draft model.

* **Refactor**
  * Centralized shared attention utilities and moved attention implementation registration to a shared registry; expanded public exports to include DFlash symbols.

* **Bug Fixes**
  * Improved validation and clearer errors for model registry lookups.

* **Tests**
  * Added unit tests for DFlash metrics and utilities.


src/speculators/models/dflash/attention.py
```diff
@@ -0,0 +1,171 @@
...
+        enable_gqa=enable_gqa,
+        scale=scaling,
+    )
+    attention_output: torch.Tensor = cast("torch.Tensor", flex_attention_output)
+    attention_output = attention_output.transpose(1, 2).contiguous()
+    return attention_output, None
+
+
+ALL_ATTENTION_FUNCTIONS = AttentionInterface()  # Singleton class used for registry
+ALL_ATTENTION_FUNCTIONS.register("simple_flex_attention", flex_attention_forward)
```
fynnsu: This is a duplicate of the function defined in `src/speculators/models/eagle3/attention.py`. We are also registering this function twice with the same name. Maybe we can move this do a different location?

src/speculators/models/dflash/config.py
```diff
@@ -0,0 +1,87 @@
...
+
+    draft_vocab_size: int = Field(
+        default=32000,
+        description="Size of draft model vocabulary for speculation",
+    )
+
+    num_hidden_layers: int = Field(
+        default=3,
+        description="Number of hidden layers in the DFlash model",
+    )
```
fynnsu: For Eagle3, we encode this information using `transformer_layer_config.num_hidden_layers`. We should probably do the same here.

src/speculators/models/dflash/config.py
```diff
@@ -0,0 +1,87 @@
...
+        default=8,
+        description=(
+            "Default size of the draft block predicted with a forward pass of the model"
+        ),
+    )
+
+    target_hidden_size: int | None = Field(
+        default=None,
+        description="Hidden size of the target model (if different from draft model)",
+    )
```
fynnsu, line 62: Is this something we support?

src/speculators/train/trainer.py
```diff
@@ -113,16 +125,37 @@ def setup_model(self):
...
+                # from rank 0 to all ranks
+                set_model_state_dict(
+                    self.model,
+                    full_state_dict,
+                    options=StateDictOptions(
+                        full_state_dict=True,
+                        broadcast_from_rank0=True,
+                        strict=False,
+                    ),
+                )
```
fynnsu: This logic is no longer needed.

src/speculators/train/trainer.py
```diff
@@ -39,7 +49,7 @@ class TrainerConfig(NamedTuple):
     local_rank: int = 0
     train_call_kwargs: dict = {}
     val_call_kwargs: dict = {}
-    scheduler_type: Literal["linear", "cosine", "none"] = "linear"
+    scheduler_type: Literal["linear", "cosine", "none"] = "cosine"
```
fynnsu: Unnecessary change

src/speculators/train/trainer.py
```diff
@@ -202,10 +235,11 @@ def train_epoch(self, epoch: int):
...
             metrics = {k: v.item() for k, v in metrics.items()}
-            metric_logger.info(
-                {"train": metrics, "epoch": epoch, "lr": current_lr},
-                extra={"step": self.global_step},
-            )
+            if self.local_rank == 0:
+                metric_logger.info(
+                    {"train": metrics, "epoch": epoch, "lr": current_lr},
+                    extra={"step": self.global_step},
+                )
```
fynnsu: Not needed (we already filter in the logger).

fynnsu (commented):
Some initial comments

brian-dellabetta (commented):
Thanks for covering this in live code review. LGTM. The structure makes sense to me, only question would be if there are tests that can be added for any of the new code.

src/speculators/models/dflash/attention.py
```diff
@@ -0,0 +1,134 @@
...
+from speculators.models.attention import ALL_ATTENTION_FUNCTIONS  # noqa: F401
+
+
+def create_anchor_block_mask_mod(
+    lengths: torch.Tensor,
+    total_seq_len: int,
+    anchor_positions: torch.Tensor,
+    block_size: int,
+):
+    """
```
rahul-tuli, line 14: Our code uses rst format for docstrings, could you update this for consistency, also for future maybe it makes sense to add a `CLAUDE.md` file  in the main repo with this rule?

src/speculators/models/dflash/attention.py
```diff
@@ -0,0 +1,134 @@
...
+                    dtype=torch.long,
+                ),
+            ]
+        ).contiguous()
+
+    if ((anchor_positions < 0) | (anchor_positions >= total_seq_len)).any():
+        bad = anchor_positions[
+            (anchor_positions < 0) | (anchor_positions >= total_seq_len)
+        ]
+        raise ValueError(f"anchor_positions out of range: {bad.tolist()}")
```
rahul-tuli: nit: Looks like a good place we could use the walrus operator

```suggestion
    if ((mask := (anchor_positions < 0) | (anchor_positions >= total_seq_len))).any():
        raise ValueError(f"anchor_positions out of range: {anchor_positions[mask].tolist()}")
```

src/speculators/models/dflash/attention.py
```diff
@@ -0,0 +1,134 @@
...
+
+    anchor_docs = document_ids[anchor_positions]
+    if (anchor_docs == -1).any():
+        bad = anchor_positions[anchor_docs == -1]
+        raise ValueError(f"anchor_positions include padding locations: {bad.tolist()}")
```
rahul-tuli: nit: Same as above use walrus operator?

src/speculators/models/dflash/core.py
```diff
@@ -0,0 +1,862 @@
...
+
+
+@torch.no_grad()
+def compute_accuracy(
+    logits: torch.Tensor,  # shape: [1, total_seq_len - ttt_step, draft_vocab_size]
+    targets: torch.Tensor,  # shape: [1, total_seq_len - ttt_step, draft_vocab_size]
+    loss_mask: torch.Tensor | None,  # shape: [1, total_seq_len - ttt_step]
+    block_size: int = 1,
+):
+    # Note: logits, targets, and loss_mask are already aligned for the current ttt_step
```
rahul-tuli: docstring?

src/speculators/models/dflash/core.py
```diff
@@ -0,0 +1,862 @@
...
+def build_kv_position_ids(
+    base_position_ids: torch.Tensor,  # [B, total_seq_len]
+    anchor_positions: torch.Tensor,  # [B, n] or [n] (indices into base seq)
+    block_size: int,
+) -> torch.Tensor:
+    """
+    Construct position_ids for KV = [base | anchor_blocks].
+
+    Appended block for anchor a gets positions:
+        base_position_ids[..., a] + [0..block_size-1]
```
rahul-tuli: params missing?

src/speculators/models/dflash/core.py
```diff
@@ -0,0 +1,862 @@
...
+                f"anchor_positions batch {anchor_positions.shape[0]} != {B}"
+            )
+    else:
+        raise ValueError(
+            f"anchor_positions must be [n] or [B, n], got {anchor_positions.shape}"
+        )
+
+    n = anchor_positions.shape[1]
+
+    # Position id at each anchor: [B, n]
```
rahul-tuli: nit: remove inline comments that describe the code, keep ones that explain why?

src/speculators/models/dflash/core.py
```diff
@@ -0,0 +1,862 @@
...
+def gather_anchor_spans(
+    input_ids: torch.Tensor,
+    anchor_positions: torch.Tensor,
+    block_size: int,
+) -> torch.Tensor:
+    """
+    input_ids: [T]
+    anchor_positions: [n] (positions into input_ids)
+    returns: [n*block_size] = input_ids[anchor_i + 0 .. anchor_i +
+        block_size-1] concatenated
```
rahul-tuli: nit: inconsistent docstring

src/speculators/models/dflash/core.py
```diff
@@ -0,0 +1,862 @@
...
+        Loss value and metrics dictionary.
+    """
+    s_loss = loss_function(logits, targets, loss_mask)
+
+    s_full_acc, per_position_acc = compute_accuracy(
+        logits, targets, loss_mask, block_size
+    )
+    # s_accept_rate, per_position_accept = compute_acceptance_rate(
+    #     logits, targets, loss_mask, block_size
+    # )
```
rahul-tuli: nit: remove?

src/speculators/models/dflash/core.py
```diff
@@ -0,0 +1,862 @@
...
+    s_metrics: dict[str, Any] = {}
+    s_metrics["loss"] = s_loss.detach().clone()
+    s_metrics["full_acc"] = s_full_acc
+    # s_metrics[f"accept_rate"] = s_accept_rate
```
rahul-tuli: nit: remove commented code

src/speculators/models/dflash/core.py
```diff
@@ -0,0 +1,862 @@
...
+    for pos in range(len(per_position_acc)):
+        s_metrics[f"position {pos} acc"] = per_position_acc[pos]
+        # s_metrics[f"position {pos} accept"]=per_position_accept[pos]
```
rahul-tuli: nit: remove commented code

src/speculators/models/dflash/core.py
```diff
@@ -0,0 +1,862 @@
...
+        cls,
+        verifier_config: "PretrainedConfig",
+        t2d: torch.Tensor | None = None,
+        d2t: torch.Tensor | None = None,
+        **kwargs,
+    ) -> "DFlashDraftModel":
+        """Create DFlash model from training arguments.
+
+        Args:
+            verifier_config: Verifier model configuration. This should be a config
```
rahul-tuli, line 113: If the num_hidden_layers should match the number of draft layers, seems weird we call it the verifier config? maybe we should make it explicit in the verifier config and call it something like num_draft_hidden_layers?

src/speculators/models/dflash/core.py
```diff
@@ -0,0 +1,862 @@
...
+        | None = None,  # shape: [1, total_seq_len, hidden_size]
+        **kwargs,
+    ):
+        device = hidden_states.device
+        total_seq_len = hidden_states.shape[1]
+
+        if lengths is None:
+            lengths = torch.tensor([total_seq_len], dtype=torch.long, device=device)
+        if position_ids is None:
+            # MEGAN FLAG CHECK THAT THIS SHOULD BE +1
```
rahul-tuli: nit: remove

rahul-tuli (commented):
The code looks good to me pending nits; Two asks from my side:

- Update docstrings to be consistent with rest of the repo
- Remove Commented out code from checked in files
- Remove unnecessary inline comments

src/speculators/models/dflash/core.py
shanjiaz: These should probably be removed.

dsikka (commented):
Can we add tests?

src/speculators/models/dflash/config.py
```diff
@@ -0,0 +1,90 @@
...
+
+from speculators import SpeculatorModelConfig
+
+__all__ = [
+    "DFlashSpeculatorConfig",
+]
+
+
+@SpeculatorModelConfig.register("dflash")
+class DFlashSpeculatorConfig(SpeculatorModelConfig):
```
dsikka, line 17: Couldn't we share a lot of these fields with Eagle and just define them in the base model?

src/speculators/models/dflash/core.py
dsikka: we should also add correct docstrings

src/speculators/models/dflash/core.py
```diff
@@ -0,0 +1,849 @@
...
+def apply_rotary_pos_emb(q, k, cos, sin, position_ids=None, unsqueeze_dim=1):  # noqa: ARG001
+    """Apply rotary position embeddings (local implementation)."""
+
+    cos = cos.unsqueeze(unsqueeze_dim)
+    sin = sin.unsqueeze(unsqueeze_dim)
+    q_len = q.size(-2)
+    q_embed = (q * cos[..., -q_len:, :]) + (_rotate_half(q) * sin[..., -q_len:, :])
+    k_embed = (k * cos) + (_rotate_half(k) * sin)
+    return q_embed, k_embed
+
```
dsikka: instead of stuffing everything here and above into this file, maybe move to a separate utils?

dsikka (changes requested):
Can you add a sample checkpoint to your PR description? Like the link that was used for the evals?

I think we need a good docs explaining dflash training - for example, looking at your code, I am still not very sure what an anchor is.

src/speculators/models/dflash/core.py
shanjiaz: yep!

src/speculators/models/dflash/config.py
shanjiaz: Yes we can. Maybe in a follow up PR? Didn't want to change too much of eagle3 code.

dsikka (commented):
Tests?

Do we have a sample checkpoint that we're sharing with Summer that you can also post here?

src/speculators/models/dflash/core.py
fynnsu: Yes, same with the qwen layer definitions below.

fynnsu:
@mergify refresh

scripts/launch_vllm.py
```diff
@@ -32,6 +32,15 @@ def parse_args():
             "Note: if set, you must also pass the same value into the training process"
         ),
     )
+    parser.add_argument(
+        "--include-last-layer",
+        action=argparse.BooleanOptionalAction,
+        default=True,
```
rahul-tuli, line 38: nit: I feel we should default this to False maybe?

scripts/launch_vllm.py
```diff
@@ -45,22 +54,24 @@ def main():
...
+
+    config = AutoConfig.from_pretrained(args.model)
+    if hasattr(config, "text_config"):
+        config = config.text_config
+    num_hidden_layers = config.num_hidden_layers
+
     if args.target_layer_ids:
         target_layer_ids = args.target_layer_ids
+        if args.include_last_layer and num_hidden_layers not in target_layer_ids:
+            target_layer_ids.append(num_hidden_layers)
```
rahul-tuli, line 67: Could we confirm if `num_hidden_layers` is correct? shouldn't it be `num_hidden_layers - 1` ?

nit: Also a thought, instead of adding an extra flag for the last layer, this functionality could in principle be supported by target_layer_ids, to make it easier for users of this script, we could probably allow negative indexing in target_layer_ids (so users don't need to know the num_hidden_layers upfront?)

scripts/launch_vllm.py
shanjiaz: I think we want users to pass in only the hidden layers that they want to extract by default, not including the last layer. See explanation below.

scripts/launch_vllm.py
shanjiaz: Yes, but we hit another error down stream. `target_layer_ids` is saved in the config and will be consumed by vllm for inference. vLLM is expecting the three layer list for for inference and will create the wrong shape if we include the last layer here. Hidden states is the only place that we would need the last layer. I agree this is pretty nasty but probably not within the scope of this PR to fix.

scripts/launch_vllm.py
shanjiaz, line 78: @rahul-tuli I think `num_hidden_layers` is the last layer index we use.

scripts/launch_vllm.py
rahul-tuli: Resolved https://github.com/vllm-project/speculators/pull/354/changes#r3073980856

src/speculators/models/dflash/core.py
rahul-tuli: +1

src/speculators/models/dflash/core.py
```diff
@@ -0,0 +1,591 @@
...
+    _keys_to_ignore_on_save: ClassVar[list[str]] = [  # type: ignore[misc,assignment]
+        "verifier_lm_head.weight",
+        "verifier_norm.weight",
+    ]
+
+    def __init__(
+        self,
+        config: DFlashSpeculatorConfig,
+        t2d: torch.Tensor,
+        d2t: torch.Tensor,
```
rahul-tuli: I think this might not work well with transformers, could you follow the same approach as eagle3 and make these class vars https://github.com/vllm-project/speculators/blob/206e043f327d9bfe51e5db8ddc6bc5108af3abfa/src/speculators/models/eagle3/core.py#L171-L172  and also provide a from_pretrained override https://github.com/vllm-project/speculators/blob/206e043f327d9bfe51e5db8ddc6bc5108af3abfa/src/speculators/models/eagle3/core.py#L591?

src/speculators/models/dflash/utils.py
```diff
@@ -0,0 +1,90 @@
...
+    Returns:
+        tuple: (anchors, anchor_valid)
+            - anchors: Selected anchor indices [1, num_anchors]
+            - anchor_valid: Boolean mask for valid anchors [1, num_anchors]
+    """
+    if loss_mask.ndim != 2:  # noqa: PLR2004
+        raise ValueError(f"Expected [B, T], got {loss_mask.shape}")
+
+    if block_size <= 0:
+        raise ValueError(f"Expected block size >= 0, got {block_size}")
```
rahul-tuli:
```suggestion
        raise ValueError(f"Expected block size > 0, got {block_size}")
```

scripts/launch_vllm.py
```diff
@@ -45,22 +54,24 @@ def main():
     if "--" in vllm_args:
         vllm_args.remove("--")
 
+    # Import here so that it isn't required if target_layer_ids passed explicitly
```
rahul-tuli:
```suggestion
```

Won't this run regardless?

shanjiaz: yep good catch it's an old comment.

src/speculators/models/dflash/core.py
shanjiaz: I think I want to add the from_pretrained method to the base model definition instead, will create a follow up PR. I moved t2d and d2t to class variables.

scripts/launch_vllm.py
rahul-tuli: Yeah the explanation doesn't make sense to me, Since for target_layer_ids the explanation already says if not provided we default to `"[2, num_hidden_layers // 2, num_hidden_layers - 3, num_hidden_layers].`

If user passes in target_layer_ids and they do not supply the last layer, I feel it is user error. Will resolve this comment and let you make the final call, imo it unnecessarily complicates the script.

src/speculators/convert/eagle/eagle_legacy_model.py
```diff
@@ -125,42 +124,6 @@ def check_add_architectures(self) -> Self:
...
-            return value
-
-        raise ValueError(
-            "transformer_layer_config must be a PretrainedConfig instance or a "
-            "dictionary that can be converted to a PretrainedConfig."
-        )
-
 
 @SpeculatorModel.register("eagle")
 class EagleSpeculator(SpeculatorModel):
```
rahul-tuli, line 166: this is okay for now, but at some point we should move to using just one definition of Eagle3Speculator, could you create a ticket for that? Maybe a good first issue?

src/speculators/models/dflash/core.py
```diff
@@ -0,0 +1,334 @@
...
+        num_draft_layers = tl_config.num_hidden_layers
+        self.layers = nn.ModuleList(
+            [
+                Qwen3DFlashDecoderLayer(config.transformer_layer_config, layer_idx)  # type: ignore[arg-type]
+                for layer_idx in range(num_draft_layers)
+            ]
+        )
+
+        # Load actual verifier config to get the real verifier layer count
+        from transformers import AutoConfig  # noqa: PLC0415
```
rahul-tuli: Should we move this to the top with rest of the imports?

src/speculators/models/dflash/metrics.py
```diff
@@ -0,0 +1,199 @@
...
+    m = loss_mask.to(logits.dtype).view(B, T)
+
+    ce = ce * w * m
+
+    denom = (m * w).sum(dim=1) + 1e-5
+    return (ce.sum(dim=1) / denom).mean()
+
+
+@torch.no_grad()
+def compute_acceptance_rate(
```
rahul-tuli, line 105: Should this be a shared util? or is it kept here because of block size param?

src/speculators/models/eagle3/data.py
```diff
@@ -0,0 +1,40 @@
...
+    # [y0, y1, y2, y3, y4, y5, y6, y7, y8, y9]
+    loss_mask = batch["loss_mask"]  # shape: [seq_len]
+    # [l0, l1, l2, l3, l4, l5, l6, l7, l8, l9]
+    lengths = batch["lengths"]  # shape: [1]
+    # [10]
+    position_ids = batch["position_ids"]  # shape: [seq_len]
+    # [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
+
+    # Need to align (x1, g0, y1, l1)
+    # todo: verify loss mask shift is correct
```
rahul-tuli: nit:
```suggestion
```

src/speculators/models/eagle3/data.py
```diff
@@ -0,0 +1,40 @@
+# ruff: noqa: ERA001
+from speculators.train.data import BatchType
+
```
rahul-tuli, line 3: nit: add `__all__`

src/speculators/config.py
```diff
@@ -239,6 +240,14 @@ def __pydantic_schema_base_type__(cls) -> type["SpeculatorModelConfig"]:
         default=version("speculators"),
         description="Version of the speculators library",
     )
+    transformer_layer_config: PretrainedConfig = Field(
```
rahul-tuli: Why are we adding transformer_layer_config logic here to the base config?Doesn't this add an assumption that all drafter models will be transformer layers? Which is True currently but might change in the future?

src/speculators/models/dflash/config.py
rahul-tuli: Yeah I think if needed we should extract a BaseEagleConfig class for shared things

src/speculators/model.py
```diff
@@ -112,6 +114,10 @@ def from_pretrained(
         :return: A SpeculatorModel instance of the appropriate subclass, loaded with
             the pretrained weights and configuration.
         """
+        # Extract vocab mapping tensors before passing to HF's from_pretrained
+        t2d = kwargs.pop("t2d", None)
```
rahul-tuli: Not a fan of this approach to mix this logic in the base speculator model, we should provide a override for this in the DFlash model itself, no?

src/speculators/model.py
```diff
@@ -174,6 +182,9 @@ def from_pretrained(
             weights_only=weights_only,
             **kwargs,
         )
+        model.load_vocab_mappings(t2d, d2t)
+        model.load_verifier_weights()
```
rahul-tuli: Same comment as above, this forces us to implement these methods which for example are not needed for MTP, if needed we should have extracted/created another class for shared functionality b/w eagle3 and dflash

rahul-tuli (commented):
This diff works for now, but it has some opportunity for cleanup, I'm okay with approving as long as we follow up with clean up PRs

src/speculators/models/dflash/metrics.py
shanjiaz: Yeah this is specific to dflash.

src/speculators/model.py
shanjiaz: Will update!

src/speculators/convert/eagle/eagle_legacy_model.py
shanjiaz: Yep I'll create a ticket for this.

src/speculators/model.py
shanjiaz: Could MTP potentially also use reduced vocab in the future? I personally think it makes more sense for these shared methods to live here in the base model since dflash, eagle3, peagle and potentially other algorithms in the future would probably all need these methods.

tests/e2e/smoke/test_offline_training_dflash.py
fynnsu, line 1: Why do we need a new file for this? It actually seems to be testing `eagle3` as well, so can we add this parameterize logic to that file?

src/speculators/models/dflash/model_definitions.py
```diff
@@ -0,0 +1,207 @@
...
+    unsqueeze_dim=1,
+):
+    """Apply rotary position embeddings (local implementation)."""
+
+    cos = cos.unsqueeze(unsqueeze_dim)
+    sin = sin.unsqueeze(unsqueeze_dim)
+    q_len = q.size(-2)
+    q_embed = (q * cos[..., -q_len:, :]) + (_rotate_half(q) * sin[..., -q_len:, :])
+    k_embed = (k * cos) + (_rotate_half(k) * sin)
+    return q_embed, k_embed
```
fynnsu, line 44: Import from transformers

src/speculators/models/dflash/utils.py
```diff
@@ -0,0 +1,69 @@
...
+
+    Args:
+        loss_mask: Binary mask indicating valid positions [1, total_seq_len]
+        n: Number of anchors to select per batch item
+        block_size: Block size (last block_size positions excluded)
+
+    Returns:
+        tuple: (anchors, anchor_valid)
+            - anchors: Selected anchor indices [1, num_anchors]
+            - anchor_valid: Boolean mask for valid anchors [1, num_anchors]
```
fynnsu: Update shapes to match true output shape: `[num_anchors]`

src/speculators/train/data.py
```diff
@@ -454,8 +415,16 @@ def _get_raw_data(self, index):
...
+def create_collate_fn(
+    max_len: int,
+    hidden_size: int,
+    preprocess: Callable[[BatchType], BatchType] | None = None,
+):
     def collate_fn(batch: list[BatchType | None]) -> BatchType:
+        # Apply per-sample preprocessing (e.g. shift_batch for Eagle3)
+        if preprocess:
+            batch = [preprocess(b) if b is not None else b for b in batch]
+
```
fynnsu: Move down to skip filter

tests/e2e/run_vllm.py
```diff
@@ -103,6 +101,9 @@ def extract_metrics(
             num_accepted_tokens += metric.value
         elif metric.name == "vllm:spec_decode_num_accepted_tokens_per_pos":
             assert isinstance(metric, Vector)
+            # Grow acceptance_counts to fit the metric values
+            while len(acceptance_counts) < len(metric.values):
+                acceptance_counts.append(0)
```
fynnsu: re-init variable with correct number of entries instead of calling append in a loop

scripts/launch_vllm.py
shanjiaz: Yes it's true that `target_layer_ids` default to `[2, num_hidden_layers // 2, num_hidden_layers - 3, num_hidden_layers]` in this script and that's ok since training would default to `[2, num_hidden_layers // 2, num_hidden_layers - 3]` and vllm can still serve fine. We want use to only pass in the layers required by the algorithm only as expected behavior.

src/speculators/models/dflash/model_definitions.py
shanjiaz: these are different from huggingface ones.

src/speculators/models/dflash/core.py
shanjiaz: yep!

rahul-tuli (approved):

fynnsu (approved):

dsikka (approved):

PR #365, opened 2026-03-26
fix(datagen): fix VllmHiddenStatesGenerator for hybrid KV cache models and large datasets

rahul-tuli:
## Purpose

The offline data generation pipeline (`VllmHiddenStatesGenerator`) does not work with hybrid attention models such as Qwen3. This PR fixes that, along with several other bugs discovered in the process.

## Problem

Qwen3 uses a hybrid attention architecture that mixes full attention layers (`FullAttentionSpec`) and sliding-window attention layers (`ChunkedLocalAttentionSpec`). The KV cache initialization in `VllmHiddenStatesGenerator` assumed all layers share a single uniform attention spec — valid for dense models like Llama, but not for hybrid models. Attempting to run data generation with a Qwen3 model failed immediately on startup:

```
TypeError: unify_hybrid_kv_cache_specs() does not support heterogeneous specs
```

Beyond the hybrid KV cache crash, the generator had additional bugs that surfaced during investigation:

**Scheduler starvation on large datasets.** The scheduler accepted the first 32 requests and completed them, but never marked them as finished — so their slots were never freed and no further requests were ever admitted. On a dataset of 7,000+ samples, only 32 hidden states were captured and the rest were silently dropped. No error was raised.

**Wrong result ordering for batches of 11+ sequences.** Request IDs were formatted as `req_0_N` and sorted lexicographically. For N ≥ 10, `"req_0_10"` sorted before `"req_0_2"`, returning results in the wrong order relative to the input — silently producing misaligned training data.

**Stale request metadata between scheduler steps.** `_current_request_metadata` was only reset between chunks, not after each forward pass. In a step with no prefill tokens, the stale dict from the previous step was reused, mapping the wrong tokens to the wrong requests.

**No GPU resource cleanup guarantee.** The vLLM executor was only shut down if the caller explicitly deleted the object. An exception mid-generation left GPU memory held until process exit.

**Package import error.** `reload_schemas()` called Pydantic's `model_rebuild()` without providing `torch` to the type evaluation namespace, causing:

```
NameError: name 'torch' is not defined
```

This made the `speculators` package unimportable in any context where the Pydantic schema rebuild ran before torch types were resolved.

## Changes

- **`vllm_hidden_states_generator.py`:** Replace `unify_hybrid_kv_cache_specs()` + `_get_kv_cache_groups_uniform_spec()` with `get_kv_cache_configs()` + `generate_scheduler_kv_cache_config()`, matching vLLM's own `EngineCore._initialize_kv_caches()`. Split `generate()` into `MAX_NUM_SEQS=32`-sized chunks processed by a new `_generate_chunk()` method so the scheduler slot is freed between batches. Sort results by integer index instead of string. Add `__enter__`/`__exit__`/`__del__` for guaranteed executor cleanup with a double-call guard.
- **`custom_worker.py`:** Clear `_current_request_metadata` immediately after consuming it in `_store_captured_states`.
- **`utils/pydantic_utils.py`:** Pass `_types_namespace={"torch": torch}` to `model_rebuild()`.
- **`scripts/response_regeneration/script.py`:** Add `gsm8k` dataset, wire `--subset` into `load_dataset` (was previously a no-op), make `--dataset` choices auto-derived from `DATASET_CONFIGS`.

## Testing

Verified on 4× H100 80GB (GPUs 2,5,6,7) with `Qwen/Qwen3-Next-80B-A3B-Instruct`.

**Data generation:**
```bash
CUDA_VISIBLE_DEVICES=2,5,6,7 python scripts/data_generation_offline.py \
    --target-model-path Qwen/Qwen3-Next-80B-A3B-Instruct \
    --train-data-path <path-to-jsonl> \
    --output-dir <output-dir> \
    --tensor-parallel-size 4
```

**Response regeneration — gsm8k:**
```bash
CUDA_VISIBLE_DEVICES=2,5,6,7 bash scripts/response_regeneration/run_all.sh \
    --model Qwen/Qwen3-Next-80B-A3B-Instruct \
    --gpus 2,5,6,7 \
    --tp-size 4 \
    --dataset gsm8k \
    --outfile gsm8k_qwen3next.jsonl
```

**Response regeneration — existing datasets:**
```bash
CUDA_VISIBLE_DEVICES=2,5,6,7 bash scripts/response_regeneration/run_all.sh \
    --model Qwen/Qwen3-Next-80B-A3B-Instruct \
    --gpus 2,5,6,7 \
    --tp-size 4 \
    --dataset magpie \
    --limit 1000
```

### DIFF
--- a/scripts/response_regeneration/script.py
+++ b/scripts/response_regeneration/script.py
@@ -22,6 +22,12 @@ DATASET_CONFIGS = {
         "prompt_field": "prompt",
         "default_split": "train_sft",
     },
+    "gsm8k": {
+        "id": "openai/gsm8k",
+        "subset": "main",
+        "prompt_field": "question",
+        "default_split": "train",
+    },
 }
 
 
@@ -43,8 +49,8 @@ def parse_args():
     parser.add_argument(
         "--dataset",
         default="ultrachat",
-        choices=["magpie", "ultrachat"],
-        help="Dataset to process (magpie or ultrachat)",
+        choices=list(DATASET_CONFIGS),
+        help="Dataset to process",
     )
     parser.add_argument(
         "--split",
@@ -54,7 +60,7 @@ def parse_args():
     parser.add_argument(
         "--subset",
         default=None,
-        help="(unused) kept for symmetry with other scripts",
+        help="Dataset subset/config name (overrides dataset default if set)",
     )
     parser.add_argument("--limit", type=int, default=None, help="Stop after N rows")
     parser.add_argument(
@@ -227,6 +233,7 @@ async def main():
     dataset_config = DATASET_CONFIGS[args.dataset]
     dataset_id = dataset_config["id"]
     prompt_field = dataset_config["prompt_field"]
+    subset = args.subset or dataset_config.get("subset")
 
     # Use dataset-specific default split if not provided
     split = args.split if args.split is not None else dataset_config["default_split"]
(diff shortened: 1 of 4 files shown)


fynnsu:
Closing as VllmHiddenStatesGenerator has been deprecated and removed.

PR #366, opened 2026-03-26
feat(mtp): add MTP finetuning support to training infrastructure

rahul-tuli:
## Purpose

Enable `scripts/train.py --speculator-type mtp` to finetune the native MTP head
of models like Qwen3-Next on domain-specific data, improving speculative decoding
acceptance rates without any external speculator model.

This follows the [FastMTP](https://arxiv.org/abs/2509.18362) approach: the MTP head
is small (~100M-400M params), so it trains on pre-extracted hidden states without
loading the full verifier (~80B params) during training.

## Results (GSM8K math reasoning, `num_speculative_tokens=3`)

| Checkpoint | Pos0 | Pos1 | Pos2 | MeanAccepted |
|---|---|---|---|---|
| Base (original MTP head) | 0.873 | 0.778 | 0.675 | 2.01 |
| Epoch 0 finetuned | 0.934 | 0.891 | 0.838 | 2.46 (+22.5%) |

Finetuned checkpoint: [`inference-optimization/Qwen3-Next-80B-A3B-Instruct-GSM8K-MTP-finetuned`](https://huggingface.co/inference-optimization/Qwen3-Next-80B-A3B-Instruct-GSM8K-MTP-finetuned)

## Description

Connect the existing MTP model architecture (PR #452) to the speculators training
pipeline. No new trainer, dataset class, or collate function — MTP plugs into the
same infrastructure used by EAGLE3 and DFlash.

### Source changes

- **`src/speculators/models/mtp/core.py`** — Add `layers` property for FSDP wrapping
  compatibility and `**kwargs` to `from_training_args()` to absorb extra training args.
- **`src/speculators/models/mtp/data.py`** (new) — `shift_batch_mtp()` renames
  `verifier_last_hidden_states` to `hidden_states`. MTP uses only the last hidden
  layer, not the three intermediate layers used by EAGLE3. No token shifting — the
  MTP forward pass handles offset slicing internally.
- **`src/speculators/train/data.py`** — Fix `create_collate_fn` to apply the
  preprocess function to the fallback empty sample.
- **`scripts/train.py`** — MTP-aware config branch (uses verifier's actual config
  instead of reconstructing a LlamaConfig), preprocess wiring, and two new CLI
  arguments (`--num-speculative-steps`, `--step-weight-beta`).

### Key design decisions

- **MTP uses the verifier's actual config** — `create_transformer_layer_config()`
  constructs a fresh LlamaConfig by copying individual fields, which drops
  architecture-specific attributes like `layer_types` for Qwen3-Next hybrid
  attention. MTP needs the full verifier config.
- **No draft vocab mapping** — MTP uses the full verifier vocabulary via frozen
  `embed_tokens` and `lm_head`. No d2t/t2d mapping, no mask token.
- **`--from-pretrained` is the expected path** — Users extract the pretrained MTP
  head (convert), finetune it, then stitch it back.
- **Only last hidden layer needed** — The example script passes
  `--target-layer-ids <last_layer>` to `launch_vllm.py` to save extraction time.

## Related Issue

Depends on PR #452 (MTP model architecture, converter, stitcher).

I have filled in:


scripts/train.py
```diff
@@ -647,6 +676,22 @@ def parse_args():
     parser.add_argument("--scheduler-warmup-steps", type=int, default=None)
     parser.add_argument("--scheduler-total-steps", type=int, default=None)
     parser.add_argument("--scheduler-num-cosine-cycles", type=float, default=0.5)
+    # MTP-specific arguments
```
rahul-tuli: These CLI should probably be cleaned up to only show algorithm specific args, but that is out of scope for this PR, it will be done in a follow up

tests/e2e/smoke/test_mtp_finetuning.py
fynnsu, line 1: With the smoke tests, we're trying to reproduce the user process as much as possible and making sure everything connects correctly, the shapes all line up, etc.

This test kinda replaces a number of key components (e.g. the dataset, the setup logic from train.py, etc). It also doesn't verify that the final model is runnable in vllm.

Could we add a smoke test for the full e2e flow?

fynnsu: Also ideally we would have an e2e regression test as well, to verify the model is learning something meaningful.

shanjiaz: @rahul-tuli if size is the concern, maybe we could use some of the smaller model that has mtp layers? like this one: https://huggingface.co/Qwen/Qwen3.5-2B/blob/main/model.safetensors.index.json

docs/user_guide/tutorials/train_mtp_offline.md
```diff
@@ -0,0 +1,274 @@
...
+
+## Step 0: Setup Your Environment
+
+Create two virtual environments (recommended to keep separate so dependencies don't conflict):
+
+```bash
+# Speculators venv (for data prep and training)
+uv venv speculators_venv
+source speculators_venv/bin/activate
+uv pip install "speculators>=0.5.0"
```
shanjiaz: Speculators 0.5.0 probably won't be able to support mtp right?

tests/e2e/regression/test_mtp_online_acceptance.py
```diff
@@ -0,0 +1,44 @@
...
+        verifier="Qwen/Qwen3.5-9B",
+        training_data_repo="inference-optimization/Qwen3.5-9B-responses",
+        num_speculative_tokens=3,
+        target_layer_ids=[32],
+        max_samples=5000,
+        seq_length=8192,
+        epochs=3,
+        lr=3e-4,
+        prompts=prompts,
+        # TODO: placeholder thresholds — calibrate from first successful run
```
shanjiaz: Is this addressed?

docs/user_guide/tutorials/train_mtp_offline.md
rahul-tuli:
```suggestion
uv pip install "speculators>=0.6.0"
```

rahul-tuli: Good catch, updated typo

tests/e2e/regression/test_mtp_online_acceptance.py
rahul-tuli: Should be now, was experimenting with hyperparams

src/speculators/models/mtp/model_definitions.py
```diff
@@ -103,7 +103,7 @@ def forward(
     ) -> torch.Tensor:
         hidden_normed = self.hidden_layernorm(hidden_states)
         embed_normed = self.token_layernorm(token_embeddings)
-        proj = self.input_proj(torch.cat([hidden_normed, embed_normed], dim=-1))
+        proj = self.input_proj(torch.cat([embed_normed, hidden_normed], dim=-1))
```
rahul-tuli, line 106: Bug I missed before

docs/user_guide/algorithms/decision_guide.md
```diff
@@ -10,25 +10,23 @@ Speculators currently supports three speculative decoding algorithms: **Eagle-3*
...
 
 ## Current Support
 
-|                     | Eagle-3       | P-EAGLE             | DFlash              |
-| ------------------- | ------------- | ------------------- | ------------------- |
-| **Draft layers**    | Llama-style   | Llama-style         | Qwen3-style         |
-| **Verifier models** | Any supported | Any supported       | Any supported       |
-| **Speculators**     | Mature        | Newer, growing fast | Newer, growing fast |
-| **vLLM**            | Mature        | Newer, growing fast | Newer, growing fast |
+| | Eagle-3 | P-EAGLE | DFlash | MTP | | ------------------- | ------------- | ------------------- | ------------------- | -------------------------------- | | **Draft layers** | Llama-style | Llama-style | Qwen3-style | Native MTP layers | | **Verifier models** | Any supported | Any supported | Any supported | Models with native MTP only | | **Training mode** | From scratch | From scratch | From scratch | Finetune existing MTP head | | **Speculators** | Mature | Newer, growing fast | Newer, growing fast | Newer, growing fast | | **vLLM** | Mature | Newer, growing fast | Newer, growing fast | Newer, growing fast |
```
fynnsu, line 26: Table formatting is broken (check docs preview).

docs/user_guide/algorithms/index.md
fynnsu, line 1: Commenting here because I don't have a better spot for this.

You need to add `mtp.md` and your mtp tutorial to `docs/.nav.yml` for it to show up in the sidebar.

docs/index.md
```diff
@@ -31,8 +31,7 @@ Speculators standardizes this process by providing a productionized end-to-end f
 - **Standardized, Extensible Format:** Provides a Hugging Face-compatible format for defining speculative models, with tools to convert from external research repositories into a standard speculators format for easy adoption.
 - **Seamless vLLM Integration:** Built for direct deployment into vLLM, enabling low-latency, production-grade inference with minimal overhead.
 
-!!! tip
-    Read more about Speculators features in this [vLLM blog post](https://blog.vllm.ai/2025/12/13/speculators-v030.html).
+!!! tip Read more about Speculators features in this [vLLM blog post](https://blog.vllm.ai/2025/12/13/speculators-v030.html).
```
fynnsu: This seems to have broken the formatting.

scripts/train.py
```diff
@@ -640,6 +663,7 @@ def parse_args():
             "'ce' = cross-entropy."
         ),
     )
+    parser.add_argument("--step-weight-beta", type=float, default=0.6)
```
fynnsu: Can you add a help here

docs/user_guide/tutorials/train_mtp_offline.md
```diff
@@ -0,0 +1,274 @@
...
+uv venv speculators_venv
+source speculators_venv/bin/activate
+uv pip install "speculators>=0.6.0"
+```
+
+```bash
+# vLLM venv (for serving the target model)
+uv venv vllm_venv
+source vllm_venv/bin/activate
+uv pip install "vllm>=0.18"
```
rahul-tuli:
```suggestion
uv pip install "vllm>=0.22.0"
```

docs/user_guide/algorithms/decision_guide.md
rahul-tuli: Should be fixed now, Thanks for catching this!

docs/index.md
rahul-tuli: Fixed!

shanjiaz (commented):
Can you take a look at failed unit & link checks?

scripts/train.py
```diff
@@ -291,30 +292,44 @@ def main(args: argparse.Namespace):
...
+            verifier_config = verifier_config.text_config
+        d2t, t2d, draft_vocab_size = None, None, verifier_config.vocab_size
+        transformer_layer_config = verifier_config
+        args.mask_token_id = None
+        if not args.from_pretrained:
+            logger.warning(
+                "MTP training without --from-pretrained: "
+                "the recommended workflow is to convert the native MTP head "
+                "first and pass the converted checkpoint via --from-pretrained."
+            )
```
fynnsu: Does this still run if you don't use `--from-pretrained`? If it just crashes later, we should probably just make this an error.

src/speculators/models/mtp/core.py
```diff
@@ -99,7 +98,6 @@ def load_verifier_weights(self) -> None:
         super().load_verifier_weights()
         del self.verifier_lm_head
 
-    @conditional_torch_compile
```
fynnsu, line 102: Why are we removing torch compile here?

src/speculators/models/mtp/data.py
```diff
@@ -0,0 +1,18 @@
...
+
+def shift_batch_mtp(batch: BatchType) -> BatchType:
+    """Rename verifier_last_hidden_states to hidden_states for MTP.
+
+    No token-level shifting — the MTP forward pass handles alignment
+    internally via per-step offset slicing of input_ids.
+    """
+    return {
+        "input_ids": batch["input_ids"],
+        "hidden_states": batch["verifier_last_hidden_states"],
```
fynnsu, line 14: Hmm this is technically okay, but we're not still generating 4 layer's hidden states are we? If MTP only requires the last layer we should be launching vllm in a way that only extracts the last layer, and then only loading that 1 layer in during training. Otherwise we're wasting time on data saving/loading.

src/speculators/train/data.py
```diff
@@ -470,7 +470,10 @@ def collate_fn(batch: list[BatchType | None]) -> BatchType:
             # Match the configured `dtype` so the placeholder doesn't crash
             # downstream layers loaded at a different precision (e.g. bf16
             # weights vs fp32 default placeholders).
-            batch = [create_empty_sample(hidden_size, dtype=dtype)]
+            empty = create_empty_sample(hidden_size, dtype=dtype)
+            if preprocess:
+                empty = preprocess(empty)
+            batch = [empty]
```
fynnsu, line 507: I'm pretty sure the eagle3 `shift_batch` fn assumes the seq_len is at least 1. If needed we can update that function, but we should make sure this doesn't break that.

tests/e2e/smoke/test_mtp_finetuning.py
```diff
@@ -0,0 +1,190 @@
...
+    # Step 2: Convert MTP head
+    logger.info("Converting %s MTP head to speculators format", verifier)
+    converter = MTPConverter()
+    converter.convert(
+        input_path=verifier,
+        output_path=str(converted_path),
+        base_model=verifier,
+        num_speculative_steps=num_speculative_tokens,
+        validate=False,
+    )
```
fynnsu, line 115: I still think this should be done differently (not with the python api).

tests/unit/models/test_mtp_data.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def test_shapes_unchanged(self):
+        seq_len, hidden_size = 32, 128
+        batch = _make_batch(seq_len=seq_len, hidden_size=hidden_size)
+        result = shift_batch_mtp(batch)
+        assert result["input_ids"].shape == (seq_len,)
+        assert result["hidden_states"].shape == (seq_len, hidden_size)
+        assert result["loss_mask"].shape == (seq_len,)
+        assert result["lengths"].shape == (1,)
+        assert result["position_ids"].shape == (seq_len,)
```
fynnsu: This again feels like a lot of tests for a function that just renames one field of a dict and drops a few others. Maybe can be combined.

fynnsu (commented):
Okay, a few more comments

fynnsu:
> Can you take a look at failed unit & link checks?

I think the failed link checker is okay. It is failing on https://github.com/vllm-project/speculators/blob/main/examples/train/mtp_qwen3next_gsm8k_offline.sh not existing, but it will exist as soon as the pr lands.

The unit tests need to be fixed tho.

tests/e2e/smoke/test_mtp_finetuning.py
fynnsu: But if you prefer to do so in a later pr (opened later today or Monday), that's okay.

scripts/train.py
rahul-tuli: It technically works, user might see edge cases we don't foresee, Updating to an error for now

src/speculators/models/mtp/core.py
rahul-tuli: Added back! and verified it works

src/speculators/models/mtp/data.py
rahul-tuli: We're only extracting 1 layer, see tests/example script

src/speculators/train/data.py
rahul-tuli: Good catch, verified

tests/e2e/smoke/test_mtp_finetuning.py
rahul-tuli: Done in a follow up!

tests/unit/models/test_mtp_data.py
rahul-tuli: Done!

src/speculators/train/data.py
```diff
@@ -111,11 +111,41 @@ def standardize_data_v1(data: dict[str, Any]) -> dict[str, Any]:
...
+    contains multimodal content.  Text-only conversations always go through
+    the Completions API with the pre-truncated ``input_ids``.
+
+    This matters for models like Qwen3.5-0.8B whose ``AutoProcessor`` returns
+    a ``ProcessorMixin`` (``Qwen3VLProcessor``), causing preprocessing to
+    populate the ``messages`` column even for purely text-only datasets.
+    Text-only EAGLE-3 models (e.g. Llama) use a plain tokenizer, so
+    ``messages`` is never created and this guard is a no-op.
+    """
+    out_dict: dict = {"input_ids": dataset_item["input_ids"].tolist()}
```
fynnsu, line 146: I'm not entirely clear on why this is needed now? I get the point about some conversations being untruncated if they're in the messages category. But then presumably their corresponding input ids will be of length `seq_length` (after truncation). Therefore, the batch sampler will assign them to their own batch. Then in the loading logic we call the `slice_and_pad_to_length` function on all the data, which will trim it down to length. So at worst we've wasted a bit of compute/data movement generating some hidden states.

fynnsu (approved):

src/speculators/train/data.py
rahul-tuli: Good question, this was a very subtle bug, the issue isn't about `slice_and_pad_to_length`, it's that the hidden states never get generated in the first place

When `messages` is present in the `ClientItem`, [`generate_hidden_states`](https://github.com/vllm-project/speculators/blob/main/src/speculators/data_generation/vllm_client.py#L229-L247) routes through `client.chat.completions.create()` ([line 241](https://github.com/vllm-project/speculators/blob/main/src/speculators/data_generation/vllm_client.py#L241)) instead of `client.completions.create()` ([line 233](https://github.com/vllm-project/speculators/blob/main/src/speculators/data_generation/vllm_client.py#L233)). The Chat Completions API **ignores `input_ids`** and re-tokenizes from the raw `messages` text. vLLM then validates the re-tokenized prompt length against `max_model_len` *before* generating any hidden states.

Since preprocessing only truncates `input_ids` (not `messages`), the re-tokenized length can far exceed `seq_length` — up to 8285 tokens in our test data. With `max_model_len = seq_length + 1 = 513`, vLLM rejects the request outright with a `VLLMValidationError`. No hidden states are ever produced, so `slice_and_pad_to_length` never gets a chance to run.

This was causing 232 validation errors in the MTP smoke test (29/50 samples failing × 4 attempts each). With the guard, zero errors — only multimodal conversations (which genuinely need `client.chat.completions.create` to carry image/video references) go through that path.

### Why Eagle-3 never needed this

Eagle-3 text-only models (e.g. Llama) use a plain `PreTrainedTokenizerBase`, not a `ProcessorMixin`. In [`_preprocess_batch`](https://github.com/vllm-project/speculators/blob/main/src/speculators/data_generation/preprocessing.py#L494-L496), the `messages` column is only created when `isinstance(processor, ProcessorMixin)` — which is `False` for plain tokenizers. So for Eagle-3 text-only flows, `messages` is never in the dataset, `build_client_item` never includes it, and every request goes through the Completions API with pre-truncated `input_ids`. The guard is a no-op.

The issue only surfaces for models whose `AutoProcessor` returns a `ProcessorMixin` even for text-only usage — like Qwen3.5, where `AutoProcessor.from_pretrained("Qwen/Qwen3.5-0.8B")` returns a `Qwen3VLProcessor`. This causes the `messages` column to be created for every sample, including purely text-only ones, which then get routed through Chat Completions and fail validation.

For multimodal Eagle-3 flows (e.g. sharegpt4v_coco), `messages` is present and the content is list-typed (e.g. `[{"type": "text", ...}, {"type": "image_url", ...}]`), so `_has_multimodal_content` returns `True` and `messages` is correctly forwarded — no change in behavior.

shanjiaz (approved):

src/speculators/train/data.py
fynnsu: Okay, but I think we can also just use a larger `max_model_len`. I think I might've set it to `seq_length+1` in the smoke tests or something cause that was all that was needed there, but we can up it to a larger number. Or just not set it at all and have it get set to the default for the model.

PR #373, opened 2026-03-30
fix(e2e): pass --legacy-data for .pt-format dataset in finetuning sanity test

rahul-tuli:
## Problem

`tests/e2e/vllm/test_finetuning_sanity.py::test_finetuning_weight_sanity` was failing with:

```
FileNotFoundError: Directory .../snapshots/4764848000dd06e30dd831ddb867ffae4d7fa6d7
is neither a Dataset directory nor a DatasetDict directory.
```

## Root Cause

The test downloads `nm-testing/sharegpt_llama3_8b_hidden_states` via `snapshot_download` and passes the path to `train.py`. The snapshot contains 50 `data_N.pt` files in the legacy v1 format — there is no `dataset_info.json` or `dataset_dict.json`.

Without `--legacy-data`, `train.py` uses `Eagle3ArrowDataset`, which calls `datasets.load_from_disk()`. That function requires Arrow metadata files at the root → **FileNotFoundError**.

## Fix

Add `--legacy-data` to the subprocess args in the test. This routes training through `Eagle3SampleFileDataset`, which correctly loads `.pt` files via `standardize_data_v1`.

## Follow-up

Uploading a proper Arrow-format dataset to replace `nm-testing/sharegpt_llama3_8b_hidden_states` is tracked separately, which will allow removing `--legacy-data` from this test.

## Test plan
- [ ] CI e2e test `test_finetuning_weight_sanity` passes with this fix https://github.com/neuralmagic/llm-compressor-testing/actions/runs/23753746293


docs/developer/index.md
```diff
@@ -1,6 +1,6 @@
----
-weight: -3
----
+______________________________________________________________________
```
shanjiaz: Is this change needed?

tests/unit/convert/test_eagle3_converter.py
```diff
@@ -262,12 +262,14 @@ def test_nm_testing_2layer_eagle3_model_config(self, tmp_path):
 
         converter.convert(
             checkpoint_path,
-            tmp_path / checkpoint_path.split("/")[-1],
+            tmp_path / checkpoint_path.rsplit("/", maxsplit=1)[-1],
```
shanjiaz: Is this to fix format?

tests/e2e/vllm/test_finetuning_sanity.py
```diff
@@ -86,6 +86,7 @@ def test_finetuning_weight_sanity(tmp_path: Path):
             "2048",
             "--num-workers",
             "2",
+            "--legacy-data",
```
shanjiaz, line 89: I think we have to update test_gen_and_train_acceptance.py and test_train.py as well. There were 3 failed tests.

docs/developer/index.md
rahul-tuli: Absolutely not, removing! Apologies

tests/unit/convert/test_eagle3_converter.py
rahul-tuli: Yes

tests/e2e/vllm/test_finetuning_sanity.py
rahul-tuli: Done!

PR #388, opened 2026-04-09
Expand E2E testing

fynnsu:

## Purpose

Refactors E2E testing into reusable components, so that it is easier to add new tests. Also adds regression tests for the new online/offline datagen + training + evaluation system.

## Description

Factors out some of the e2e logic into:
- `run_data_generation_offline2`
- `run_training`
- `run_prepare_data`
- `launch_vllm_server`

Which can then we reused to create the different e2e workflows. Added a bunch of optional arguments to those e2e workflows allowing both smoke and regression tests to use the same workflow functions.

Adds new 5k sample regression tests for the online/offline e2e flows. 

## Tests

Will manually trigger e2e test. 

I have filled in:

### DIFF
--- a/tests/e2e/smoke/test_offline_training.py
+++ b/tests/e2e/smoke/test_offline_training.py
@@ -9,109 +9,85 @@ Exercises the full offline pipeline:
   6. Validate the trained checkpoint via vLLM inference (run_vllm_engine)
 """
 
-import subprocess
-import sys
 from pathlib import Path
 
 import pytest
-from loguru import logger
 
 from tests.e2e.utils import (
-    SCRIPTS_DIR,
-    launch_vllm_server,
-    prepare_data,
+    launch_vllm_server_context,
+    run_data_generation_offline2,
+    run_prepare_data,
+    run_training,
     run_vllm_engine,
-    stop_vllm_server,
 )
 
 MODEL = "Qwen/Qwen3-0.6B"
-VLLM_PORT = 8322
-
-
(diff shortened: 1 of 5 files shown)


tests/e2e/smoke/test_offline_training.py
```diff
@@ -9,109 +9,89 @@
...
 
 @pytest.mark.e2e
 @pytest.mark.slow
-def test_offline_training(
-    tmp_path: Path, prompts: list[list[dict[str, str]]], vllm_server
+def test_offline_smoke(tmp_path: Path, prompts: list[list[dict[str, str]]]):
+    run_offline_e2e(tmp_path, MODEL, prompts=prompts, vllm_gpu_util=0.9)
+
+
+def run_offline_e2e(
```
shanjiaz, line 33: Should we move this function to `utils.py` as well so smoke and regression tests look symmetric? Or perhaps create a `training_test_utils.py` to avoid overcomplicating `utils.py`? Not a blocker.

shanjiaz (commented):
Thanks for adding the regression tests!

shanjiaz (approved):

tests/e2e/smoke/test_offline_training.py
fynnsu: Yeah, probably but also this is not bad either since it localizes the test logic next to the test (for smoke tests).

rahul-tuli (approved):

PR #389, opened 2026-04-09
[Tests][CI/CD] Add regression testing for training; enable cadence set-up

dsikka:
## Purpose

- Introduce a `requires_cadence` decorator to gate tests by run frequency (`commit` / `nightly` / `weekly`) via a `CADENCE` environment variable. Currently only applied to regression tests so that we can split when certain regression tests run / don't all run together
- Apply `@requires_cadence("nightly")` to all existing e2e regression tests 
- Add regression tests that isolate training acceptance from data generation. These tests currently runs on a weekly cadence but we can also just alternate the nights when each set of regression tests run for the time being (i.e instead of nightly vs weekly, run group_1 vs group_2 on alternating nights - but this might be overkill)
- All other tests remain unchanged
- Add small updates to the verbosity of our CI/CD so that we can see why a particular test is skipped
- Update dflash target layer handling to write back to the config if using the default target layers - this is required for vLLM 

I also found it unclear as to what the required / expected dataset directory structure was, the requirement for an arrow dataset, and what happens if you don't provide token_freq.pt - I tried adding some details to make this clear but we should spend time highlighting these nuances. 

## Testing
Test flow: https://github.com/neuralmagic/llm-compressor-testing/actions/runs/24734129413
- All new tests pass


dsikka:
@mergify refresh

tests/e2e/regression/test_training_only_acceptance.py
```diff
@@ -0,0 +1,178 @@
...
+    d2t, t2d = build_vocab_mappings_from_distribution(
+        token_freq_dict=token_freq_dict,
+        draft_vocab_size=draft_vocab_size,
+        target_vocab_size=target_vocab_size,
+    )
+    np.save(d2t_path, d2t.cpu().numpy())
+    np.save(t2d_path, t2d.cpu().numpy())
+
+
+def _resolve_repo(repo_id: str, repo_type: str = "dataset") -> Path:
```
shanjiaz, line 11: Can we move this to `tests.e2e.utils`? Could be useful for other tests that uses pre-generated data

shanjiaz (approved):

tests/e2e/regression/configs/training/eagle3_llama3_8b_sharegpt.json
```diff
@@ -0,0 +1,14 @@
+{
+    "name": "eagle3_llama3_8b_sharegpt",
+    "verifier_model": "meta-llama/Llama-3.1-8B-Instruct",
+    "hidden_states_repo": "inference-optimization/llama3-8b-sharegpt-5k-gen",
```
fynnsu: This uses our old data system. We should update this with a new dataset in the new data format.

tests/e2e/regression/test_training_only_acceptance.py
```diff
@@ -0,0 +1,178 @@
...
+            and item.suffix == ".pt"
+            or item.is_file()
+            and item.name == "sample_lengths.json"
+        ):
+            (data_dir / item.name).symlink_to(item)
+    return data_dir
+
+
+def _run_training(args: dict) -> subprocess.CompletedProcess:
+    cmd = ["python", "scripts/train.py"]
```
fynnsu: Can we use `tests/e2e/utils.py` `run_training` instead of creating a new one? we can update the logic if needed but good to have only one of these.

tests/e2e/regression/test_training_only_acceptance.py
```diff
@@ -0,0 +1,178 @@
...
+        d2t_path=d2t_path,
+        t2d_path=t2d_path,
+        draft_vocab_size=config["draft_vocab_size"],
+        target_vocab_size=config["target_vocab_size"],
+    )
+    return d2t_path, t2d_path
+
+
+def _make_training_data_dir(hidden_states_dir: Path, tmp_path: Path) -> Path:
+    """Return a directory containing only the top-level .pt training files.
```
fynnsu: I'm not sure why we need this. The `token_freq` file shouldn't be placed in the same subdirectory as the training samples. We should also be switching to the new data format anyways but even if we want to support the old format, we should just make sure the huggingface checkpoint has the right file structure, rather than correcting it on download.

tests/e2e/regression/test_training_only_acceptance.py
```diff
@@ -0,0 +1,178 @@
...
+            snapshot_download(
+                repo_id=repo_id, repo_type=repo_type, local_files_only=True
+            )
+        )
+    except LocalEntryNotFoundError:
+        return Path(snapshot_download(repo_id=repo_id, repo_type=repo_type))
+
+
+def _resolve_vocab_mappings(config: dict, tmp_path: Path) -> tuple[Path, Path]:
+    """Return (d2t_path, t2d_path).
```
fynnsu: This is fine, although not technically needed as `train.py` will now create the vocab mappings if they don't exist yet. You just have to pass in `--draft-vocab-size` to the training script.

fynnsu: And not sure why we need both this function and `_build_vocab_mappings` above.

fynnsu (changes requested):
Some suggestions below. 

I think the config system looks good and enforces some standardization but I'm also not entirely sold on need this vs just setting up runs like this:

https://github.com/vllm-project/speculators/blob/10272113d059f8d2c81ae55c6e4df5ee74bf4800/tests/e2e/regression/test_eagle3_online_acceptance.py#L16-L27

That's just a python based "config" instead of a json one, and also allows more flexibility if needed.

tests/e2e/regression/configs/training/eagle3_llama3_8b_sharegpt.json
shanjiaz: Could we actually add handling for the new data format? (safetensors & pt)

shanjiaz (changes requested):

tests/e2e/regression/configs/training/eagle3_llama3_8b_sharegpt.json
dsikka: Yes - this is what I was asking about during standup, if we could share a dataset from the new system for both dflash and eagle 3

tests/e2e/regression/test_training_only_acceptance.py
dsikka: yup

dsikka:
@coderabbitai rereview

examples/ONLINE_TRAINING.md
```diff
@@ -52,11 +52,37 @@ CUDA_VISIBLE_DEVICES=4,5,6,7 torchrun --standalone --nproc_per_node 4 scripts/tr
...
+    t2d.npy                      #  ⎤ Vocab mappings
     d2t.npy                      #  ⎦
     
     checkpoints/                 # Training checkpoints (loadable by vLLM)
         0/
         1/
         ...
 ```
+
+> **Note:** If using the offline data generation system (`scripts/data_generation_offline2.py`), hidden states are pre-generated and cached to disk rather than generated on-demand during training. In that case, the data directory will also contain a `hidden_states/` subdirectory:
```
shanjiaz: Seems a little weird to talk about offline in ONLINE_TRAINING.md but we could probably move both to docs instead of readmes.

tests/e2e/regression/configs/training/eagle3_qwen3_8b_sharegpt.json
```diff
@@ -0,0 +1,12 @@
+{
+    "name": "eagle3_qwen3_8b_sharegpt",
+    "verifier_model": "Qwen/Qwen3-8B",
+    "hidden_states_repo": "inference-optimization/Qwen3-8b-sharegpt-5k",
+    "training": {
+        "lr": 3e-4,
+        "total_seq_len": 8192,
+        "epochs": 5
+    },
```
shanjiaz: Can we add max_samples:5000 here? Otherwise I think we'll use the default value of 50.

examples/ONLINE_TRAINING.md
dsikka: yeah I didn't want to start a new doc as Fynn is restructuring and we dont have any other doc which references the new system so no good options

tests/e2e/regression/configs/training/eagle3_qwen3_8b_sharegpt.json
dsikka: I dont think that's true. The max_samples is used for data preparation but we already have prepared data so all 5k samples are used.

0.4 for first position also doesn't make sense for 50 samples

shanjiaz: Makes sense!

shanjiaz (approved):

rahul-tuli (approved):
LGTM! Thanks for taking this on.

tests/utils.py
```diff
@@ -0,0 +1,13 @@
+import os
+from collections.abc import Callable
+
+import pytest
+
+
+def requires_cadence(cadence: str | list[str]) -> Callable:
+    cadence = [cadence] if isinstance(cadence, str) else cadence
+    current_cadence = os.environ.get("CADENCE", "commit")
```
fynnsu, line 9: We never actually use "commit" cadence right? Should we have this default to nightly?

Also are we already setting the cadence in llm-compressor-testing or do we need a pr to update there as well?

src/speculators/models/dflash/core.py
```diff
@@ -71,15 +71,17 @@ def __init__(
...
             self.target_layer_ids = config.aux_hidden_state_layer_ids
         else:
+            # Eagle3 defaults; write back so they are persisted in config.json
             self.target_layer_ids = [
                 2,
                 num_verifier_layers // 2,
                 num_verifier_layers - 3,
             ]
+            # set defaults to config if not provided - vLLM will fail otherwise
+            config.aux_hidden_state_layer_ids = self.target_layer_ids
```
fynnsu, line 84: This is fine for now, but we should standardize how we do this across both eagle3 and dflash.

@coderabbitai can you create an issue for this?

Currently in `eagle3/core.py` we resolve `target_layer_ids` in the `from_training_args` function and then pass that in when creating the `Eagle3SpeculatorConfig`. We should probably do the same for DFlash. We can also combine the logic for getting the `num_verifier_layers` into a util function that is used by both.

fynnsu (commented):
Soft approval pending cadence questions.

tests/utils.py
dsikka: No but we should use it as a way to enable e2e testing on a per commit testing. 

The llm-compressor-testing PR has landed. This is how the linked test runs were done.

fynnsu (approved):

PR #411, opened 2026-04-13
Fix optimizer resume bug

shanjiaz:

## Purpose
Fix optimizer resume dtype error.    

## Description

When we resume distributed training from the previous `best_checkpoint`, there is a type mismatch bug.
```
RuntimeError: aten._foreach_mul_.Scalar: got mixed torch.Tensor and DTensor, need to convert all torch.Tensor to DTensor before calling distributed operators!
```
This is because `init_best_val_loss_from_checkpoint_best()` was called inside `setup_model()`,  we ran a forward pass in validation to calculate the `best_val_loss`. This forward pass converts model parameters into DTensors with some other bookkeeping params and `set_optimizer_state_dict` then crashes because it was expecting clean tensors. 

There's another issue with `step` data type now because `convert_float_dtype` casts all floating-point tensors to bfloat16 when saving/loading, including Adam's step counters. PyTorch's `_group_tensors_by_device_and_dtype` only exempts CPU step tensors that are float32 or float64, so bfloat16 step tensors cause a  device/dtype grouping error at optimizer.step().   

Now we're saving `best_val_loss` in a json along with best checkpoint and loading it directly instead of running a full validation forward pass to avoid this type mismatch and casting step dtype back to float32.

## Tests

Added `/home/shanjiaz/speculators/tests/e2e/smoke/test_resume_optimizer.py` and ran locally to make sure it captures this bug properly. Had to update `tests/unit/train/test_checkpoint.py` as well because the flow is now different. 

I have filled in:

### DIFF
--- a/src/speculators/train/checkpointer.py
+++ b/src/speculators/train/checkpointer.py
@@ -111,20 +111,24 @@ class BaseCheckpointer:
     def best_path(self) -> Path:
         return self.path / "checkpoint_best"
 
-    def best_val_loss_path(self) -> Path:
-        return self.path / "checkpoint_best" / "metrics.json"
+    def val_metrics_path(self, epoch: int) -> Path:
+        return self.path / str(epoch) / "val_metrics.json"
 
-    def save_best_val_loss(self, val_loss: float):
-        self.path.mkdir(parents=True, exist_ok=True)
-        self.best_val_loss_path().write_text(json.dumps({"best_val_loss": val_loss}))
+    def save_val_metrics(self, epoch: int, val_metrics: dict[str, float]):
+        path = self.val_metrics_path(epoch)
+        path.parent.mkdir(parents=True, exist_ok=True)
+        path.write_text(json.dumps(val_metrics))
 
     def load_best_val_loss(self) -> float | None:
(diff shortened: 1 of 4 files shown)


src/speculators/train/checkpointer.py
```diff
@@ -110,6 +111,23 @@ def scheduler_path(self, epoch: int):
     def best_path(self) -> Path:
         return self.path / "checkpoint_best"
 
+    def best_val_loss_path(self) -> Path:
+        return self.path / "best_val_loss.json"
+
```
fynnsu, line 116: It might be better to store this in `self.path / "checkpoint_best" / "metrics.json"` where checkpoint_best is a symlink to the real epoch file. 

We can then just save this metrics.json for every epoch when we save the checkpoint (since we're computing it for every one anyways. That way we're consistent for all epochs and also get more information if we ever want to go back and see how the different epochs compare.

shanjiaz: sure!

fynnsu (approved):
Thanks for updating

rahul-tuli (approved):

PR #412, opened 2026-04-14
Add timeouts/retries to all vllm requests

fynnsu:

## Purpose

Improve the robustness of our online/offline data generation with better handling for failed requests to vllm.

We had a nightly e2e CI run that just hung on data_generation_offline2.py. This pr ensures every vllm requests has appropriate timeouts with retry logic and handling when that fails.

## Description

Requests to vLLM (in both online and offline training) will now have a timeout (default 15s) and max retries (default 3).

The data_generation_offline2.py script was also made more robust. It now has a `fail-on-error`  flag that will cause any issue with a sample that can't be handled by the timeout/retries to stop the process. 

Without the flag, the failing samples will be skipped. Except if `--max-consecutive-errors` fail in a row, the script will still crash. This is better handling for cases where the vllm server process goes down/becomes unreachable. The script will attempt a few samples, but if they're all failing it will terminate early. 

## Related Issue

https://github.com/neuralmagic/llm-compressor-testing/actions/runs/24380242770/job/71202006249#step:10:32430

## Tests

Verified manually that the script crashes when vllm becomes unreachable. 

I have filled in:

### DIFF
--- a/src/speculators/data_generation/vllm_client.py
+++ b/src/speculators/data_generation/vllm_client.py
@@ -1,4 +1,5 @@
 import asyncio
+import functools
 import logging
 import time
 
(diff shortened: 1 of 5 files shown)


src/speculators/train/data.py
```diff
@@ -292,8 +297,14 @@ def _maybe_generate_hs(self, index: int) -> dict[str, torch.Tensor] | None:
...
+            hs_filepath = generate_hidden_states(
+                self.client,  # type:ignore[arg-type]
+                self.model,  # type:ignore[arg-type]
+                input_ids,
+                timeout=self.request_timeout,
+                max_retries=self.max_retries,
+            )
+        except Exception as e:  # noqa: BLE001
             warnings.warn(str(e), stacklevel=1)
             return None
```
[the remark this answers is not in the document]
fynnsu: No, I want training to not crash even if there are other issues with the data gen.

shanjiaz (approved):
Talked about this in person. Diff looks good.

src/speculators/data_generation/vllm_client.py
```diff
@@ -23,30 +33,102 @@ def extract_output(completion, token_ids) -> str:
...
+    max_retries: int = DEFAULT_MAX_RETRIES,
 ) -> str:
-    completion = client.completions.create(
-        model=model,
-        prompt=token_ids,
-        max_tokens=1,
-        extra_body={"return_token_ids": True},
-    )
-    return extract_output(completion, token_ids)
+    total_attempts = max_retries + 1
```
rahul-tuli: This works, but it feels like the function is doing too much, I think the retry logic would be cleaner as a decorator? something like
```
@retry(timeout=..., max_retries=...)
def generate_hidden_states(...)
``` 

That way the generate_hidden_states states as is?

rahul-tuli (commented):
One suggestion, otherwise LGTM!

src/speculators/data_generation/vllm_client.py
dsikka: I also agree that this is likely better suited for a decorator

fynnsu:
@mergify refresh

src/speculators/data_generation/vllm_client.py
fynnsu: Updated

src/speculators/data_generation/vllm_client.py
```diff
@@ -22,31 +99,54 @@ def extract_output(completion, token_ids) -> str:
     return completion.kv_transfer_params.get("hidden_states_path")
 
 
+@with_retries
```
rahul-tuli, line 102: So beautiful!

rahul-tuli (approved):

rahul-tuli:
@coderabbit re-review

Issue #356, opened 2026-03-23
[RFC]: Add Multi-node Training Suppor


Liccol:
### Motivation.

# RFC: Multi-node Training Support for Speculators

## Introduction

This RFC proposes the addition of multi-node training support . The changes aim to enable efficient distributed training across multiple nodes, allowing the project to scale to larger models and datasets.

## Motivation

As the Speculators project continues to evolve, there is a growing need to support training on multi-node clusters. Single-node training with multiple GPUs has limitations in terms of memory and computational capacity, especially for larger models. Multi-node training enables:

- Scaling to larger model sizes beyond the memory capacity of a single node
- Faster training times by distributing the workload across multiple nodes
- Access to more computational resources for experiments and production deployments

### Multi-node Training Parameters

Add new parameters to `scripts/gen_and_train.py` to support multi-node training:

| Parameter       | Description                  | Default Value         |
|-----------------|------------------------------|-----------------------|
| `nproc_per_node`| Number of processes per node | `torch.accelerator.device_count()` |
| `nnodes`        | Total number of nodes        | 1                     |
| `node_rank`     | Rank of the current node     | 0                     |
| `master_addr`   | IP address of the master node| None                  |
| `master_port`   | Port of the master node      | 12345                 |

### Rank Parameter Correction

- Correct the usage of `local_rank` to `rank` in `scripts/train.py` for the `setup_dataloader` function
- Update `MultipackDistributedBatchSamplerV2` calls to use global rank instead of local rank for proper data sharding
- Fix documentation in `src/speculators/train/distributed_batch_sampler.py` to clarify the use of global rank

### Command Execution Optimization

- Enhance `build_torchrun_command` to handle unset parameters gracefully
- Implement logic to automatically select the appropriate command format based on the number of nodes:
  - Single-node training: `torchrun --standalone --nproc_per_node={device_count}`
  - Multi-node training: Full torchrun command with all distributed parameters

### Documentation Updates

- Add detailed documentation for multi-node training parameters in `scripts/README.md`
- Include multi-node training startup examples
- Add documentation for scheduler-related parameters

### Single-node Training

```bash
torchrun --standalone --nproc_per_node=4 scripts/train.py
```

### Multi-node Training

```bash
# Node 0
torchrun --nnodes=2 --nproc_per_node=4 --node_rank=0 --master_addr=192.168.1.100 --master_port=12345 scripts/train.py

# Node 1
[...]

PR #299, opened 2026-02-24
add multinode train support

Liccol:
This PR adds multi-node training support. The main changes include adding multi-node training parameters, correcting rank parameter usage, and improving README.

# Changes
## 1. Multi-node Training Parameters
- Added 5 new multi-node training parameters in scripts/gen_and_train.py:
  - nproc_per_node: Number of processes per node
  - nnodes: Number of nodes
  - node_rank: Current node rank
  - master_addr: Master node address
  - master_port: Master node port
## 2. Rank Parameter Correction
- Corrected some local_rank usage, such as MultipackDistributedBatchSamplerV2 using global rank for data sharding
## 3. README Updates
- Added detailed documentation for multi-node training parameters in scripts/README.md
- Added multi-node training startup example commands
- Added documentation for scheduler-related parameters
# Testing
## Single-node Training
```
python scripts/data_generation_offline.py --train-data-path sharegpt --turn-dropout --seq-length 2048 --target-model-path /home/data/weights/Qwen3-30B-A3B/ --token-freq-path /home/data/ckpt/qwen3-30b-eagle-chat-temp07-3w-2node/vocab_mapping/token_freq_sharegpt.pt --output-dir /home/data/ckpt/qwen3-30b-eagle-chat-temp07-3w-2node/gen/sharegpt

python scripts/build_vocab_mapping.py --draft-vocab-size 32000 --target-vocab-size 151936 --token-freq-path /home/data/ckpt/qwen3-30b-eagle-chat-temp07-3w-2node/vocab_mapping/token_freq_sharegpt.pt --output-path /home/data/ckpt/qwen3-30b-eagle-chat-temp07-3w-2node/vocab_mapping

[...]

PR #427, opened 2026-04-15
Fix local_rank usage in distributed batch sampler for multi-node training

ianliuy:
## Summary

Fix data duplication bug in multi-node distributed training caused by passing `local_rank` instead of global `rank` to `MultipackDistributedBatchSamplerV2`.

In multi-node setups, `local_rank` is per-node (e.g., 0–3 on each node), so processes on different nodes with the same `local_rank` receive identical data shards — every GPU trains on duplicated data.

## What's broken?

When training with `nnodes > 1`, data sharding is incorrect. For example, in a 2-node × 4-GPU setup:
- Node 0, GPU 0: `local_rank=0, rank=0` → gets shard 0
- Node 1, GPU 0: `local_rank=0, rank=4` → **also gets shard 0** (should get shard 4)

Additionally, tqdm progress bars appear on every node's GPU 0 instead of just the global rank 0 process.

Single-node training is unaffected because `local_rank == rank` when `nnodes=1`.

## Changes

- **`scripts/train.py`**: `setup_dataloader()` now accepts and passes global `rank` (instead of `local_rank`) to `MultipackDistributedBatchSamplerV2` for correct data sharding. Added rank logging after distributed setup.
- **`src/speculators/train/trainer.py`**: Added `self.rank` (from `RANK` env var) and use it for tqdm progress bar guards instead of `self.local_rank`.
- **`src/speculators/train/utils.py`**: Include global rank in distributed setup log message.

## Context

Rebased version of #299 by @Liccol, which has been stale with merge conflicts. All review feedback from @fynnsu on that PR has been incorporated.

Part of #356.

## Testing

- Single-node: no behavioral change (`local_rank == rank` when `nnodes=1`)
- Multi-node: each rank now correctly receives unique data shards via `MultipackDistributedBatchSamplerV2`
- Lint: `ruff check` and `ruff format` pass on all changed files

cc @fynnsu

### DIFF
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -55,7 +55,7 @@ def set_seed(seed: int, deterministic: bool = False):
 def setup_dataloader(
     dataset: BaseEagle3Dataset,
     world_size: int,
-    local_rank: int,
+    rank: int,
     hidden_size: int,
     num_workers: int = 12,
     prefetch_factor: int = 4,
@@ -64,7 +64,7 @@ def setup_dataloader(
     Args:
         file_list: List of file paths to load data from.
         world_size: Number of processes in the distributed training.
-        local_rank: Rank of the current process.
+        rank: Global rank of the current process.
         add_noise: Whether to add noise to the data.
         noise_std: Standard deviation for noise augmentation.
         num_workers: Number of dataloader workers.
@@ -76,7 +76,7 @@ def setup_dataloader(
         batch_max_length=args.total_seq_len,
         lengths=dataset.approx_lengths,
         num_replicas=world_size,
-        rank=local_rank,
+        rank=rank,
     )
     return DataLoader(
         dataset,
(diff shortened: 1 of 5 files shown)


src/speculators/train/trainer.py
fynnsu, line 40: We should just replace this with `rank`.
That way we can also pass the value in (the way we currently do with local_rank).

src/speculators/train/trainer.py
```diff
@@ -58,6 +59,7 @@ def __init__(
     ):
         self.model = model
         self.config = config
+        self.rank = int(os.environ.get("RANK", "0"))
         self.local_rank = config.local_rank
```
fynnsu: Use `self.rank = config.rank` and remove the `self.local_rank` attribute if it's no longer used.

scripts/train.py
```diff
@@ -213,6 +213,9 @@ def main(args: argparse.Namespace):
 
     # Setup distributed training
     local_rank, world_size, rank, is_distributed = maybe_setup_distributed()
+    logger.info(
+        f"Using RANK: {rank}, LOCAL_RANK: {local_rank}, WORLD_SIZE: {world_size}"
+    )
```
fynnsu: We have the updated `logger.info` call at the end of `maybe_setup_distributed()` so we don't have to log this again here.

fynnsu (changes requested):
Hi, thanks for submitting the pr. I've added a few small requests below.

fynnsu:
Hi @ianliuy, sorry this kind of got pushed to the side because I didn't have good way to test it. Could you please fix DCO check for your last commit? I think we can merge it then

Liccol:
Hi @ianliuy , thank you for continuing my PR, plz help fix DCO and merge it, so we can use it directly on main branch~

fynnsu:
I think #619 largely fixed this

PR #431, opened 2026-04-16
fix: resolve Pydantic v2 + transformers MRO conflicts

### LINKED ISSUE #370: transformers >= 4.57 + Pydantic v2: runtime crashes and mypy errors in speculators

ianliuy:
## Summary

Fix three runtime crashes, one type annotation mismatch, one forward-compatibility issue, and one integration test failure caused by newer transformers' `PretrainedConfig` changes interacting with Pydantic v2.

All fixes use standard Python MRO resolution patterns for dual inheritance and are backwards-compatible with older transformers versions.

Closes #370.

---

### 1. Import-time crash: `PydanticUndefinedAnnotation: name 'torch' is not defined`

**Root cause:** `reload_schemas()` calls `model_rebuild(force=True)`. Newer transformers added `dtype: Union[str, "torch.dtype"] | None` to `PretrainedConfig`. Pydantic evaluates this forward reference but `torch` is not in the evaluation namespace.

**Fix:** Override `reload_schema()` on `SpeculatorModelConfig` to pass `_types_namespace={"torch": torch}` — this is the [official Pydantic v2 API](https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_rebuild) for providing external types during schema rebuild. Also rebuilds registered subclasses (parent rebuild does not propagate).

**Backwards compat:** Passing extra keys to `_types_namespace` is harmless on older transformers that lack the `torch.dtype` annotation.

### 2. Config construction crash: `AttributeError: '__pydantic_fields_set__'`

**Root cause:** Newer transformers' `PretrainedConfig.__init_subclass__` applies `@dataclass(repr=False)` to subclasses lacking `__init__` in `cls.__dict__`. This replaces the inherited Pydantic `__init__` with a dataclass wrapper that calls `setattr()` before Pydantic initializes `__pydantic_fields_set__`.

**Fix:** `__init_subclass__()` hook injects the parent's `__init__` into each subclass's `__dict__` **before** calling `super().__init_subclass__()`. The transformers check `"__init__" in cls.__dict__` then passes, skipping the dataclass wrapping.

**Backwards compat:** Older transformers do not apply `@dataclass` in `__init_subclass__`; the injection is a no-op.

### 3. `save_pretrained` crash: `TypeError: BaseModel.validate() missing 1 required positional argument`

**Root cause:** Pydantic's `BaseModel.validate(cls, value)` classmethod appears earlier in MRO than the `validate()` instance method that newer transformers' `@strict` decorator adds. `save_pretrained()` calls `self.validate()` and hits the wrong one.

**Fix:** Explicit `validate()` instance method on `SpeculatorModelConfig` that delegates to `PretrainedConfig.validate(self)` with a `hasattr` guard.

**Backwards compat:** Older transformers have no `validate()` on `PretrainedConfig`; the `hasattr` guard correctly skips delegation.

---

### 4. `base_model_pp_plan` type annotation

Newer transformers widened `PretrainedConfig.base_model_pp_plan` from `tuple[list[str]]` to `Sequence[list[str]]`. Aligned our annotation.

### 5. `cache_position` in `Eagle3FirstLayerMixin.forward()`

Newer transformers removed `cache_position` from the explicit parameter list of `LlamaDecoderLayer.forward()` / `Qwen3DecoderLayer.forward()` (it now flows through `**kwargs`). Removed it from our mixin — it flows through `**kwargs` in both old and new transformers.

### 6. Integration test `test_verifier_config_from_verifier_config`

Newer transformers strictly validates `rope_scaling` and requires `rope_theta` inside it. The Hub config for `RedHatAI/Llama-3.1-8B-Instruct` predates this requirement. Switched to `get_config_dict()` (raw fetch, no validation) and dropped `rope_scaling` since the test only needs `architectures`.

---

## What this PR does NOT do

- Does **not** use dict unpacking to bypass mypy `[call-arg]` errors on `LlamaConfig(...)` — those are upstream typing issues in transformers' `@strict` decorator and should be resolved there.
- Does **not** add any `# type: ignore` suppressions for upstream issues.

## Files changed (4)

| File | Changes |
|------|---------|
| `src/speculators/config.py` | `__init_subclass__`, `validate()`, `reload_schema()`, `base_model_pp_plan` type |
| `src/speculators/models/eagle3/model_definitions.py` | Remove `cache_position` from mixin |
| `tests/integration/test_config.py` | Use `get_config_dict` for test |
| `pyproject.toml` | Remove transformers upper bound |

### DIFF
--- a/tests/integration/test_config.py
+++ b/tests/integration/test_config.py
@@ -23,6 +23,8 @@ def test_verifier_config_from_verifier_config():
             "RedHatAI/Llama-3.1-8B-Instruct",
             cache_dir=tmp_dir,
         )
+    # TODO(ianliuy): Remove this workaround once the upstream config includes
+    # the rope_scaling fields required by newer transformers validation.
     config_dict.pop("rope_scaling", None)
     pretrained_config = PretrainedConfig(**config_dict)
 
--- a/tests/integration/test_config.py
+++ b/tests/integration/test_config.py
@@ -17,10 +17,14 @@ from speculators import (
 @pytest.mark.smoke
 def test_verifier_config_from_verifier_config():
     with tempfile.TemporaryDirectory() as tmp_dir:
-        pretrained_config = PretrainedConfig.from_pretrained(
-            pretrained_model_name_or_path="RedHatAI/Llama-3.1-8B-Instruct",
+        # Use get_config_dict (raw fetch, no validation) to avoid strict
+        # rope_scaling validation added in newer transformers.
+        config_dict, _ = PretrainedConfig.get_config_dict(
+            "RedHatAI/Llama-3.1-8B-Instruct",
             cache_dir=tmp_dir,
         )
+    config_dict.pop("rope_scaling", None)
+    pretrained_config = PretrainedConfig(**config_dict)
 
     config = VerifierConfig.from_config(
         pretrained_config, name_or_path="RedHatAI/Llama-3.1-8B-Instruct"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,7 +51,7 @@ dependencies = [
     "setuptools",
     "torch>=2.9.0,<=2.11.0",
     "tqdm>=4.66.3,<=4.67.3",
-    "transformers>=4.56.1,<5.4.0",
+    "transformers>=4.56.1",
     "typer>=0.12.0",
 ]
 
--- a/src/speculators/config.py
+++ b/src/speculators/config.py
@@ -19,7 +19,6 @@ Classes:
 """
 
 import os
-from collections.abc import Sequence
 from importlib.metadata import version
 from typing import Any, ClassVar
 
@@ -237,7 +236,7 @@ class SpeculatorModelConfig(PydanticClassRegistryMixin, PretrainedConfig):
     is_composition: ClassVar[bool] = False  # type: ignore[misc]
     attribute_map: ClassVar[dict[str, str]] = {}  # type: ignore[misc]
     base_model_tp_plan: ClassVar[dict[str, Any] | None] = None  # type: ignore[misc]
-    base_model_pp_plan: ClassVar[dict[str, Sequence[list[str]]] | None] = None  # type: ignore[misc]
+    base_model_pp_plan: ClassVar[dict[str, tuple[list[str]]] | None] = None  # type: ignore[misc]
     _auto_class: ClassVar[str | None] = ""  # type: ignore[misc]
 
     # Speculator model instance attributes
@@ -271,7 +270,7 @@ class SpeculatorModelConfig(PydanticClassRegistryMixin, PretrainedConfig):
         # ensure we always update the transformers version
         self.transformers_version = version("transformers")
 
-    def validate(self) -> None:
+    def validate(self) -> None:  # type: ignore[override]
         """Resolve MRO conflict between BaseModel.validate (classmethod) and
         PretrainedConfig.validate (instance method added in transformers v5).
 
--- a/src/speculators/config.py
+++ b/src/speculators/config.py
@@ -19,9 +19,11 @@ Classes:
 """
 
 import os
+from collections.abc import Sequence
 from importlib.metadata import version
 from typing import Any, ClassVar
 
+import torch
 from pydantic import BaseModel, ConfigDict, Field
 from transformers import PretrainedConfig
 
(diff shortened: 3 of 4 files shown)


rahul-tuli:
@fynnsu can you take a look, it looks good to me

tests/integration/test_config.py
```diff
@@ -17,10 +17,14 @@
...
     with tempfile.TemporaryDirectory() as tmp_dir:
-        pretrained_config = PretrainedConfig.from_pretrained(
-            pretrained_model_name_or_path="RedHatAI/Llama-3.1-8B-Instruct",
+        # Use get_config_dict (raw fetch, no validation) to avoid strict
+        # rope_scaling validation added in newer transformers.
+        config_dict, _ = PretrainedConfig.get_config_dict(
+            "RedHatAI/Llama-3.1-8B-Instruct",
             cache_dir=tmp_dir,
         )
+    config_dict.pop("rope_scaling", None)
```
shanjiaz, line 28: Can you add a TODO here so we can revisit? Thanks!

shanjiaz (commented):
@ianliuy Thanks so much for your contribution. We really appreciate it! Just a quick note for next time, we have a PR description template that helps streamline the review process, so it’d be great if you could use that when opening future PRs. It looks like the quality checks are currently failing—would you mind running `mypy --check-untyped-defs` and resolving any issues that come up? Also, if you’ve run any tests with your changes, it would be really helpful to include those results as well. Thanks again for your work on this!

fynnsu:
Closing as #370 was resolved.

PR #434, opened 2026-04-17
Add guidellm-based performance benchmarking utilities

anmarques:
Adds a complete perf-benchmark pipeline under examples/evaluate/ that:
- Estimates output token distribution via guidellm throughput mode and derives per-subset max_tokens (first power-of-2 >= median)
- Runs guidellm sweep benchmarks with the derived max_tokens cap
- Parses sweep results into CSV (rps, latency, ITL, TTFT, output TPS)
- Provides multi-version comparison plots (plot_perf_compare.py) and pairwise speedup visualizations with bwr colormap (plot_perf_speedup.py)
- Supports configurable sampling params (--gen-kwargs) and data column mapping (--data-column-mapper) for different datasets

I have filled in:

- [X] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [X] (Optional) The necessary documentation update.

## Release Notes

* **New Features**
  * Added comprehensive performance benchmarking suite with automated pipeline orchestration for profiling LLM throughput and latency metrics
  * Added visualization tools for performance comparisons across versions and speedup analysis
  * Added support for collecting and analyzing speculative decoding metrics from vLLM servers

* **Documentation**
  * Added detailed README documenting end-to-end benchmarking workflow, CLI options, CSV output schema, and usage examples


dsikka (commented):
I think it would also be useful to consolidate these with the existing scripts so that we have one set of go-to benchmarking utils cc @rahul-tuli

docs/user_guide/tutorials/evaluating_performance.md
```diff
@@ -1,3 +1,228 @@
...
+## Prerequisites
+
+```bash
+cd scripts/evaluate
+pip install -r requirements.txt
+```
+
+## Quick Start
+
+See the example script at [`examples/evaluate/example_humaneval_qwen3_8b_dflash.sh`](../../../examples/evaluate/example_humaneval_qwen3_8b_dflash.sh) for a complete end-to-end example that:
```
fynnsu: Bad link. Maybe just link to where the file will be once merged.

docs/user_guide/tutorials/evaluating_performance.md
```diff
@@ -1,3 +1,228 @@
...
+    --data-column-mapper '{"text_column":"input"}'
+```
+
+## Pipeline Steps
+
+### Step 1: Output Length Estimation
+
+For each subset, runs `guidellm benchmark --profile throughput` to generate responses for all prompts and record output token counts.
+
+Script: `run_perf_benchmark.sh`
```
fynnsu: I'm a little confused by these "Script" items in the pipeline steps. Aren't all of these steps run when you call `run_perf_benchmark.sh`?

docs/user_guide/tutorials/evaluating_performance.md
```diff
@@ -1,3 +1,228 @@
...
+
+For each subset, runs `guidellm benchmark --profile sweep` with the computed `max_tokens` injected into `--backend-args`.
+
+Script: `run_perf_benchmark.sh`
+
+### Step 4: Extract Metrics to CSV
+
+Parses all sweep JSONs and extracts median metrics for each sweep point. Throughput-mode entries are excluded (they represent max-load saturation).
+
+When `VLLM_URL` environment variable is set, the pipeline also:
```
fynnsu: Little bit confused by this `VLLM_URL` env variable. Aren't we providing `--target ...` to the `run_perf_benchmark.sh` ? Isn't that kind of the vllm url? 

Do we need `VLLM_URL` to be passed as well?

rahul-tuli (approved):
Thanks @shanjiaz, lgtm!

scripts/evaluate/run_perf_benchmark.sh
```diff
@@ -0,0 +1,422 @@
...
+    if ! curl -s -f --max-time 30 "${METRICS_URL}" > "${CURRENT_FILE}"; then
+        echo "[ERROR] Failed to fetch current metrics from ${METRICS_URL}" >&2
+        exit 1
+    fi
+
+    # Parse and display acceptance rates (delta from baseline)
+    PARSE_ARGS=(--acceptance-only --current-metrics "${CURRENT_FILE}" --baseline-metrics "${BASELINE_FILE}")
+    [[ -n "${OUTPUT_DIR}" ]] && PARSE_ARGS+=(--output "${OUTPUT_DIR}/acceptance.csv")
+
+    python "${SCRIPT_DIR}/parse_sweep.py" "${PARSE_ARGS[@]}"
```
anmarques: The parsing of acceptance rates must happen for each subset individually. This will process acceptance rates after the loop over subsets, which will report the overall acceptance rates.

anmarques (commented):
I think we should expose acceptance rates computation via a different end point, even if we reuse the codebase (it actually has a separate block within run_perf_benchmark so I think it makes sense to keep separate)

scripts/evaluate/run_perf_benchmark.sh
shanjiaz: Thanks for pointing it out! This call is just to create the final csv file. Around line 355 we capture baseline metrics before the current dataset sweep and find diff between two runs to calculate the final result. I have tested the full sweep too, the result seems comparable to when we run each dataset separately. Will keep this in mind as I keep working on refactoring.

fynnsu (approved):

PR #438, opened 2026-04-20
Stablize tests using random

shanjiaz:

## Purpose
This test might be a little flaky. Adding seed to stablize it. 

## Description
Added `torch.manual_seed(42)`

## Tests
Tested locally. 

I have filled in:

### DIFF
--- a/tests/unit/models/test_dflash_metrics.py
+++ b/tests/unit/models/test_dflash_metrics.py
@@ -154,6 +154,7 @@ class TestLossFunction:
     def test_different_gamma(self):
         """Different gamma values should produce different losses."""
         B, T, V = 1, 16, 10
+        torch.manual_seed(42)
         logits = torch.randn(B, T, V)
         target_ids = torch.randint(0, V, (B, T))
         loss_mask = torch.ones(B, T)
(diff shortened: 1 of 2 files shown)


tests/unit/models/test_dflash_metrics.py
```diff
@@ -154,6 +154,7 @@ def test_loss_mask_zeros_out_positions(self):
     def test_different_gamma(self):
         """Different gamma values should produce different losses."""
         B, T, V = 1, 16, 10
+        torch.manual_seed(42)
```
rahul-tuli: instead of repeating this line, could we make this into an autouse fixture? will help stabilize other tests as well, something like:

```python
import pytest
import numpy as np
import torch

@pytest.fixture(autouse=True)
def seed_everything(seed: int = 42):
    np.random.seed(seed)
    torch.manual_seed(seed)
    yield
```

rahul-tuli (dismissed):

tests/unit/models/test_dflash_metrics.py
shanjiaz: sure

fynnsu (approved):

PR #439, opened 2026-04-20
Improve data_generation_offline2.py shutdown and print vllm logs

fynnsu:

## Purpose

Currently, we're timing out because `data_generation_offline2.py` failing doesn't seem to always exit correctly. This pr makes it exit more forcefully if there are any issues.

We also update vllm the print logs (with uvicorn logs disabled) during generation. This means we won't print all 5000 status ok logs for all the requests, but will display any errors that may occur. 

This also might be fixing the time out issue we were having which seems like it might've been an issue with the logs being captured by the testing process. 

## Tests

E2E test run. Note this still failed but no longer gets stuck and hangs. It's failing on dflash related tests, unrelated to data gen. 
https://github.com/neuralmagic/llm-compressor-testing/actions/runs/24676474803/job/72161733494

I have filled in:

### DIFF
--- a/scripts/data_generation_offline2.py
+++ b/scripts/data_generation_offline2.py
@@ -15,6 +15,7 @@ Usage:
 import argparse
 import asyncio
 import logging
+import os
 import shutil
 import sys
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


shanjiaz (approved):
Thanks for fixing!

PR #440, opened 2026-04-20
[Dflash][Tests] Add DFlash Regression test

dsikka:
## Purpose

- Enable DFlash regression test case

## Description

- Add a DFlash regression test case for Qwen3 8b
- vLLM also requires that the list of target layers are set in the model's config.json but we do not write to the config if using the default list. Add a step to write

## Tests

Workflow testing dFlash: https://github.com/neuralmagic/llm-compressor-testing/actions/runs/24690434074/job/72210647415

I have filled in:


src/speculators/models/dflash/core.py
```diff
@@ -71,15 +71,17 @@ def __init__(
...
             self.target_layer_ids = config.aux_hidden_state_layer_ids
         else:
+            # Eagle3 defaults; write back so they are persisted in config.json
             self.target_layer_ids = [
                 2,
                 num_verifier_layers // 2,
                 num_verifier_layers - 3,
             ]
+            # set defaults to config if not provided - vLLM will fail otherwise
+            config.aux_hidden_state_layer_ids = self.target_layer_ids
```
shanjiaz, line 84: 🎉

shanjiaz (approved):

PR #452, opened 2026-04-21
feat(mtp): add MTP model architecture, converter, and stitcher

rahul-tuli:
## Purpose

Add Multi-Token Prediction (MTP) speculator support — model architecture, checkpoint conversion, and weight stitching. This enables finetuning MTP heads for Qwen3-Next, Qwen3.5, and Qwen3.5-MoE architectures using the speculators framework.

### Model (`src/speculators/models/mtp/`)

- **`MTPDraftModel`** — single MTP layer speculator with input projection (2H→H), using `DraftVocabMixin` for standardized embed_tokens/lm_head lifecycle (frozen copies from verifier, saved with checkpoint but not stitched back). Registered via `@SpeculatorModel.register("mtp")`. `conditional_torch_compile` on forward pass.
- **`MTPConfig`** — uses the `transformer_layer_config: PretrainedConfig` pattern to hold all architecture dimensions. Includes `compute_step_weights()` for exponential-decay per-step loss weights (FastMTP, arXiv:2509.18362). `num_nextn_predict_layers` kept for vLLM compatibility.
- **`MTPLayerMixin` + `_create_mtp_layer_class`** — factory pattern creating architecture-specific MTP decoder layers. Handles hybrid attention models via `_last_full_attention_idx` to select a matching full-attention layer config.
- **`model_definitions`** — concrete layer classes for `qwen3_next`, `qwen3_5_text`, `qwen3_5_moe_text`, with informative version errors when transformers is too old.

### Converter (`src/speculators/convert/mtp/`)

- **`MTPConverter`** — extracts MTP weights from native checkpoints (sharded safetensors). Remaps keys from `mtp.*` to `mtp_layers.0.*` format. Handles MoE expert weight fusing (per-expert → packed 3D tensors). embed_tokens/lm_head loaded from verifier at runtime, not extracted.

### Stitch CLI (`scripts/stitch_mtp.py`)

- Merges finetuned MTP weights back into a verifier checkpoint for vLLM deployment. Skips frozen weights (embed_tokens, lm_head). Handles MoE expert unfusing with expert count and orphaned key validation. Regenerates safetensors shard index.

### Shared infrastructure

- `requires_transformers_version` decorator in `tests/conftest.py` for version-gated test skipping.
- `list_checkpoint_keys()` added to `speculators.utils.loading`.
- Qwen3.5 `RMSNorm` and `RotaryEmbedding` registered in `base_components`.

### Supported models

| Model | `model_type` | Min transformers |
|---|---|---|
| Qwen3-Next | `qwen3_next` | 4.57.0 |
| Qwen3.5 | `qwen3_5_text` | 5.2.0 |
| Qwen3.5-MoE | `qwen3_5_moe_text` | 5.2.0 |

## Tests

- **Unit tests** — config validation, model forward pass, converter key remapping (`tests/unit/models/test_mtp_*.py`, `tests/unit/convert/test_mtp_converter.py`)
- **E2E smoke test** — convert → load → forward → perturb weights → stitch round-trip (`tests/e2e/smoke/test_mtp_conversion_roundtrip.py`)
- All Qwen3.5 tests gated on `transformers>=5.2.0`
- Test model: `Qwen/Qwen3.5-0.8B` (small footprint for CI)

```
python -m pytest tests/unit/models/test_mtp_config.py tests/unit/models/test_mtp_model.py tests/unit/convert/test_mtp_converter.py tests/e2e/smoke/test_mtp_conversion_roundtrip.py -v
```

I have filled in:


src/speculators/models/mtp/config.py
```diff
@@ -0,0 +1,109 @@
...
+
+    transformer_layer_config: PretrainedConfig = Field(
+        default_factory=Qwen2Config,
+        description=(
+            "Underlying transformer architecture config "
+            "(e.g., Qwen2Config, Qwen3NextConfig)"
+        ),
+    )
+
+    num_nextn_predict_layers: int = Field(
```
rahul-tuli, line 46: Since we only support 1 head, I can remove this based on opinions, please let me know!

src/speculators/convert/mtp/stitcher.py
fynnsu, line 1: How is the stitcher being used currently?

Or will this be a "convert" step that converts the checkpoint from a separate mtp head back into an full model+mtp checkpoint?

src/speculators/models/mtp/config.py
```diff
@@ -0,0 +1,109 @@
...
+        default=1,
+        description=(
+            "Number of MTP prediction heads in the checkpoint. vLLM "
+            "reads this field to create the correct number of "
+            "speculator head instances."
+        ),
+    )
+
+    step_weight_beta: float = Field(
```
fynnsu: This seems like a training parameter? We probably shouldn't include this in the model config.

src/speculators/models/mtp/config.py
```diff
@@ -0,0 +1,109 @@
...
+    @property
+    def vocab_size(self) -> int:
+        return self.transformer_layer_config.vocab_size  # type: ignore[return-value]
+
+    @property
+    def num_speculative_steps(self) -> int:
+        return self.speculators_config.proposal_methods[0].speculative_tokens  # type: ignore[union-attr,attr-defined]
+
+    @staticmethod
+    def compute_step_weights(beta: float = 0.6, num_steps: int = 3) -> list[float]:
```
fynnsu: Should this just be a util fn?

src/speculators/models/mtp/core.py
```diff
@@ -0,0 +1,217 @@
...
+    """
+
+    config_class: ClassVar[type[MTPConfig]] = MTPConfig  # type: ignore[misc]
+
+    def __init__(self, config: MTPConfig) -> None:
+        super().__init__(config=config)
+        tc = config.transformer_layer_config
+        self._model_definitions = mtp_model_classes[resolve_model_type(tc.model_type)]
+        self.mtp_layers = nn.ModuleList(
+            [self._model_definitions.first_layer_class(tc, layer_idx=0)]
```
fynnsu, line 81: So this is always just a single layer?

src/speculators/models/mtp/core.py
```diff
@@ -0,0 +1,217 @@
...
+            lm_head_weight.detach().clone(), requires_grad=False
+        )
+
+    def forward(
+        self,
+        input_ids: torch.Tensor,
+        hidden_states: torch.Tensor,
+        attention_mask: torch.Tensor | None = None,
+        position_ids: torch.Tensor | None = None,
+        labels: torch.Tensor | None = None,
```
fynnsu: Where are the labels coming from?

fynnsu (commented):
Hi @rahul-tuli, I added some initial comments. 

Could you also provide a smoke test for this? I'm not sure I fully understand the data format / how the model will be called, and it would help to see a working example.

src/speculators/convert/mtp/converter.py
```diff
@@ -0,0 +1,318 @@
...
+            source_config = source_config["text_config"]
+
+        config = self._build_config(source_config, base_model, num_speculative_steps)
+        saved_path = self._save(config, weights, output_path)
+        logger.success(f"Saved to: {saved_path}")
+
+        if validate:
+            self._validate(saved_path)
+
+    def _list_checkpoint_keys(self, checkpoint_dir: Path) -> list[str]:
```
shanjiaz: Can we move this to one of the util files? Feels like it could be reused.

src/speculators/convert/mtp/converter.py
```diff
@@ -0,0 +1,318 @@
...
+            non_expert[f"{prefix}.down_proj"] = down
+            logger.debug(
+                f"Fused {num_experts} experts at {prefix}: "
+                f"gate_up_proj={gate_up.shape}, "
+                f"down_proj={down.shape}"
+            )
+
+        return non_expert
+
+    def _extract_weights(
```
shanjiaz, line 172: Can we rewrite `load_model_layers()` in `utils/loading.py` or something to support this? Maybe out of the scope for this PR but feels like we have multiple helper functions serving very similar purpose.

src/speculators/convert/mtp/stitcher.py
shanjiaz: I think this "stitches" the mtp head back to the verifier and produce a single checkpoint with updated MTP head?

rahul-tuli: Helen is correct, it the flow is:

Original Checkpoint (with mtp layers) -> (convert step) trainable MTPDraftModel -> finetune -> (stitch) re-attach the finetuned MTPDraftModel into the Original Checkpoint

src/speculators/convert/mtp/converter.py
rahul-tuli: Agree with you on that, we'll do it in a follow up PR since that touches eagle3 too

src/speculators/convert/mtp/converter.py
rahul-tuli: Yes, moved!

src/speculators/models/mtp/config.py
rahul-tuli: Moved to trainer_kwargs, put it here initially because we are anyway re-stitching the weights back into the original model and ship that, not this config

src/speculators/models/mtp/core.py
rahul-tuli: Yes, kept in a ModuleList to match vllm.

src/speculators/models/mtp/core.py
rahul-tuli: MTP finetuning uses hard-label cross-entropy rather than soft distillation like EAGLE3, so the targets are simply the tokens themselves — input_ids sliced at per-step offsets. I get how this could be confusing.

The labels parameter has been removed entirely; 
the forward pass now derives targets directly from input_ids as 
`step_targets = input_ids[:, step + 2 : step + 2 + valid_len]`. 

No separate label tensor is needed — loss_mask handles position exclusion for prompt tokens and padding by setting masked positions to -100, which cross_entropy(ignore_index=-100) skips. 

This is the fundamental difference from EAGLE3: EAGLE3 computes soft targets at training time from `verifier_last_hidden_states` via the verifier's `lm_head` using KL divergence, whereas MTP simply predicts the literal next token with CE loss, making `input_ids` both the embedding source and the prediction target.

rahul-tuli:
> Hi @rahul-tuli, I added some initial comments.
> 
> Could you also provide a smoke test for this? I'm not sure I fully understand the data format / how the model will be called, and it would help to see a working example.

A verification script is attached with the PR description

src/speculators/convert/mtp/stitcher.py
fynnsu: Okay, I get the idea here, but this is not being used anywhere in code currently. How will the user interact with this, where will the stitching be called, etc.

src/speculators/models/mtp/model_definitions.py
```diff
@@ -0,0 +1,210 @@
...
+from transformers.models.qwen3.modeling_qwen3 import (
+    Qwen3DecoderLayer,
+    Qwen3RMSNorm,
+)
+
+from speculators.models import base_components
+
+__all__ = ["MTPLayerMixin", "mtp_model_classes", "resolve_model_type"]
+
+_MODEL_TYPE_ALIASES: dict[str, set[str]] = {}
```
fynnsu: Where is this getting populated?

fynnsu (commented):
Okay, I think this is almost there. My main concerns are around the user flow:
- Do users need to use the python api to stitch the weights back together?
- If we only use labels for training instead of computing targets from the models last layer, then model results will be poor unless users do response regen. Ideally we would support distillation loss functions like the other algorithms do. But if you think this is okay, then we can just use the input ids.

shanjiaz (commented):
Great job on this! I have a similar concern as Fynn suggested, stitching is a required step for mtp finetuning and users need the python API as documented in the user guide. This workflow might be a little different from our other models. Also would be great to add e2e test for this workflow.

rahul-tuli:
> Okay, I think this is almost there. My main concerns are around the user flow:
> 
> * Do users need to use the python api to stitch the weights back together?
> * If we only use labels for training instead of computing targets from the models last layer, then model results will be poor unless users do response regen. Ideally we would support distillation loss functions like the other algorithms do. But if you think this is okay, then we can just use the input ids.

1) Yes right now users will explicitly need to call the `MTPStitcher().stitch(...)` explicitly, we will be landing docs in a follow up to make this clear, can also add a cli if needed in a follow up, didn't want to add it here to keep the diff small

2) Hard-label CE is the approach from the FastMTP paper (Eq. 2), That said, I agree distillation loss would be a valuable follow-up. It could be added as an alternative loss mode without changing the current architecture the MTP model's forward pass would just need a `soft_targets` parameter alongside the existing `input_ids` path. I'd prefer to land the hard-label path first and add distillation as a follow-up. Does that work?

src/speculators/models/mtp/model_definitions.py
rahul-tuli: Good catch, it's not populated and the alias loop is dead code. I'll remove it and simplify `resolve_model_type` to just check `mtp_model_classes` directly.

rahul-tuli:
> Okay, I think this is almost there. My main concerns are around the user flow:
> 
> * Do users need to use the python api to stitch the weights back together?
> * If we only use labels for training instead of computing targets from the models last layer, then model results will be poor unless users do response regen. Ideally we would support distillation loss functions like the other algorithms do. But if you think this is okay, then we can just use the input ids.

1) Added the CLI for stitching in scripts
2) Response regeneration is a hard requirement right now for FastMtp style finetuning according to the paper, we can add support for distillation in a follow up PR

rahul-tuli:
> Great job on this! I have a similar concern as Fynn suggested, stitching is a required step for mtp finetuning and users need the python API as documented in the user guide. This workflow might be a little different from our other models. Also would be great to add e2e test for this workflow.

Added smoke test that verifies the conversion round trip, hopefully its cleaner and makes more sense now?

tests/e2e/smoke/test_mtp_conversion_roundtrip.py
```diff
@@ -0,0 +1,189 @@
+"""E2E test: convert → load → forward → perturb → stitch round-trip for MTP.
```
shanjiaz, line 1: Thanks for adding this test! This looks more like unit test for each step, could we maybe follow the existing pattern of e2e smoke test? Most importantly I think we need a step at the end to serve in vllm and test that the pathway works.

src/speculators/convert/mtp/converter.py
```diff
@@ -0,0 +1,299 @@
...
+import re
+from pathlib import Path
+
+import torch
+from loguru import logger
+from safetensors import safe_open
+from transformers import PretrainedConfig
+
+from speculators.config import SpeculatorsConfig, VerifierConfig
+from speculators.convert.eagle.utils import (
```
shanjiaz, line 23: nit: maybe move these shared function to a convert/utils.py file?

rahul-tuli: Will do that in a follow up to keep that change small and focused

tests/e2e/smoke/test_mtp_conversion_roundtrip.py
fynnsu: Yeah, I think this makes sense. It should allow us to confirm that the roundtrip doesn't somehow break the final checkpoint.

I think one of the Qwen3.5 models wouldn't be good for this since they all have mtp heads. `Qwen/Qwen3.5-2B` or `Qwen/Qwen3.5-0.8B` would be quite small and easy to test.

fynnsu:
Looks like the `transformers<5` tests are failing because the model isn't available. Can you add a skip check to these tests to make sure they're only run if a valid transformers version is available?

src/speculators/models/mtp/core.py
```diff
@@ -0,0 +1,242 @@
...
+    def layers(self) -> nn.ModuleList:
+        """Expose mtp_layers for FSDP wrapping compatibility."""
+        return self.mtp_layers
+
+    def load_verifier_weights(self) -> None:
+        """Re-set NaN sentinel before loading — meta-device init may clear
+        it. Deletes verifier_lm_head after loading since MTP does not use it.
+        """
+        with torch.no_grad():
+            self.embed_tokens.weight.fill_(torch.nan)
```
shanjiaz, line 97: Just curious, is this needed? I imagine other models would have the same issue

src/speculators/convert/mtp/converter.py
fynnsu: Agreed. I think there is a lot of overlap here with the existing utilities in `utils/loading.py` and in `scripts/stitch_mtp.py`.

src/speculators/convert/mtp/converter.py
```diff
@@ -0,0 +1,299 @@
...
+
+        single = checkpoint_dir / "model.safetensors"
+        if single.exists():
+            weights: dict[str, torch.Tensor] = {}
+            with safe_open(str(single), framework="pt") as f:
+                for key in needed & set(f.keys()):
+                    weights[self._remap_key(key)] = f.get_tensor(key)
+            return weights
+
+        pytorch = checkpoint_dir / "pytorch_model.bin"
```
fynnsu: Do you have a specific model that still uses pytorch bin weights? Otherwise I don't think we need this, and in general we're currently only supporting safetensors format models.

src/speculators/models/__init__.py
```diff
@@ -2,13 +2,16 @@
...
 from .eagle3 import Eagle3DraftModel, Eagle3SpeculatorConfig
+from .mtp import MTPConfig, MTPDraftModel
 from .peagle import PEagleDraftModel, PEagleSpeculatorConfig
 
 __all__ = [
     "DFlashDraftModel",
     "DFlashSpeculatorConfig",
     "Eagle3DraftModel",
     "Eagle3SpeculatorConfig",
+    "MTPConfig",
```
fynnsu: We should probably use `MTPSpeculatorConfig` to match the others

src/speculators/utils/loading.py
```diff
@@ -9,6 +9,35 @@
...
+
+    single = checkpoint_dir / "model.safetensors"
+    if single.exists():
+        with safe_open(str(single), framework="pt") as f:
+            return list(f.keys())
+
+    pytorch = checkpoint_dir / "pytorch_model.bin"
+    if pytorch.exists():
+        state_dict = torch.load(str(pytorch), map_location="cpu", weights_only=True)
+        return list(state_dict.keys())
```
fynnsu: Again, I wouldn't bother support pytorch.

tests/unit/models/test_mtp_config.py
```diff
@@ -0,0 +1,206 @@
...
+
+class TestConstruction:
+    def test_default_initialization(self):
+        config = MTPConfig()
+        assert config.speculators_model_type == "mtp"
+        assert config.architectures == ["MTPDraftModel"]
+        assert config.num_nextn_predict_layers == 1
+        assert config.model_type == "speculator_model"
+
+    def test_derived_hidden_size(self, mtp_config, qwen3_5_pretrained_config):
```
fynnsu: Not really sure we need all of these tests. The defaults make some sense, but most these other ones just seem to be testing either the passthrough properties added to the config.

The last one `test_transformer_layer_config_preserved` in particular doesn't seem to be testing much.

tests/unit/models/test_mtp_config.py
```diff
@@ -0,0 +1,206 @@
...
+            mtp_config.transformer_layer_config.num_attention_heads
+            == qwen3_5_pretrained_config.num_attention_heads
+        )
+
+
+# ===== Validation =====
+
+
+class TestValidation:
+    def test_num_nextn_predict_layers_rejects_zero(self):
```
fynnsu: This also seem unnecessarily verbose. We have 3 tests to check that `num_nextn_predict_layers=1` is required?

tests/unit/models/test_mtp_config.py
```diff
@@ -0,0 +1,206 @@
...
+class TestRegistration:
+    def test_registered_as_mtp(self):
+        assert SpeculatorModelConfig.registry is not None
+        assert "mtp" in SpeculatorModelConfig.registry
+        assert SpeculatorModelConfig.registry["mtp"] == MTPConfig
+
+    def test_discoverable_in_registered_classes(self):
+        registered = SpeculatorModelConfig.registered_classes()
+        class_names = [cls.__name__ for cls in registered]
+        assert "MTPConfig" in class_names
```
fynnsu: Again, technically fine to test for this but feels sorta unnecessary.

tests/e2e/utils.py
```diff
@@ -397,3 +401,23 @@ def run_vllm_engine(
             assert acci >= thresholdi, (
                 f"Acceptance {acci} at token {i} is less than threshold {thresholdi}"
             )
+
+
+def run_mtp_engine(
```
fynnsu: I think we can do without this function and just call `run_vllm_engine` directly, especially since we only have one call site so far.

tests/e2e/smoke/test_mtp_conversion_roundtrip.py
```diff
@@ -0,0 +1,205 @@
...
+    prompts = [
+        [{"role": "user", "content": "Write a binary search function in python"}],
+    ]
+    run_mtp_engine(
+        model_path=str(stitched_path),
+        tmp_path=tmp_path,
+        prompts=prompts,
+        num_speculative_tokens=1,
+        max_model_len=256,
+        enforce_eager=True,
```
fynnsu, line 206: If you add the `acceptance_thresholds` arg and provide a list of acceptance rates, we can check the per-position acceptance rates.

Since this test is just running roundtrip conversion and the existing mtp modules are usually relatively good, we'd expect the acceptance rates to be good too. Adding the acceptance rates check allows us to confirm the roundtrip didn't somehow break the weights.

tests/unit/models/test_mtp_model.py
```diff
@@ -0,0 +1,144 @@
...
+        assert total_loss.dim() == 0
+        assert torch.isfinite(total_loss)
+        assert total_loss >= 0
+
+    def test_metrics_keys(self, forward_result):
+        _, _, metrics = forward_result
+        expected_keys = {f"loss_step_{k}" for k in range(NUM_STEPS)}
+        assert set(metrics.keys()) == expected_keys
+        for key in expected_keys:
+            assert math.isfinite(metrics[key])
```
fynnsu: Maybe we could combine this into one test? The current implementation will run create a new `forward_result` for each test (which in turn initializes a new `mtp_model`). To reduce the overhead, a single test would only require a single model setup / run.

tests/unit/models/test_mtp_model.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+# ===== Loss masking =====
+
+
+class TestLossMasking:
+    def test_zero_mask_ignores_all_targets(self, mtp_model, seed):
+        """All-zero loss_mask sets every target to -100. cross_entropy(mean)
+        with all-ignored targets returns NaN (0/0), confirming the mask is
+        applied to every position."""
```
fynnsu: We might not want this behaviour. I think some of the other loss functions intentionally add a small number to the denominator to avoid division by 0. As it stands, if one batch ends up being bad for some reason and contains all zero loss_mask values, the whole training run will break.

tests/unit/models/test_mtp_model.py
```diff
@@ -0,0 +1,144 @@
...
+        loss_mask = torch.zeros(BATCH, SEQ_LEN)
+        with torch.no_grad():
+            _, total_loss, _ = mtp_model(
+                input_ids=input_ids,
+                hidden_states=hidden_states,
+                loss_mask=loss_mask,
+            )
+        assert total_loss.isnan()
+
+    def test_partial_mask_reduces_loss(self, mtp_model, seed):
```
fynnsu: Misleading fn name.

tests/unit/models/test_mtp_model.py
```diff
@@ -0,0 +1,144 @@
...
+import torch
+
+from speculators.models.mtp.model_definitions import (
+    mtp_model_classes,
+    resolve_model_type,
+)
+
+BATCH = 1
+SEQ_LEN = 10
+NUM_STEPS = 3
```
fynnsu: The number of steps is actually the models config. `NUM_STEPS` here redefines the value but isn't actually tied to the value used by `mtp_model` in any way.

Instead tests that use `NUM_STEPS` should get the true value from the config instead.

tests/unit/models/test_mtp_model.py
```diff
@@ -0,0 +1,144 @@
...
+        assert metrics["loss_step_1"] == 0.0
+        assert metrics["loss_step_2"] == 0.0
+
+
+# ===== Short sequence truncation =====
+
+
+class TestShortSequenceTruncation:
+    def test_short_sequence_fewer_logits(self, mtp_model, seed):
+        hidden_size = mtp_model.config.hidden_size
```
fynnsu: What is being tested here?

tests/unit/models/test_mtp_model.py
```diff
@@ -0,0 +1,144 @@
...
+        assert len(logits_list) < NUM_STEPS
+        assert len(logits_list) == 1
+
+
+# ===== resolve_model_type =====
+
+
+class TestResolveModelType:
+    @pytest.mark.parametrize("model_type", list(mtp_model_classes.keys()))
+    def test_registered_type_resolves(self, model_type):
```
fynnsu:
```python
def resolve_model_type(model_type: str) -> str:
    """Resolve a model_type to a canonical type in mtp_model_classes."""
    if model_type in mtp_model_classes:
        return model_type
    raise _unsupported_model_type_error(model_type)
```

I'm not sure that this test is super useful.

scripts/stitch_mtp.py
```diff
@@ -0,0 +1,345 @@
...
+        )
+
+    weights = _unfuse_moe_experts(weights)
+    native_weights = {_remap_key(k): v for k, v in weights.items()}
+    console.print(f"  Remapped [cyan]{len(native_weights)}[/] keys to native format")
+
+    with _spinner() as progress:
+        progress.add_task("Copying verifier checkpoint", total=None)
+        if output_path.exists():
+            shutil.rmtree(output_path)
```
shanjiaz: If output_path resolves to the same directory as verifier_path, the rmtree deletes the source before the copy.

scripts/stitch_mtp.py
```diff
@@ -0,0 +1,345 @@
...
+        shutil.copytree(verifier_path, output_path)
```
shanjiaz:
```suggestion
        shutil.copytree(verifier_path, output_path, dirs_exist_ok=True)
```

shanjiaz: `copytree` would copy every shard even though only few of them containing MTP weights would be modified, could we maybe hardlink unchanged shards instead?

tests/e2e/smoke/test_mtp_conversion_roundtrip.py
rahul-tuli: Not really, acceptance_thresholds don't make sense here since we are synthetically perturbing the weights (not actually finetuning them); However acceptance thresholds will be tested as a part of #366

src/speculators/models/mtp/core.py
rahul-tuli: Good question, other models (Eagle3, etc.) don't hit this because they save `embed_tokens.weight` and `lm_head.weight` in the checkpoint (not in their [`_keys_to_ignore_on_save`](https://github.com/vllm-project/speculators/blob/6f5ba52/src/speculators/models/eagle3/core.py#L40-L43)). So by the time [`load_verifier_weights()`](https://github.com/vllm-project/speculators/blob/6f5ba52/src/speculators/model.py#L159) runs, `from_pretrained` already loaded real values from the safetensors; thus the NaN sentinel is irrelevant.

MTP can't do that. It always uses full vocab ([`core.py:75`](https://github.com/vllm-project/speculators/blob/6f5ba52/src/speculators/models/mtp/core.py#L75)), so saving embed/lm_head would just duplicate large verifier tensors in every checkpoint. We exclude them via [`_keys_to_ignore_on_save`](https://github.com/vllm-project/speculators/blob/6f5ba52/src/speculators/models/mtp/core.py#L54-L57) and reload from the verifier at runtime instead. The catch is `from_pretrained` with meta-device init (`device_map="auto"`): HF creates the model on meta device, then materializes parameters from the checkpoint. Weights in the checkpoint get real values; weights *not* in the checkpoint get materialized as zeros — bypassing the `_is_hf_initialized` flag entirely. So the NaN that [`_init_vocab()`](https://github.com/vllm-project/speculators/blob/6f5ba52/src/speculators/model.py#L74-L76) wrote is gone, `isnan().any()` returns `False`, and [`load_verifier_weights()`](https://github.com/vllm-project/speculators/blob/6f5ba52/src/speculators/model.py#L159) silently skips loading. The reset at [`core.py:97-99`](https://github.com/vllm-project/speculators/blob/6f5ba52/src/speculators/models/mtp/core.py#L97-L99) ensures the sentinel is intact before we call `super()`.

Note this only affects the `from_pretrained` path `from_training_args` constructs normally and the `_is_hf_initialized` flag protects the NaN through `post_init()`. See [`PreTrainedModel.from_pretrained()`](https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py) for the meta-device materialization logic.

So in principle if we were to not save `embed_tokens` and `lm_head` in eagle3 checkpoints they would also need this reset for the `from_pretrained` pathway to work correctly.

Does that make sense?

src/speculators/convert/mtp/converter.py
rahul-tuli: There was an old qwen3next checkpoint in HF that used pytorch bin weights, but I think it makes sense to only support safetensors for now. Cleaned it accordingly!

src/speculators/models/__init__.py
rahul-tuli: Done!

src/speculators/utils/loading.py
rahul-tuli: Removed!

tests/e2e/utils.py
rahul-tuli: Done!

tests/unit/models/test_mtp_model.py
rahul-tuli: Done!

tests/unit/models/test_mtp_model.py
rahul-tuli: Updated!

tests/unit/models/test_mtp_model.py
rahul-tuli: Makes sense, Fixed in commit `6fc2f4d`.  When all positions are masked, returns `0.0` instead of `NaN`

tests/unit/models/test_mtp_model.py
rahul-tuli: Updated!

scripts/stitch_mtp.py
rahul-tuli: Updated

scripts/stitch_mtp.py
rahul-tuli: Removed `rmtree` and added `dirs_exist_ok`, we can do this optimization (to hardlink) in a follow up

tests/unit/models/test_mtp_config.py
rahul-tuli: Updated to reduce/remove unnecessary tests

src/speculators/convert/mtp/converter.py
rahul-tuli: Good call, there is overlap between _extract_weights here and load_model_layers() in utils/loading.py. Consolidating them would touch the Eagle3 converter too though, so I'd rather keep this PR focused and do the refactor as a follow-up. Filing this as out-of-scope in the PR, making a note to add a follow up for this

src/speculators/models/mtp/core.py
```diff
@@ -137,11 +141,29 @@ def forward(
         batch_size, seq_len = input_ids.shape
         num_steps = self.config.num_speculative_steps
 
+        if step_weights is not None and len(step_weights) < num_steps:
+            raise ValueError(
+                f"step_weights has {len(step_weights)} entries but "
+                f"num_speculative_steps={num_steps}; expected at least "
```
fynnsu: Arguably this should be an `==` check since these two values should match.

src/speculators/models/mtp/core.py
```diff
@@ -137,11 +141,29 @@ def forward(
...
+        if attention_mask is not None and attention_mask.dim() == _2D:
+            causal_mask = create_causal_mask(
+                config=self.config.transformer_layer_config,
+                inputs_embeds=hidden_states,
+                attention_mask=attention_mask,
+                past_key_values=None,
+                position_ids=position_ids,
+            )
+        else:
+            causal_mask = attention_mask
```
fynnsu: Hmm, I'm a little confused as to what the use case is where the user would provide this attention mask instead of us creating it.

But also this check seems to be "if the user explicitly passed in an attention mask but it has 2d dim, we're just going to silently replace it. Otherwise the user must pass a valid mask in". 

So if I wanted an attention mask to be created for me, I'd have to pass in a dummy 2d tensor to trigger the creation?

src/speculators/models/mtp/core.py
```diff
@@ -157,9 +179,10 @@ def forward(
             )
             step_pos_ids = position_ids[:, :valid_len]
             step_pos_emb = self.rotary_emb(step_hidden, step_pos_ids)
-            step_attn_mask = (
-                attention_mask[:, :valid_len] if attention_mask is not None else None
-            )
+            if causal_mask is not None:
+                step_attn_mask = causal_mask[:, :, :valid_len, :valid_len]
+            else:
+                step_attn_mask = None
```
fynnsu, line 181: So this seems to suggest the attention mask can be `None`? Won't that result in full bi-directional attention, instead of causal attention?

src/speculators/models/mtp/core.py
fynnsu, line 189: Also, if this is doing something similar to Eagle3's train-time testing (which it sorta looks like it is based on the `for` loop over steps), then we'll need a more advanced attention mask, plus potentially some kind of kv cache to handle this.

Like the Eagle3 attention on the very first step is just regular causal. But then step 2 looks something like:
```
1 0 0 0 1 0 0 0
1 1 0 0 0 1 0 0
1 1 1 0 0 0 1 0
1 1 1 1 0 0 0 1
```
where the left bit is causal attention on the original sequence and the right half includes the previously predicted token for each position.
And then step 3 uses:
```
1 0 0 0 1 0 0 0 1 0 0 0
1 1 0 0 0 1 0 0 0 1 0 0
1 1 1 0 0 0 1 0 0 0 1 0
1 1 1 1 0 0 0 1 0 0 0 1
```

src/speculators/models/mtp/core.py
rahul-tuli: oops, good catch, updated!

src/speculators/models/mtp/core.py
rahul-tuli: You're right, the branching was confusing. Simplified -- we now call `create_causal_mask` unconditionally, which handles `None`, 2D, and 4D inputs internally. This matches the reference which always creates the mask:
[`modeling_mimo.py` L118-L131](https://github.com/Tencent-BAC/FastMTP/blob/5875b708b34007eeab3cdf27bc8aca517f87eb21/transformers-4.54.0/src/transformers/models/mimo/modeling_mimo.py#L118-L131)

src/speculators/models/mtp/core.py
rahul-tuli: Good catch, this was a real bug with `eager` attention. Fixed by the same change as comment 2 -- `create_causal_mask` is now called unconditionally. When `attention_mask=None`, it still produces a causal mask (no padding, but causal). The reference does the same:
[`modeling_mimo.py` L118-L131](https://github.com/Tencent-BAC/FastMTP/blob/5875b708b34007eeab3cdf27bc8aca517f87eb21/transformers-4.54.0/src/transformers/models/mimo/modeling_mimo.py#L118-L131)

When `create_causal_mask` returns `None`, it's an optimization signal ("the backend will use `is_causal=True` internally"), not "no mask." Our existing `if causal_mask is not None: slice` handles this correctly.

src/speculators/models/mtp/core.py
rahul-tuli: I'm a little confused, This is MTP (teacher forcing), not Eagle3 (autoregressive draft). At each step, we feed ground-truth `input_ids` shifted by the step index, we never attend to our own prior predictions. The inter-step dependency is through the `hidden_states` argument, not attention.
The reference confirms the same causal mask and position_ids are reused across all MTP steps with no modification:
[`modeling_mimo.py` L154-L173](https://github.com/Tencent-BAC/FastMTP/blob/5875b708b34007eeab3cdf27bc8aca517f87eb21/transformers-4.54.0/src/transformers/models/mimo/modeling_mimo.py#L154-L173)
Specifically:
- Same mask every step: [`L165`](https://github.com/Tencent-BAC/FastMTP/blob/5875b708b34007eeab3cdf27bc8aca517f87eb21/transformers-4.54.0/src/transformers/models/mimo/modeling_mimo.py#L165)
- Same position_ids every step: [`L166`](https://github.com/Tencent-BAC/FastMTP/blob/5875b708b34007eeab3cdf27bc8aca517f87eb21/transformers-4.54.0/src/transformers/models/mimo/modeling_mimo.py#L166)
- Same RoPE embeddings every step: [`L169`](https://github.com/Tencent-BAC/FastMTP/blob/5875b708b34007eeab3cdf27bc8aca517f87eb21/transformers-4.54.0/src/transformers/models/mimo/modeling_mimo.py#L169)
No expanding mask or KV cache needed.

shanjiaz (approved):

fynnsu (approved):

fynnsu:
E2E job: https://github.com/neuralmagic/llm-compressor-testing/actions/runs/26844522282

PR #460, opened 2026-04-21
Add Gemma4 support

dsikka:
## Summary
- Resolve `hidden_act` by falling back to `hidden_activation` for models like Gemma 4 
- Add a CLI arg to override explicitly when required (such as in the case of serving Dflash models in vLLM, which requires its Qwen3 layers to use silu)
- Gemma 4 is in transformers v5.4 - to support it: guard `SpeculatorModelConfig.__init__` from being overwritten by transformers ≥ 5.x subclass injection; add `validate()` stub to avoid conflict with Pydantic v2; fix `model_rebuild()` calls to pass `_types_namespace={"torch": torch}` so Pydantic can resolve `torch.dtype` forward refs introduced in transformers 5.4+. These changes are backwards compatible with older transformers 

Tests:
- All CI / CD tests pass
- Weekly (for regression testing and smoke tests): https://github.com/neuralmagic/llm-compressor-testing/actions/runs/24750705038 all pass 

With these changes, I was able to successfully run and serve a gemma 4 dflash speculator with 5k samples 
```bash
chg run -g 1 -- torchrun --standalone --nproc_per_node=1 scripts/train.py \
  --verifier-name-or-path google/gemma-4-31B-it \
  --data-path /mnt/nvme_stripe/playground/dsikka/gemma4_prepped \
  --hidden-states-path /mnt/nvme_stripe/playground/dsikka/gemma4_prepped \
  --on-missing generate \
  --on-generate delete \
  --scheduler-type cosine \
  --draft-vocab-size 32000 \
  --max-anchors 3072 \
  --target-layer-ids 1 17 29 47 58 \
  --speculator-type dflash \
  --num-layers 5 \
  --draft-hidden-act silu \
  --run-name flash_5layers_5epochs \
  --lr 0.0006 \
  --epochs 5
```


dsikka:
@coderabbitai rereview completely

fynnsu (approved):

shanjiaz (approved):

Issue #292, opened 2026-02-18
[RFC] Add P-EAGLE support in training


shanjiaz:
## Overview

This RFC proposes implementing training support for P-EAGLE (Parallel EAGLE), a parallel speculative decoding method that extends EAGLE-3 with multi-token prediction, potentially offering 2-3x speedups over sequential EAGLE-3 drafting.

## Background

P-EAGLE inherits EAGLE-3's architecture but introduces parallel prediction. The approach uses the same lightweight decoder architecture as EAGLE-3 but generates multiple token predictions in parallel through Conditional-On-Distribution (COD) sampling, rather than sequential test-time training steps.

"Training a parallel token prediction model requires extending each sequence of length n to accommodate K parallel prediction depths, where depth k predicts the token $k+1$ positions ahead. Without optimization, this creates $n\times K$ total positions with $\mathcal{O}((nK)^2)$ attention complexity, causing out-of-memory failures at long sequences." P-EAGLE addresses this through COD sampling and sequence partitioning to make training tractable.

**Paper**: [P-EAGLE: Parallel EAGLE Speculative Decoding](https://arxiv.org/pdf/2602.01469)

## Key Technical Differences from Eagle-3

**Parallel Prediction Groups**: P-EAGLE organizes predictions into `para_num` parallel groups (typically $K=8$) created via COD sampling. Each depth $k$ predicts the token $k+1$ positions ahead" - depth $0$ predicts $i+1$, depth 1 predicts $i+2$, up to depth $7$ predicting $i+8$. All depths are processed in a single forward pass. EAGLE-3 uses sequential `ttt_steps` where each step depends on previous predictions through autoregressive generation.

**Conditional Drop-token (COD) Sampling**: COD reduces the number of positions at each prediction depth (group) through geometric decay: depth $0$ retains all n positions, depth $1$ randomly retains $n \times r$ positions, depth $2$ retains $n \times r^2$, and so on, where $r \in (0,1)$ is the retention rate (`down_sample_ratio`). This reduces total positions from $n \times K$ to $n \times (1 + r + r^2 + ⋯ + r^{K-1})$, significantly reducing attention memory from $\mathcal{O}((nK)^2)$ to $\mathcal{O}((n\Sigma r^i)^2)$. P-EAGLE further filters sampled positions to only include valid training positions where `loss_mask == 1`.

[...]

PR #480, opened 2026-04-27
P-eagle training support

shanjiaz:
## Purpose
P-EAGLE inherits EAGLE-3's architecture but introduces parallel prediction. The approach uses the same lightweight decoder architecture as EAGLE-3 but generates multiple token predictions in parallel through Conditional-On-Distribution (COD) sampling, rather than sequential test-time training steps. This PR implements the P-EAGLE algorithm based on existing EAGLE3 infrastructure.
 

## Description

  - Model definition: `speculators/models/peagle/core.py` — P-EAGLE draft model with parallel multi-token prediction via COD sampling, flex attention for cross-depth masking, and a learnable `mask_hidden` parameter for padding unsampled positions
  - Metrics: Extracted loss and accuracy computation into `peagle/metrics.py`. Includes count-based normalization for correct distributed averaging when COD causes different ranks to sample different depths.
  - Trainer: Added `normalize_counted_metrics` to `train/utils.py` to handle per-position accuracy averaging across ranks, with minimal changes to the shared trainer.
  - Training script: Added P-EAGLE config/args to `scripts/train.py` and an example training script for Qwen3-8B

  ## Related Issue

  https://github.com/vllm-project/speculators/issues/292

  ## Tests
  - Added a training example for Qwen3-8B with ShareGPT 5k samples
  - Will post more validation results.

I have filled in:


scripts/train.py
```diff
@@ -615,6 +616,40 @@ def parse_args():
         default=256,
         help="Maximum anchor positions for DFlash training (default: 256)",
     )
+    # P-EAGLE specific parameters
+    parser.add_argument(
+        "--para-depths",
```
fynnsu: Is this name already a standard somewhere? I don't love it.

scripts/train.py
```diff
@@ -615,6 +616,40 @@ def parse_args():
...
+        help="Geometric decay ratio for COD sampling in P-EAGLE (default: 0.7)",
+    )
+    parser.add_argument(
+        "--down-sample-ratio-min",
+        type=float,
+        default=0.2,
+        help="Minimum retention ratio for COD sampling in P-EAGLE (default: 0.2)",
+    )
+    parser.add_argument(
+        "--max-seq-len",
```
fynnsu: How does this interact with the existing `total-seq-len` parameter? Are they the same/could they be?

scripts/train.py
```diff
@@ -615,6 +616,40 @@ def parse_args():
...
+        type=int,
+        default=2048,
+        help="Maximum sequence length for attention mask construction (default: 2048)",
+    )
+    parser.add_argument(
+        "--prediction-loss-weight",
```
fynnsu: What is this doing? It seems to just get multiply by the loss. But weighting only really makes sense if there is something else to weight against.

e.g. `loss = loss_1 * weight + loss_2`

This just seems to be doing `loss = loss_1 * weight`.

src/speculators/models/peagle/attention.py
```diff
@@ -0,0 +1,161 @@
...
+        return valid_query & valid_key & hierarchical & same_sample & attention_allowed
+
+    return peagle_mask_mod
+
+
+# Compile flex_attention for efficient block-sparse attention
+_compiled_flex_attention = torch.compile(flex_attention)
+
+
+def peagle_flex_attention_forward(
```
fynnsu: This looks like it's the same as the flex attention implementation for Eagle-3 / DFlash, so we can just reuse that.

src/speculators/models/peagle/attention.py
```diff
@@ -0,0 +1,161 @@
...
+    )
+    depth_assignments = torch.full((total_length,), -1, dtype=torch.long, device=device)
+
+    for _concat_idx, full_idx in enumerate(all_indices):
+        original_pos = full_idx % seq_length
+        depth = full_idx // seq_length
+
+        original_positions[full_idx] = original_pos
+        depth_assignments[full_idx] = depth
+
```
fynnsu: Can we use tensor ops here.

src/speculators/models/peagle/config.py
```diff
@@ -0,0 +1,82 @@
...
+    # (matches p-eagle-train)
+    embed_requires_grad: bool = Field(
+        default=True,
+        description=(
+            "Whether embedding layer weights require gradients during "
+            "training (True for P-EAGLE)"
+        ),
+    )
+
+    prediction_loss_weight: float = Field(
```
fynnsu: Alongside the previous question of is this even needed, I also don't think it should be in the model config (if it is needed). Training hyperparameters should not be included in the final model configs.

src/speculators/models/peagle/core.py
```diff
@@ -0,0 +1,347 @@
...
+    _keys_to_ignore_on_load_missing: ClassVar[list[str]] = [  # type: ignore[misc]
+        *Eagle3DraftModel._keys_to_ignore_on_load_missing,
+        "mask_hidden",
+    ]
+
+    def __init__(
+        self,
+        config: PEagleSpeculatorConfig,
+        t2d: torch.Tensor | None = None,
+        d2t: torch.Tensor | None = None,
```
fynnsu: Can we update this to match the Eagle-3 / Dflash inits

src/speculators/models/peagle/core.py
```diff
@@ -0,0 +1,347 @@
...
+            intensors: Input tensor [batch, n, hidden_size]
+            target_len: Target sequence length to pad to
+            hidden: Learnable padding parameter (self.mask_hidden)
+            index_list: List to append current length to (for tracking)
+
+        Returns:
+            Padded tensor [batch, target_len, hidden_size]
+        """
+        batch_size, current_len, hidden_size = intensors.shape
+        index_list.append(current_len)
```
fynnsu: This seems strange

src/speculators/models/peagle/core.py
```diff
@@ -0,0 +1,347 @@
...
+            device=device,
+        )
+
+        layer_outputs = self.layers[0](
+            layer_input,
+            attention_mask=attention_mask,
+            position_ids=position_ids,
+            position_embeddings=position_embeddings,
+            **kwargs,
+        )
```
fynnsu: Could we not include this in the loop below with some better variable names?

src/speculators/models/peagle/core.py
```diff
@@ -0,0 +1,347 @@
...
+                ],
+                default_proposal_method="greedy",
+                verifier=VerifierConfig.from_config(
+                    verifier_config, name_or_path=kwargs["verifier_name_or_path"]
+                ),
+            ),
+        )
+
+        model = cls(config=config, t2d=t2d, d2t=d2t)
+        model.load_verifier_weights()
```
fynnsu, line 236: Add
```python
model.load_vocab_mappings(t2d, d2t)
```

src/speculators/train/trainer.py
```diff
@@ -217,6 +217,7 @@ def train_epoch(self, epoch: int):
                         dist.reduce(v, dst=0, op=dist.ReduceOp.AVG)
```
fynnsu: We're still reducing using an AVG. If we're using counts instead we probably want to use a sum here, no?

src/speculators/train/trainer.py
fynnsu: Same here. Also we need to make sure this works with the existing averaging logic across val batches.

fynnsu (commented):
Some initial comments. 

I didn't look in a lot of detail at the hard math parts (i.e. attention mask creation, metric computation, etc). Will do that in a follow up, but the thing I noticed is that we're not always using tensor computations. As much as possible we want to use tensors and avoid lists / python loops, to improve performance and interact better with torch compile.

scripts/train.py
shanjiaz: Good catch! This is safe to remove. It's from p-eagle's prediction loss weight and hidden states loss weight but we're not actually doing hidden states loss.

src/speculators/train/trainer.py
shanjiaz: Yep, have to make the counts changes.

scripts/train.py
shanjiaz: How about depth?

src/speculators/models/peagle/core.py
shanjiaz: yep!

scripts/train.py
shanjiaz: removed `max-seq-len`.

scripts/train.py
fynnsu: depth is maybe better, but I'm honestly not sure what this variable is representing.

"Parallel prediction groups" - what does this mean? Is this the max num tokens ahead we speculate?

src/speculators/models/dflash/core.py
```diff
@@ -158,14 +158,6 @@ def from_training_args(
             ),
         )
 
-        # Create identity mappings if t2d/d2t not provided (no vocab reduction)
-        if t2d is None or d2t is None:
-            vocab_size = kwargs["draft_vocab_size"]
-            # t2d: all tokens in target vocab are in draft vocab
-            t2d = torch.ones(vocab_size, dtype=torch.bool)
-            # d2t: identity mapping (zero offset for all tokens)
-            d2t = torch.zeros(vocab_size, dtype=torch.long)
```
fynnsu, line 167: This is handled in train.py? That's why we're removing? Or something else?

src/speculators/models/dflash/metrics.py
```diff
@@ -51,15 +51,18 @@ def compute_metrics(
...
         pred_ids, target_ids, loss_mask, pos_idx, block_size
     )
 
     metrics: dict[str, Any] = {}
-    metrics["loss"] = loss.detach().clone()
-    metrics["full_acc"] = full_acc
+    metrics["loss sum"] = loss.detach().clone()
+    metrics["loss count"] = torch.tensor(1.0, device=logits.device)
+    # Position 0 is the anchor — intentionally excluded from accuracy
+    metrics["full_acc sum"] = correct_per_pos[1:].sum()
```
fynnsu: I don't link the inconsistency of the labels using "_" and " " as word separators. Personally I would prefer all underscores but I'm also okay with all spaces. Just think we should be consistent.

scripts/train.py
```diff
@@ -615,6 +623,40 @@ def parse_args():
...
+        type=float,
+        default=0.7,
+        help="Geometric decay ratio for COD sampling in P-EAGLE (default: 0.7)",
+    )
+    parser.add_argument(
+        "--down-sample-ratio-min",
+        type=float,
+        default=0.2,
+        help="Minimum retention ratio for COD sampling in P-EAGLE (default: 0.2)",
+    )
```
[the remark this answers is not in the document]
fynnsu: @shanjiaz this should probably be corrected

src/speculators/models/peagle/config.py
```diff
@@ -0,0 +1,67 @@
...
+        description="Number of parallel prediction groups (num_depths)",
+        ge=1,
+        le=16,
+    )
+
+    down_sample_ratio: float = Field(
+        default=0.7,
+        description="Geometric decay ratio for COD sampling (retention rate r)",
+        gt=0.0,
+        lt=1.0,
```
fynnsu: Should we allow `down_sample_ratio=1.0` (i.e. no downsampling)?

src/speculators/models/peagle/core.py
```diff
@@ -0,0 +1,329 @@
...
+    P-EAGLE (Parallel EAGLE) draft model for speculative decoding.
+
+    P-EAGLE extends EAGLE-3 with parallel multi-token prediction using
+    Conditional-On-Distribution (COD) sampling for memory-efficient training.
+
+    Attributes:
+        mask_hidden: Learnable padding parameter [1, 1, 3*hidden_size]
+        fc: Projection layer from 3*hidden_size to hidden_size
+        layers: Transformer decoder layers from EAGLE-3
+        num_depths: Number of parallel prediction depths (groups)
```
fynnsu: Remove or update

src/speculators/models/peagle/core.py
```diff
@@ -0,0 +1,329 @@
...
+        config: PEagleSpeculatorConfig,
+    ):
+        super().__init__(config=config)
+
+        self.num_depths = config.num_depths
+        self.down_sample_ratio = config.down_sample_ratio
+        self.down_sample_ratio_min = config.down_sample_ratio_min
+        self.mask_token_id = config.mask_token_id
+
+        self.lm_head.weight.requires_grad = False
```
fynnsu: We freeze the lm_head? Always? So this just uses the verifier's weights?

src/speculators/models/peagle/core.py
```diff
@@ -0,0 +1,329 @@
...
+        self.lm_head.eval()
+
+        # Learnable mask_hidden parameter for padding unsampled positions
+        self.mask_hidden = torch.nn.Parameter(torch.randn(1, 1, 3 * self.hidden_size))
+
+    def _pad_hidden_states(self, intensors, target_len, hidden, index_list):
```
fynnsu: Is this method being used somewhere?

src/speculators/models/peagle/core.py
```diff
@@ -0,0 +1,329 @@
...
+
+        # Concatenate
+        return torch.cat((intensors, padding_tensor), dim=1)
+
+    def forward(
+        self,
+        hidden_states: torch.Tensor,
+        input_ids: torch.Tensor,
+        lengths: torch.Tensor | None = None,
+        attention_mask: torch.Tensor | None = None,
```
fynnsu: Doesn't look like this is being used.

src/speculators/models/peagle/core.py
```diff
@@ -0,0 +1,329 @@
...
+
+        if cum_len < seq_length:
+            pad_id = len(lengths) - 1 if len(lengths) > 0 else 0
+            sample_ids_list.append(
+                torch.full(
+                    (seq_length - cum_len,), pad_id, dtype=torch.long, device=device
+                )
+            )
+
+        sample_ids = torch.cat(sample_ids_list, dim=0).unsqueeze(0)  # [1, seq_len]
```
fynnsu: This can be vectorized the same way we do for Eagle3.

Copied from `eagle3/attention.py::create_combined_mask_mode`:

```python

    document_ids = torch.repeat_interleave(
        torch.arange(lengths.shape[0], device=lengths.device, dtype=torch.long), lengths
    )
    # Pad ids with -1 to indicate padding
    document_ids = torch.cat(
        [
            document_ids,
            -1
            * torch.ones(
                total_seq_len - document_ids.shape[0],
                device=lengths.device,
                dtype=torch.long,
            ),
        ]
    ).contiguous()

```

src/speculators/models/peagle/core.py
```diff
@@ -0,0 +1,329 @@
...
+
+        target_len = input_ids.shape[1] * num_depths
+
+        batch_size, current_len, hidden_size = hidden_states.shape
```
fynnsu: So during data loading we combine multiple samples into a batch and then truncate/pad to `total_seq_len`. So at this stage `current_len` should be `total_seq_len` (fixed).

I think the below could be cleaned up a bit given that. Like shouldn't `pad_len` almost always be greater than `0` (like unless `num_depths=1`, is that a valid config?)

src/speculators/models/peagle/core.py
```diff
@@ -0,0 +1,329 @@
...
+        single_pos = torch.arange(0, seq_length, dtype=torch.long, device=device)
+        position_ids = single_pos.repeat(num_depths).unsqueeze(0)
+        inputs_embeds = inputs_embeds.to(hidden_states_tensor.dtype)
+
+        # Project concatenated hidden states (3*hidden_size) -> hidden_size
+        hidden_states_tensor = self.fc(hidden_states_tensor)
+
+        layer_input = torch.cat([inputs_embeds, hidden_states_tensor], dim=-1)
+
+        all_indices_list = []
```
fynnsu: I'm a little confused as to what this list stores in the end. Also can we replace this with tensor ops?

src/speculators/models/peagle/core.py
```diff
@@ -0,0 +1,329 @@
...
+        hidden = layer_input
+        for layer in self.layers:
+            output = layer(
+                hidden,
+                attention_mask=attention_mask,
+                position_ids=position_ids,
+                position_embeddings=position_embeddings,
+                **kwargs,
+            )
+            hidden = output if isinstance(output, torch.Tensor) else output[0]
```
fynnsu: Why might this change? Is this to support multiple layer architectures or something?

src/speculators/models/peagle/data.py
```diff
@@ -0,0 +1,71 @@
...
+        next_candidates = (sampled_idx + 1) % seq_length
+        if filter_position_zero:
+            next_candidates = next_candidates[next_candidates != 0]
+        mask = torch.isin(next_candidates, all_valid_indices)
+        prev_indices = next_candidates[mask]
+
+        sample_indices.append(sampled_idx)
+
+    total_additional_length = sum(len(idx) for idx in sample_indices[1:])
+    return sample_indices, total_additional_length
```
fynnsu: Hmm, so I'm not sure I have a good understanding of the data structure we're using to store this information.

It seems to be something like a list of tensors of indices of positions. Where the different tensors correspond to different depths.Are the positions shifted to indicate the corresponding target token?

e.g. 
```
[[a, b, c], [a+1, c+1], [c+2]]
```

Or is it more like
```
[[a, b, c], [a, c], [c]]
```

It also seems like we're setting a target length in core.py of `depth * total_seq_len` which we will always fall significantly short of (unless `down_sample_ratio=1`) and then we pad to that length. We should be able to determine what the expanded length will be after the sampling and use that directly instead of padding and only pad if we don't have enough `loss_mask=1` tokens.

fynnsu: I'd also much prefer a tensor based approach for all of these where we have 1 or 2 tensors to encode the state, instead of lists of uneven tensors.

Unevenly sized tensors (with shapes dependent on data), typically do not work well with torch.compile.

src/speculators/models/peagle/metrics.py
```diff
@@ -0,0 +1,80 @@
...
+    ).unsqueeze(0)
+
+    if loss_mask is not None:
+        orig_positions = all_indices % seq_length
+        sampled_loss_mask = loss_mask[:, orig_positions]
+    else:
+        sampled_loss_mask = torch.ones(
+            1, total_sampled, device=device, dtype=torch.bool
+        )
+
```
fynnsu: I think we can assume `loss_mask is not None` for this function. We already guarantee that in p-eagle core, by initializing it if it is.

src/speculators/models/peagle/metrics.py
```diff
@@ -0,0 +1,80 @@
...
+        "loss count": torch.tensor(1.0, device=device),
+        "full_acc sum": correct_per_pos.sum(),
+        "full_acc count": total_per_pos.sum(),
+    }
+    for depth in range(num_depths):
+        metrics[f"position {depth} acc sum"] = correct_per_pos[depth]
+        metrics[f"position {depth} acc count"] = total_per_pos[depth]
+
+    first_depth_len = len(sample_indices[0])
+    draft_tokens = torch.argmax(logits[:, :first_depth_len], dim=-1)
```
fynnsu: What's going on here with the `:first_depth_len` indexing?

src/speculators/models/attention.py
```diff
@@ -10,6 +10,8 @@
 from torch.nn.attention.flex_attention import BlockMask, flex_attention
 from transformers.modeling_utils import AttentionInterface
 
+_compiled_flex_attention = torch.compile(flex_attention)
```
fynnsu: Why are we doing this instead of `torch.compile`ing the whole `forward` method like we do for the other implementations?

src/speculators/models/metrics.py
```diff
@@ -21,22 +21,21 @@ def compute_accuracy_single_step(
...
 
     Returns:
-        Tuple of (full_accuracy, conditional_accuracy) where conditional accuracy
-        is accuracy given all previous steps were also correct.
+        Tuple of (full_correct, full_total, cond_correct, cond_total) as raw
+        counts suitable for distributed reduction before computing ratios.
     """
     correct = pred_ids == target_ids
-    cond_denom: torch.Tensor | int = correct.numel()
+    cond_total = torch.tensor(correct.numel(), dtype=torch.float, device=correct.device)
```
fynnsu, line 28: In some places we are calling this "count" and some places use "total" for the denominator. Can we make this consistent.

src/speculators/train/trainer.py
```diff
@@ -255,15 +256,19 @@ def val_epoch(self, epoch: int) -> dict[str, float] | None:
 
             if self.is_distributed:
                 for m in metrics.values():
-                    dist.all_reduce(m, op=dist.ReduceOp.AVG)
+                    dist.all_reduce(m, op=dist.ReduceOp.SUM)
```
fynnsu, line 277: Okay so now we always use `ReduceOp.SUM`. Are we handling metrics that aren't "counted" correctly?

I think currently this is just the loss function, which doesn't matter a ton, but arguably we should be doing something like:

```
loss = loss / world_size
```

Maybe we can update `normalize_counted_metrics` so that counted metrics are reduced using their counts and the other metrics are divided by the world size so that we're still taking the average.

src/speculators/train/utils.py
```diff
@@ -119,6 +138,6 @@ def apply_fully_sharded(model: torch.nn.Module):
     for layer in model.layers:  # type: ignore[union-attr]
         fully_shard(layer, mp_policy=mp_policy)
 
-    fully_shard(model)
+    fully_shard(model, mp_policy=mp_policy)
```
fynnsu, line 154: This is a little risky. Why was this added? I think we previously removed the mp_policy from here because it was causing nan values in deeper models.

pyproject.toml
```diff
@@ -140,10 +140,26 @@ follow_imports = 'silent'
...
+disable_error_code = ["arg-type", "dict-item"]
+
+[[tool.mypy.overrides]]
+module = "speculators.models.peagle.attention"
+disable_error_code = ["union-attr"]
+
+[[tool.mypy.overrides]]
+module = "speculators.models.peagle.core"
+disable_error_code = ["assignment", "attr-defined"]
+
```
fynnsu: These are a lot of new module level overrides. We should really be avoiding adding these unless we need to because it limits the effectiveness of the type checker. Unless we need a ton of these, it would be better to just use the inline ignore for specific cases. e.g. `# type: ignore[assignment]`

pyproject.toml
```diff
@@ -249,12 +265,30 @@ select = [
...
+    "SLF001", # allow private member access for model configuration
 ]
 
 "examples/**/*.py" = [
     "INP001", # allow implicit namespace packages in examples
 ]
 
+"src/speculators/models/peagle/core.py" = [
+    "SLF001", # accessing parent class private attributes for inheritance
+    "S101", # assert for training validation is acceptable
```
fynnsu: We probably shouldn't have asserts in the code base. Explicitly raise an error instead.

pyproject.toml
```diff
@@ -249,12 +265,30 @@ select = [
...
+    "PLR2004", # magic value 2 in ndim comparison is clear
```
fynnsu: Don't do a module level ignore for a single issue like this. Use inline comment ignores instead `noqa`.

pyproject.toml
```diff
@@ -249,12 +265,30 @@ select = [
...
+    "C901", # forward() complexity is acceptable
+    "PLR0912", # forward() branch count is acceptable
+    "PLR0915", # forward() statement count is acceptable
+]
```
fynnsu: Is it?

pyproject.toml
```diff
@@ -249,12 +265,30 @@ select = [
...
+
+"src/speculators/models/peagle/attention.py" = [
+    "S101", # assert for type narrowing is acceptable
+]
+
+"src/speculators/train/trainer.py" = [
+    "C901", # train/val epoch complexity is acceptable
+    "PLR0912", # train/val epoch branch count is acceptable
+]
```
fynnsu: Same with these.

examples/train/peagle_qwen3_8b_sharegpt_online_5k.sh
```diff
@@ -0,0 +1,100 @@
...
+    --verifier-name-or-path "$MODEL" \
+    --data-path "$OUTPUT_DIR" \
+    --vllm-endpoint "http://localhost:${VLLM_PORT}/v1" \
+    --save-path "$OUTPUT_DIR/checkpoints" \
+    --epochs "$EPOCHS" \
+    --lr "$LR" \
+    --total-seq-len "$SEQ_LENGTH" \
+    --speculator-type "$SPECULATOR_TYPE" \
+    --num-layers "$NUM_LAYERS" \
+    --para-depths "$PARA_DEPTHS" \
```
fynnsu: This name is outdated right?

We should probably run this (once everything else is resolved) to make sure it works as expected.

src/speculators/models/peagle/attention.py
```diff
@@ -0,0 +1,100 @@
...
+    assert sample_ids is not None, "sample_ids must be provided"
```
fynnsu: If sample ids must be provided, then they shouldn't be an optional arg to the function.

src/speculators/models/peagle/attention.py
```diff
@@ -0,0 +1,100 @@
...
+        parent_pos = q_pos - (q_depth - kv_depth)
+        cross_depth_to_intermediate = kv_is_intermediate & (kv_pos == parent_pos)
+
+        cross_depth_mask = cross_depth_to_depth0 | cross_depth_to_intermediate
+
+        attention_allowed = (same_depth & same_depth_mask) | (
+            (~same_depth) & cross_depth_mask
+        )
+
+        return valid_query & valid_key & hierarchical & same_sample & attention_allowed
```
fynnsu: This seems complex. I will need to review further, but any chance this logic can be simplified. Also have you verified these masks by printing them out, making sure they make sense?

fynnsu (commented):
This review is a bit more in depth. I think a lot of the structure is there, but there's still more clean up to be done, particularly around simplifying and vectorizing the implementation.

src/speculators/models/peagle/config.py
shanjiaz: sure

src/speculators/models/peagle/core.py
shanjiaz: You meant the whole PEagleDraftModel?

src/speculators/models/peagle/core.py
shanjiaz: I think we always do? https://github.com/vllm-project/speculators/blob/main/src/speculators/model.py#L71

src/speculators/models/peagle/core.py
shanjiaz: removed!

src/speculators/models/peagle/core.py
shanjiaz: removed!

src/speculators/models/attention.py
shanjiaz: You're right! reverted.

src/speculators/models/dflash/core.py
shanjiaz: I think load_vocab_mappings should handle this. like eagle3?

src/speculators/models/metrics.py
shanjiaz: yep!

src/speculators/train/trainer.py
shanjiaz: Sure!

src/speculators/models/peagle/core.py
shanjiaz: This is safe to remove.

src/speculators/models/peagle/metrics.py
shanjiaz: I think it's only used by eagle3 in training. I can delete it and just return None

src/speculators/models/peagle/core.py
fynnsu: I think I meant the Attributes part of the docstring. I don't think we really need it

fynnsu (commented):
Generally the changes look good. I didn't go into a ton of detail with the review but I think this is close to landing if everything works and the remaining comments are addressed.

fynnsu (approved):
LGTM! Would be good to update smoke and regression tests too!

src/speculators/models/peagle/data.py
```diff
@@ -0,0 +1,79 @@
...
+        down_sample_ratio_min: Minimum retention ratio floor
+        filter_position_zero: Whether to filter out position 0 from candidates
+
+    Returns:
+        Tuple of:
+            anchor_pos: The starting position in the original sequence the current
+                sampling chain started from.
+            depth: Which COD sampling round each element belongs to
+    """
+    loss_mask = loss_mask.squeeze()
```
rahul-tuli: nit: as far as I understand this is to remove the batch dimension? If so could we add the dimension to make it explicit?
```suggestion
    loss_mask = loss_mask.squeeze(0)
```

src/speculators/models/peagle/core.py
```diff
@@ -0,0 +1,252 @@
...
+        )  # [1, total_sampled, vocab_size]
+
+        if verifier_last_hidden_states is None:
+            raise ValueError("verifier_last_hidden_states required for training")
+        with torch.no_grad():
+            targets = self.verifier_lm_head(
+                self.verifier_norm(verifier_last_hidden_states)
+            )
+
+        targets = targets[:, orig_positions, :]  # [1, total_sampled, vocab_size]
```
rahul-tuli, line 166: Should we add an assert at the top that batch size should be 1?

Something like:
```python
  if input_ids.shape[0] != 1:
      raise ValueError( 
          f"P-EAGLE currently requires batch_size=1, got {input_ids.shape[0]}"
      )   
```

src/speculators/models/peagle/core.py
```diff
@@ -0,0 +1,252 @@
...
+                position_embeddings=position_embeddings,
+                **kwargs,
+            )
+
+        logits = self.lm_head(
+            self.norm(hidden_states)
+        )  # [1, total_sampled, vocab_size]
+
+        if verifier_last_hidden_states is None:
+            raise ValueError("verifier_last_hidden_states required for training")
```
rahul-tuli: Let's move this to the top of the function, doesn't make sense to go through all the computation and then hit this error

examples/train/peagle_qwen3_8b_sharegpt_online_5k.sh
```diff
@@ -0,0 +1,112 @@
...
+
+set -euo pipefail
+
+# ============ Configuration ============
+MODEL="Qwen/Qwen3-8B"
+DATASET="sharegpt"                # sharegpt, ultrachat, or path to custom data
+OUTPUT_DIR="./output/peagle_qwen3_8b_sharegpt"
+VLLM_PORT=8108
+MAX_SAMPLES=5000
+SEQ_LENGTH=4196
```
rahul-tuli: is `4196` intentional or should it be`4096`?

rahul-tuli (commented):
Minor comments, otherwise LGTM! Great job on this.

src/speculators/models/peagle/core.py
shanjiaz: I think we're adding batching in the future. It's in the comment.

src/speculators/models/peagle/core.py
shanjiaz: sure!

examples/train/peagle_qwen3_8b_sharegpt_online_5k.sh
shanjiaz: will update

src/speculators/models/peagle/core.py
rahul-tuli: I strongly feel we should raise a `NotImplementedError` then rather than putting these assumptions in comments

rahul-tuli (approved):
LGTM!

PR #491, opened 2026-04-28
Feat/fp8 connector

shubhra:
## Summary
Optional FP8 (`float8_e4m3fn`) quantization of hidden states during vLLM extraction via a custom **KV connector**  Enabled with **`launch_vllm.py --fp8-quantize`** (`FP8HiddenStatesConnector`)

## Files
- **`fp8_utils.py`** — quant / dequant (per-token scaling)
- **`fp8_hidden_states_connector.py`** — connector: FP8 + scales → safetensors
- **`launch_vllm.py`** — `--fp8-quantize` wires `kv_connector_module_path`
- **`train/data.py`** — `ArrowDataset`: dequantize when `hidden_states_scales` present in safetensors

## Tested
Qwen3-8B, vLLM 0.18.0, temp=`default`

### FP8 (ckpt 2)
| Dataset | k=1 | k=2 | k=3 | k=4 | k=5 |
|---------|-----|-----|-----|-----|-----|
| HumanEval | 1.78 | 2.34 | 2.74 | 3.05 | 3.21 |
| math_reasoning | 1.81 | 2.42 | 2.87 | 3.19 | 3.39 |
| qa | 1.69 | 2.11 | 2.39 | 2.57 | 2.64 |
| question | 1.72 | 2.19 | 2.50 | 2.70 | 2.81 |
| rag | 1.69 | 2.15 | 2.41 | 2.61 | 2.67 |
| summarization | 1.62 | 1.95 | 2.13 | 2.23 | 2.28 |
| translation | 1.65 | 2.04 | 2.26 | 2.39 | 2.46 |

### BF16 (ckpt 2)
| Dataset | k=1 | k=2 | k=3 | k=4 | k=5 |
|---------|-----|-----|-----|-----|-----|
| HumanEval | 1.78 | 2.36 | 2.75 | 3.02 | 3.19 |
| math_reasoning | 1.81 | 2.42 | 2.87 | 3.18 | 3.41 |
| qa | 1.68 | 2.13 | 2.41 | 2.60 | 2.69 |
| question | 1.72 | 2.19 | 2.50 | 2.68 | 2.83 |
| rag | 1.70 | 2.14 | 2.42 | 2.60 | 2.66 |
| summarization | 1.62 | 1.95 | 2.14 | 2.23 | 2.28 |
| translation | 1.65 | 2.04 | 2.26 | 2.38 | 2.46 |

### DIFF
--- a/src/speculators/data_generation/fp8_hidden_states_connector.py
+++ b/src/speculators/data_generation/fp8_hidden_states_connector.py
@@ -29,12 +29,8 @@ from speculators.data_generation.fp8_utils import (
 if TYPE_CHECKING:
     from vllm.v1.attention.backend import AttentionMetadata
 
-ExampleHiddenStatesConnector = _eh_mod.ExampleHiddenStatesConnector
-ExampleHiddenStatesConnectorMetadata = _eh_mod.ExampleHiddenStatesConnectorMetadata
-extract_from_kv_cache = _eh_mod.extract_from_kv_cache
 
-
-class FP8HiddenStatesConnector(ExampleHiddenStatesConnector):
+class FP8HiddenStatesConnector(_eh_mod.ExampleHiddenStatesConnector):
     """Quantizes hidden states to float8_e4m3fn with per-token scaling
     before saving to safetensors.
 
@@ -58,11 +54,13 @@ class FP8HiddenStatesConnector(ExampleHiddenStatesConnector):
         assert isinstance(attn_metadata, CacheOnlyAttentionMetadata)
 
         connector_metadata = self._get_connector_metadata()
-        assert isinstance(connector_metadata, ExampleHiddenStatesConnectorMetadata)
+        assert isinstance(
+            connector_metadata, _eh_mod.ExampleHiddenStatesConnectorMetadata
+        )
 
         os.makedirs(self._storage_path, exist_ok=True)
         for request in connector_metadata.requests:
-            hidden_states = extract_from_kv_cache(
+            hidden_states = _eh_mod.extract_from_kv_cache(
                 kv_layer, request.slot_mapping, request.token_ids.shape[0]
             )
             cpu_hs = hidden_states.detach().cpu()
--- a/src/speculators/data_generation/fp8_hidden_states_connector.py
+++ b/src/speculators/data_generation/fp8_hidden_states_connector.py
@@ -12,7 +12,6 @@ Launch example::
 from __future__ import annotations
 
 import os
-from concurrent.futures import Future, ThreadPoolExecutor
 from typing import TYPE_CHECKING, Any, Optional
 
 import safetensors.torch
@@ -48,16 +47,6 @@ class FP8HiddenStatesConnector(ExampleHiddenStatesConnector):
         hidden_states_scales – fp32 [seq_len, 1]
         token_ids            – int64 [seq_len]
     """
-
-    def __init__(
-        self,
-        vllm_config: "VllmConfig",
-        role: KVConnectorRole,
-        kv_cache_config: Optional["KVCacheConfig"] = None,
-    ):
-        super().__init__(vllm_config, role, kv_cache_config)
-        self._executor = ThreadPoolExecutor(max_workers=4)
-        self._pending: list[Future] = []
 
     def save_kv_layer(
         self,
@@ -97,13 +86,4 @@ class FP8HiddenStatesConnector(ExampleHiddenStatesConnector):
                 SCALES_KEY: scales,
                 "token_ids": request.token_ids.detach().cpu(),
             }
-            self._pending.append(
-                self._executor.submit(
-                    safetensors.torch.save_file, tensors, request.filename
-                )
-            )
-
-    def wait_for_save(self):
-        for f in self._pending:
-            f.result()
-        self._pending.clear()
+            safetensors.torch.save_file(tensors, request.filename)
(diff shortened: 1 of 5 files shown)


fynnsu:
Hi @shubhra, a couple things:
1. Can you give me a sense of what kind of performance change I can expect when using this? Does it speed up online/offline training? Or are the advantages mostly in the reduced disk size when doing offline training?
2. I like the approach of an out-of-tree HiddenStatesConnector, but I'm a little bit concerned about the packaging of this. We typically suggest using a separate virtual environment for launch_vllm (with just `vllm`) vs training (with just `speculators`) to reduce the number of potential dependency mismatches and other issues. These changes would require vllm to be installed alongside speculators, so that it can access the new connector. Do you have any thoughts on this?

shubhra:
> Hi @shubhra, a couple things:
> 
> 1. Can you give me a sense of what kind of performance change I can expect when using this? Does it speed up online/offline training? Or are the advantages mostly in the reduced disk size when doing offline training?
> 2. I like the approach of an out-of-tree HiddenStatesConnector, but I'm a little bit concerned about the packaging of this. We typically suggest using a separate virtual environment for launch_vllm (with just `vllm`) vs training (with just `speculators`) to reduce the number of potential dependency mismatches and other issues. These changes would require vllm to be installed alongside speculators, so that it can access the new connector. Do you have any thoughts on this?


1. This won't really speed up training (online or offline): training still sees bf16 after load, and you add a small dequant step when reading FP8 files. The big saving is from smaller hidden-state files (~50% smaller), which matters most for offline datagen. Datagen itself isn’t going to be any faster (quant work vs smaller writes depends on the machine).

2. Training can stay speculators-only (no vLLM); nothing changes there. Only the process that runs vLLM to generate hidden states needs both: FP8 uses a connector that lives only in speculators for now, not the built-in bf16 one inside vLLM — so the vLLM process has to be able to import speculators. If we want vLLM-only, the next step is to add this connector upstream in vLLM next to `ExampleHiddenStatesConnector`. Happy to do that on the vllm side as well.

orestis-z (commented):
The FP8 quantization/dequantization logic is correct: per-token scaling with clamped denominators, proper flatten-before-quant / reshape-after-dequant symmetry between the connector and the data loader. No correctness issues found. Note that CI is still red from lint (EN DASH in docstrings, per the CodeRabbit comment) -- a `make style` pass should clear it.

src/speculators/data_generation/fp8_hidden_states_connector.py
```diff
@@ -0,0 +1,80 @@
...
+            flat = cpu_hs.reshape(cpu_hs.shape[0], -1)
+            fp8_flat, scales = quantize_tensor_to_fp8(flat)
+            fp8_hs = fp8_flat.reshape(original_shape)
+
+            tensors: dict[str, torch.Tensor] = {
+                "hidden_states": fp8_hs,
+                SCALES_KEY: scales,
+                "token_ids": request.token_ids.detach().cpu(),
+            }
+            safetensors.torch.save_file(tensors, request.filename)
```
speculatorsbot, line 80: This `save_kv_layer` override was written against the original vLLM 0.18.0 connector API ([`9433acb8d`](https://github.com/vllm-project/vllm/commit/9433acb8d)), where `save_kv_layer` was the primary extraction hook and `ExampleHiddenStatesConnectorMetadata` had a `requests: list[ReqMeta]` field with `slot_mapping` on each `ReqMeta`.

Since then, the upstream connector has been significantly refactored (notably [`04bf5a35f`](https://github.com/vllm-project/vllm/commit/04bf5a35f) and [`ebc6ef971`](https://github.com/vllm-project/vllm/commit/ebc6ef971)):

1. `save_kv_layer` is now a **no-op** in the base class -- hidden states are cached by `CacheOnlyAttentionLayer` during forward, and extraction + disk write happens later in `get_finished` via `_submit_async_write`.
2. `ExampleHiddenStatesConnectorMetadata.requests` was replaced with `pending_saves: list[PendingSave]`, and `ReqMeta.slot_mapping` was replaced with `PendingSave.block_ids` (slot mapping is now computed at write time from block IDs).
3. Writes are now async: DtoH copy on a dedicated CUDA stream, then disk write in a thread pool, with file-level flocking for client synchronization.

As-is, this code will raise `AttributeError: 'ExampleHiddenStatesConnectorMetadata' object has no attribute 'requests'` at runtime. And even if the attribute names were patched, the base class's `get_finished` would still run `_submit_async_write`, overwriting the FP8 file with BF16 data (since `get_finished` is not overridden here).

For the rebase, I'd suggest overriding `_submit_async_write` instead of `save_kv_layer`. That method receives a `PendingSave`, extracts from the KV cache, does the DtoH copy, and submits to the thread pool. The FP8 connector could call `super()._submit_async_write(...)` with a modified `_write_tensors` that adds quantization before `save_file`, or override `_submit_async_write` directly to inject quantization after the DtoH copy lands on CPU.

speculatorsbot (commented):
The FP8 quantization math (per-token scaling, flatten/reshape symmetry between writer and reader) is sound, and the benchmark numbers show minimal degradation vs BF16. However, the connector was built against vLLM 0.18.0's `ExampleHiddenStatesConnector` API, which has since been substantially refactored. This goes beyond what a git rebase can fix -- the connector needs to be re-architected against the current API. See inline comment for details.

Also, the new quantization modules (`fp8_utils.py`, `fp8_hidden_states_connector.py`) have no unit tests. The round-trip quant/dequant logic, edge cases (zero rows, single-token sequences), and the connector's file format should have test coverage.

PR #344, opened 2026-03-13
add multimodal preprocessing support

shx2005:
## Purpose

Enable multimodal-aware(currently single image-text pair) preprocessing for the EAGLE offline data-generation pipeline while preserving backward compatibility for existing text-only workflows.

## Description
This PR adds multimodal preprocessing support and refactors related preprocessing internals:

- Added `--multimodal` flag to offline generation entrypoint and passed it through to preprocessing.
- Added helpers for:
  - multimodal content normalization,
  - image reference extraction,
  - singleton batch flattening with clearer error context.
- Generalized assistant-mask/pattern logic to work with either tokenizer or processor callers.
- Kept compatibility for existing callers:
  - `is_multimodal=None` defaults to text mode.
- Updated tests to new helper/function names and added a regression test for default text fallback behavior.

## Related Issue

https://github.com/vllm-project/speculators/issues/290

## Tests

- Ran an existing pure-text Qwen example end-to-end to ensure this change does not affect the original text-only path.
- Ran training in my adapted VL code path to verify multimodal preprocessing outputs are consumable and behavior is correct.

Notes:

- The data generation and training workflow in speculators will be updated soon and this PR intentionally focuses on organizing and stabilizing the preprocessing part only.

I have filled in:

- [x] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [x] The test plan/results, such as providing test command and pasting the results.
- [ ] (Optional) The necessary documentation update.
- [x] I (a human) have written or reviewed the code in this pr to the best of my ability.

PR #495, opened 2026-04-30
Support multi-modal datasets

### LINKED ISSUE #290: [RFC]: VL model (Qwen3-VL) image-text pair input support
# [RFC] VL Model (Qwen3-VL) Image-Text Pair Input Support

This document aims to bring multimodal (image-text) support to Speculators.

Current focus: single image-text pairs, with Qwen3-VL support.

We plan to deliver this in four steps:
1. Consolidate vLLM-side changes (work in progress in vLLM PR `#34608`: <https://github.com/vllm-project/vllm/pull/34608>).
2. Add training integration.
3. Add an evaluation pipeline.
4. Provide end-to-end examples and expand the README.

---

## Results Preview
We use Qwen3-VL-4B (the same model used to train Speculators) to re-annotate 4k high-quality samples from LLaVA-CoT-100k (filter requirement: minimum word count > 10) and train for 10 epochs on 2x H800 PCIe, taking approximately 2.5 hours.
Key metrics over 50 validation samples at temperature=0 (Baseline vs Speculator):

| Metric | Baseline | Speculator | Δ  |
| --- | --- | --- | --- |
| Latency avg (ms) | 6310.57 | 4401.96 | -30.24% |
| Requests/s | 0.1579 | 0.2261 | +43.14% |
| Elapsed (s) | 316.56 | 221.15 | -30.14% |
| Prompt tokens (total) | 14,052 | 14,052 | +0.00% |
| Completion tokens (total) | 47,996 | 33,407 | -30.40% |
| All tokens (total) | 62,048 | 47,459 | -23.51% |
| All tokens/s | 196.00 | 214.60 | +9.49% |
| Completion tokens/s | 151.62 | 151.06 | -0.37% |

Weighted per-position acceptance rates:
```text
[0.45  0.156 0.054]
```

Conditional acceptance rates:
```text
[0.45  0.346 0.344]
```

**Notes:**
- Training gains are limited, mainly due to limited data (only 4k samples) and training on VL-only data without mixing in text-only data (see: https://arxiv.org/pdf/2505.14260).
- Completion token totals differ significantly between runs, so a direct latency comparison is not meaningful (the runs are not strictly comparable in output length).

Overall, this indicates the end-to-end code path and pipeline logic are working as expected.

## Architecture / Data Flow Overview
- Draft training target remains hidden-state learning over sequence tensors (`input_ids`, `hidden_states`, `loss_mask`).
- For VL samples, the pipeline additionally stores prefill-time decoder `input_embeds` from vLLM in the `.pt` file (key `input_embeds`), which `standardize_data_v1` maps to `input_embeds_override` at training time.
- In `Eagle3DraftModel.forward(...)`, `input_embeds_override` is consumed only at `ttt_step == 0`, so the draft sees the same merged vision-text embeddings the verifier produced at prefill.
- For `ttt_step > 0`, the model continues with the standard token-id drafting path, so no additional multimodal-specific operation is required.
- vLLM request/runtime is responsible for generating and persisting image-conditioned features, and draft forward consumes these saved features (via `input_embeds_override` at step-0, and optional `multimodal_embeddings` path).
- Data generation/preprocessing is responsible for request construction, prompt-token expansion alignment, and loss-mask alignment.
- This design keeps current runtime integration stable and remains aligned with upcoming `sample-generate` convergence.

## Training Logs

Image

Image

Image

### `examples/convert_pokemon_gpt4o_to_jsonl.py` (new) 
Behavior changes:
- CLI updates: adds `--dataset`, `--split`, `--output`, `--image-root`, `--image-output-dir`.
- Adds a VL dataset conversion utility from HF dataset format to Speculators VL JSONL format.
- Normalizes image references from multiple source forms: `path`, `url`, `file://`, and byte-backed dict entries.
- Attaches image items to the first user turn and strips inline `<image>` markers from text content.
- Supports image path management via `--image-output-dir` (materialize/copy images to local files) and `--image-root` (store relative names in JSONL and resolve at generation time).

Key functions:
- `_resolve_image_ref`, `_extract_image_refs`, `_build_messages`, `_write_image_bytes`, `_copy_image_file`.

### `scripts/data_generation_offline.py`
Behavior changes:
- CLI updates: adds `--data-mode {text,vl}`, `--image-field`, `--image-root`; validates image args only when `--data-mode vl`.
- `data-mode=text`: no behavior change to existing text-only generation path.
- `data-mode=vl`:
  1. `_resolve_image_ref(...)` resolves image refs (`str`/`dict`, local path, URL, `file://`) into canonical `PIL.Image(RGB)`.
  2. `_build_multimodal_inputs(...)` converts `multi_modal_data` into vLLM-compatible payload (`{"image": [image]}`), with strict single-image contract.
  3. Calls `generator.generate(..., prompt_texts=..., multimodal_inputs=...)` for multimodal requests.
  4. `_align_loss_mask_to_vllm(...)` aligns loss masks to multimodal-expanded token IDs and masks image-only expanded positions as non-loss tokens.
- Persists `input_embeds` in output `.pt`: these are the exact prefill decoder input embeddings captured from vLLM (already containing the model-consumed multimodal embedding result for VL samples), then mapped to `input_embeds_override` during training so `ttt_step==0` can consume exact runtime inputs instead of recomputing from token IDs.
- Preprocessing is explicitly switched by `is_multimodal` based on `data-mode`.

Key functions:
- `_resolve_image_ref`, `_build_multimodal_inputs`, `_align_loss_mask_to_vllm`.

### `scripts/train.py`
Behavior changes:
- CLI updates: adds `--data-format` (e.g. `text-v1`, `vl-v1`) and keeps `--data-format-version` as a deprecated compatibility option.
- Unconditionally calls `_extract_text_config(...)` to derive draft layer shapes from the text backbone config; for text-only verifiers it returns the config unchanged, for VL verifiers (e.g. Qwen3-VL) it extracts `.text_config` to avoid shape mismatches from multimodal wrapper fields.
- Builds verifier metadata via `VerifierConfig.from_config(...)` from the loaded HF config; this is needed to keep the real verifier architecture at runtime (instead of a hardcoded text-only assumption).

Key functions:
- `setup_dataloader`, `_extract_text_config`, `create_transformer_layer_config`, `main`, `parse_args`.

### `src/speculators/data_generation/custom_worker.py`
Behavior changes:
- Adds a VL-specific `deepstack_input_embeds` accumulation branch so Qwen3-VL hidden-state capture follows the same vision-text merged forward path as the verifier model.
- `HiddenStatesWorkerExtension` now buffers/resets and returns captured decoder `input_embeds` (for VL step-0 override) via RPC, in addition to intermediate hidden states.

Key additions:
- `_store_input_embeds`, `_get_captured_input_embeds`.

### `src/speculators/data_generation/vllm_hidden_states_generator.py`
Behavior changes:
- Builds a dedicated VL request path with `InputProcessor` (`prompt_text` + `multi_modal_data`) and per-image unique `multi_modal_uuids`.
- Extends `generate(...)` to accept `prompt_texts` + `multimodal_inputs` while preserving the original text-only path.
- For VL samples, returns vLLM-expanded prompt `input_ids` for downstream mask/shape alignment, and enforces multimodal request validity (`mm_features` must exist).
- Collects worker-captured `input_embeds` via RPC, slices per sample, checks total-token consistency (`expected == captured`), and adds `get_multimodal_prompt_token_count(...)` plus stricter mismatch protection, because image placeholders can easily drift from actual image-embedding expansion.

Key functions:
- `_process_multimodal_inputs`, `_validate_multimodal_engine_request`, `_extract_and_validate_prompt_token_ids`, `get_multimodal_prompt_token_count`, `generate`.

### `src/speculators/data_generation/preprocessing.py`
Behavior changes:
- Refactors preprocessing into explicit text and multimodal paths, each with its own assistant-mask/pattern detection flow.
- Normalizes multimodal conversation content (including inline image token cleanup) and extracts structured `multi_modal_data` image references.
- Improves shape robustness with singleton-batch flattening; multimodal preprocessing now emits `input_ids`, `loss_mask`, `multi_modal_data`, and `prompt`.
- Makes mode selection explicit in `load_and_preprocess_dataset(...)` via `is_multimodal`, and dispatches tokenizer/processor + dataset mapping accordingly (`AutoTokenizer` for text, `AutoProcessor` for multimodal).

Key functions:
- `_normalize_content`, `_get_conversations_from_examples`, `_flatten_singleton_batch`, `_get_tokenizer_from_processor`, `_get_image_ref`, `_extract_multimodal_data_from_conversation`.
- `_supports_assistant_mask_text`, `_detect_assistant_pattern_text`, `_supports_assistant_mask_multimodal`, `_detect_assistant_pattern_multimodal`.
- `_preprocess_batch_text`, `_preprocess_batch_multimodal`.
- `build_eagle3_dataset`, `load_and_preprocess_dataset`.

### `src/speculators/models/eagle3/core.py`
Behavior changes:
- Adds `_merge_multimodal_embeddings(...)` to masked-merge multimodal embeddings into token embeddings.
- Extends `Eagle3DraftModel.forward(...)` with VL-related inputs:
  - `input_embeds_override`: used only at `ttt_step == 0`, with strict shape-prefix validation against `input_ids`.
  - `multimodal_embeddings` + `is_multimodal`: used to scatter multimodal features into embeddings for non-override path.

### `src/speculators/train/data.py`
Behavior changes:
- `shift_batch(...)` now shifts `input_embeds_override` in lockstep with all sequence-aligned tensors.
- `standardize_data_v1(...)` maps serialized `input_embeds` to training-time key `input_embeds_override`.
- Adds `get_standardize_fn(data_format)` to resolve `v0/v1/text-v1/vl-v1` and numeric aliases.
- `Eagle3SampleFileDataset.__getitem__(...)` applies hidden-state dtype policy to `input_embeds_override` as well.

---

DarkLight1337:

## Purpose

FIX https://github.com/vllm-project/speculators/issues/290

By forwarding the conversations directly into Chat Completions API, and also observing that `AutoProcessor` returns tokenizer instance for text-only models, we can simplify the code a lot compared to #344.

cc @shx2005

## Description

- Enable preprocessing of multimodal datasets. No user flag is required; we detect this automatically based on `isinstance(..., ProcessorMixin)`.
- Add `--trust-remote-code` flag to data preparation and training scripts.
- Add `hf_name` and `filter_fn` options to `DatasetConfig`.
- Add support for ShareGPT4V dataset.
  - Only samples from COCO are supported right now.
  - You need to download the images separately since it's not handled by HF Datasets. Use the `COCO_DIR` environment variable to control where COCO images are read from.
  - To avoid copying the images when storing the preprocessed dataset and HTTP transfer to vLLM, we express image inputs in terms of file URLs instead of base64-encoded images. For vLLM to access those files, you should pass `--allowed-media-domain-paths /path/to/coco` when serving it.
- Disable caching for dataset normalization as it makes debugging `normalize_fn` more difficult. To reduce overhead, this step is now executed after `raw_dataset.select`.
- Use `--enforce-eager` for vLLM in e2e tests by default to reduce startup time.
- Add `torchaudio` and `torchvision` to dependencies.
- Fix whitespace issues in the help text of the scripts.

## Related Issue

#290

## Tests

Add integration and e2e tests to ensure MM support. Note:
  - e2e smoke tests use a dummy COCO image so they can be run in CI.
  - e2e regression tests are only run if the real COCO images are downloaded.

I have filled in:


shx2005:
@DarkLight1337 Thanks a lot for working on this and for cc'ing me!

I haven't had a chance to look through this PR carefully yet, but I'll review it in more detail shortly. I also just organized a related implementation in #497, where I have been working on the full multimodal E2E training flow and am currently wrapping up the final testing/validation.

There may be some overlap between the two PRs. I'll take a closer look at the concrete implementations in both PRs and see how we can best align them, combine the useful parts, or avoid duplicated work.

Thanks again for the help and for pushing this forward!

DarkLight1337:
> I haven't had a chance to look through this PR carefully yet, but I'll review it in more detail shortly. I also just organized a related implementation in #497, where I have been working on the full multimodal E2E training flow and am currently wrapping up the final testing/validation.
> 
> There may be some overlap between the two PRs. I'll take a closer look at the concrete implementations in both PRs and see how we can best align them, combine the useful parts, or avoid duplicated work.
> 
> Thanks again for the help and for pushing this forward!

Thanks, sounds good! I'm able to run preprocessing and training E2E in this PR now, =using these commands one by one:

```
MODEL=Qwen/Qwen3-VL-2B-Instruct
DATA_OUTPUT=./data/sharegpt4v_coco
NUM_SAMPLES=1000

COCO_DIR=/media/data2/coco python scripts/prepare_data.py --model $MODEL --data sharegpt4v_coco --max-samples $NUM_SAMPLES --output "$DATA_OUTPUT" --overwrite
python scripts/launch_vllm.py $MODEL --allowed-local-media-path / --max-model-len -1
python scripts/data_generation_offline.py --preprocessed-data "$DATA_OUTPUT" --validate-outputs
python scripts/train.py \
  --verifier-name-or-path $MODEL \
  --data-path "$DATA_OUTPUT" \
  --hidden-states-path "$DATA_OUTPUT/hidden_states" \
  --save-path "$DATA_OUTPUT/checkpoints" \
  --draft-vocab-size 32000 \
  --epochs 5 \
  --lr 1e-4 \
  --total-seq-len 8192 \
  --on-missing raise
```

Training metrics from the logs:
```
[11:09:54] INFO     val/loss_0_epoch=2.270, val/full_acc_0_epoch=0.392, val/cond_acc_0_epoch=0.392, val/loss_1_epoch=2.908, val/full_acc_1_epoch=0.077, val/cond_acc_1_epoch=0.195,            trainer.py:264
                    val/loss_2_epoch=3.135, val/full_acc_2_epoch=0.009, val/cond_acc_2_epoch=0.122, val/loss_epoch=8.313, epoch=0                                                                            
[11:10:21] INFO     val/loss_0_epoch=1.684, val/full_acc_0_epoch=0.481, val/cond_acc_0_epoch=0.481, val/loss_1_epoch=2.438, val/full_acc_1_epoch=0.140, val/cond_acc_1_epoch=0.290,            trainer.py:264
                    val/loss_2_epoch=2.792, val/full_acc_2_epoch=0.028, val/cond_acc_2_epoch=0.204, val/loss_epoch=6.915, epoch=1                                                                            
[11:10:46] INFO     val/loss_0_epoch=1.583, val/full_acc_0_epoch=0.500, val/cond_acc_0_epoch=0.500, val/loss_1_epoch=2.379, val/full_acc_1_epoch=0.163, val/cond_acc_1_epoch=0.326,            trainer.py:264
                    val/loss_2_epoch=2.783, val/full_acc_2_epoch=0.042, val/cond_acc_2_epoch=0.255, val/loss_epoch=6.745, epoch=2                                                                            
[11:11:11] INFO     val/loss_0_epoch=1.424, val/full_acc_0_epoch=0.522, val/cond_acc_0_epoch=0.522, val/loss_1_epoch=2.212, val/full_acc_1_epoch=0.180, val/cond_acc_1_epoch=0.344,            trainer.py:264
                    val/loss_2_epoch=2.606, val/full_acc_2_epoch=0.048, val/cond_acc_2_epoch=0.268, val/loss_epoch=6.242, epoch=3                                                                            
[11:11:35] INFO     val/loss_0_epoch=1.513, val/full_acc_0_epoch=0.515, val/cond_acc_0_epoch=0.515, val/loss_1_epoch=2.321, val/full_acc_1_epoch=0.172, val/cond_acc_1_epoch=0.332,            trainer.py:264
                    val/loss_2_epoch=2.728, val/full_acc_2_epoch=0.047, val/cond_acc_2_epoch=0.269, val/loss_epoch=6.561, epoch=4
```

Could you try running your setup on my PR and see if you get similar results as those on your PR?

I have gone through your PR briefly. I think we can break this down into the following components:
- Dataset processing (mainly from my PR)
- Changes related to `allow_prefix_truncation` (from your PR)
- Changes related to `hidden_states_dtype` (from your PR)

src/speculators/data_generation/preprocessing.py
```diff
@@ -428,17 +527,18 @@ def load_raw_dataset(train_data_path: str, num_proc: int = 8) -> HFDataset:
         )
 
     config = DATASET_CONFIGS[train_data_path]
-    raw_dataset = load_dataset(config.hf_path, split=config.split)
+    raw_dataset = load_dataset(config.hf_path, config.hf_name, split=config.split)
 
-    if config.normalize_fn is not None:
```
DarkLight1337, line 433: I disabled caching for this as it made debugging the `normalize_fn` more difficult. To reduce the overhead I have moved this step after `raw_dataset.select`.

shx2005:
> > I haven't had a chance to look through this PR carefully yet, but I'll review it in more detail shortly. I also just organized a related implementation in #497, where I have been working on the full multimodal E2E training flow and am currently wrapping up the final testing/validation.
> > There may be some overlap between the two PRs. I'll take a closer look at the concrete implementations in both PRs and see how we can best align them, combine the useful parts, or avoid duplicated work.
> > Thanks again for the help and for pushing this forward!
> 
> Thanks, sounds good! I'm able to run preprocessing and training E2E in this PR now, =using these commands one by one:
> 
> ```
> MODEL=Qwen/Qwen3-VL-2B-Instruct
> DATA_OUTPUT=./data/sharegpt4v_coco
> NUM_SAMPLES=1000
> 
> COCO_DIR=/media/data2/coco python scripts/prepare_data.py --model $MODEL --data sharegpt4v_coco --max-samples $NUM_SAMPLES --output "$DATA_OUTPUT" --overwrite
> python scripts/launch_vllm.py $MODEL --allowed-local-media-path / --max-model-len -1
> python scripts/data_generation_offline.py --preprocessed-data "$DATA_OUTPUT" --validate-outputs
> python scripts/train.py \
>   --verifier-name-or-path $MODEL \
>   --data-path "$DATA_OUTPUT" \
>   --hidden-states-path "$DATA_OUTPUT/hidden_states" \
>   --save-path "$DATA_OUTPUT/checkpoints" \
>   --draft-vocab-size 32000 \
>   --epochs 5 \
>   --lr 1e-4 \
>   --total-seq-len 8192 \
>   --on-missing raise
> ```
> 
> Training metrics from the logs:
> 
> ```
> [11:09:54] INFO     val/loss_0_epoch=2.270, val/full_acc_0_epoch=0.392, val/cond_acc_0_epoch=0.392, val/loss_1_epoch=2.908, val/full_acc_1_epoch=0.077, val/cond_acc_1_epoch=0.195,            trainer.py:264
>                     val/loss_2_epoch=3.135, val/full_acc_2_epoch=0.009, val/cond_acc_2_epoch=0.122, val/loss_epoch=8.313, epoch=0                                                                            
> [11:10:21] INFO     val/loss_0_epoch=1.684, val/full_acc_0_epoch=0.481, val/cond_acc_0_epoch=0.481, val/loss_1_epoch=2.438, val/full_acc_1_epoch=0.140, val/cond_acc_1_epoch=0.290,            trainer.py:264
>                     val/loss_2_epoch=2.792, val/full_acc_2_epoch=0.028, val/cond_acc_2_epoch=0.204, val/loss_epoch=6.915, epoch=1                                                                            
> [11:10:46] INFO     val/loss_0_epoch=1.583, val/full_acc_0_epoch=0.500, val/cond_acc_0_epoch=0.500, val/loss_1_epoch=2.379, val/full_acc_1_epoch=0.163, val/cond_acc_1_epoch=0.326,            trainer.py:264
>                     val/loss_2_epoch=2.783, val/full_acc_2_epoch=0.042, val/cond_acc_2_epoch=0.255, val/loss_epoch=6.745, epoch=2                                                                            
> [11:11:11] INFO     val/loss_0_epoch=1.424, val/full_acc_0_epoch=0.522, val/cond_acc_0_epoch=0.522, val/loss_1_epoch=2.212, val/full_acc_1_epoch=0.180, val/cond_acc_1_epoch=0.344,            trainer.py:264
>                     val/loss_2_epoch=2.606, val/full_acc_2_epoch=0.048, val/cond_acc_2_epoch=0.268, val/loss_epoch=6.242, epoch=3                                                                            
> [11:11:35] INFO     val/loss_0_epoch=1.513, val/full_acc_0_epoch=0.515, val/cond_acc_0_epoch=0.515, val/loss_1_epoch=2.321, val/full_acc_1_epoch=0.172, val/cond_acc_1_epoch=0.332,            trainer.py:264
>                     val/loss_2_epoch=2.728, val/full_acc_2_epoch=0.047, val/cond_acc_2_epoch=0.269, val/loss_epoch=6.561, epoch=4
> ```
> 
> Could you try running your setup on my PR and see if you get similar results as those on your PR?
> 
> I have gone through your PR briefly. I think we can break this down into the following components:
> 
> * Dataset processing (mainly from my PR)
> * Changes related to `allow_prefix_truncation` (from your PR)
> * Changes related to `hidden_states_dtype` (from your PR)

@DarkLight1337 Thanks, this sounds good to me!

I’ll run my setup on this PR and compare the results with #497. Since there is some overlap in the preprocessing path, I’ll take a closer look at both implementations and see how we can best combine the useful pieces.

If this PR works well as the generic dataset-processing base, I’m happy to align around it and port over any remaining Qwen3-VL / vLLM 0.20 E2E pieces from #497, including the prompt-token alignment, allow_prefix_truncation, and hidden_states_dtype changes.

After that, I can help split things into smaller PRs for easier review.

src/speculators/data_generation/preprocessing.py
```diff
@@ -107,44 +117,110 @@ def _normalize_conversation(
...
+    if part_type == "text":
+        return {"type": "text", "text": part["text"]}
+
+    for modality in ("image", "video", "audio"):
+        if part_type == modality:
+            if local_path := part.get("path"):
+                file_url = f"file://{local_path}"
+                return {"type": f"{modality}_url", f"{modality}_url": {"url": file_url}}
+            if url := part.get("url"):
+                return {"type": f"{modality}_url", f"{modality}_url": {"url": url}}
```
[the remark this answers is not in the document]
DarkLight1337: Fixed

src/speculators/train/data.py
```diff
@@ -258,12 +258,16 @@ def _maybe_generate_hs(self, index: int) -> dict[str, torch.Tensor] | None:
...
+        item = self.data[index]
+        input_ids = item["input_ids"].tolist()
+        messages = self.data.get("_vllm_messages")
+
         try:
             hs_filepath = generate_hidden_states(
                 self.client,  # type:ignore[arg-type]
                 self.model,  # type:ignore[arg-type]
                 input_ids,
+                messages=messages,
```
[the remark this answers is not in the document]
DarkLight1337: Fixed

src/speculators/data_generation/vllm_client.py
```diff
@@ -114,15 +117,27 @@ async def generate_hidden_states_async(
...
+            timeout=timeout,
+        )
+    else:
+        coro = client.chat.completions.create(
+            model=model,
+            messages=messages,
+            max_tokens=1,
+            extra_body={"return_token_ids": True},
+            timeout=timeout,
+        )
```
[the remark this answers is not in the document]
DarkLight1337: Fixed

DarkLight1337:
@coderabbitai review

DarkLight1337:
Finally got CI green

shx2005:
@DarkLight1337 Thanks, and congrats on getting CI green! Great work pushing this through.

I took a quick look and also tried a small local porting run, though I haven’t reviewed everything in detail yet. I think this PR can serve as a good base for the generic multimodal dataset-processing work.

After this PR is merged, I’ll manually resolve the conflicts in #497, continue from the updated main branch, and port the remaining Qwen3-VL / vLLM 0.20 E2E correctness pieces from #497.

Those follow-up pieces mainly include prompt-token alignment, hidden-state dtype handling, Qwen3-VL image-token/loss-mask alignment, online generation, and any needed vLLM launch/runtime fixes.

Thanks again for working on this and for coordinating here. Please let me know if you prefer a different split.

DarkLight1337:
Yes, that sounds good

shx2005:
> @DarkLight1337 Thanks, and congrats on getting CI green! Great work pushing this through.
> 
> 
> 
> I took a quick look and also tried a small local porting run, though I haven’t reviewed everything in detail yet. I think this PR can serve as a good base for the generic multimodal dataset-processing work.
> 
> 
> 
> After this PR is merged, I’ll manually resolve the conflicts in #497, continue from the updated main branch, and port the remaining Qwen3-VL / vLLM 0.20 E2E correctness pieces from #497.
> 
> 
> 
> Those follow-up pieces mainly include prompt-token alignment, hidden-state dtype handling, Qwen3-VL image-token/loss-mask alignment, online generation, and any needed vLLM launch/runtime fixes.
> 
> 
> 
> Thanks again for working on this and for coordinating here. Please let me know if you prefer a different split.
> 
> 

@shanjiaz Maybe we can handle this PR first.

tests/integration/datagen/test_regex_patterns.py
```diff
@@ -29,49 +33,91 @@
     "openai/gpt-oss-20b",
 ]
 
+if Version(TRANSFORMERS_VERSION) >= Version("5.5.0"):
+    # Multimodal
+    MODELS.append("google/gemma-4-E2B-it")
```
DarkLight1337, line 39: I used gemma4 instead of qwen3-vl here as qwen3-vl doesn't have an assistant mask pattern

DarkLight1337:
Just realized that the messages weren't actually being sent to vLLM. Fixed now

DarkLight1337:
Now getting `Prompt token IDs mismatch` when running offline data processing, going to investigate further later.

DarkLight1337:
It has been fixed. Now to add a test...

DarkLight1337:
Done setting up the nightly tests.

src/speculators/data_generation/preprocessing.py
```diff
@@ -387,39 +541,48 @@ def build_eagle3_dataset(
...
-            turn_dropout,
-            minimum_valid_tokens,
-        ),
-        batched=True,
-        num_proc=num_proc,
-        batch_size=1000,
-        remove_columns=original_cols,
-        keep_in_memory=True,  # skip caching
-    )
+    # Avoid CPU contention for MM processing:
```
DarkLight1337, line 576: This is very important for MM processing, I got around 10x improvement in preprocessing throughput (50 examples/s -> 500 examples/s for Qwen3.5)

fynnsu:
Hi @DarkLight1337, sorry I haven't had time to do a proper review, it's unfortunately a busy time right now but I'll try to get to this soon. In the meantime the main concern I have with multi-modal is how the loss mask calculation will be handled. 

Currently, we have a system where:
1. `prepare_data.py` tokenizes/applies chat template to the whole dataset while using heuristics/templates to get an assistant token mask. These are both saved to disk in a new processed dataset.
2. Then during online training or offline data gen we pass the token ids directly to the `/completions` endpoint which ensures that the hidden states we receive will line up with the loss mask we generated during step 1. 

Under the current system, we need to use `/completions` + `token_ids` to ensure the loss mask lines up perfectly. Otherwise it's possible vllm will use a slightly different chat template or tokenize differently leading to misaligned sequences. I haven't had time to dive into your implementation but I saw you're switching to the `/chat/completions` endpoint since it has better handling for mulit-modal inputs. Could you explain how the new system handles the loss mask creation / alignment?

I think the current system is also not perfect and we will likely need to update it at some point to use the vllm `render` endpoint for templating/tokenization to ensure alignment with vllm. So I don't think this multimodal implementation needs to be perfect right now, just want to make sure we aren't creating a silent loss mask alignment issue.

DarkLight1337:
By passing `add_generation_prompt=False` to vLLM Chat Completions API, we get the same tokens as what the local `processor.apply_chat_template` call returns.

The existing `extract_output` function verifies that the tokens returned by vLLM match the tokens returned from the local chat template. As long as this holds true, the loss mask should be correct right?

fynnsu:
> By passing `add_generation_prompt=False` to vLLM Chat Completions API, we get the same tokens as what the local `processor.apply_chat_template` call returns.

Yeah, my concern is whether this is always the case. If vLLM has custom logic for applying the chat template that doesn't exist in the transformers implementation then there will be a misalignment. I'll admit, I don't have a great sense of how common this is, but my understanding is this logic often undergoes a number of small fixes when new models are merged, which might not always get reflected in transformers immediately.

> The existing `extract_output` function verifies that the tokens returned by vLLM match the tokens returned from the local chat template. As long as this holds true, the loss mask should be correct right?

Yep, that's true. I'm not sure if we currently check for this during online training, but we could add a check there too if needed. I'm alright with implementing this as is and just having it fail loudly if there's a mismatch for some reason. Just want to make sure we're checking it somehow.

DarkLight1337:
> Yeah, my concern is whether this is always the case. If vLLM has custom logic for applying the chat template that doesn't exist in the transformers implementation then there will be a misalignment. I'll admit, I don't have a great sense of how common this is, but my understanding is this logic often undergoes a number of small fixes when new models are merged, which might not always get reflected in transformers immediately.

For `tokenizer_mode="hf"` (which is used for vast majority of models) we always use HF's implementation. So I think this is fine for now. I agree with using Renderer + `/inference/v1/generate` to better cover the edge cases in the future, though there is the downside of having to also start up vLLM for the preprocessing stage.

DarkLight1337:
> I'm not sure if we currently check for this during online training, but we could add a check there too if needed.

The online training code calls `generate_hidden_states` (non-async) which in turn calls `extract_output` as well.

fynnsu:
Sounds good!

And then the only other question I have is regarding what the new dataset looks like?

Currently when running `prepare_data.py` we save an arrow dataset with `token_ids, loss_mask, seq_len` columns. If we're using multimodal do we still store those + another column with the raw unformatted conversation (list of dicts)?

Or something else?

DarkLight1337:
> And then the only other question I have is regarding what the new dataset looks like?
>
> Currently when running prepare_data.py we save an arrow dataset with token_ids, loss_mask, seq_len columns. If we're using multimodal do we still store those + another column with the raw unformatted conversation (list of dicts)?

The dataset contains an additional column `messages` which stores the vLLM-compatible Chat Completions messages. As mentioned in the PR description, we avoid storing any multimodal data in the arrow dataset by having the messages reference file URLs.

DarkLight1337:
Could we move this forward?

shx2005:
> Could we move this forward?

👀 Could anyone help review this? I don’t have the required permission to move it forward. Thanks!

shanjiaz:
> Could we move this forward?

Taking a look today!

src/speculators/train/utils.py
```diff
@@ -73,7 +75,10 @@ def resolve_mask_token_id(
         logger.info(f"Using explicit mask_token_id={mask_token_id}")
         return mask_token_id
 
-    tokenizer = AutoTokenizer.from_pretrained(verifier_name_or_path)
+    tokenizer = AutoTokenizer.from_pretrained(
```
fynnsu: Is it okay that this doesn't use `AutoProcessor.from_pretrained(...).tokenizer` like in `data_generation/preprocessing.py`?

tests/e2e/utils.py
```diff
@@ -65,9 +68,12 @@ def launch_vllm_server(
     model: str,
     port: int,
     hidden_states_path: str,
+    *,
     max_model_len: int = 513,
     gpu_memory_utilization: float = 0.5,
     target_layer_ids: list[int] | None = None,
+    enforce_eager: bool = True,
```
fynnsu: Could you make the regression tests use `enforce_eager=False` for launch_vllm_server? The regression tests run a longer training test where the performance will be more affected by compilation.

tests/e2e/utils.py
```diff
@@ -277,6 +307,10 @@ def run_vllm_engine(
     model_path: str,
     tmp_path: Path,
     prompts: list[list[dict[str, str]]],
+    max_model_len: int = 1024,
+    gpu_memory_utilization: float = 0.8,
+    enforce_eager: bool = True,
```
fynnsu: Could you change the default for `run_vllm_engine`'s `enforce_eager` arg to `False`?

This stage is for actually testing the produced draft checkpoint in vllm, so we probably want both smoke and regression tests to run the model without `enforce_eager` so we see the default behaviour.

tests/e2e/utils.py
```diff
@@ -305,9 +339,10 @@ def run_vllm_engine(
         json.dumps(
             {
                 "model": model_path,
-                "max_model_len": 1024,
-                "gpu_memory_utilization": 0.8,
-                "enforce_eager": True,
+                "max_model_len": max_model_len,
+                "gpu_memory_utilization": gpu_memory_utilization,
+                "enforce_eager": enforce_eager,
+                "allowed_local_media_path": allowed_local_media_path,
```
fynnsu: This is throwing an error when `allowed_local_media_path` is `None`:
```
2026-05-21 18:39:08.286 | INFO     | tests.e2e.utils:run_vllm_engine:353 - run_vllm.py command:
    ['/home/runner/_work/llm-compressor-testing/llm-compressor-testing/vllm-venv/bin/python', '/home/runner/_work/llm-compressor-testing/llm-compressor-testing/speculators/tests/e2e/run_vllm.py', '--sampling-params-args', '{"temperature": 0, "top_p": 0.9, "max_tokens": 50, "ignore_eos": true}', '--llm-args', '{"model": "/tmp/pytest-of-runner/pytest-0/test_convert_run_vllm_engine_e2/random-weights-llama3.1.8b-2layer-eagle3-unconverted", "max_model_len": 1024, "gpu_memory_utilization": 0.8, "enforce_eager": false, "allowed_local_media_path": null}', '--prompts', '[[{"role": "user", "content": "Write a binary search function in python"}], [{"role": "user", "content": "Explain how speculative decoding works"}], [{"role": "user", "content": "Code a transformer block function"}]]', '--results-file', '/tmp/pytest-of-runner/pytest-0/test_convert_run_vllm_engine_e2/results.json']

...

    pydantic_core._pydantic_core.ValidationError: 1 validation error for ModelConfig
    allowed_local_media_path
      Input should be a valid string [type=string_type, input_value=None, input_type=NoneType]
        For further information visit https://errors.pydantic.dev/2.13/v/string_type

```

fynnsu: Not sure if it's also an issue for `launch_vllm_server` because it's passed differently there but would be good to check.

fynnsu (commented):
Hi @DarkLight1337, thanks for opening the pr and sorry that we've been a bit behind on reviewing this. Generally looks good, just a few comments below. Once those are resolved we should be able to merge this.

I'm still working on the fix for the chunked prefill (prompt length mismatch) issue, but I think we can merge this independently. As a side note, I think that may only be a more recent issue on vllm main/nightly because we only started seeing it in our CI recently. You might have better luck with vllm 0.20 or maybe even 0.21 if that supports your VL model.

scripts/data_generation_offline.py
```diff
@@ -397,7 +401,7 @@ async def generate_and_save_hidden_states(args, dataset):
         if args.model and args.model != model_id:
             raise ValueError(
                 f"An explicit model name was passed ({args.model}) which doesn't match"
-                "found model_id {model_id}."
+                f"found model_id {model_id}."
```
shanjiaz: small nit
```suggestion
                f" found model_id {model_id}."
```

src/speculators/data_generation/vllm_client.py
```diff
@@ -114,39 +134,71 @@ async def generate_hidden_states_async(
         client: The async OpenAI client.
         model: The model ID.
         token_ids: The input token IDs.
```
shanjiaz: nit: We should probably update this to client_item.

shanjiaz (commented):
This looks good, again thank you so much for putting up this fix! Tested with vllm 0.21.0 on Qwen/Qwen3-VL-2B-Instruct, preprocessed data and hidden states seem normal. Pending approval given the issues Fynn mentioned are fixed.

src/speculators/data_generation/preprocessing.py
```diff
@@ -250,14 +357,121 @@ def _create_loss_mask_from_offsets(
...
+
+    processor_kwargs: dict = {
+        "return_offsets_mapping": True,
+        "max_length": max_length,
+        "truncation": True,
+        "add_special_tokens": False,
+    }
+
+    if isinstance(processor, ProcessorMixin):
+        if Version(TRANSFORMERS_VERSION) >= Version("5.4.0"):
```
shanjiaz, line 419: lol transformers sorry you had to do this.

src/speculators/train/utils.py
DarkLight1337: Good point, updated

tests/e2e/utils.py
DarkLight1337: I have changed the default to False, setting it to True explicitly only for the smoke tests

tests/e2e/utils.py
DarkLight1337: Reverted

tests/e2e/utils.py
DarkLight1337: Fixed both locations to only set the arg if `allowed_local_media_path is not None`

scripts/data_generation_offline.py
DarkLight1337: Fixed

src/speculators/data_generation/vllm_client.py
DarkLight1337: Updated the docstring

DarkLight1337:
I have addressed your comments, please take another look!

shanjiaz (approved):
Thanks for updating! Looks good to me.

DarkLight1337:
@fynnsu could you take another look as well?

fynnsu (approved):
Thank you!

PR #501, opened 2026-05-04
Graceful shutdown on interrupt

fynnsu:

## Purpose

With this pr, when a training job is interrupted (e.g. using `Ctrl+C` or `kill <PID>` (sends `SIGTERM`) or `kill -INT <PID>` (sends `SIGINT`)), it will attempt to save a checkpoint of the current state to `CHECKPOINT_SAVE_PATH/interrupted`.

This can be useful to ensure failing jobs or long running jobs can be stopped without losing all progress. Note: this approach is not guaranteed to work in all scenarios, so the recommended approach is to wait for the checkpoints to be saved as normal. 

## Description

Add signal handlers for the `SIGTERM` and `SIGINT` events which we wrap the main train/val/checkpointing loop in.

When one of these signals is received it raises a `TrainingInterruptedError` (note: all processes receive the signal) which gets caught by the wrapper and then executes a checkpoint save.

## Tests

I have manually tested this works on a 1 gpu and 2 gpu setup.

I have filled in:

### DIFF
--- a/src/speculators/train/checkpointer.py
+++ b/src/speculators/train/checkpointer.py
@@ -1,4 +1,5 @@
 import json
+import logging
 import shutil
 from abc import abstractmethod
 from pathlib import Path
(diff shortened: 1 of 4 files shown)


shanjiaz (approved):
This is fine, but is there a way to add testing for this?

fynnsu:
@shanjiaz, added a test and merging.

Issue #493, opened 2026-04-28
Add configurable checkpointing frequency during training


dsikka:
Add a configurable checkpointing heuristic that allows saving checkpoints more frequently than once per epoch. For long-running training jobs or datasets with large epochs, this means significant work can be lost if training is interrupted.

# Suggested Heuristic

Introduce a new parameter `checkpoint_interval` that supports additional modes e.g. `N` steps - checkpoint every N training steps/iterations

AayushSaini101:
Hey @dsikka can i take this issue? I have done some contributions in the semantic router, happy to explore this required feature thanks cc: @rahul-tuli

dsikka:
@shanjiaz have you started this? If not, I can assign it to [AayushSaini101](https://github.com/AayushSaini101)

shanjiaz:
> [@shanjiaz](https://github.com/shanjiaz) have you started this? If not, I can assign it to [AayushSaini101](https://github.com/AayushSaini101)

Sure, please do! Thank you.

dsikka:
Thanks - assigned to you @AayushSaini101

shanjiaz:
@AayushSaini101 Hey! How's it going? Let use know if you need anything.

AayushSaini101:
> [@AayushSaini101](https://github.com/AayushSaini101) Hey! How's it going? Let use know if you need anything.

hey sorry for late response,i was involved in some priority task, thanks for raising the request

shanjiaz:
> > [@AayushSaini101](https://github.com/AayushSaini101) Hey! How's it going? Let use know if you need anything.
> 
> hey sorry for late response,i was involved in some priority task, thanks for raising the request

No worries! Reassigned because this request is time-sensitive. We're actively training. We'll have more good first issues coming soon!

PR #509, opened 2026-05-06
Save multiple checkpoints per epoch

shanjiaz:

## Purpose
Save multiple checkpoints per epoch

## Description
updated `checkpoint_freq` to float, setting it to 0.5 would mean save twice per epoch to the same directory by overwriting. We skip the last saving right before validation since it's handled separately. (Or should it?) When checkpoint_freq > 1 it should be an int and we save per multiple epochs. 

## Related Issue

https://github.com/vllm-project/speculators/issues/493

## Tests
Tested locally with an example. 

I have filled in:

### DIFF
--- a/src/speculators/train/trainer.py
+++ b/src/speculators/train/trainer.py
@@ -28,6 +28,7 @@ root_logger = logging.getLogger("speculators")
 metric_logger = logging.getLogger("speculators.metrics")
 
 warnings.filterwarnings("ignore", category=TqdmExperimentalWarning)
+MIN_STEP_PCT = 0.25
 
 
 class TrainerConfig(NamedTuple):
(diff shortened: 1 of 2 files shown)


src/speculators/train/trainer.py
```diff
@@ -228,6 +234,14 @@ def train_epoch(self, epoch: int):
                 )
             self.global_step += 1
 
+            if (
+                step_interval is not None
+                and not self.config.save_best
+                and local_step % step_interval == 0
+                and num_steps - local_step >= step_interval
```
fynnsu: Let's change this to something like:
```python
and num_steps - local_step >= step_interval * MIN_STEP_PCT
```

or something where `MIN_STEP_PCT` is `0.10` or `0.25` or something. 

This way we can avoid saving twice back to back, but still have multiple saves if the user specifices a save frequence of `0.4` or something.

fynnsu (commented):
Looks good. One small comment

src/speculators/train/trainer.py
shanjiaz: sure!

fynnsu (approved):

PR #510, opened 2026-05-09
Sweep concurrencies regeneration

MeganEFlynn:
## Purpose
We currently set the concurrency for the number of requests we're sending during response regeneration to 64, but it appears as though tuning that parameter can result in much better throughput

## Description

This PR adds a shell script to scripts that sweeps the concurrency value and reports the results.  It also adds shuffling to the dataset, since some datasets are ordered by topic.  

## Related Issue

NA

## Tests

Run with: 
./scripts/response_regeneration/sweep_concurrency.sh --model google/gemma-4-31B-it --tp-size 2 --dp-size 4 --dataset tulu3

Results: 

=========================================
Concurrency     Samples         Samples/sec
-----------     -------         -----------
256             525             8.75
512             677             11.28
64              186             3.10
1024            692             11.53           <-- best
128             336             5.60
2048            652             10.86          
4096            620             10.33
8192            535             8.91

I have filled in:

### DIFF
--- a/scripts/response_regeneration/script.py
+++ b/scripts/response_regeneration/script.py
@@ -281,7 +281,10 @@ async def main():
         dataset = dataset.shuffle(
             seed=args.shuffle_seed, buffer_size=args.shuffle_buffer_size
         )
-        print(f"Shuffling with seed={args.shuffle_seed}, buffer={args.shuffle_buffer_size}")
+        print(
+            f"Shuffling with seed={args.shuffle_seed},"
+            f" buffer={args.shuffle_buffer_size}"
+        )
 
     queue: asyncio.Queue = asyncio.Queue(maxsize=args.concurrency * 4)
     semaphore = asyncio.Semaphore(args.concurrency)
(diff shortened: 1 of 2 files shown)


scripts/response_regeneration/script.py
fynnsu, line 72: Can we update the default concurrency value as well? Or do we think this is an okay default? Most people will probably not bother to run the sweep.

scripts/response_regeneration/script.py
```diff
@@ -93,6 +93,23 @@ def parse_args():
...
+        type=int,
+        default=42,
+        help="Random seed for shuffling (default: 42)",
+    )
+    parser.add_argument(
+        "--shuffle-buffer-size",
+        type=int,
+        default=10000,
+        help="Buffer size for streaming shuffle (default: 10000)",
+    )
```
fynnsu, line 112: I think we can just chose a reasonable value and hard code it in. We don't need to expose this to the user.

fynnsu (commented):
Thanks for adding this. 

We should also add a note about the new script in the response regen tutorial: https://docs.vllm.ai/projects/speculators/en/latest/user_guide/tutorials/response_regeneration/

And lastly we should update the cli reference page to include the new cli arguments:
https://docs.vllm.ai/projects/speculators/en/latest/cli/response_regeneration/ (and maybe mention the sweep_concurrency.sh script).

PR #516, opened 2026-05-12
Add HF_TOKEN env variable to dev flow

dsikka:
Add HF_TOKEN to dev flow

### DIFF
--- a/.github/workflows/development.yml
+++ b/.github/workflows/development.yml
@@ -6,6 +6,7 @@ on:
 
 env:
   UV_TORCH_BACKEND: "auto"
+  HF_TOKEN: ${{ secrets.HF_RED_HAT_READ_ONLY }}
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}


shanjiaz (approved):
Thank you!

fynnsu (changes requested):
This isn't going to work. Secrets can't be accessed by workflows triggered by a `pull_request` trigger for security reasons.

dsikka:
> This isn't going to work. Secrets can't be accessed by workflows triggered by a `pull_request` trigger for security reasons.

I don't think that's correct. This is only true for forks / external contributors. Not from within the same repo. For any community member, until we get buildkite enabled, we'll have to manually approve the PR.

fynnsu:
> I don't think that's correct. This is only true for forks / external contributors. Not from within the same repo. For any community member, until we get buildkite enabled, we'll have to manually approve the PR.

Okay, you're right about the internal contributions having access, so we can merge this. I don't think the community prs will access the token, even if they've been approved by maintainers, but that should be okay since that's the current behaviour.

fynnsu (approved):

PR #517, opened 2026-05-12
Fix speculator config for models with explicit head_dim

MeganEFlynn:
## Purpose

Models like Laguna XS and Qwen3.6-27B have hidden_size (5120) not divisible by num_attention_heads (24) because they use an explicit head_dim (256). LlamaConfig's validate_architecture rejects this, so recompute num_attention_heads as hidden_size // head_dim for the speculator.

## Description

This PR changes the way we calculate the attention heads so that the initialization doesn't fail when we use we use a hidden state dim that isn't divisible by the number of attention heads in the model, due to a limitation in the llama config. 

## Related Issue

NA

## Tests

Using this PR makes Qwen 3.6 27B run whereas it previously failed

I have filled in:

### DIFF
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -139,18 +139,34 @@ def create_transformer_layer_config(
             "nor 'hidden_activation'"
         )
 
+    head_dim = getattr(verifier_config, "head_dim", None)
+    num_attention_heads = verifier_config.num_attention_heads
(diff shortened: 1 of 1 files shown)


fynnsu (approved):

fynnsu:
Closing in favor of #558

PR #518, opened 2026-05-13
Add tqdm progress bar for response regeneration

eldarkurtic:
Right now it is not very clear (without manually checking logs) at which stage of the regeneration pipeline the script is as there are no progress bars. This PR adds a simple tqdm progress bar which gives us ETA and the current state. It looks like this:
```bash
=========================================
Response Regeneration Pipeline
=========================================

Step 1: Starting vLLM server on port 8005
  Model: /home/eldarkurtic/hf_models/Qwen/Qwen3-8B
  GPUs: 4,5,6,7
  Data parallel size: 4
  Tensor parallel size: 1
  Command: vllm serve /home/eldarkurtic/hf_models/Qwen/Qwen3-8B --host 127.0.0.1 --port 8005 --data-parallel-size 4 --tensor-parallel-size 1

Step 2: Waiting for server to be ready...
  (Large models may take several minutes to load)
  Still waiting... (5 retries)
  Still waiting... (10 retries)
  Still waiting... (15 retries)
  Still waiting... (20 retries)
  Server ready (after 23 retries)

Step 3: Running response regeneration...
Arguments: --model /home/eldarkurtic/hf_models/Qwen/Qwen3-8B --dataset magpie --limit 5000 --endpoint http://127.0.0.1:8005/v1/chat/completions

Using endpoint: http://127.0.0.1:8005/v1/chat/completions
Using model: /home/eldarkurtic/hf_models/Qwen/Qwen3-8B
Using dataset: Magpie-Align/Magpie-Llama-3.1-Pro-300K-Filtered
Split: train
Prompt field: instruction
Output file: magpie_Qwen3-8B.jsonl

Generating responses: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 5000/5000 [49:29<00:00,  1.68sample/s, errors=0, ok=5000]

```

### DIFF
--- a/scripts/response_regeneration/script.py
+++ b/scripts/response_regeneration/script.py
@@ -233,7 +233,7 @@ async def worker(
             queue.task_done()
 
 
-async def main():
+async def main():  # noqa: PLR0915
     """Main async function to process dataset through vLLM endpoints."""
     args = parse_args()
 
--- a/scripts/response_regeneration/script.py
+++ b/scripts/response_regeneration/script.py
@@ -10,6 +10,7 @@ from typing import Any
 
 import aiohttp
 from datasets import load_dataset
+from tqdm import tqdm
 
 DATASET_CONFIGS = {
     "magpie": {
@@ -154,6 +155,8 @@ async def worker(
     args,
     out_fh,
     endpoint: str,
+    progress,
+    stats: dict[str, int],
 ):
     """Worker that pulls items from queue and sends them to the vLLM endpoint."""
     while True:
@@ -206,6 +209,7 @@ async def worker(
             }
             out_fh.write(json.dumps(output, ensure_ascii=False) + "\n")
             out_fh.flush()
+            stats["ok"] += 1
         except Exception as e:  # noqa: BLE001
             error_output = {
                 "id": item.get("uuid") or f"sample_{idx}",
@@ -218,7 +222,14 @@ async def worker(
             }
             out_fh.write(json.dumps(error_output, ensure_ascii=False) + "\n")
             out_fh.flush()
+            stats["errors"] += 1
         finally:
+            progress.update(1)
+            progress.set_postfix(
+                ok=stats["ok"],
+                errors=stats["errors"],
+                refresh=False,
+            )
             queue.task_done()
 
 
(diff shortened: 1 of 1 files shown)


scripts/response_regeneration/script.py
```diff
@@ -218,11 +222,18 @@ async def worker(
             }
             out_fh.write(json.dumps(error_output, ensure_ascii=False) + "\n")
             out_fh.flush()
+            stats["errors"] += 1
         finally:
+            progress.update(1)
+            progress.set_postfix(
+                ok=stats["ok"],
+                errors=stats["errors"],
+                refresh=False,
```
fynnsu, line 230: `refresh=False`, is this being refreshed elsewhere? (maybe implicitly by `progress.update`?)

Otherwise it seems like this won't actually update the display.

scripts/response_regeneration/script.py
```diff
@@ -218,11 +222,18 @@ async def worker(
...
+            )
             queue.task_done()
 
 
-async def main():
+async def main():  # noqa: PLR0915
```
fynnsu: Shouldn't be needed after #519

scripts/response_regeneration/script.py
```diff
@@ -276,43 +287,72 @@ async def main():
...
+            progress_total = args.limit if args.limit is not None else None
+            progress = tqdm(
+                total=progress_total,
+                desc="Generating responses",
+                unit="sample",
+                dynamic_ncols=True,
+            )
+            workers = []
+
+            try:
```
fynnsu: Rather than using `try-finally` and manually closing the tqdm object, we can instead use:
```python
with tqdm(total=100) as pbar:
    for i in range(10):
        sleep(0.1)
        pbar.update(10)
```

Reference: https://github.com/tqdm/tqdm#manual

We can also combine this with the file open `with` statement above to avoid the extra level of indentation used here.

scripts/response_regeneration/script.py
```diff
@@ -276,43 +287,72 @@ async def main():
...
-                    }
-                )
-                processed_count += 1
-
-            # Signal workers to stop
-            for _ in range(len(workers)):
-                await queue.put(None)
-            await asyncio.gather(*workers)
+            stats = {"ok": 0, "errors": 0}
+            progress_total = args.limit if args.limit is not None else None
```
fynnsu: Can we do something like this?

```python
progress_total = args.limit if args.limit is not None else len(dataset)
```

scripts/response_regeneration/script.py
```diff
@@ -276,43 +287,72 @@ async def main():
...
+                            "uuid": uuid,
+                            "prompt": prompt,
+                        }
+                    )
+                    processed_count += 1
+
+                if progress.total is None or processed_count < progress.total:
+                    progress.total = processed_count
+                    progress.refresh()
+
```
fynnsu: Shouldn't need this if we set total correctly when initializing.

fynnsu (changes requested):

scripts/response_regeneration/script.py
eldarkurtic: yes, `progress.update` is doing it.  explicit`refresh=False` just disables the second update during the same iteration

scripts/response_regeneration/script.py
eldarkurtic: unfortunately we cant use `len(dataset)` when dataset has `streaming=True`

eldarkurtic:
addressed initial reviews, @fynnsu please re-review

fynnsu (approved):
Looks good! Thanks

PR #525, opened 2026-05-15
Add vendored DSv4 chat template for preprocessing pipeline

rahul-tuli:
## Purpose

DeepSeek-V4-Flash's tokenizer does not provide a Jinja `apply_chat_template`, which causes `prepare_data.py` to fail with a `ValueError`. This PR adds a vendored encoding function and model-type dispatch so that DSv4 conversations can be preprocessed end-to-end.

## Description

**New file: `src/speculators/data_generation/chat_templates.py`**
- Vendors DSv4's `encode_messages()` and encoding dependencies from `deepseek-ai/DeepSeek-V4-Flash` (`encoding/encoding_dsv4.py`, commit `fd53f944`). Parsing/decoding functions removed.
- Adds `dsv4_format_conversation()` wrapper: formats normalized conversations with `thinking_mode="chat"`.
- Adds `DSV4_ASSISTANT_PATTERN` regex: captures assistant content between `</think>` and `<｜end▁of▁sentence｜>` for loss-mask creation.

**Modified: `src/speculators/data_generation/preprocessing.py`**
- Replaces the hard `ValueError` for missing `chat_template` with explicit model-type dispatch: loads `AutoConfig`, checks `config.model_type == "deepseek_v4"`, dispatches to vendored encoder. Other models without a chat template still raise `ValueError`.
- Adds `format_fn` parameter threaded through `_preprocess_batch` → `build_eagle3_dataset` → `load_and_preprocess_dataset`. When provided, uses `format_fn(conv)` instead of `apply_chat_template(tokenize=False)` in the regex-based loss-masking path.
- Dispatch is explicit and deterministic (model-type check, no `try/except`).

All changes are backward-compatible — models with `apply_chat_template` are completely unaffected.

## Related Issue

Part of the DFlash DSv4 code changes (Diff 2: Chat Template) — see `local/PRDs/dflash-dsv4-code-changes/prd.md`.

## Tests

All 25 existing preprocessing tests pass (1 skipped, as before):
```
================= 25 passed, 1 skipped, 26 warnings in 40.55s ==================
```

Verified end-to-end with the DSv4 tokenizer:
- `dsv4_format_conversation()` produces correct DSv4 format with special tokens
- `DSV4_ASSISTANT_PATTERN` correctly captures assistant content (single-turn and multi-turn)
- `_preprocess_batch` with `format_fn` produces correct `input_ids` and `loss_mask` (assistant content trainable, user/system masked)
- `load_and_preprocess_dataset` auto-detects `model_type == "deepseek_v4"` and dispatches correctly

I have filled in:

### DIFF
--- a/src/speculators/data_generation/chat_templates.py
+++ b/src/speculators/data_generation/chat_templates.py
@@ -1,0 +1,557 @@
+"""
+Vendored DeepSeek-V4 chat template encoding.
+
+Source: deepseek-ai/DeepSeek-V4-Flash, encoding/encoding_dsv4.py
+Vendored commit: fd53f944496234770ba80e15004f9b6d269a71f5
+
+Trimmed to encoding-only; parsing/decoding functions removed.
+"""
+
+import copy
+import json
+import re
+from typing import Any
+
+__all__ = [
+    "DSV4_ASSISTANT_PATTERN",
+    "dsv4_format_conversation",
+]
+
+# ============================================================
+# Special Tokens
+# ============================================================
+
+bos_token: str = "<｜begin▁of▁sentence｜>"  # noqa: S105
+eos_token: str = "<｜end▁of▁sentence｜>"  # noqa: S105
+thinking_start_token: str = "<think>"  # noqa: S105
+thinking_end_token: str = "</think>"  # noqa: S105
+dsml_token: str = "｜DSML｜"  # noqa: S105
+
+USER_SP_TOKEN = "<｜User｜>"  # noqa: S105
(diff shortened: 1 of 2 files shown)


fynnsu:
Hmm, that's annoying that they didn't include a jinja template in the release. 

I see there's an [open pr by HF staff to add one to the upstream model](https://huggingface.co/deepseek-ai/DeepSeek-V4-Flash/discussions/16).

 Maybe a more general fix would be to allow users to manually pass in a chat template file. Then to run dsv4 you could just copy the linked chat template to a local file and manually pass it in. Also not perfect, but I think better than copying the dsv4 templating script into our codebase.

PR #527, opened 2026-05-16
fix(train/data): default create_empty_sample to bfloat16 to match training dtype

coolthor:
## Problem

When `extract_hidden_states` + `ExampleHiddenStatesConnector` is used to train an EAGLE-3 drafter against a bf16 verifier, intermittent vLLM extraction time-outs cause the training loop to substitute an empty sample built by `create_empty_sample()`.

In the original implementation:

```python
return {
    "hidden_states": torch.empty(0, 3 * hidden_size),
    "input_ids": torch.empty(0),
    "verifier_last_hidden_states": torch.empty(0, hidden_size),
    ...
}
```

`torch.empty` without an explicit `dtype` defaults to `float32`. Downstream EAGLE-3 layers (`fc`, `verifier_lm_head`) hold `bfloat16` weights when training a bf16 verifier, and the first time the empty sample reaches one of those layers PyTorch raises:

```
RuntimeError: ... expected Float, got BFloat16
```

The whole training run dies at a random late-epoch step.

## Repro

```
1. Train an EAGLE-3 drafter against any bf16 verifier (e.g. `google/gemma-4-26B-A4B-it`).
2. Use `extract_hidden_states` + `ExampleHiddenStatesConnector` as the data path.
3. Let it run long enough that some vLLM extractions time out (high concurrency or large samples accelerate this).
4. Eventually a step substitutes the empty sample and the run crashes at `self.fc(hidden_states)` with the dtype mismatch above.
```

## Fix

- Add a `dtype: torch.dtype = torch.bfloat16` keyword argument to `create_empty_sample()` (covers the common case of training against a bf16 verifier).
- Apply it to the `hidden_states` and `verifier_last_hidden_states` placeholders so they match downstream weight dtype.
- Pin `input_ids` to `torch.long` (it previously fell back to float, which would surface separately for any consumer that does `embedding(input_ids)`).

Callers that train against a non-bf16 verifier (fp32, fp16, etc.) can override the default explicitly.

## Tested with

- `coolthor/Huihui-gemma-4-26B-A4B-it-abliterated-FP8-Dynamic` verifier + RedHatAI EAGLE-3 drafter, 50k Magpie samples, 1 epoch on a single DGX Spark GB10.
- Before the patch: ~9h crash at step ~9485 with the BFloat16 / Float dtype mismatch.
- After the patch: 11h run completed cleanly, pos 3 acceptance climbs to 72.7%.

Writeup with full numbers: <https://ai-muninn.com/en/blog/dgx-spark-eagle3-finetune-abliterated-round1>

### DIFF
--- a/src/speculators/train/data.py
+++ b/src/speculators/train/data.py
@@ -126,7 +126,7 @@ class BaseDataset(Dataset):
         self,
         max_len: int,
         transform: TransformTensors | None = None,
-        hidden_states_dtype=torch.float,
+        hidden_states_dtype=torch.bfloat16,
     ):
         self.max_len = max_len
         self.transform = transform
@@ -193,7 +193,7 @@ class ArrowDataset(BaseDataset):
         on_generate: Literal["cache", "delete"] = "delete",
         split_ratio: float = 1.0,
         transform: TransformTensors | None = None,
-        hidden_states_dtype=torch.float,
+        hidden_states_dtype=torch.bfloat16,
         model: str | None = None,
         request_timeout: float | None = DEFAULT_REQUEST_TIMEOUT,
         max_retries: int = DEFAULT_MAX_RETRIES,
@@ -364,7 +364,7 @@ class SampleFileDataset(BaseDataset):
         datapath: str | None = None,
         file_list: list[str] | None = None,
         transform: TransformTensors | None = None,
-        hidden_states_dtype=None,
+        hidden_states_dtype: torch.dtype = torch.bfloat16,
     ):
         """Initialize the SampleFileDataset.
         Args:
--- a/src/speculators/train/data.py
+++ b/src/speculators/train/data.py
@@ -64,7 +64,7 @@ def split_files(datapath: str, ratio: float = 0.9, seed: int = 0):
 StandardizeFnSig = Callable[[dict[str, Any]], dict[str, Any]]
 
 
-def create_empty_sample(hidden_size: int):
+def create_empty_sample(hidden_size: int, dtype: torch.dtype = torch.bfloat16):
     # data structure: {
     #     "hidden_states": [seq_len, 3 * hidden_size],
     #     "input_ids": [seq_len],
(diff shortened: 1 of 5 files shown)


coolthor:
Thanks @coderabbitai — good catch on `create_collate_fn` being the only call site and not having a `dtype` kwarg of its own. Applied Option 1 in commit e8b6576: added `dtype: torch.dtype = torch.bfloat16` to `create_collate_fn` and threaded it through to the `create_empty_sample` call. The bf16 default keeps existing training paths unchanged; non-bf16 callers can now override cleanly.

src/speculators/train/data.py
```diff
@@ -64,7 +64,7 @@ def split_files(datapath: str, ratio: float = 0.9, seed: int = 0):
 StandardizeFnSig = Callable[[dict[str, Any]], dict[str, Any]]
 
 
-def create_empty_sample(hidden_size: int):
+def create_empty_sample(hidden_size: int, dtype: torch.dtype = torch.bfloat16):
```
shanjiaz, line 69: Instead of setting the default to bfloat16, could we infer dtype from the model? It's not super intuitive to override the default.

fynnsu: We can actually set this explicitly when creating the collate function. It just isn't being piped in yet. 

This is where `create_collate_fn` is called: https://github.com/vllm-project/speculators/blob/4f80f5dcd0b99030d5895975fea1cfd73aca6663/scripts/train.py#L99

We explicitly set the hidden state dtype in `scripts/train.py` here: 
https://github.com/vllm-project/speculators/blob/4f80f5dcd0b99030d5895975fea1cfd73aca6663/scripts/train.py#L319

This is then set as an attribute on the dataset object. So if you either get `hidden_states_dtype` from the dataset object passed into `setup_dataloader` or pass it into the fn directly, you can then pipe it into the `create_collate_fn` call.

fynnsu (changes requested):
One small comment. Otherwise looks good!

shanjiaz (approved):
Looks good!

PR #528, opened 2026-05-18
Fix AttributeError in ArrowDataset on default split_ratio

aman-source:
## Purpose

Fix a crash in `ArrowDataset` when constructed with the default `split_ratio=1.0`. `_map_to_file_idx` (and therefore `__getitem__`) reads `self.start_file_idx` unconditionally, but `__init__` only sets that attribute inside the branches handling a non-trivial split. With `split_ratio == 1.0`, the attribute is never assigned and the first sample access raises `AttributeError`.

The bug is latent in current in-repo usage because `scripts/train.py` is the only caller and always passes `split_ratio=0.9` or `-0.1`. It surfaces for any user constructing `ArrowDataset` directly with default arguments.

## Description

Initialize `self.start_file_idx = 0` once before the `if/elif/else` block in `ArrowDataset.__init__`. The existing assignments inside the non-trivial-split branches harmlessly override it, so behavior is unchanged on the current code paths.

Single-line fix in `src/speculators/train/data.py`.

## Related Issue

None — bug found while reading the code.

## Tests

Added `test_arrow_dataset_default_split_ratio_does_not_crash` in `tests/unit/train/test_data.py`. The test constructs a real `ArrowDataset` with the default `split_ratio=1.0` against a small on-disk fixture and asserts `_map_to_file_idx` returns sensible values instead of raising.

Verified locally:

​```
pytest tests/unit/train/test_data.py -v
# new test passes; existing tests still pass

ruff check src/speculators/train/data.py tests/unit/train/test_data.py
ruff format --check src/speculators/train/data.py tests/unit/train/test_data.py
# all clean
​```

I have filled in:

### DIFF
--- a/src/speculators/train/data.py
+++ b/src/speculators/train/data.py
@@ -191,6 +191,7 @@ class ArrowDataset(BaseDataset):
             hidden_states_dtype: The dtype of the hidden states.
         """
         self.data = load_from_disk(datapath)
+        self.start_file_idx = 0
         if split_ratio == 1.0:
             pass
         elif 1.0 > split_ratio > 0:
--- a/tests/unit/train/test_data.py
+++ b/tests/unit/train/test_data.py
@@ -4,9 +4,11 @@ import json
 from pathlib import Path
 
 import torch
+from datasets import Dataset
 
 from speculators.models.eagle3.data import shift_batch
 from speculators.train.data import (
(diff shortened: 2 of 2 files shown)


fynnsu (approved):

fynnsu:
Thanks for the fix!

PR #532, opened 2026-05-19
Retrieve thinking msg from the metadata reasoning_content field

Chibukach:
## Purpose

Recent VLLM output separate thinking content in the metadata instead of <think> </think> tags
## Description

This PR adds support for retrieving thinking content from the "reasoning_content" field inside the metadata during a model response generation 

## Tests

Tested locally on Qwen/Qwen3-8B and the output prints <think> thinking_msg  </think> 

```
2026-05-19 11:29:50,075 - speculators.data_generation.preprocessing - INFO -   HIGHLIGHTED TEXT (BLUE = trainable, GREY = masked)
2026-05-19 11:29:50,089 - speculators.data_generation.preprocessing - INFO -   <|im_start|>user
Here is the problem:
Suppose that Alice and Bob are participating in an online auction.  They can either bid or pass.  Assuming that each person acts independently and that the probability of bidding is 1/2 for each participant,  then the probability that both Alice and Bob bid is $\frac{1}{4}$.  Can you explain this answer.

The answer is indeed $\frac{1}{4}$.  But, I don't understand what is meant by independent events.  Does this mean that their actions (bidding or passing) are independent of each other.
Yes, exactly.  In this case, "independent events" means that Alice's decision to bid or pass does not affect Bob's decision to bid or pass, and vice versa.

Thank you.  So, the probability that both Alice and Bob bid is calculated by multiplying their individual probabilities of bidding, which is 1/2, so 1/2 * 1/2 = 1/4, which I can understand.
That's correct.  The probability of independent events occurring together is the product of the individual probabilities.  This is a basic rule of probability, and it applies to any situation where the events are independent.

Okay, I think I have a good understanding of this problem now.  Is there anything else I should know about independent events?<|im_end|>
<|im_start|>assistant
<think>
Here's a thinking process that leads to the suggested response:

1.  **Analyze the User's Request:**
    *   **Context:** The user is asking about a probability problem involving Alice and Bob in an auction.
    *   **Current Understanding:** They understand *why* the answer is 1/4 (multiplying probabilities) and *what* independence means in this context (one person's choice doesn't affect the other's).
    *   **Current Question:** "Is there anything else I should know about independent events?"
    *   **Goal:** Provide a helpful, educational, and concise expansion on the concept of independent events without overwhelming the user. I need to reinforce their correct understanding while adding valuable nuance or common pitfalls.

2.  **Determine Key Concepts to Cover:**
    *   **Definition Reinforcement:** Briefly reiterate the formal definition ($P(A \cap B) = P(A) \cdot P(B)$).
    *   **Distinction from Mutually Exclusive:** This is the most common confusion point. Students often mix up "independent" (doesn't affect probability) with "mutually exclusive" (can't happen together).
    *   **Conditional Probability:** Briefly mention how independence relates to conditional probability ($P(A|B) = P(A)$).
    *   **Real-world Nuance:** Mention that true independence is often an assumption in models (like the auction example) but might not hold in reality (e.g., if they talk to each other).
    *   **More than two events:** Briefly note that this extends to 3+ events.

3.  **Drafting the Response (Iterative Process):**

    *   *Draft 1 (Internal Monologue):* Okay, they get the basics. I should tell them about mutually exclusive events because people always confuse that. If events are mutually exclusive, they can't both happen, so probability is 0. But independent events *can* both happen. Also, maybe mention conditional probability. If A and B are independent, knowing B happened doesn't change the chance of A. Oh, and maybe warn them that in real life, independence is an assumption.

    *   *Draft 2 (Closer to final):* You have a great grasp. Here are a few extra things. First, don't confuse independent with mutually exclusive. Mutually exclusive means they can't both happen. Independent means one doesn't change the other's odds. Second, think about conditional probability. If they are independent, P(Alice bids | Bob bids) is still 1/2. Third, remember this is a model. In real auctions, maybe they know each other. Finally, this works for more than two people too.

    *   *Refining for Tone and Clarity:* The tone should be encouraging and instructional. I should use formatting (bullet points) to make it readable. I need to make sure the distinction between "Independent" and "Mutually Exclusive" is very clear, as that's the biggest learning opportunity here.

4.  **Finalizing the Content (incorporating into the final output):**
    *   **Acknowledge and Validate:** Start by confirming their understanding is solid.
    *   **Key Point 1: Independent vs. Mutually Exclusive:** This is crucial. Explain the difference clearly.
        *   Independent: Can both happen, doesn't affect odds.
        *   Mutually Exclusive: Cannot both happen.
    *   **Key Point 2: Conditional Probability:** Explain $P(A|B) = P(A)$. This is the mathematical definition of independence.
    *   **Key Point 3: Assumption vs. Reality:** Contextualize the auction example. In game theory, decisions might *not* be independent (strategic interaction), but in this probability problem, we assume they are.
    *   **Key Point 4: Extension:** Mention it works for 3+ events.

5.  **Review against Constraints:**
    *   The user is asking a follow-up question in a dialogue format.
    *   Keep it conversational but informative.
    *   Ensure mathematical accuracy.

6.  **Final Polish:** (Self-Correction during drafting) *Wait, I should be careful not to introduce Game Theory too deeply, or it might confuse the probability concept. Just mention that in real auctions, independence might not hold, but for this math problem, it does.*

    *Let's structure the answer:*
    1. Validation.
    2. The "Mutually Exclusive" trap.
    3. The Conditional Probability perspective.
    4. Real-world context (Assumptions).
    5. Closing encouragement.

7.  **Final Output Generation** (similar to the provided good response).
</think>

You have a solid grasp of the basics! Since you are interested in deepening your understanding, here are a few key concepts and common pitfalls regarding independent events that are worth knowing:

**1. Independent vs. Mutually Exclusive**
This is the most common point of confusion in probability.
*   **Independent Events:** The occurrence of one does *not* change the probability of the other. They **can** happen at the same time (like Alice and Bob both bidding).
*   **Mutually Exclusive Events:** The two events **cannot** happen at the same time. (For example, if a coin flip results in "Heads," it cannot also be "Tails").
*   **Why it matters:** If events are mutually exclusive, the probability of both happening is **0**. If they are independent, you **multiply** their probabilities.

**2. The Conditional Probability Perspective**
Another way to define independence is using conditional probability (the probability of A happening *given* that B has already happened).
*   If Alice and Bob are independent, knowing that Bob bid does not change Alice's odds.
*   Mathematically: $P(\text{Alice bids} \mid \text{Bob bids}) = P(\text{Alice bids})$.
*   If knowing Bob bid made Alice *more* likely to bid (perhaps they are competing rivals), they would **not** be independent.

**3. Independence is Often an Assumption**
In textbook problems like this auction scenario, independence is usually given as a rule to make the math solvable. However, in real-world scenarios (especially in economics or game theory), actions are rarely truly independent.
*   **Real World:** If Alice sees Bob bid, she might be more likely to bid to win the item, or less likely to bid to save money.
*   **Math Problem:** We assume "independence" to isolate the probability calculation from psychological or strategic factors.

**4. It Extends to More Than Two Events**
The rule you used ($P(A) \times P(B)$) works for any number of independent events.
*   If Charlie also joins the auction with a 1/2 probability of bidding, the chance that **all three** bid is $\frac{1}{2} \times \frac{1}{2} \times \frac{1}{2} = \frac{1}{8}$.

You are asking the right questions to build a strong foundation in probability. Keep distinguishing between *what* happens (the events) and *how* they influence each other (the dependence)!<|im_end|>
```

I have filled in:


src/speculators/data_generation/preprocessing.py
```diff
@@ -267,17 +271,42 @@ def _preprocess_batch(
...
+    if len(metadatas) != len(conversations):
+        log.warning(
+            f"Metadata/conversation length mismatch: "
+            f"{len(metadatas)} vs {len(conversations)}; padding metadata"
+        )
+        if len(metadatas) < len(conversations):
+            metadatas = metadatas + [{} for _ in range(len(conversations) - len(metadatas))]
+        else:
+            metadatas = metadatas[: len(conversations)]
+
```
fynnsu: This feels pretty unsafe. We're just assuming that if we have less metadata that the metadata lines up with first part of the conversation?

The conversation contains system, user, and assistant messages. Presumably we only have reasoning content for assistant responses.

src/speculators/data_generation/preprocessing.py
```diff
@@ -267,17 +271,42 @@ def _preprocess_batch(
...
+    for idx, (conv, metadata) in enumerate(zip(conversations, metadatas, strict=True)):
+        if not conv or not isinstance(conv, list):
+            continue
+        if not isinstance(metadata, dict):
+            continue
```
fynnsu: If we're constructing things this way we don't need to check for it. Also if the sample structure doesn't meet our expectations it would be better to fail with an error than just silently skip it.

fynnsu (changes requested):
Hi @Chibukach, my main issue with this is I don't fully understand the structure of the reasoning content you're trying to support. 

I think I saw an example from Alex where we had a single reasoning_content value for a full conversation, is this the same as that?

The code seems to be assuming we have:
```
{
  "conversations": list of dicts,
  "metadata": list of dicts of same length
}
```
Is that correct?

shanjiaz:
@Chibukach Thanks so much for putting up this change! Just trying to understand this issue better. Could you maybe provide an example json file for what we actually get when we run response regeneration?  I tested on vllm 0.21.0 and this is what I got:
``` json
{
    "id": "cf428475-a0ef-5fd8-a35e-25abbf57e895",
    "conversations": [
        {
            "from": "human",
            "value": "Has the SoundCloud Generation been broke down into further age-defined groups any?"
        },
        {
            "from": "gpt",
            "value": "<think>\nOkay, the user is asking if the SoundCloud Generation has been broken down into more specific age groups. \n</think>\n\nThe term \"SoundCloud Generation\" ... strict demographic segmentation."
        }
    ],
    "metadata": {
        "idx": 2,
        "finish_reason": "stop",
        "latency_s": 6.904,
        "usage": {
            "prompt_tokens": 23,
            "total_tokens": 973,
            "completion_tokens": 950,
            "prompt_tokens_details": null
        },
        "endpoint": "http://127.0.0.1:8000/v1/chat/completions"
    }
}
```
This looks like what we had before. Perhaps you could provide more information on which model & which version of vllm that requires this change? Thank you thank you!

Chibukach:
@shanjiaz This PR addresses the preprocessing step to create after the response generation located at `ctest-subset-Qwen3.5-397B-A17B-FP8-dynamic-speculator-dataset`. 

The issue is that when you do the response generation with Qwen with the reasoning parser, the `scripts/prepare_data.py ` sample prints empty "<think> </think>"  tags. This PR tries to address that. You can reproduce with this script 


```
hf download inference-optimization/ctest-subset-Qwen3.5-397B-A17B-FP8-dynamic-speculator-dataset --repo-type dataset --local-dir ./Dataset

python scripts/prepare_data.py --model Qwen/Qwen3.5-9B --data ./Dataset/magpie.jsonl --data ./Dataset/ultrachat.jsonl --output Qwen3_5_preprocessed  --max-samples 12 --assistant-pattern "<\|im_start\|>assistant\s*([\s\S]*?)<\|im_end\|>" 
```

src/speculators/data_generation/preprocessing.py
Chibukach: Fair point, there's only 1 metadata per conversation. I've removed these checks.

shanjiaz:
@Chibukach Could you fix quality and rebase? Thanks!

src/speculators/data_generation/preprocessing.py
```diff
@@ -267,17 +271,24 @@ def _preprocess_batch(
...
     if not conversations:
         log.warning(f"No conversations key found. Keys: {list(examples.keys())}")
         return results
 
-    for idx, conv in enumerate(conversations):
+    if len(metadatas) != len(conversations):
+        raise ValueError(
+            f"Metadata/conversation length mismatch: "
+            f"{len(metadatas)} vs {len(conversations)}"
+        )
```
speculatorsbot, line 284: `metadatas = examples.get("metadata", None)` returns `None` when the dataset lacks a `metadata` column, and the immediately following `len(metadatas)` on line 280 raises `TypeError: object of type 'NoneType' has no len()`. This crashes every dataset that does not include per-example metadata, which is the common case.

The `metadata` column should be treated as optional -- fall back to empty dicts so existing datasets continue to work:

```suggestion
    metadatas = examples.get("metadata", None)
    if metadatas is None:
        metadatas = [{} for _ in conversations]
    elif len(metadatas) != len(conversations):
        raise ValueError(
            f"Metadata/conversation length mismatch: "
            f"{len(metadatas)} vs {len(conversations)}"
        )
```

speculatorsbot (commented):
The approach of extracting `reasoning_content` from per-conversation metadata addresses a real need when vLLM's reasoning parser separates thinking content from the response body. However, there are crash-level issues and missing test updates that need to be resolved before this can merge.

**Critical issues:**

1. **`TypeError` on datasets without a `metadata` column** -- `len(None)` crashes when the column is absent (see inline comment).
2. **Existing tests broken** -- `_normalize_conversation` now requires `metadata: dict` as a mandatory positional parameter with no default, but no tests were updated to pass it. Every test calling `_normalize_conversation(conv)` will fail with `TypeError: _normalize_conversation() missing 1 required positional argument: 'metadata'`. The `metadata` parameter should default to `{}` (or `None`) so existing callers keep working.

**Other items:**

3. **DCO sign-off missing** -- All five commits lack the `Signed-off-by` line required by the [DCO check](https://github.com/vllm-project/speculators/blob/main/CONTRIBUTING.md). Use `git commit --amend -s` (or squash and re-sign) to add it.
4. **Rebase needed** -- The underlying preprocessing code on `main` has evolved significantly since this branch was created (multimodal processor support, tool handling, removal of `turn_dropout`, new `reasoning_content` handling in `_normalize_conversation`). The `needs-rebase` label is accurate. Note that `main` already handles per-turn `reasoning_content` via `turn.get("reasoning_content")` at [line 114](https://github.com/vllm-project/speculators/blob/310e0ff/src/speculators/data_generation/preprocessing.py#L114), so the rebase will need to reconcile the per-turn and per-conversation-metadata approaches.

src/speculators/data_generation/preprocessing.py
```diff
@@ -97,6 +98,9 @@ def _normalize_conversation(
         if "thinking" in turn and turn["thinking"]:
             normalized_turn["thinking"] = turn["thinking"]
 
+        if role == "assistant" and "reasoning_content" in metadata and metadata["reasoning_content"]:
+            normalized_turn["reasoning_content"] = metadata["reasoning_content"]
```
speculatorsbot, line 102: This applies the same `reasoning_content` from the per-conversation metadata to every assistant turn in the conversation. In a multi-turn conversation, earlier assistant turns (from the original dataset, not regenerated by vLLM) would incorrectly receive the reasoning content that belongs to the final regenerated turn.

Consider restricting this to only the **last** assistant turn, or passing `reasoning_content` as a per-turn field in the conversation data rather than per-conversation metadata.

Issue #541, opened 2026-05-20
Support configurable loss function via training args


shanjiaz:
### Motivation.

Currently, the loss function used during training is hardcoded per model:
- eagle3 uses kl_div_loss (src/speculators/models/eagle3/metrics.py:91)
- peagle uses kl_div_loss (src/speculators/models/peagle/metrics.py)
- dflash uses ce_loss (src/speculators/models/dflash/metrics.py:47)

The loss_function() in src/speculators/models/metrics.py:162 already accepts a loss_fn callable parameter, but there's no way to select between kl_div_loss and ce_loss (or other loss functions) from the training script or CLI args.

### Proposed Change.

- Add a --loss-fn argument to scripts/train.py (e.g., kl, ce)
- Thread it through get_trainer_kwargs() so each model's compute_metrics call uses the user-specified loss function instead of the hardcoded one

christinaexyou:
hi i'm interested in working on this - @shanjiaz can i be assigned to this issue ?

PR #542, opened 2026-05-21
feat(training): support configurable loss fn via training args

christinaexyou:

## Purpose
This PR addresses #541

## Description
It adds a --loss-fn argument and input validation  to `scripts/train.py`. If defined, it passes the resolved loss function to get_trainer_kwargs() so each model's compute_metrics call uses arg.loss_fn. If args.loss_fn=None, it falls back to the previously hardcoded loss fn.

## Tests

I added unit tests here `tests/unit/train/test_cli_args.py` to test that arg.loss_fn is passed correctly to each model's get_trainer_kwargs()  and that we maintain backward compatibility

I have filled in:


christinaexyou:
I noticed that there were no tests to verify CLI args so I created a new file but lmk if you want me to rename/move the tests :)

scripts/train.py
```diff
@@ -568,6 +569,17 @@ def parse_args():
     parser.add_argument("--mask-token-id", type=int, default=None)
     parser.add_argument("--ttt-steps", type=int, default=3)
     parser.add_argument("--ttt-step-loss-decay", type=float, default=1.0)
+    parser.add_argument(
+        "--loss-fn",
+        type=str,
+        default=None,
```
shanjiaz: Let's add kl as default.

src/speculators/models/peagle/core.py
```diff
@@ -237,17 +240,15 @@ def from_training_args(
         return model
 
     @staticmethod
-    def get_trainer_kwargs(**kwargs) -> tuple[dict, dict]:  # noqa: ARG004
+    def get_trainer_kwargs(**kwargs) -> tuple[dict, dict]:
```
shanjiaz, line 243: lol thanks for fixing

shanjiaz (commented):
Looks great! Pending quality check.

scripts/train.py
shanjiaz: I'm thinking maybe we should just default to kl. Just confirmed with research, let's default to kl and could you remove the hard-coded defaults for each algorithm? Thanks!

src/speculators/models/dflash/core.py
```diff
@@ -194,6 +196,7 @@ def forward(
         verifier_last_hidden_states: torch.Tensor,  # shape: [1, total_seq_len, hidden_size] # noqa: E501
         lengths: torch.Tensor | None = None,  # shape: [batch_size]
         position_ids: torch.Tensor | None = None,  # shape: [1, total_seq_len]
+        loss_fn=ce_loss,
```
shanjiaz: can we remove the default ce_loss here?

src/speculators/models/eagle3/core.py
```diff
@@ -130,6 +131,7 @@ def forward(  # noqa: C901
         ttt_steps: int = 3,
         ttt_step_loss_decay: float = 1.0,
         use_off_policy_tokens: bool = False,
+        loss_fn=kl_div_loss,
```
shanjiaz, line 134: same here.

src/speculators/models/peagle/core.py
```diff
@@ -55,6 +56,7 @@ def forward(
         position_ids: torch.Tensor | None = None,
         loss_mask: torch.Tensor | None = None,
         verifier_last_hidden_states: torch.Tensor | None = None,
+        loss_fn=kl_div_loss,
```
shanjiaz, line 59: also here.

src/speculators/models/dflash/core.py
fynnsu: I think it's okay to have default loss functions here. But we should just always use `kl_div_loss` as the default value.

The reason I think this is useful to have is for testing / other callers of this method so that they don't always have to specify it.

scripts/train.py
```diff
@@ -574,6 +575,17 @@ def parse_args():
     parser.add_argument("--mask-token-id", type=int, default=None)
     parser.add_argument("--ttt-steps", type=int, default=3)
     parser.add_argument("--ttt-step-loss-decay", type=float, default=1.0)
+    parser.add_argument(
+        "--loss-fn",
+        type=str,
+        default="kl",
+        choices=["kl", "ce"],
```
fynnsu: I kinda feel like we should use "kl_div" over "kl" but maybe that's just me. Don't feel super strongly either way tho

fynnsu (commented):
Thanks! Should be almost good to merge

src/speculators/models/dflash/core.py
shanjiaz: sure!

christinaexyou: @shanjiaz @fynnsu ok i removed the default loss functions and set the `loss_fn` param to `None`. my reason to for doing so is to make it obvious that `loss_fn` is configurable in the `forward` method.

scripts/train.py
christinaexyou: done

shanjiaz: Dealer's choice. @christinaexyou could you fix quality again? You can run `make quality` locally.

src/speculators/models/dflash/core.py
shanjiaz: Sorry I think Fynn has a point 😬, could be default to `kl_div_loss` for all three? Thanks!

shanjiaz (approved):
Great job! Thanks for contributing.

fynnsu (approved):

src/speculators/models/dflash/metrics.py
```diff
@@ -28,13 +30,15 @@ def compute_metrics(
...
         gamma: Temperature for exponential decay in loss weighting
+        loss_fn: Loss function
 
     Returns:
         Tuple of (loss, metrics_dict) where metrics_dict contains:
             - loss: Scalar loss value
             - full_acc: Overall accuracy
             - position {i} acc: Accuracy at position i within blocks
     """
+    loss_fn = loss_fn or kl_div_loss
```
fynnsu: Yeah, the failing quality check is right. This doesn't make sense now that `loss_fn` is always defined. We can just remove this line (and the others like it).

https://github.com/vllm-project/speculators/actions/runs/26540646793/job/78184414079?pr=542#step:6:24

PR #424, opened 2026-04-14
Add support for async hidden states connector

fynnsu:
PLEASE FILL IN THE PR DESCRIPTION HERE ENSURING ALL CHECKLIST ITEMS (AT THE BOTTOM) HAVE BEEN CONSIDERED.

## Purpose

https://github.com/vllm-project/vllm/pull/37374 updates the ExampleHiddenStatesConnector to write the hidden states to disk in an async manner (and also in `.pt` files instead of `.saftensors`). This pr adds support for this while maintaining support for the previous implementation.

## Description

The new system creates lock files for each hidden state file before starting to write the hidden states. On the client side, we block on obtaining the lock and once it's ready the hidden states are fully written to disk and can be read.

We only employ this behavior if the lock file exists, otherwise we assume the hidden states are fully written. This supports the existing system, and also works for offline or cached data gen results.

We clean up (delete) lock files after obtaining them.

## Tests

Will run smoke / regression tests on this pr to ensure compatibility with existing system. I have also tested this locally with the upstream pr. 

I have filled in:

- [x] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [x] The test plan/results, such as providing test command and pasting the results.
- [ ] (Optional) The necessary documentation update.
- [x] I (a human) have written or reviewed the code in this pr to the best of my ability.

PR #566, opened 2026-06-01
Fix waiting for lock acquisition during online training

fynnsu:

## Purpose

#424 accidentally added lock waiting logic but only for the first load attempt of the hidden states file. When that first load (intended for pre-existing data) fails, the generate hidden states is called and new hidden states are generated. Unfortunately the new hidden states were loaded without waiting for lock. 

## Description

Make `_maybe_load_hs_file` a helper function so that it can be used by both loads. 

## Tests

Tested locally that this fixes these errors (which subsequently led to training crashing):
```
speculators/src/speculators/train/data.py:300: UserWarning: Failed to load/cache hidden states for sample 29: No such file or directory:
/tmp/pytest-of-fynnsu/pytest-55/test_online_smoke_Qwen_Qwen3_V0/hidden_states/chatcmpl-b549e54beee9b9c2-93723e8e.safetensors
```

I have filled in:

### DIFF
--- a/src/speculators/train/data.py
+++ b/src/speculators/train/data.py
@@ -178,6 +178,17 @@ class BaseDataset(Dataset):
         return data
 
 
+def _maybe_load_hs_file(file_path: Path) -> dict[str, torch.Tensor] | None:
+    lock_path = str(file_path) + ".lock"
+    if Path(lock_path).exists():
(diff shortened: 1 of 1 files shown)


shanjiaz (approved):
Good catch! We should probably allow smoke tests on PRs.

Issue #570, opened 2026-06-03
[Bug]: draft model inference not working due to lack of M-RoPE support


Chibukach:
### Your current environment

Please provide the following information about your environment if applicable:

- vLLM main - commit: https://github.com/vllm-project/vllm/commit/a4ac746405f4ddbef553098507210c072b5ba39e  
- Speculators main - commit: 927c9c8ea3548a6073ea6eba0bef9d01b7d3bdc7
- CUDA - 13.2
- PyTorch - 2.11.0
- Transformers - 5.9.0
- Hardware - A100-SXM4-80GB
- Model - ctest-Qwen3.6-27B-speculator-regen.dflash

### 🐛 Describe the bug

The current speculators copies the draft-arch config from llama by default then overrides m-rope in  rope_parameters from the verifier config. This is not supported by default so rope_parameters still needs to be overridden when using transformers version is greater than 5 but all the information such as `mrope_interleaved` and `mrope_section` should not be copied over. 

#### Command:
```
CUDA_VISIBLE_DEVICES=0,1 vllm serve inference-optimization/ctest-Qwen3.6-27B-speculator-regen.dflash -tp 2  --max-model-len 16384
```
#### Error:
```
(Worker pid=587950) ERROR 06-03 07:59:13 [multiproc_executor.py:868]     raise NotImplementedError(
(Worker pid=587950) ERROR 06-03 07:59:13 [multiproc_executor.py:868] NotImplementedError: Speculative Decoding with draft models or parallel drafting does not support M-RoPE yet
```

#### Fix:
Working Fix by explicitly selecting `rope_theta` and `rope_type`. Alternatively, copy existing config and exclude `mrope_interleaved` and `mrope_section`.
```
if version.parse(transformers.__version__) >= version.parse("5.0.0"):
         if hasattr(verifier_config, "rope_parameters"):
-            config.rope_parameters = deepcopy(verifier_config.rope_parameters)
+            config.rope_parameters= {"rope_theta": verifier_config.rope_parameters.get("rope_theta", 10000.0), "rope_type": verifier_config.rope_parameters.get("rope_type",  "default")}
```

fynnsu:
#598

fynnsu:
Closed by #598

PR #571, opened 2026-06-03
fix m-rope inference bug

Chibukach:
## Purpose

This PR fixes m-rope not supported bug. Speculator inference does not work without it. 

```
(Worker pid=587950) ERROR 06-03 07:59:13 [multiproc_executor.py:868]     raise NotImplementedError(
(Worker pid=587950) ERROR 06-03 07:59:13 [multiproc_executor.py:868] NotImplementedError: Speculative Decoding with draft models or parallel drafting does not support M-RoPE yet
```

## Description

- update rope_parameters to exclude m-rope key-value pairs 

## Related Issue

- https://github.com/vllm-project/speculators/issues/570

## Tests

Trained and inferenced dflash sliding window attention model 

I have filled in:


scripts/train.py
```diff
@@ -687,6 +693,12 @@ def parse_args():
         default=8,
         help="Block size for DFlash model (default: 8)",
     )
+    parser.add_argument(
+        "--disable-m-rope",
+        action=argparse.BooleanOptionalAction,
+        default=True,
+        help="Whether to support m-rope positional encoding (default: True)",
+    )
```
rahul-tuli, line 701: Could we invert and simplify this?
I'm not a fan of double negation flags like `--no-disable-m-rope`

```suggestion
    parser.add_argument(
        "--m-rope",
        action="store_true",
        help="Enable m-rope positional encoding.",
    )
```

fynnsu (changes requested):
I'm not really a fan of this approach of just deleting rope_parameters.

If this is needed because the llama draft layers don't support the params, then we should automatically remove the params when working with llama draft layers.

fynnsu:
Closing in favor of #598

PR #574, opened 2026-06-04
Preserve rope_scaling when building Eagle3 transformer config

SuperMarioYL:
## Summary

`Eagle3Converter._create_transformer_config_from_eagle` builds the drafter's `LlamaConfig` through `build_llama_config_rope_kwargs(...)`, but only forwards `rope_theta` — it never passes `rope_scaling`. As a result, any Eagle3 checkpoint converted for a **rope-scaled** target silently loses its RoPE scaling and falls back to default RoPE.

This is inconsistent with the Eagle v1 converter, which already forwards it:

```python
# src/speculators/convert/eagle/eagle_converter.py
**build_llama_config_rope_kwargs(
    rope_theta=eagle_config.get("rope_theta", 10000.0),
    rope_scaling=eagle_config.get("rope_scaling"),   # <-- forwarded
),

# src/speculators/convert/eagle/eagle3_converter.py (before this PR)
**build_llama_config_rope_kwargs(
    rope_theta=eagle_config.get("rope_theta", 10000.0),
    # rope_scaling dropped
),
```

`build_llama_config_rope_kwargs` already fully supports `rope_scaling` (both the transformers v4 `rope_scaling` and v5 `rope_parameters` layouts), so the value was simply not being threaded through.

## Impact

Drafters converted for common long-context targets — Llama‑3.1 / Llama‑3.3 (`rope_type: "llama3"`), linear/dynamic NTK, YaRN — get a transformer config with **default RoPE instead of the scaled RoPE**, i.e. wrong positional encoding at long context. The conversion succeeds with no error, so the regression is silent.

## Fix

Forward `rope_scaling` from the Eagle config, mirroring the v1 converter (one line).

## Tests

Added to `tests/unit/convert/test_eagle3_converter.py`:
- `test_config_preserves_rope_scaling` — a `llama3` scaling config survives conversion (asserts on `rope_parameters` for transformers v5 and `rope_scaling` for v4).
- `test_config_no_rope_scaling_when_absent` — regression guard: when the Eagle config has no `rope_scaling`, none is introduced.

The first test fails on `main` (`rope_type` stays `"default"` instead of `"llama3"`) and passes with this change. Full `tests/unit/convert/test_eagle3_converter.py` passes, and `ruff check`, `ruff format --check`, and `mypy --check-untyped-defs` are clean.


tests/unit/convert/test_eagle3_converter.py
fynnsu, line 1: The code change looks okay, but I think we can do without the test.

fynnsu (commented):
Thanks for the pr!

tests/unit/convert/test_eagle3_converter.py
SuperMarioYL: Done — dropped the two converter tests as suggested. The rope_scaling forwarding fix in the converter is unchanged. Thanks for the review!

fynnsu (approved):

PR #581, opened 2026-06-08
[Feature] Add Virtual Width Network Eagle3 to Speculators

YzTongNiar:
## Purpose

We add a customed speculative model "Virtual Width Network Eagle3" with architecture similar to ZJ's work 'Virtual Width Network' (https://huggingface.co/papers/2511.11238)  

We impelmented and trained the 'vwn eagle3' based on 'speculators' library  (https://github.com/vllm-project/speculators); the training datasets includes 'ultrachat' (https://huggingface.co/datasets/openbmb/UltraChat) and 'sharegpt' (https://huggingface.co/collections/bunnycore/sharegpt-datasets).

We used the same setting to train a vwn-eagle3 and an eagle3 model to compare their performace on longbenchv2, gsm8k, sharegpt, and humaneval; as well as some man-made fixed-length datasets. The performance is concluded as following:

Datasets testing result:

| Dataset | Context-Length | E2E Throughput Gain|
| ------ | ------ | ------ | 
| gsm8k   | 1.5k | 7.9%|
| humaneval | 2.5k | 11.94%|
| sharegpt  | 5k | 2.84%|
| longbench_v2| 128k | 18.23%|

Fixed length testing results:

|Input Lenght | Output Length | E2E Throughput Gain|
| ------ | ------ | ------ | 
| 2k   | 2k | 9.83%|
| 3.5k | 1.5k | 2.38%|
| 32k  | 1k | 18.66%|

## Tests

The codes are based on speculators v0.4.1; we are rebaseing it to the main currently. We raised this PR simply for notification and comments usage, we will rebase it very soon.

## Checklist

I have filled in:

- [ Yes ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ No ] The test plan/results, such as providing test command and pasting the results.
- [ No ] (Optional) The necessary documentation update.
- [ Yes ] I (a human) have written or reviewed the code in this pr to the best of my ability.

### DIFF
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -248,8 +248,49 @@ def parse_args():
     parser.add_argument(
         "--speculator-type",
         type=str,
-        default="eagle3",
-        help="Type of speculator model to train (e.g., eagle3)",
+        default="vwn_eagle3",
+        help="Type of speculator model to train (e.g., eagle3, vwn_eagle3)",
+    )
+    parser.add_argument(
+        "--vwn-m",
+        type=int,
+        default=1,
+        help=(
+            "VWN partition count for original hidden vector. "
+            "Must divide hidden_size evenly. "
+            "Ablation: set --vwn-m 1 --vwn-r 1.0 to disable width expansion."
+        ),
+    )
+    parser.add_argument(
+        "--vwn-r",
+        type=float,
+        default=1,
+        help=(
+            "VWN expansion ratio: D' = r * D. "
+            "Must satisfy r * m = integer. "
+            "Recommended: 1.5 (m=2) or 8.0 (m=8)."
+        ),
+    )
+    parser.add_argument(
+        "--vwn-expand",
+        type=bool,
+        default=True,
+        help=(
+            "VWN expansion switch "
+            "set to False when r=1."
+
+        ),
+    )
+    parser.add_argument(
+        "--pre-vwn-version",
+        type=int,
+        default=1,
+        choices=[0, 1],
+        help=(
+            "Version of pre_vwn_layer to use. "
+            "0: VwnEagle3PreVwnLayer (default, full fusion). "
+            "1: VwnEagle3PreVwnLayerV1 (ablation, simplified fusion)."
+        ),
     )
     parser.add_argument("--data-path", type=str, default="./data")
     parser.add_argument("--save-path", type=str, default="./checkpoints")
--- a/src/speculators/models/eagle3/config.py
+++ b/src/speculators/models/eagle3/config.py
@@ -82,3 +82,94 @@ class Eagle3SpeculatorConfig(SpeculatorModelConfig):
                 ).__class__
             return config_class(**value)
         return value
+# ──────────────────────────────────────────────────────────────────────────────
+# VwnConfig: VWN 参数容器 + 约束校验
+# ──────────────────────────────────────────────────────────────────────────────
+
+@dataclass
+class VwnConfig:
+    """VWN 宽度扩展参数。
+
+    约束（来自设计文档）：
+        D' = r * D
+        r  = n / m   →  r * m 必须为整数
+        m  能整除 D
+
+    """
+
+    m: int = 2
+    r: float = 1.5
+    expand: bool = True
+
+    def __post_init__(self):
+        n_float = self.r * self.m
+        self.n: int = round(n_float)
+        if self.r == 1:
+            self.expand = False
+
+    def expanded_hidden_size(self, hidden_size: int) -> int:
+        """返回 D' = r * D，同时校验整除约束。"""
+        if hidden_size % self.m != 0:
+            raise ValueError(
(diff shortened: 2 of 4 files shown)


fynnsu:
Hey, thanks for the pr. I saw your note on rebasing, which would be great!

My main feedback at this stage is that it looks like you're mainly changing the architecture of the draft model. This should be done by adding a `--draft-arch` option (currently we support `llama` and `qwen3`), instead of new speculator type. 

These architectures are currently defined in `src/speculators/models/base_components.py` and `src/speculators/models/eagle3/model_definitions.py`. 

And then when running training you would use `--speculator-type eagle3 --draft-arch vwn` (or whatever the architecture is called).

I will also note that currently vLLM only supports running inference with llama-style eagle3 drafters. There is some work in-progress to maybe add qwen3 arch drafters, and if VWN is notably better then it should be possible to add drafters with that architecture as well, but that will need to be done on the vLLM side. 

Let me know if you have any questions.

scripts/train.py
```diff
@@ -248,8 +248,49 @@ def parse_args():
...
+    )
+    parser.add_argument(
+        "--vwn-r",
+        type=float,
+        default=1,
+        help=(
+            "VWN expansion ratio: D' = r * D. "
+            "Must satisfy r * m = integer. "
+            "Recommended: 1.5 (m=2) or 8.0 (m=8)."
+        ),
```
speculatorsbot, line 272: **Inconsistent defaults between CLI and config class.**

The CLI defaults here (`--vwn-m 1`, `--vwn-r 1`) produce **no width expansion**: `r=1` causes `VwnConfig.__post_init__` to set `expand=False`, making the VWN wrapper a no-op. Meanwhile, `VwnEagle3SpeculatorConfig` ([config.py](https://github.com/vllm-project/speculators/blob/fabf9c41334279870ca7ca385d3a19d2b21d6baf/src/speculators/models/eagle3/config.py#L137-L145)) defaults to `m=2, r=1.5` -- the meaningful expansion case from the paper.

This means:
- Creating a VWN model via CLI with defaults -> no expansion (effectively plain Eagle3 with extra unused parameters)
- Creating a VWN model programmatically via config -> actual 1.5x expansion

These should be aligned. If expansion must be explicitly opted into, the config defaults should match (`m=1, r=1`). If VWN models should expand by default, the CLI defaults should be `m=2, r=1.5`.

The help text for `--vwn-m` even describes `m=1, r=1` as the "ablation" case to "disable width expansion" -- making the ablation config the default seems unintentional.

src/speculators/models/eagle3/config.py
```diff
@@ -82,3 +82,94 @@ def validate_transformer_config(cls, value: Any) -> PretrainedConfig:
...
+        if hidden_size % self.m != 0:
+            raise ValueError(
+                f"hidden_size ({hidden_size}) 必须能被 m ({self.m}) 整除。"
+            )
+        d_prime = round(self.r * hidden_size)
+        if d_prime % self.n != 0:
+            raise ValueError(
+                f"扩维后 D'={d_prime} 必须能被 n={self.n} 整除。"
+            )
+        return d_prime
```
speculatorsbot, line 121: **Missing `D' % m == 0` validation causes runtime crash for some hyperparameter combinations.**

The grouped projections in `VwnEagle3FirstLayerMixin` reshape tensors via `view(-1, self.d_prime // self.vwn.m)`, which requires `D'` to be divisible by `m`. This method validates `D % m == 0` (line 112) and `D' % n == 0` (line 117), but never checks `D' % m == 0`.

For the default `m=2, n=3`, this happens to work because common hidden sizes produce an even `D'`. But for other valid configurations it crashes:

```
m=3, n=4 (r=4/3):
  D=3072 -> D'=4096
  D % m = 3072 % 3 = 0   (passes)
  D' % n = 4096 % 4 = 0  (passes)
  D' % m = 4096 % 3 = 1  (FAILS at runtime)
```

The `view(-1, 4096 // 3)` would attempt to reshape a tensor of size `...x4096` into groups of 1365, and `4096 / 1365 != integer`, so PyTorch raises a RuntimeError.

Suggested fix -- add a third check before the return:
```python
if d_prime % self.m != 0:
    raise ValueError(
        f"D'={d_prime} must be divisible by m={self.m} "
        f"for grouped projections. Choose m, r such that "
        f"round(r * hidden_size) is divisible by m."
    )
```

speculatorsbot (commented):
The maintainer (@fynnsu) has already provided the key design feedback: VWN should be integrated as a new `--draft-arch` option rather than a new speculator type. I agree with that direction and will not repeat it here. The CodeRabbit review also covers several implementation issues (missing `dataclass` import, missing `@SpeculatorModel.register`, `type=bool` argparse bug, debug code, hard-coded paths).

Below are two additional **correctness issues** in the VWN implementation that I did not see raised elsewhere and that would apply regardless of how the architecture is ultimately integrated.

PR #585, opened 2026-06-08
feat(mtp): auto-extract native MTP weights during training

rahul-tuli:
## Purpose

Eliminate the explicit conversion step before MTP training. Users no longer need to run `speculators convert` or `convert_model()` before training -- the training script now auto-extracts native MTP weights from the verifier model in-memory during initialization.

**Before:** `convert → save to disk → train --from-pretrained ./converted`
**After:** `train` (auto-extracts internally)

### Changes

- **`MTPConverter.convert_to_state_dict()`** — new method that performs the full extraction pipeline (download → verify → extract → remap → fuse MoE) and returns the state dict without disk writes. `convert()` now delegates to it internally.
- **`MTPDraftModel.from_training_args(load_native_weights=True)`** — when True, calls `convert_to_state_dict()` and loads weights via `load_state_dict(strict=False)`. When False (resume), skips extraction.
- **`train.py`** — adds `--num-speculative-steps` CLI arg (default 3), removes the ValueError requiring `--from-pretrained` for MTP, adds resume detection to skip extraction when checkpoint dirs exist.
- **Tutorial** — removes "Step 1: Convert Native MTP Head", renumbers steps, updates commands and notes.
- **Example script** — removes inline conversion block and `CONVERTED_DIR`, renumbers steps.
- **E2E smoke test** — removes explicit `MTPConverter().convert()` call, replaces `--from-pretrained` with `--num-speculative-steps`.

### Backward compatibility

- `--from-pretrained` still works for MTP (bypasses `from_training_args()` entirely)
- `speculators convert --algorithm mtp` CLI and `convert_model()` API unchanged
- Standalone conversion roundtrip test (`test_mtp_conversion_roundtrip.py`) unmodified

## Tests

- All 328 unit tests pass (`python -m pytest tests/unit/ -x -q`)
- `ruff check` and `ruff format` pass on all changed files
- E2E smoke test (`test_mtp_finetuning.py`) updated to exercise auto-conversion flow
- E2E roundtrip test unchanged (regression coverage for standalone conversion)

NOTE: 
- [ ] PR #593 must land before this diff
## Checklist

I have filled in:


scripts/train.py
```diff
@@ -274,6 +274,16 @@ def parse_vocab_mappings(args: argparse.Namespace):
...
+def _mtp_will_resume(args: argparse.Namespace) -> bool:
+    """Check if an MTP training run will resume from an existing checkpoint."""
+    if args.from_pretrained:
+        return False
+    save_path = Path(args.save_path)
+    if args.no_resume_from_checkpoint or not save_path.exists():
+        return False
+    return any(d.is_dir() and d.name.isdigit() for d in save_path.iterdir())
+
+
```
fynnsu: I understand the reasoning behind avoid this extra write step, but I think checkpoint resumption is rare enough that it's better to keep the code simple, at a slight setup cost when it happens. 

This is also what happens when loading `--from-pretrained` with the other draft model variants.

fynnsu (commented):
I think it would be better to keep this a little simpler. Otherwise, the change to ux looks good.

scripts/train.py
rahul-tuli: Done!

fynnsu (approved):

shanjiaz (approved):

Issue #544, opened 2026-05-25
[Bug]: qwen3-8b dflash


Jim2016713:
### Your current environment

Please provide the following information about your environment if applicable:

- vLLM 0.18.0
- Speculators 0.5.0
- CUDA 13.0
- PyTorch 2.9
- Transformers 5.5.4
- Hardware A100
- Model qwen3-8b

### 🐛 Describe the bug

[14:41:10] INFO     Started distributed with local_rank=1, world_size=2                                                                                                                                                                   utils.py:39
[14:41:10] INFO     Started distributed with local_rank=0, world_size=2                                                                                                                                                                   utils.py:39
           INFO     Loading vocab mappings from '/output/dflash_qwen3_8b_onestepmodel/d2t.npy' and 'output/dflash_qwen3_8b_onestepmodel/t2d.npy'                                                      train.py:161
[14:41:12] WARNING  Added <|MASK|> to tokenizer, mask_token_id=151669 (tokenizer len=151670, vocab_size=151936)                                                                                                                           utils.py:85
2026-05-25 14:41:24.498 | INFO     | speculators.utils.loading:_resolve_file:83 - Loading from local directory: /output/e2e_78_bl140_260410/v3-20260509-033321/checkpoint-968-merged
2026-05-25 14:41:24.584 | INFO     | speculators.utils.loading:_resolve_file:83 - Loading from local directory: /output/e2e_78_bl140_260410/v3-20260509-033321/checkpoint-968-merged
2026-05-25 14:41:24.614 | INFO     | speculators.utils.loading:_resolve_file:83 - Loading from local directory: /output/e2e_78_bl140_260410/v3-20260509-033321/checkpoint-968-merged
2026-05-25 14:41:24.644 | INFO     | speculators.utils.loading:_resolve_file:83 - Loading from local directory: /doutput/e2e_78_bl140_260410/v3-20260509-033321/checkpoint-968-merged
2026-05-25 14:41:24.742 | INFO     | speculators.utils.loading:_resolve_file:83 - Loading from local directory: /output/e2e_78_bl140_260410/v3-20260509-033321/checkpoint-968-merged
2026-05-25 14:41:24.772 | INFO     | speculators.utils.loading:_resolve_file:83 - Loading from local directory: /output/e2e_78_bl140_260410/v3-20260509-033321/checkpoint-968-merged
[...]

Issue #531, opened 2026-05-19
[Bug][NPU]: Running simple example on NPU seems to be broken


menogrey:
### Your current environment

Please provide the following information about your environment if applicable:
I use vllm-ascend latest nightly docker image.
- vLLM:0.20.2 and vllm-ascend:0.19.1rc2.dev101+ga45cdf9b9
- Speculators:latest
- CUDA:None
- PyTorch:torch==2.10.0+cpu and torch_npu==2.10.0
- Transformers==5.5.3
- Hardware:Ascend A3 and aarch64 CPU
- Model:Qwen3-8B

### 🐛 Describe the bug

Run example script and found vllm server ready, but step 3: training is broken.
```
bash examples/train/eagle3_qwen3_8b_sharegpt_online_5k.sh
```

```
[rank0]: Traceback (most recent call last):
[rank0]:   File "/workspace/speculators/scripts/train.py", line 681, in <module>
[rank0]:     main(args)
[rank0]:   File "/workspace/speculators/scripts/train.py", line 380, in main
[rank0]:     trainer.run_training()
[rank0]:   File "/workspace/speculators/src/speculators/train/graceful_shutdown.py", line 118, in wrapper
[rank0]:     return fn(self, *args, **kwargs)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/workspace/speculators/src/speculators/train/trainer.py", line 340, in run_training
[rank0]:     self.train_epoch(epoch)
[rank0]:   File "/workspace/speculators/src/speculators/train/trainer.py", line 209, in train_epoch
[rank0]:     _draft_tokens, loss, metrics = self.model(
[rank0]:                                    ^^^^^^^^^^^
[rank0]:   File "/usr/local/python3.11.15/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1776, in _wrapped_call_impl
[rank0]:     return self._call_impl(*args, **kwargs)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/usr/local/python3.11.15/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1882, in _call_impl
[rank0]:     return inner()
[rank0]:            ^^^^^^^
[rank0]:   File "/usr/local/python3.11.15/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1830, in inner
[...]

PR #589, opened 2026-06-09
feat(dflash): selectable attention backend (sdpa/eager) with dense mask

Sawyer117:
## What this does
  Lets you pick the attention backend for the DFlash draft via `--draft-attn-impl`.
  The default doesn't change (still flex) — this just adds `sdpa`/`eager` for cases
  where flex attention can't run.

  ## Why
  On Ascend NPU, `flex_attention` flat-out errors (`FlexAttention is only supported on
  CUDA, CPU or HPU devices`, #531), and DFlash hardcodes flex — so there was no way to
  train it on NPU. The `@torch.compile`'d forward is also unhappy on NPU, and flex +
  compile has been flaky on CUDA too (#544). So I added a path that doesn't need flex.

  ## How
  - Added `build_anchor_block_dense_mask` — it builds the same anchor-block mask that
    `create_anchor_block_mask_mod` already builds, just as a plain dense boolean tensor
    instead of a flex `BlockMask`. Same rules (a query sees context before its anchor +
    its own block, sliding-window supported), with a unit test checking it matches the
    flex version exactly.
  - Added a `--draft-attn-impl {simple_flex_attention,sdpa,eager}` flag. Default is
    flex, so CUDA behavior is unchanged. `sdpa`/`eager` use the dense mask instead of a
    flex `BlockMask`.
  - With `sdpa`, the op becomes a fused FlashAttention kernel on NPU, so the full
    attention scores never get materialized — `max_anchors` can stay large.

  ## Tested
  - Unit test: dense mask matches the flex `mask_mod` exactly (full attention + sliding
    window).
  - End-to-end: trained Qwen3-8B DFlash on Ascend (CANN 9.0.0 / vllm-ascend 0.20.2rc1)
    with `--draft-attn-impl sdpa` — runs, loss ~1.3, position_1 acc ~0.40 and climbing.

  ## Notes
  - flex is block-sparse and faster when it works, so it stays the default; sdpa/eager
    is the fallback for when flex isn't available.
  - EAGLE3 and PEAGLE hardcode flex the same way; this PR is DFlash-only, but the same
    approach drops in for them as a follow-up.

### DIFF
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -726,6 +726,14 @@ def parse_args():
         default=256,
         help="Maximum anchor positions for DFlash training (default: 256)",
     )
+    parser.add_argument(
+        "--draft-attn-impl",
+        type=str,
+        default="simple_flex_attention",
+        choices=["simple_flex_attention", "sdpa", "eager"],
+        help="Attention implementation for the DFlash draft layers. "
+        "Use 'sdpa' or 'eager' for non-flex backends.",
+    )
     # P-EAGLE specific parameters
     parser.add_argument(
         "--num-depths",
--- a/src/speculators/models/dflash/core.py
+++ b/src/speculators/models/dflash/core.py
@@ -2,7 +2,7 @@ from typing import ClassVar
 
 import torch
 from torch import nn
-from torch.nn.attention.flex_attention import create_block_mask
+from torch.nn.attention.flex_attention import create_block_mask, create_mask
 from transformers import PretrainedConfig
 from transformers.models.qwen3.modeling_qwen3 import (
     Qwen3RMSNorm,
@@ -49,6 +49,12 @@ class DFlashDraftModel(DraftVocabMixin, SpeculatorModel):
             config.transformer_layer_config._attn_implementation = (  # noqa: SLF001
                 "simple_flex_attention"
             )
+        self._attn_impl = config.transformer_layer_config._attn_implementation  # noqa: SLF001
+        self._create_mask_fn = (
+            create_block_mask
+            if self._attn_impl == "simple_flex_attention"
+            else create_mask
+        )
         super().__init__(config=config)
         self._init_vocab(config)
 
@@ -144,6 +150,10 @@ class DFlashDraftModel(DraftVocabMixin, SpeculatorModel):
             kwargs["verifier_name_or_path"],
         )
 
+        verifier_config._attn_implementation = kwargs.get(  # noqa: SLF001
+            "draft_attn_impl", "simple_flex_attention"
+        )
+
         config = DFlashSpeculatorConfig(
             transformer_layer_config=verifier_config,
             draft_vocab_size=kwargs["draft_vocab_size"],
@@ -195,6 +205,32 @@ class DFlashDraftModel(DraftVocabMixin, SpeculatorModel):
             )
         return self.config.mask_token_id
 
+    def _create_attention_mask(
+        self,
+        lengths: torch.Tensor,
+        total_seq_len: int,
+        anchor_positions: torch.Tensor,
+        device: torch.device,
+        sliding_window: int | None = None,
+        sliding_window_non_causal: bool = False,
+    ):
+        mask_mod, q_len, kv_len = create_anchor_block_mask_mod(
+            lengths=lengths.to(device),
+            total_seq_len=total_seq_len,
+            anchor_positions=anchor_positions,
+            block_size=self.block_size,
+            sliding_window=sliding_window,
+            sliding_window_non_causal=sliding_window_non_causal,
+        )
+        return self._create_mask_fn(
+            mask_mod,
+            B=None,
+            H=None,
+            Q_LEN=q_len,
+            KV_LEN=kv_len,
+            device=device,
+        )
+
     @torch.compiler.disable
     def _build_attention_mask(self, loss_mask, lengths, device):
         total_seq_len = loss_mask.shape[1]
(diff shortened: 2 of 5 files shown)


scripts/train.py
```diff
@@ -726,6 +726,16 @@ def parse_args():
         default=256,
         help="Maximum anchor positions for DFlash training (default: 256)",
     )
+    parser.add_argument(
+        "--draft-attn-impl",
+        type=str,
+        default=None,
```
menogrey: default="simple_flex_attention" may be better?

src/speculators/models/dflash/core.py
```diff
@@ -205,30 +226,51 @@ def _build_attention_mask(self, loss_mask, lengths, device):
...
-            mask_mod, q_len, kv_len = create_anchor_block_mask_mod(
-                lengths=lengths.to(device),
-                total_seq_len=total_seq_len,
-                anchor_positions=anchor_positions,
-                block_size=self.block_size,
-                sliding_window=None,
-            )
-            full_attn_mask = create_block_mask(
-                mask_mod, B=None, H=None, Q_LEN=q_len, KV_LEN=kv_len, device=device
-            )
```
fynnsu, line 217: I think we can use:
```python
from torch.nn.attention.flex_attention import create_mask, create_block_mask

# in __init__
self._create_mask_fn = create_block_mask if self._attn_impl == "simple_flex_attention" else create_mask

# in forward
mask_mod, q_len, kv_len = create_anchor_block_mask_mod(
    lengths=lengths.to(device),
    total_seq_len=total_seq_len,
    anchor_positions=anchor_positions,
    block_size=self.block_size,
    sliding_window=None,
)

full_attn_mask = self._create_mask_fn(mask_mod, B=None, H=None, Q_LEN=q_len, KV_LEN=kv_len, device=device)
```

The `create_mask` function from pytorch creates a dense mask from the given mask_mod function and has pretty much the same signature as `create_block_mask`.

src/speculators/models/dflash/core.py
```diff
@@ -205,30 +226,51 @@ def _build_attention_mask(self, loss_mask, lengths, device):
...
+            else:
+                mask_mod, q_len, kv_len = create_anchor_block_mask_mod(
+                    lengths=lengths.to(device),
+                    total_seq_len=total_seq_len,
+                    anchor_positions=anchor_positions,
+                    block_size=self.block_size,
+                    sliding_window=self.sliding_window,
+                    sliding_window_non_causal=self.sliding_window_non_causal,
+                )
+                sliding_window_attn_mask = create_block_mask(
```
fynnsu: Same deal with sliding window.

src/speculators/models/dflash/core.py
```diff
@@ -144,6 +158,13 @@ def from_training_args(
             kwargs["verifier_name_or_path"],
         )
 
+        # Attention backend for the draft layers (--draft-attn-impl); defaults to
+        # flex in __init__. Set 'sdpa'/'eager' on backends without flex_attention
+        # (e.g. Ascend NPU).
+        draft_attn_impl = kwargs.get("draft_attn_impl")
```
fynnsu: Can be simplified if we just set a valid default value for `--draft-attn-impl`

src/speculators/models/dflash/core.py
```diff
@@ -44,11 +47,22 @@ def __init__(
...
-        # Forcibly override config settings
+        """Build the DFlash draft model from its config.
+
+        Resolves the draft attention backend (`config.transformer_layer_config.
+        _attn_implementation`): flex by default, or the dense-mask `sdpa`/`eager`
+        path on backends without `flex_attention` (e.g. Ascend NPU).
+        """
+        # Default to flex attention. On backends without flex_attention (e.g. Ascend
+        # NPU), pass --draft-attn-impl sdpa (or eager) to use a materialized dense
+        # mask instead; see build_anchor_block_dense_mask.
```
fynnsu: Remove

scripts/train.py
fynnsu: Agreed

fynnsu (changes requested):
Added some comments. This is worth adding but I think we can simplify this implementation a bit. Please see notes below. 

Also, it would be great if you could add this to the other algorithms (eagle3, peagle, mtp), using the approach described below? But I can also do that in a follow up if you don't have time.

Sawyer117:
> Added some comments. This is worth adding but I think we can simplify this implementation a bit. Please see notes below.
> 
> Also, it would be great if you could add this to the other algorithms (eagle3, peagle, mtp), using the approach described below? But I can also do that in a follow up if you don't have time.

Hi @fynnsu thank you very much for the reply! I agree with all the comments and will fix them tomorrow, meanwhile if you want to fix it for eagle3/peagle/mtp, feel free to go ahead🙏

fynnsu:
> I agree with all the comments and will fix them tomorrow, meanwhile if you want to fix it for eagle3/peagle/mtp, feel free to go ahead🙏

Sounds good! I'll add this as a follow up, once this pr lands.

scripts/train.py
Sawyer117: totally agree, just fixed.

src/speculators/models/dflash/core.py
Sawyer117: much appreicated,   I followed this pattern, but wrapped the final call in `_create_attention_mask()` so the same logic can be reused for both full attention and sliding-window attention.

  `self._create_mask_fn` is still the backend selector:
  - `simple_flex_attention` -> `create_block_mask`
  - `sdpa` / `eager` -> `create_mask`

  Then `_create_attention_mask()` builds `mask_mod, q_len, kv_len` and calls `self._create_mask_fn(...)`. So it’s the same structure as your suggestion, just factored into a helper to avoid duplicating the block for the sliding-window path.

src/speculators/models/dflash/core.py
Sawyer117: fixed as mentioned above

src/speculators/models/dflash/core.py
Sawyer117: just fixed

src/speculators/models/dflash/core.py
Sawyer117: done

Sawyer117:
Hi @fynnsu anything else from my side that needs to be done for the PR to be merged?

fynnsu (approved):

fynnsu:
> Hi @fynnsu anything else from my side that needs to be done for the PR to be merged?

No, I think you should be good. I just pushed a couple extra tests, and should be almost ready to merge.

PR #591, opened 2026-06-09
Fix empty sample creation

Chibukach:
## Purpose

The purpose of this PR is to fix the creation of an empty sample if an error occurs such as if the vllm hidden-state extraction times out. The current implementation uses `3 * hidden_size`  which seems to be hardcoded for an Eagle implementation whereas Dflash should be 5. The proposed solution creates the empty batch from `num_layers * hidden_size` instead, in order to be agnostic of the speculative decoding algorithm. 

```
[rank0]:E0608 10:44:59.043000 4122456 torch/_subclasses/fake_tensor.py:2908] [1/1] RuntimeError: a and b must have same reduction dim, but got [8192, s33] X [10240, 2048].
```
## Checklist

I have filled in:

### DIFF
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -102,7 +102,11 @@ def setup_dataloader(
         prefetch_factor=prefetch_factor,
         pin_memory=True,
         collate_fn=create_collate_fn(
-            args.total_seq_len, hidden_size, dataset.hidden_states_dtype, preprocess
+            args.total_seq_len,
+            hidden_size,
+            dataset.hidden_states_dtype,
+            args.num_layers,
+            preprocess,
         ),
         persistent_workers=True,
     )
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -104,8 +104,8 @@ def setup_dataloader(
         collate_fn=create_collate_fn(
             args.total_seq_len,
             hidden_size,
-            dataset.hidden_states_dtype,
             args.num_layers,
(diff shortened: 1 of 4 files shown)


rahul-tuli:
Looks good pending tests

shanjiaz (approved):
@Chibukach would really love to get this in! Could you run `make quality`, should be able to see the errors locally. Thank you so much!

rahul-tuli (approved):

scripts/train.py
```diff
@@ -102,7 +102,11 @@ def setup_dataloader(
         prefetch_factor=prefetch_factor,
         pin_memory=True,
         collate_fn=create_collate_fn(
-            args.total_seq_len, hidden_size, dataset.hidden_states_dtype, preprocess
+            args.total_seq_len,
+            hidden_size,
+            args.num_layers,
```
shanjiaz, line 107: Took another look, I think this is actually the number of layers for draft model but what we need would be number of hidden states layers going into the model here. I think we should switch to `draft_model.config.aux_hidden_state_layer_ids` for dflash and `draft_model.config.eagle_aux_hidden_state_layer_ids` for peagle and eagle. Maybe generalize it in a function? This is not the prettiest solution, but will do for now.

shanjiaz (changes requested):

fynnsu:
Closing since #597 has landed

PR #597, opened 2026-06-10
Fix empty batch

shanjiaz:
## Purpose
`create_empty_sample` is now hardcoding 3 layers and causing shape mismatches. We should use correct layer numbers populated by `resolve_target_layer_ids`, surfaced this function to train.py. This was originally PR [591](https://github.com/vllm-project/speculators/pull/591)

## Tests
Ran unit and integration tests.

## Checklist

I have filled in:


scripts/train.py
```diff
@@ -334,12 +340,19 @@ def main(args: argparse.Namespace):
...
+    # Resolve target layer IDs before model creation
+    target_layer_ids = resolve_target_layer_ids(
+        args.target_layer_ids, args.verifier_name_or_path
+    )
+    num_target_layers = len(target_layer_ids)
+
     if args.from_pretrained:
         draft_model = model_class.from_pretrained(
             args.from_pretrained, t2d=t2d, d2t=d2t
         )
```
fynnsu, line 345: I think this doesn't cover the case where `--from-pretrained` is used with a non-default num_target_layers.

i.e. Let's say `--from-pretrained` is used with a dflash checkpoint that uses 5 layers as input but the user doesn't bother passing target-layer-ids in. Then `resolve_target_layer_ids` will default to 3 layers and we'll get an error.

fynnsu: Maybe we can extract the selected value from the loaded model instead?

Like add:

```python
class DFlashDraftModel:
    ...
    @property
    def target_layer_ids(self):
         ...

```

In train.py
```python
target_layer_ids = draft_model.target_layer_ids
```

Adding a property or similar would help with the issue where different model types call the target layer ids different things in their config files.

fynnsu (changes requested):

scripts/train.py
shanjiaz: I see. Sure.

src/speculators/models/dflash/core.py
```diff
@@ -72,13 +72,6 @@ def __init__(
         self.uses_full_attn = bool(num_draft_layers - len(self.sliding_window_indices))
         self.sliding_window_non_causal = config.sliding_window_non_causal
 
-        if config.aux_hidden_state_layer_ids is None:
-            raise ValueError(
-                "aux_hidden_state_layer_ids must be set in DFlashSpeculatorConfig. "
-                "Use DFlashDraftModel.from_training_args() to resolve defaults."
-            )
```
fynnsu, line 79: We should keep this error, no?

fynnsu (approved):

src/speculators/models/dflash/core.py
shanjiaz: We could, but this would only happen if users create a faulty config.json.

PR #602, opened 2026-06-12
Fix e2e test failures from stale vocab mapping files in HF cache

rahul-tuli:
## Purpose

Fix e2e regression test failures caused by stale vocab mapping files (`d2t.npy`, `t2d.npy`) persisting in the shared HF cache on CI runners.

`train.py` caches generated vocab mappings into `data_path`, which on CI is a persistent HF snapshot directory (~218GB). When a later test or CI run uses the same dataset with a different `draft_vocab_size`, it picks up the stale files and fails with a `ValueError`. We can't delete or re-download the dataset between runs due to its size.

Adds a `purge_newfiles` context manager decorator that timestamps the directory after resolving the HF repo and deletes only top-level files with a newer mtime on exit. The cached dataset stays intact; only generated artifacts are removed.

Why this approach:
- **Preserves the ~218GB cached dataset** — only files created during the `with` block are removed
- **Timestamp-based** — no dependency on HF's internal cache format (symlinks vs copies)
- **Explicit cleanup boundary** — visible at the call site via `with` syntax
- **Reusable** — any `Path`-returning function can be decorated with `@purge_newfiles`

## Tests

- Existing tests pass, and delete any generated *.npy files in the requested dataset folder 
- [ ] Weekly regression tests (`test_eagle3_qwen3_8b_sharegpt`, `test_dflash_qwen3_8b_sharegpt`, `test_peagle_qwen3_8b_sharegpt`) pass after stale files are cleared from CI runner cache
- [ ] Log output shows "Purged generated artifact: d2t.npy" / "t2d.npy" after training
- [ ] HF-managed files (`.arrow`, `.json`, `.pt`) survive cleanup

## Checklist

I have filled in:

### DIFF
--- a/tests/e2e/regression/test_training_only_acceptance.py
+++ b/tests/e2e/regression/test_training_only_acceptance.py
@@ -4,10 +4,11 @@ import pytest
 from huggingface_hub import snapshot_download
 from huggingface_hub.utils import LocalEntryNotFoundError
 
-from tests.e2e.utils import run_training, run_vllm_engine
+from tests.e2e.utils import purge_newfiles, run_training, run_vllm_engine
 from tests.utils import requires_cadence
 
 
+@purge_newfiles
 def _resolve_repo(repo_id: str, repo_type: str = "dataset") -> Path:
     """Return a local Path for a repo, downloading from HuggingFace if needed.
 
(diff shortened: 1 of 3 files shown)


tests/e2e/regression/test_training_only_acceptance.py
```diff
@@ -4,10 +4,11 @@
 from huggingface_hub import snapshot_download
 from huggingface_hub.utils import LocalEntryNotFoundError
 
-from tests.e2e.utils import run_training, run_vllm_engine
+from tests.e2e.utils import purge_newfiles, run_training, run_vllm_engine
 from tests.utils import requires_cadence
 
 
+@purge_newfiles
```
rahul-tuli, line 11: Instead of always setting this on `_resolve_repo` we can also inline this at the call-site instead to make it more explicit;

That would look like
```python
with purge_newfiles(_resolve_repo(
        "inference-optimization/Qwen3-8b-sharegpt-5k"
    )) as hidden_states_dir
```

Let me know if that's prefered, I'm okay with both

tests/e2e/utils.py
```diff
@@ -29,6 +31,34 @@
...
+    @contextmanager
+    def wrapper(*args, **kwargs):
+        path = fn(*args, **kwargs)
+        cutoff = time.time()
+        try:
+            yield path
+        finally:
+            if path.is_dir():
+                for f in path.iterdir():
+                    if f.is_file() and f.stat().st_mtime > cutoff:
```
dsikka, line 55: why do we need a time cutoff?

dsikka (commented):
I think this is fine but I want to make sure the 5k samples we're pulling aren't deleted / are cached?

tests/e2e/utils.py
rahul-tuli: cutoff time is the simplest discriminator that gives us what we want, some other (not so good approaches would be) 
- hardcoded filenames (fragile if train.py starts writing other files), 
- symlinks (All files generated by HF are symlinks, generated files would be regular files, but this behavior can change)
- snapshotting the full directory listing (i.e keeping a record of all existing files, more state to carry around). One float does the job.

The cutoff works like this:
  1. _resolve_repo() downloads/resolves the dataset path
  2. We record time.time() — everything that exists at this point is "original"
  3. Training runs, train.py writes new files into that directory
  4. On exit, any file with mtime > cutoff that was created in the monitored folder during the with block, we consider them as "extra" or "generated files" and delete it

dsikka (approved):

shanjiaz (approved):

PR #603, opened 2026-06-12
feat(train): mid-epoch checkpoint save and resume

shubhra:
## Purpose
Previously, fractional checkpoints (checkpoint_freq < 1) were named by epoch index only (e.g. 0/, 1/) and resume always jumped to the *next* epoch, discarding any training progress within the interrupted epoch.

This change adds proper mid-epoch checkpoint state tracking so training can resume from the exact batch where it was interrupted:

- Save training_state.json alongside each checkpoint recording the current epoch, local_step within that epoch, and global_step. This allows the trainer to distinguish a mid-epoch save (local_step > 0) from an end-of-epoch save (local_step == 0).

- On resume, if training_state.json indicates a mid-epoch checkpoint, stay in the same epoch and fast-skip the already-processed batches by slicing the batch sampler's pre-generated batch list. This avoids calling __getitem__ (and thus vLLM) for skipped batches, making resume essentially instant regardless of how many batches are skipped.

- Create a human-readable epoch{N}_step{S} symlink pointing to the checkpoint folder whenever a mid-epoch checkpoint is saved, and an epoch{N}_end symlink for end-of-epoch checkpoints. The symlink is updated on each new checkpoint for that epoch so there is always exactly one readable pointer per epoch.

- Skip symlinks in BaseCheckpointer._get_previous_epoch so the new descriptive symlinks do not interfere with epoch detection.

The training_state.json is not written for interrupted checkpoints (epoch="interrupted") since those are emergency saves that require manual intervention to use.

## Tests
Ran the targeted resume regression test:
- Command:
  - `/home/shubhra/fp8_experiments/venv/bin/python test_mid_epoch_resume.py`
  
What this validates:
- `training_state.json` is written at fractional checkpoints with correct `epoch`, `local_step`, and `global_step`.
- Mid-epoch resume restores `current_epoch`, `_resume_local_step`, and `global_step` correctly.
- Fast-skip behavior works: already-processed batches are skipped without replay.
- Resume continues from the next unprocessed batch and keeps `global_step` continuity.
- Descriptive symlink creation/update works (e.g. `epoch0_step3`).

## Checklist

I have filled in:

- [ x] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [x ] The test plan/results, such as providing test command and pasting the results.
- [ x] I (a human) have written or reviewed the code in this pr to the best of my ability.

### DIFF
--- a/src/speculators/train/checkpointer.py
+++ b/src/speculators/train/checkpointer.py
@@ -105,6 +105,8 @@ class BaseCheckpointer:
             return -1
         last_checkpoint_num = -1
         for d in self.path.iterdir():
+            if d.is_symlink():
+                continue  # skip descriptive symlinks like epoch0_step16626
             if d.is_dir():
                 if d.name == "interrupted":
                     logger.warning(
--- a/src/speculators/train/trainer.py
+++ b/src/speculators/train/trainer.py
@@ -308,6 +308,9 @@ class Trainer:
         if hasattr(self.train_loader.batch_sampler, "set_epoch"):
             self.train_loader.batch_sampler.set_epoch(epoch)  # type: ignore[union-attr]
 
+        # Capture full-epoch step count before any resume fast-skip mutation.
+        num_steps = len(self.train_loader)
+
         # Determine how many batches to skip for mid-epoch resume.
         skip_steps = self._prepare_resume_skip(epoch)
 
@@ -315,7 +318,6 @@ class Trainer:
         if self.rank == 0:
             train_loader = tqdm(train_loader, desc=f"Epoch {epoch}")  # type: ignore[assignment]
 
-        num_steps = len(self.train_loader)
         step_interval = (
             max(1, round(num_steps * self.config.checkpoint_freq))
             if self.config.checkpoint_freq < 1
--- a/src/speculators/train/trainer.py
+++ b/src/speculators/train/trainer.py
@@ -107,8 +107,10 @@ class Trainer:
         if p.exists():
             try:
                 return json.loads(p.read_text())
-            except Exception:
-                pass
+            except json.JSONDecodeError as e:
+                root_logger.warning(f"Failed to decode training state {p}: {e}")
+            except (FileNotFoundError, PermissionError, OSError) as e:
+                root_logger.warning(f"Failed to read training state {p}: {e}")
         return {}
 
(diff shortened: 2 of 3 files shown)


src/speculators/train/checkpointer.py
shubhra, line 1: @coderabbitai review

tests/unit/train/test_mid_epoch_resume.py
```diff
@@ -0,0 +1,337 @@
...
+    cfg = TrainerConfig(
+        save_path=save_path,
+        num_epochs=epochs,
+        lr=1e-4,
+        local_rank=0,
+        is_distributed=False,
+        resume_from_checkpoint=resume,
+        checkpoint_freq=0.3,
+        log_freq=1,
+        scheduler_type="none",
```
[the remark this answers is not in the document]
shubhra: @coderabbitai review

src/speculators/train/trainer.py
shanjiaz: Could we move this before we call `_prepare_resume_skip` to get the correct `num_step`? Otherwise it might affect how we save checkpoints mid-epoch when calling
`num_steps - local_step >= step_interval * MIN_STEP_PCT`.

tests/unit/train/test_mid_epoch_resume.py
```diff
@@ -0,0 +1,223 @@
...
+                continue
+            trained_steps.append((epoch, local_step, self.global_step))
+            self.global_step += 1
+            if (
+                step_interval
+                and not self.config.save_best
+                and local_step % step_interval == 0
+                and num_steps - local_step >= step_interval * 0.1
+            ):
+                self.maybe_save_checkpoint(epoch, local_step=local_step)
```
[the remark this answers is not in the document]
shanjiaz: @shubhra I think this might be worth addressing. The mock trainers reimplement the skip/slice logic inline rather than calling the `_prepare_resume_skip` function. This means the tests verify that the mock's copy of the logic works, not the actual production code path.

src/speculators/train/trainer.py
shubhra: Updated

shanjiaz (approved):
Thanks for addressing the issues!

PR #607, opened 2026-06-15
Add Muon optimizer

eldarkurtic:
We recently found and validated that Muon optimizer performs much better than Adam in training of speculator models.
Validated on two models: decoder-only (Qwen3-8B) and an MoE (Laguna-XS.2). 
Training setup is the standard Magpie dataset, 50k samples for 1 epoch.
Validation setup: an extensive LR sweep for both Adam and Muon. 

Test example 1: 
a) Qwen3-8B, DFlash, LR sweep, metric=accuracy
image

b) Qwen3-8B, DFlash, LR sweep, metric=loss
image

Test example 2:
a) Laguna-XS.2, DFlash, LR sweep, metric=accuracy
image

b) Laguna-XS.2, DFlash, LR sweep, metric=loss
image

A few notes regarding the implementation:
- it supports single and multi-gpu training
- Muon optimizer can be used only for 2D weight matrices, so Adam is used for others
- this means we need to enable support for "a list of optimizers"

### DIFF
--- a/docs/cli/train.md
+++ b/docs/cli/train.md
@@ -108,6 +108,22 @@ torchrun --standalone --nproc_per_node=4 scripts/train.py \
 
 - **`--deterministic-cuda`** (flag) Enable deterministic CUDA operations. May impact performance.
 
+### Optimizer Arguments
+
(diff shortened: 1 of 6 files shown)


rahul-tuli (approved):
LGTM!

fynnsu (changes requested):
Blocking till e2e is done:
https://github.com/neuralmagic/llm-compressor-testing/actions/runs/27772187815

Just want to make sure the switch to multi-optimizers doesn't break anything.

fynnsu (approved):
E2E passed

PR #615, opened 2026-06-17
raise error for nan hidden states

shanjiaz:
## Purpose
We had a failure early this week due to hiddenstatesconnector sending nan hidden states. However, our training kept going. We should explicitly fair when getting garbage hidden states

## Tests
Tested locally and ran the e2e smoke tests.

Now failes directly:
``` bash
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/shanjiaz/speculators/src/speculators/train/data.py", line 327, in _maybe_generate_hs
    check_hidden_states(loaded_hs, dataset_item["input_ids"].tolist())
  File "/home/shanjiaz/speculators/src/speculators/data_generation/offline.py", line 17, in check_hidden_states
    raise ValueError("Hidden states contain NaN values")
ValueError: Hidden states contain NaN values
================================================== 1 failed, 16 warnings in 180.86s (0:03:00) ==================================================
```

## Checklist

I have filled in:

### DIFF
--- a/src/speculators/train/data.py
+++ b/src/speculators/train/data.py
@@ -16,6 +16,7 @@ from datasets import load_from_disk
 from safetensors.torch import load_file
 from torch.utils.data import Dataset
 
+from speculators.data_generation.offline import check_safetensors_file
 from speculators.data_generation.vllm_client import (
     DEFAULT_MAX_RETRIES,
     DEFAULT_REQUEST_TIMEOUT,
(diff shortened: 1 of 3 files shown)


src/speculators/train/data.py
```diff
@@ -323,14 +324,21 @@ def _maybe_generate_hs(self, index: int) -> dict[str, torch.Tensor] | None:
 
             loaded_hs = _maybe_load_hs_file(Path(hs_filepath))
 
+            check_safetensors_file(
+                Path(hs_filepath),
+                dataset_item["input_ids"].tolist(),
+            )
+
```
fynnsu, line 330: So we're actually already loading this file in the line above. And then this function loads and checks the status. I think we should update the check function so that it can take in the loaded_hs data itself. That way we don't load twice in data.py but can still re-use the check function in data_generation_offline.py.

fynnsu (changes requested):
One comment, otherwise looks good.

src/speculators/train/data.py
shanjiaz: Sure

fynnsu (approved):

PR #617, opened 2026-06-18
Add DFlash model converter

### LINKED ISSUE #616: [RFC]: Add DFlash model converter
### Motivation.

The Speculators library supports finetuning pretrained speculator checkpoints via `--from-pretrained` arg in `train.py`. However, this only works for checkpoints already in the Speculators format (`DFlashSpeculatorConfig` + safetensors). There is no way to convert an externally-trained DFlash checkpoint into the Speculators format.

### Proposed Change.

Address the TODO in dflash/model,py, "once conversion is added, need to handle the case where a non speculator config is passed in as a kwarg and auto convert." When `from_pretrained` loads a checkpoint whose config.json is not a SpeculatorModelConfig, instead of raising TypeError, it will attempt auto-conversion:

  1. In `SpeculatorModel.from_pretrained()`, when the loaded config is not a SpeculatorModelConfig, identify if it's an external DFlash checkpoint (e.g., by inspecting config fields or weight key patterns).
  2. Invoke a `DFlashConverter` that:
    - Remaps external weight names to the DFlashDraftModel state dict layout (layers, fc, norm, hidden_norm, etc.)
    - Builds a DFlashSpeculatorConfig from the source config.json + verifier model config, including DFlash-specific fields (block_size, max_anchors, mask_token_id, aux_hidden_state_layer_ids, sliding_window_non_causal)
    - Handles t2d/d2t vocab mapping tensors if present
    - Saves the converted checkpoint to a cache/temp directory
  3. Load converted checkpoint — Continue the normal from_pretrained flow with the converted config and weights, returning a ready-to-use DFlashDraftModel.

### Any Other Things.

Optionally update the pathway for EAGLE3 as well. We should unify the behavior for both EAGLE3 and DFlash. Users should be able to use `--from-pretrained` to finetune EAGLE3 models.

guan404ming:
## Purpose

Close #616

External DFlash checkpoints (e.g. z-lab/*-DFlash) had no path into the Speculators format, so they could not be loaded or finetuned via from_pretrained.

## Tests

- add mocked _build_config unit tests
- test on a real checkpoint: converted `z-lab/Qwen3-8B-DFlash-b16` with verifier `Qwen/Qwen3-8B`, reloaded via `from_pretrained` with weights bit-identical to source

## Checklist

I have filled in:

### DIFF
--- a/src/speculators/convert/dflash/converter.py
+++ b/src/speculators/convert/dflash/converter.py
@@ -1,0 +1,164 @@
+"""DFlash checkpoint converter.
+
+Converts an external DFlash checkpoint (e.g. ``z-lab/*-DFlash``) to a Speculators
+checkpoint that loads with ``DFlashDraftModel.from_pretrained(path)``.
+
+The draft transformer body (``layers.*``, ``fc``, ``hidden_norm``, ``norm``) already
+matches ``DFlashDraftModel`` so weights are copied as-is. The external checkpoint
+borrows the verifier's embedding and LM head at runtime, so ``embed_tokens`` /
+``lm_head`` / ``verifier_lm_head`` / ``verifier_norm`` are loaded from the verifier
+before saving.
+"""
+
+from pathlib import Path
+
+import torch
+from loguru import logger
+from transformers import PretrainedConfig
+
+from speculators.config import SpeculatorsConfig, VerifierConfig
+from speculators.convert.utils import (
+    ensure_checkpoint_is_local,
+    load_checkpoint_config,
+    load_checkpoint_weights,
+)
+from speculators.models.dflash import DFlashDraftModel, DFlashSpeculatorConfig
+from speculators.proposals.greedy import GreedyTokenProposalConfig
+
+__all__ = ["DFlashConverter"]
+
+# config.json keys that are not part of the draft transformer (Qwen3) config
+_NON_TRANSFORMER_KEYS = frozenset(
+    {"architectures", "auto_map", "block_size", "dflash_config", "num_target_layers"}
+)
+
+
+class DFlashConverter:
+    """Convert an external DFlash checkpoint to speculators format.
+
+    Copies the draft transformer body as-is and fills the embedding, LM head, and
+    verifier norm from the verifier model so the saved checkpoint is self-contained.
+    """
+
+    def convert(
+        self,
+        input_path: str | Path,
+        output_path: str | Path,
+        base_model: str,
+        validate: bool = True,
+        aux_hidden_state_layer_ids: list[int] | None = None,
+        cache_dir: str | Path | None = None,
+    ) -> None:
+        logger.info(f"Converting DFlash checkpoint: {input_path}")
+
+        local_checkpoint_path = ensure_checkpoint_is_local(input_path, cache_dir)
+        source_config = load_checkpoint_config(local_checkpoint_path)
+        weights = load_checkpoint_weights(local_checkpoint_path)
+        logger.info(f"Loaded {len(weights)} weights")
+
+        config = self._build_config(
+            source_config, base_model, aux_hidden_state_layer_ids
+        )
+        saved_path = self._save(config, weights, output_path)
+        logger.success(f"Saved to: {saved_path}")
+
+        if validate:
+            self._validate(saved_path)
+
+    def _build_config(
+        self,
+        source_config: dict,
+        base_model: str,
+        aux_hidden_state_layer_ids: list[int] | None,
+    ) -> DFlashSpeculatorConfig:
+        dflash = source_config.get("dflash_config", {})
+        transformer_config = {
+            k: v for k, v in source_config.items() if k not in _NON_TRANSFORMER_KEYS
+        }
+
+        verifier_config_dict, _ = PretrainedConfig.get_config_dict(base_model)
+        source_hidden = transformer_config.get("hidden_size")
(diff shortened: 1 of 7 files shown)


src/speculators/convert/dflash/converter.py
```diff
@@ -0,0 +1,164 @@
...
+        if unexpected:
+            raise ValueError(
+                "Unexpected keys in checkpoint -- the structure does not match "
+                f"DFlashDraftModel. Unexpected keys: {unexpected}"
+            )
+        critical_missing = [k for k in missing if k.startswith("layers.")]
+        if critical_missing:
+            raise ValueError(
+                f"Draft layer weights missing after load: {critical_missing}"
+            )
```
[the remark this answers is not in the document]
guan404ming: Fixed: now flags any weight not filled from the verifier, so fc/norm/hidden_norm missing also raise, not just layers.*.

src/speculators/convert/dflash/converter.py
```diff
@@ -0,0 +1,164 @@
...
+        logger.debug(f"Keys loaded from verifier at save time: {missing}")
+
+        # embed_tokens / lm_head / verifier_lm_head / verifier_norm come from the
+        # verifier; without this they would be saved as NaN.
+        model.load_verifier_weights()
+
+        model.to(dtype=next(iter(body.values())).dtype)  # type: ignore[call-arg]
```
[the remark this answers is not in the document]
guan404ming: The fix above makes this unreachable an empty/incomplete body now raises "Draft weights missing" before this line. Added a _save test covering it.

src/speculators/convert/dflash/converter.py
```diff
@@ -0,0 +1,175 @@
...
+            target_layer_ids = dflash.get("target_layer_ids")
+            if target_layer_ids is None:
+                raise ValueError(
+                    "Checkpoint config has no `dflash_config.target_layer_ids`; "
+                    "pass `aux_hidden_state_layer_ids` explicitly."
+                )
+            # z-lab reads hidden_states[layer_id + 1] (index 0 is the embedding
+            # output) while speculators uses the layer id directly.
+            # Source: z-lab utils.extract_context_feature.
+            aux_hidden_state_layer_ids = [i + 1 for i in target_layer_ids]
```
shanjiaz, line 112: good catch!

src/speculators/convert/dflash/converter.py
```diff
@@ -0,0 +1,175 @@
...
+
+        speculators_config = SpeculatorsConfig(
+            algorithm="dflash",
+            proposal_methods=[
+                GreedyTokenProposalConfig(
+                    speculative_tokens=source_config["block_size"] - 1,
```
shanjiaz, line 118: Nice!

src/speculators/convert/dflash/converter.py
```diff
@@ -0,0 +1,175 @@
...
+                f"DFlashDraftModel. Unexpected keys: {unexpected}"
+            )
+        critical_missing = [k for k in missing if k not in _VERIFIER_FILLED_KEYS]
+        if critical_missing:
+            raise ValueError(f"Draft weights missing after load: {critical_missing}")
+        logger.debug(f"Keys loaded from verifier at save time: {missing}")
+
+        # embed_tokens / lm_head / verifier_lm_head / verifier_norm come from the
+        # verifier; without this they would be saved as NaN.
+        model.load_verifier_weights()
```
shanjiaz, line 159: Could we look into replacing this with `load_model_layers` in `src/speculators/utils/loading.py` to avoid having to load the entire verifier model?

src/speculators/model.py
```diff
@@ -315,8 +328,6 @@ def from_pretrained(
             )
 
         if not isinstance(config, SpeculatorModelConfig):
-            # once conversion is added, need to handle the case where a non speculator
-            # config is passed in as a kwarg and auto convert
             raise TypeError(
```
shanjiaz, line 336: A few suggestions on the auto-conversion flow:
  - Move format detection here. Load the config dict once, check if it's speculators format, and only then call the converter. Right now we're checking format in two places (isinstance in from_pretrained and "speculators_model_type" in config_dict in maybe_convert_external_checkpoint).
  - Have the converter return the config and model directly instead of writing to a temp dir and re-reading.
What do you think?

src/speculators/convert/dflash/converter.py
guan404ming: I think that `load_verifier_weights` already uses `load_model_layers` (only embed_tokens/lm_head/norm shards). No full verifier load. Please feel free correct me if I am wrong, thanks!

src/speculators/model.py
guan404ming: Done. Gating moved into from_pretrained (load config dict once, convert only if not speculators). The isinstance check is type-validation of the config arg, not format detection, so it stays. 

I currently kept the temp-dir round-trip so both formats share one `super().from_pretrained` path (device_map/dtype/safetensors kwargs). I think if returning a live model would fork loading for one disk write+read.

src/speculators/convert/dflash/converter.py
shanjiaz: Nice!

src/speculators/model.py
shanjiaz: Sure! Reasonable.

shanjiaz (approved):

guan404ming:
Thanks @shanjiaz !

Issue #569, opened 2026-06-03
[RFC]: Explicit draft-model config contract for training and vLLM launch


imargulis:
## Motivation.
Today the draft decoder architecture is derived implicitly from the verifier, so there is no supported way to train a draft whose decoder differs from the target (e.g. a Qwen3-style DFlash draft for a Qwen3 verifier). Several sharp edges cause silent, hard-to-debug failures:

- **MoE verifiers have no sensible default draft width**, so training picks a wrong dense width or fails obscurely.
- **`--target-layer-ids` means different things** in `launch_vllm.py` vs `train.py` (auxiliary-only vs auxiliary + verifier-final). Worse, published configs disagree by an off-by-one: DFlash stores `dflash_config.target_layer_ids`, while Eagle/speculator configs store `aux_hidden_state_layer_ids` / `eagle_aux_hidden_state_layer_ids` one index higher. A mismatch silently corrupts training targets — the last auxiliary layer gets consumed as the verifier-final hidden state.
- **Nothing validates** that the hidden-state files a run consumes contain the expected number of layers, so a wrong `--target-layer-ids` (or a `--no-include-last-layer` launch) fails deep in training, if at all.

## Proposed Change.
Introduce one "draft-config contract" shared by training and serving:

- Add `--draft-config` (HF model id or local config path) to `train.py` and `launch_vllm.py`. Draft decoder parameters (intermediate size, head dim, attention heads, hidden layers) are read from it; explicit CLI flags (`--num-layers`, `--draft-hidden-act`, …) then override. The draft hidden size must match the verifier (the mismatch case is deferred — see below).
- Define the default draft width when `--draft-config` is absent: dense verifiers reuse `intermediate_size`; MoE verifiers infer a dense width from active-expert capacity (`num_experts_per_tok * moe_intermediate_size + shared_expert_intermediate_size`); otherwise `--draft-config` is required, with a clear error.
- Establish a **single canonical "capture" convention** for auxiliary layer ids — 0-based verifier decoder-layer indices, excluding the verifier-final layer. `aux_hidden_state_layer_ids` and `eagle_aux_hidden_state_layer_ids` are already canonical; `dflash_config.target_layer_ids` is `canonical - 1`, matching vLLM's speculators loader (vllm-project/vllm#40727). Resolution converts every present field to canonical *before* comparing, so the two real-world config flavors are treated as equivalent and only genuine disagreements warn.
[...]

PR #618, opened 2026-06-18
feat(train): add cli options to extend --from-pretrained for config-only init

imargulis:
## Summary
Adds two ways to define a draft model up front and a way to validate it before training:

- **`--draft-config <hf id | dir | json>`** — load just the decoder
  `transformer_layer_config` (LlamaConfig for eagle3/peagle, Qwen3Config for
  dflash); the rest of the speculator is built from the other CLI args. The draft
  `hidden_size` must match the verifier (hard error); `vocab_size` is aligned to
  the verifier (use `--draft-vocab-size` for pruning).
- **`--from-pretrained` extended** — in addition to a full checkpoint, it now
  accepts a directory containing only a `config.json` (no weights), in which case
  a fresh draft is initialized from that full speculator config.
- **`--dry-run`** — build the speculator, initialize weights, save a checkpoint to
  `--save-path`, and exit before training. The output can be validated (e.g. in
  vLLM) and fed straight back via `--from-pretrained`. Combinable with the above.

### Draft-init contract
The draft is defined in exactly one way: `--from-pretrained` (takes precedence) **or**
`--draft-config` **or** the decoder-shaping flags (`--num-layers`, `--draft-arch`,
`--draft-hidden-act`, `--sliding-window`, `--sliding-window-indices`). Conflicting
combinations are rejected at parse time. Detection is provided-based (an explicitly
passed flag conflicts even at its default value).

## Fixes folded in (surfaced while implementing/validating)
- **Verifier `architectures` propagation:** new `VerifierConfig.from_pretrained`
  reads the verifier's real `config.json`, so eagle3/dflash/peagle **and** MTP now
  record `verifier.architectures` instead of an empty list.
- **DFlash load symmetry:** `verifier_lm_head.weight` added to
  `_keys_to_ignore_on_load_missing` (excluded on save, reloaded from the verifier),
  fixing a "missing verifier lm_head" error on reload; invariant test added.
- **`--dry-run` dtype:** the saved checkpoint is cast to `--hidden-states-dtype`.
- **Draft `intermediate_size` resolution:** the *synthesized draft decoder's* MLP
  `intermediate_size` is copied from the verifier when the verifier is dense; MoE
  verifiers expose no single dense `intermediate_size`, so the draft falls back to
  `3 × hidden_size` (the Qwen3 dense / gated-MLP ratio the dflash draft already uses).
- **Vocab-mapping no-op:** `load_vocab_mappings` ignores full-vocab t2d/d2t when
  `draft_vocab_size == verifier_vocab_size` instead of raising.
- **`--draft-arch`** now treats both `llama` and `qwen3` as vLLM-supported.

## Backwards compatibility
Default behavior is unchanged: with none of the new flags, training synthesizes the
draft from the verifier + CLI flags exactly as before.

## Test plan
- [ ] Unit suite green (`tests/unit/train/test_draft_config_init.py`,
  `test_utils.py`, `test_argparse_utils.py`, `test_config.py`, `test_model.py`,
  `test_setup_model.py`, `test_loading.py`).
- [ ] `--dry-run` init → reload via `--from-pretrained` round-trips (dense + MoE).
- [ ] Config-only `--from-pretrained` initializes fresh weights.
- [ ] Conflicting flag combinations error at parse time.

### DIFF
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -686,6 +686,9 @@ def validate_draft_init_args(
     options: it is mutually exclusive with ``--draft-config`` and with the
     decoder-shaping flags, since those values come from the checkpoint.
     ``--draft-config`` is likewise incompatible with the decoder-shaping flags.
+    MTP from scratch (``--speculator-type mtp`` without ``--from-pretrained``)
+    reuses the verifier's own decoder config, so ``--draft-config`` and the
+    decoder-shaping flags do not apply and are rejected.
 
     ``provided`` is the set of decoder-shaping dests the user explicitly passed
     (see :func:`speculators.utils.argparse_utils.explicitly_provided_dests`); a flag
@@ -698,15 +701,24 @@ def validate_draft_init_args(
             parser.error(
                 "--from-pretrained loads a complete draft model and takes precedence "
                 "over all other model-definition options, so these conflict with it "
-                f"(remove them): "
-                f"{', '.join(conflicting)}"
+                f"(remove them): {', '.join(conflicting)}"
+            )
+        return
+    if args.speculator_type == "mtp":
+        # MTP-from-scratch reuses the verifier's own decoder config and extracts the
+        # native MTP head weights; --draft-config and the decoder-shaping flags do not
+        # apply, so reject them rather than silently ignoring them.
+        conflicting = shaping + (["--draft-config"] if args.draft_config else [])
+        if conflicting:
+            parser.error(
+                "--speculator-type mtp reuses the verifier's decoder config, so these "
+                f"options do not apply (remove them): {', '.join(conflicting)}"
             )
         return
     if args.draft_config and shaping:
         parser.error(
             "--draft-config defines the draft decoder, so these flags conflict with "
-            "it (remove them): "
-            f"{', '.join(shaping)}"
+            f"it (remove them): {', '.join(shaping)}"
         )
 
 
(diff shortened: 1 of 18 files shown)


imargulis:
Addressing [RFC#569](https://github.com/vllm-project/speculators/issues/569).

imargulis:
@coderabbitai review

imargulis:
@coderabbitai review

scripts/train.py
```diff
@@ -211,6 +217,52 @@ def create_transformer_layer_config(  # noqa: C901
...
+    is controlled separately via ``--draft-vocab-size``.
+    """
+    config_dict, _ = PretrainedConfig.get_config_dict(draft_config)
+    if "transformer_layer_config" in config_dict:
+        # A full speculator config was passed; use only the decoder definition.
+        config_dict = config_dict["transformer_layer_config"]
+
+    model_type = config_dict.get("model_type")
+    config_class: type[PretrainedConfig] = (
+        type(AutoConfig.for_model(model_type)) if model_type else Qwen3Config
```
fynnsu: I'd prefer to fail loudly here if the model_type is not provided.

fynnsu (approved):
Thanks for the pr! This looks good, just one small comment and please resolve failing quality tests.

fynnsu:
The buildkite failure is a known flake (unrelated)

scripts/train.py
imargulis: Agree. Done.

PR #626, opened 2026-06-20
[Features] Add Qwen3-VL Multimodal Training Support with vLLM 0.20

shx2005:
## Purpose

Add end-to-end Qwen3-VL image-text Eagle3 training support with vLLM chat-based hidden-state generation.

## Description

This PR adds a full multimodal image-text training path for Qwen3-VL speculators:

- prepares image-text datasets with `prepare_data.py --multimodal`
- expands image placeholders to the processor's actual visual-token length
- preserves normalized `messages` for vLLM chat-based hidden-state generation
- supports both offline datagen and online `--on-missing generate` training
- adds runnable Qwen3-VL LLaVA-CoT 5k and 48k training examples

Key implementation details:

- text-only samples stay on the existing token-id Completions path
- real multimodal samples use Chat Completions with OpenAI-compatible image content
- local images are sent as `file://` media URLs
- returned prompt token IDs are checked against preprocessed `input_ids`
- hidden-state files are consumed only after vLLM lock release and then aligned to the preprocessed sequence length

Validation-driven fixes:

- avoid routing text-only list content through chat retokenization
- reject truncation in the middle of expanded image token blocks
- keep checkpoint loading stable by avoiding target-layer mutation in `from_pretrained`
- avoid unnecessary offline safetensors reads when output validation is disabled

## Related Issue

https://github.com/vllm-project/speculators/issues/290

## Tests

Following `CONTRIBUTING.md`, validation was split by quality, unit, integration, e2e, and example-script checks.

Passed on current clean remote validation source:

- `make quality`
- `python -m pytest ...` targeted regression subset: `90 passed`
  - covers multimodal preprocessing, vLLM chat/message requests, hidden-state alignment, empty-batch dtype, target-layer handling, Pydantic config registration, and offline datagen no-load move regression
- `python -m pytest ...` prefix-cache launcher regression subset: `4 passed`
- `python -m pytest ...` multimodal offline e2e smoke: `1 passed`
- `python -m pytest ...` multimodal online `--on-missing generate` e2e smoke: `1 passed`
- `python -m pytest ...` resume regression e2e smoke: `1 passed`
Targeted coverage includes multimodal preprocessing, vLLM chat/message requests, hidden-state alignment, empty-batch dtype, target-layer handling, Pydantic config registration, and offline datagen no-load move regression.

Previously completed E2E validation:

- `bash examples/train/eagle3_qwen3_vl_4b_llava_cot_5k_online.sh`
- `bash examples/train/eagle3_qwen3_8b_sharegpt_online_5k.sh`

5k Qwen3-VL online run:

- 3x RTX 5090 32GB
- vLLM on GPUs 0,1; training on GPU 2
- total runtime: 4054 seconds
- `val/loss_epoch=7.803`
- `val/full_acc_0_epoch=64.2%`
- `val/full_acc_1_epoch=39.2%`
- `val/full_acc_2_epoch=23.9%`

48k Qwen3-VL online run, 1 epoch:

- `val/loss_epoch=6.894`
- `val/full_acc_0_epoch=65.5%`
- `val/full_acc_1_epoch=40.6%`
- `val/full_acc_2_epoch=25.3%`

Not counted as passed in the current clean remote validation:

- full unit suite: external HF test-resource failure; `nm-testing/testing-llama3.1.8b-2layer-eagle3` has config but no weights
- full integration suite: isolated HF network timeout while resolving `openai/gpt-oss-20b`

### DIFF
--- a/src/speculators/data_generation/preprocessing.py
+++ b/src/speculators/data_generation/preprocessing.py
@@ -2,7 +2,8 @@ import json
 from collections.abc import Callable
 from contextlib import nullcontext
 from pathlib import Path
-from typing import Literal, TypedDict
+from typing import Any, Literal, TypedDict, cast
+from urllib.parse import unquote, urlparse
 
 import torch
 from datasets import Dataset as HFDataset
@@ -15,6 +16,7 @@ from transformers import (
 
 from speculators.data_generation.configs import DATASET_CONFIGS
 from speculators.data_generation.logging_utils import PipelineLogger
+from speculators.data_generation.media import get_image_ref
 from speculators.data_generation.render_client import render_conversation
 from speculators.data_generation.torch_utils import set_default_torch_num_threads
 from speculators.train.vocab_mapping import save_token_frequency_distribution
@@ -119,6 +121,30 @@ def _normalize_conversation(
     return normalized
 
 
+def _get_conversations_from_examples(examples: dict) -> list:
+    """Extract conversations/messages from a batched dataset example."""
+    if "conversations" in examples:
+        return examples.get("conversations", [])
+    if "messages" in examples:
+        return examples.get("messages", [])
+    return []
+
+
+def _local_media_path_to_uri(media_path: str | Path) -> str:
+    """Return a canonical, percent-encoded file URI for a local media path."""
+    media_text = str(media_path)
+    parsed = urlparse(media_text)
+    if parsed.scheme == "file":
+        if parsed.netloc not in ("", "localhost"):
+            raise ValueError(f"Unsupported file URI host: {parsed.netloc}")
+        if parsed.query or parsed.fragment:
+            raise ValueError("Local file URIs must not contain query/fragment")
+        media_text = unquote(parsed.path)
+    elif parsed.scheme:
+        raise ValueError(f"Unsupported local media URI scheme: {parsed.scheme}")
+    return Path(media_text).expanduser().resolve().as_uri()
+
+
 def _adapt_part_for_vllm(part: str | dict):
     if isinstance(part, str):
         return {"type": "text", "text": part}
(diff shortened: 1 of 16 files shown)


shx2005:
@coderabbitai review

shx2005:
I noticed that the latest upstream changes include several larger training/model refactors, so I am rebasing this PR again and rerunning the validation suite against the current `main`.

I have marked the PR as draft while I work through the rebase conflicts and test results. I will move it back to ready for review once the branch is updated and the relevant tests pass.

shx2005:
Quick update: I rebased the PR onto the latest main branch and reran both the ShareGPT text workflow and the 5k VL workflow. Both completed successfully, confirming that the existing text functionality remains intact and the VL path works as expected.

shx2005:
Final results!
  image

src/speculators/data_generation/preprocessing.py
```diff
@@ -582,12 +1065,28 @@ def _preprocess_batch(
...
+            AttributeError,
+            RuntimeError,
+            OSError,
+        ) as e:
+            raise RuntimeError(
                 f"Failed to process conversation {idx} "
                 f"(assistant_pattern={assistant_pattern is not None}): {e}"
-            )
-            continue
+            ) from e
```
speculatorsbot: This changes `_preprocess_batch` error handling from log-and-skip (`log.error(...); continue`) to raise-and-crash. Since `_preprocess_batch` runs inside `dataset.map(batched=True)`, the old behavior silently filtered out bad samples within a batch, while the new behavior aborts the entire `map()` operation on the first failure.

The fail-fast approach is reasonable for multimodal preprocessing (catching missing images, broken URIs, etc. early), but this also affects the existing **text-only path** — a dataset with a few malformed conversations that previously preprocessed successfully (skipping the bad ones) would now crash.

Is this intentional? If so, it may be worth calling out in the PR description since it changes behavior for existing text-only users. If not, one option is to keep `raise` for the multimodal `ProcessorMixin` path but preserve the old `log.error + continue` for text-only tokenizer-based preprocessing.

speculatorsbot (commented):
Well-structured PR that cleanly separates text-only and multimodal paths. The loss-mask alignment logic in `_align_loss_mask_to_token_ids` is complex but correct — the resync algorithm properly preserves trainable-token supervision while tolerating vision-token expansion differences between HF preprocessing and vLLM re-tokenization. Test coverage is thorough.

One behavioral question below; otherwise recommend approving once addressed or explicitly acknowledged.

src/speculators/data_generation/preprocessing.py
shx2005: Fixed. I updated the exception handling so ProcessorMixin / multimodal preprocessing remains fail-fast, but tokenizer-based text-only preprocessing preserves the previous log-and-skip behavior inside the batch.

speculatorsbot (commented):
Previous concern about error-handling behavior change for text-only paths has been addressed — multimodal `ProcessorMixin` preprocessing now fails fast while text-only tokenizer-based preprocessing preserves the original log-and-skip behavior. LGTM on the new commit. Recommend approving (pending rebase noted in the `needs-rebase` label).

Issue #621, opened 2026-06-18
[Bug]: MTP doesn't separate sequences in the same batch when running attention


fynnsu:
### Your current environment
N/A

### 🐛 Describe the bug

The other spec decoding algorithms keep track of the lengths of the sequences that make up a batch and use that to produce document ids. These documents ids are then used in the attention mask creation to ensure each token only attends to other tokens from the same sequence.

MTP doesn't use the `lengths` tensor (soon to be replaced with `document_ids` tensor) when constructing its causal attention mask. This means later sequences in the batch will be able to attend to tokens from previous unrelated sequences.

e.g. 
```
# for a batch with these sequences [1 3 2]
# current attention mask

1 0 0 0 0 0
1 1 0 0 0 0
1 1 1 0 0 0
1 1 1 1 0 0
1 1 1 1 1 0
1 1 1 1 1 1

# target attention mask

1 0 0 0 0 0
0 1 0 0 0 0
0 1 1 0 0 0
0 1 1 1 0 0
0 0 0 0 1 0
0 0 0 0 1 1
```

We'll need to fix this + verify that other parts of the mtp `forward` correctly handle the sequence boundaries.

WindChimeRan:
**Edit: I think this is also a false alarm**, loss_mask implicitly separate chat-only data. 

Original comments: 
> Beyond attention, the same packing/boundary root cause also hits supervision: the per-step slicing pulls each step's teacher-forced input (step+1) and target (step+2) across document boundaries, which the attention mask can't fix — the same document_ids should mask those boundary targets.


https://github.com/vllm-project/speculators/blob/f8f53f4178560170bb63366764b09c0d3dac4c56/src/speculators/models/mtp/core.py#L205-L216 


**Why supervision side is a false alarm.**`loss_mask` already neutralizes it: the spilled targets land at the start of the next packed conversation, i.e. its chat-template prefix, which is masked (loss is assistant-only). Since prepare_data only emits chat-templated, assistant-masked data, every packed document begins with a masked prefix of ~10+ tokens (≫ num_speculative_steps + 1), so every spilled target is already IGNORE. It would only bite a document whose masked prefix is shorter than num_steps + 1 (response-only / full-sequence-loss data), which the pipeline can't produce — so no change needed unless a non-chat training path is added.

WindChimeRan:
For this issue specifically: 

I think MTP training attention is already document-local, through `position_ids`, not `document_ids`.

[...]

PR #627, opened 2026-06-21
test(mtp): pin packed-batch attention to document-local (#621)

WindChimeRan:
## Purpose

Regression test for https://github.com/vllm-project/speculators/issues/621. Full investigation in the issue thread.

MTP builds its mask on the dense `create_causal_mask` path, where transformers auto-derives the block-diagonal packed mask from the per-document `position_ids` the collate emits (positions reset per document, attention_mask=None). So MTP training attention is already document-local — implicitly via position_ids, not via an explicit document_ids mask.

This test pins that invariant: a packed batch with per-document position_ids must yield a causal and same-document mask in the MTP forward. It fails if positions stop carrying segment boundaries or packed-mask detection changes — the silent failure mode where loss still decreases.

## Tests

This PR is test only. 

## Checklist

I have filled in:


WindChimeRan:
@coderabbitai review

shanjiaz (approved):

PR #629, opened 2026-06-22
fix: prevent cross-document attention in MTP

### LINKED ISSUE #621: [Bug]: MTP doesn't separate sequences in the same batch when running attention

shotsan:
Fixes #621. MTP's causal_mask did not account for document boundaries, allowing later sequences in the batch to attend to previous unrelated sequences. This PR adds masking based on `document_ids` passed in kwargs to block cross-document attention.


WindChimeRan:
I think this is a no-issue? Can you check 

https://github.com/vllm-project/speculators/issues/621#issuecomment-4760631442

PR #632, opened 2026-06-22
feat: Response-regeneration refactor (RFC #584)

### LINKED ISSUE #584: [RFC]: Response-regeneration refactor — resumable, multi-split, quality-filtered training data (related to RFC#583)
### Motivation.

`scripts/response_regeneration/script.py` regenerates assistant responses in existing
datasets against a vLLM-served model, producing the conversation JSONL that
`prepare_data.py` turns into draft-training data. As the datasets and models grew, the
original tool hit several limitations:

1. **Resume is unreliable.** Resume is keyed on a per-row `id` that is the source UUID
   *or* a positional `sample_{idx}` fallback. Index-based ids shift whenever the source is
   re-ordered, filtered, or re-split, so a crashed multi-hour job could not be trusted to
   skip exactly what it had already written — it could re-do or, worse, skip the wrong rows.
   There is no single stable identity across the different source schemas the pipeline is proposed to
   ingests.

2. **One split, one file.** Multi-split datasets (e.g. `nemotron` with `stem`/`code`/`chat`/
   `math`) have no first-class handling, and there is no way to fan results out into a tidy
   per-split bundle.

3. **Rows with no utility for training.** On a failed generation it still emitts a
   *user-only* conversation (no assistant turn), and it writes responses that hit `--max-tokens`
   (`finish_reason="length"`, i.e. likely truncated) with no way to filter them. Both pollute
   the downstream SFT/draft data.

4. **Single-turn only.** There is no way to regenerate *multi-turn* assistant responses by
   walking a conversation and regenerating each assistant turn from the regenerated prefix.

5. **Fragile requests, thin observability.** vLLM calls have no bounded retry/backoff/timeout
   policy, and the run gives little accounting of what happened to each row.

### Proposed Change.

Refactor the regenerator around three goals — **safe resume**, **output quality**, and
**multi-split awareness** . The user-facing contract
changes are:

- **Stable identity + resume.** Each output row carries `metadata.primary_id`, chosen
  deterministically from the source (`id` / `uuid` / selected metadata fields, else a stable
  content hash of the source row) with a `primary_id_source` recording which. `--resume` skips
  rows already present **by `primary_id`**, and still understands the legacy key so outputs
  from the old tool can be continued. A `secondary_id` (currently the row index) is kept for
  monitoring only.

- **Output modes.** `--output-mode single` writes one JSONL (`--outfile`, or an auto-named
  file under `--output-dir`); `--output-mode bundle` writes one
  `{dataset}_{split}_{model}.jsonl` per split under a directory.

- **Turn modes.** `--turn-mode single` (default) regenerates the first turn;
  `--turn-mode multi` iteratively regenerates assistant turns from the regenerated
  conversation prefix and stops at the first undesirable generation.

- **Quality filtering.** Rows that never produce a valid assistant response are **dropped**
  (no user-only conversations leak into the training data). Length-finished responses are
  **dropped by default** because they may be truncated; `--keep-length-finished` keeps them
  (and, in multi-turn mode, writes the prefix and stops after the length-finished turn).

- **Multi-split presets.** Presets that expose multiple splits process each one; `--limit`
  counts **new** rows queued *per split* (rows skipped by resume, language filtering, or
  source validation don't count against it).

- **Robustness + accounting.** `--max-retries` / `--retry-backoff` / `--request-timeout` bound
  every vLLM request. Each sample logs exactly one terminal outcome (`skipping`,
  `new_written`, `new_written_partial`, `new_skipped`), and each split plus the whole run
  print reconciled counters where `processed = pre_existing + written + dropped`.

### Any Other Things.

- **drop length-finished by default.** Dropping by default is the safe choice for general training, but can be flipped in case of speculators.

shotsan:
Closes #584.

This PR refactors the response regeneration script to be robust and production-ready.

**Changes:**
- **Stable identity & resume**: Uses a stable `primary_id` based on uuid, id, or content hash. `--resume` reliably avoids duplicates.
- **Output modes**: Supports `--output-mode bundle` to write separate files per dataset split.
- **Turn modes**: Supports `--turn-mode multi` to iteratively regenerate multi-turn assistant responses based on the freshly generated conversation prefix.
- **Quality filtering**: Drops responses with `finish_reason == 'length'` by default (configurable via `--keep-length-finished`) and completely drops conversations with no valid generated response.
- **Robustness**: Wraps API requests in bounded exponential backoff with configurable timeouts and max retries.
- **Accounting**: Prints reconciled processing counters per split at the end.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ Big updates have landed in Speculators! To get a more in-depth look, check out t
 Some of the exciting new features include:
 
 - **P-EAGLE Training Support**: Added support for the [P-EAGLE training algorithm](https://docs.vllm.ai/projects/speculators/en/latest/user_guide/algorithms/peagle), which extends EAGLE-3's architecture with parallel multi-token prediction via Conditional-On-Distribution (COD) sampling. Rather than generating draft tokens sequentially, P-EAGLE predicts multiple tokens in a single forward pass, reducing drafting latency. The Red Hat team published a [P-EAGLE speculator for Qwen3-8B](https://huggingface.co/RedHatAI/Qwen3-8B-speculator.peagle).
-- **MTP Finetuning Support**: Added support for finetuning the native Multi-Token Prediction (MTP) heads of models like Qwen3-Next on domain-specific data, following the [FastMTP](https://arxiv.org/abs/2509.18362) approach. Because the MTP head is small (~100M–400M params), it can be trained on pre-extracted hidden states without loading the full verifier
+- **MTP Finetuning Support**: Added support for finetuning the native Multi-Token Prediction (MTP) heads of models like Qwen3-Next on domain-specific data, following the [FastMTP](https://arxiv.org/abs/2509.18362) approach. Because the MTP head is small (~100M–400M params), it can be trained on pre-extracted hidden states without loading the full verifier.
 - **Sliding Window Attention for DFlash**: Added sliding window attention support to DFlash speculators via `--sliding-window` and `--swa-causal` training flags. Sliding window attention reduces KV cache allocation for long-context sequences and can improve per-position acceptance rates compared to full attention.
 - **Qwen3-8B DFlash Speculator**: The RedHat team published a [DFlash speculator for Qwen3-8B](https://huggingface.co/RedHatAI/Qwen3-8B-speculator.dflash), achieving average speculative token acceptance lengths of up to 3.74 on `math_reasoning`.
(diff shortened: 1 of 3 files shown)


WindChimeRan (commented):
Thanks @shotsan 

Lots of useful work here. Two pieces of feedback:

1. Scope. This bundles three unrelated changes: the RFC #584 regen refactor, an MTP attention-mask feature (core.py document_ids), and README fixes. Could we split into focused PRs? 

2. Multi-turn drops the system prompt. extract_user_turns keeps only user turns, so regeneration runs with no system prompt. SpecForge  preserves it. Since the target generates without the system prompt but inference includes it, it's a train/inference discrepancy. I'll open a focused PR for the multi-turn path that preserves it.

santosh-agebold:
Closing in favor of the upstream PRs that were already merged.

shotsan:
Closing in favor of already merged PRs

PR #624, opened 2026-06-19
Fix conditional-accuracy denominator to respect loss_mask

SuperMarioYL:
## Summary

`compute_accuracy_single_step` (the shared metrics primitive used by eagle3 training) restricts the conditional-accuracy **numerator** to `loss_mask` positions but computes the **denominator** (`cond_total`) over *all* positions. When `loss_mask` excludes any position — the normal case, where prompt/padding tokens are masked out — the reported `cond_acc_{ttt_step}` metric is deflated, and it even drifts with the amount of padding while real-token accuracy is unchanged.

This is a metrics/observability fix only: the loss is masked separately and correctly (`loss_function`), so model quality and gradients are unaffected — only the logged accuracy number was wrong.

## The bug

In `compute_accuracy_single_step`:

```python
cond_total = correct.numel()                 # all positions
if prev_correct is not None:
    cond_total = prev_correct.sum().float()  # all positions
    correct = torch.logical_and(prev_correct, correct, out=prev_correct)
if loss_mask is not None:
    correct = torch.masked_select(correct, loss_mask.to(torch.bool))  # numerator masked
```

The numerator is masked; `cond_total` is not. `eagle3/metrics.py` passes a real `s_loss_mask` (sourced from the dataset) plus `s_prev_correct`, so this path is hit in normal eagle3 training.

Reproduction (4 real tokens, 3 correct; only padding count varies):

| input | `cond_total` (before) | `cond_acc` (before) | `cond_acc` (after) |
|------|----------------------|---------------------|--------------------|
| 4 real + 2 padding | 6 | 0.500 | **0.750** |
| 4 real + 4 padding | 8 | 0.375 | **0.750** |

The metric should be `0.750` in both cases.

## The fix

Count `cond_total` over `prev_correct & loss_mask`, read **before** the in-place `logical_and` mutates `prev_correct`, and over `loss_mask` alone on the first step (`prev_correct is None`). Behaviour is unchanged when `loss_mask is None`. `full_total` was already mask-aware (it is `numel()` *after* `masked_select`) and is left untouched.

## Tests

The existing `compute_accuracy_single_step` test only exercised `loss_mask=None`, so the masked path was untested. Added regression tests covering:

- `cond_total` respects `loss_mask` with `prev_correct` given;
- conditional accuracy is invariant to padding count;
- first-step (`prev_correct=None`) masked denominator;
- the denominator uses the *previously-correct* mask, not the post-AND mask (the in-place `out=prev_correct` aliasing) — a case where the two differ.

[...]

PR #636, opened 2026-06-22
fix(convert): align Eagle3 num_key_value_heads default with Eagle1

SuperMarioYL:
## Summary

The Eagle-3 converter hardcodes `num_key_value_heads` to `8` when the draft
checkpoint config omits it:

```python
num_key_value_heads=eagle_config.get("num_key_value_heads", 8),
```

This silently forces grouped-query attention with 8 KV groups regardless of the
draft model's real architecture. The Eagle-1 converter handles the same field
correctly by leaving it `None`, which lets `LlamaConfig` fall back to
`num_attention_heads` (standard multi-head attention):

```python
# eagle_converter.py
num_key_value_heads=eagle_config.get("num_key_value_heads"),
```

When a draft config omits `num_key_value_heads`, the Eagle-3 path therefore
produces an attention configuration that does not match the checkpoint (e.g.
`num_attention_heads=32` but `num_key_value_heads=8`), corrupting the KV
projection shape of the converted draft.

## Fix

Drop the hardcoded `8` so Eagle-3 mirrors Eagle-1 and the HuggingFace idiom:
an absent `num_key_value_heads` falls back to `num_attention_heads`. When the
key is present, behavior is unchanged. This follows the same converter-parity
direction as the previously merged rope_scaling fix (#574).

## Tests

Added two CPU-only regression tests to `TestEagle3ConverterFixes` (mocking
`PretrainedConfig.get_config_dict`, matching the existing
`test_config_num_hidden_layers_*` pattern — no checkpoint download required):

- `test_config_num_key_value_heads_from_config` — present value is used as-is.
- `test_config_num_key_value_heads_defaults_to_attention_heads` — absent value
  resolves to `num_attention_heads` (not `8`). This test is red before the fix
  and green after.

The existing real-model conversion test
(`nm-testing/testing-llama3.1.8b-2layer-eagle3`, whose config does include
`num_key_value_heads`) continues to pass, confirming no change to the
present-key path.

### Why not keep the `8` default?

No code in the library assumes a KV-head count of 8 — every consumer reads
`config.num_key_value_heads` dynamically. The literal was an arbitrary default
that diverges from the Eagle-1 converter; `num_attention_heads` is the
architecturally neutral, HF-standard fallback for a config that does not
specify GQA.

### DIFF
--- a/src/speculators/convert/eagle/eagle3_converter.py
+++ b/src/speculators/convert/eagle/eagle3_converter.py
@@ -163,7 +163,11 @@ class Eagle3Converter:
             intermediate_size=eagle_config.get("intermediate_size", 11008),
             num_hidden_layers=eagle_config.get("num_hidden_layers", 1),
             num_attention_heads=eagle_config.get("num_attention_heads", 32),
-            num_key_value_heads=eagle_config.get("num_key_value_heads", 8),
+            # Mirror the Eagle1 converter: when the draft config omits
+            # num_key_value_heads, leave it None so LlamaConfig falls back to
+            # num_attention_heads (standard MHA) instead of silently forcing
+            # GQA with a hardcoded group count.
+            num_key_value_heads=eagle_config.get("num_key_value_heads"),
             hidden_act=eagle_config.get("hidden_act", "silu"),
             # Ensure max_position_embeddings match between Eagle3 and target configs
             max_position_embeddings=max(
--- a/src/speculators/convert/eagle/eagle3_converter.py
+++ b/src/speculators/convert/eagle/eagle3_converter.py
@@ -163,10 +163,6 @@ class Eagle3Converter:
             intermediate_size=eagle_config.get("intermediate_size", 11008),
             num_hidden_layers=eagle_config.get("num_hidden_layers", 1),
             num_attention_heads=eagle_config.get("num_attention_heads", 32),
-            # Mirror the Eagle1 converter: when the draft config omits
-            # num_key_value_heads, leave it None so LlamaConfig falls back to
-            # num_attention_heads (standard MHA) instead of silently forcing
-            # GQA with a hardcoded group count.
             num_key_value_heads=eagle_config.get("num_key_value_heads"),
             hidden_act=eagle_config.get("hidden_act", "silu"),
             # Ensure max_position_embeddings match between Eagle3 and target configs
--- a/tests/unit/convert/test_eagle3_converter.py
+++ b/tests/unit/convert/test_eagle3_converter.py
@@ -141,6 +141,52 @@ class TestEagle3ConverterFixes:
     @patch(
         "speculators.convert.eagle.eagle3_converter.PretrainedConfig.get_config_dict"
     )
+    def test_config_num_key_value_heads_from_config(
+        self, mock_get_config, sample_eagle3_config
+    ):
+        """Test that num_key_value_heads is taken from eagle_config when present."""
+        mock_get_config.return_value = ({}, None)
+        converter = Eagle3Converter()
+
+        sample_eagle3_config["num_attention_heads"] = 32
+        sample_eagle3_config["num_key_value_heads"] = 8
+
+        llama_config = converter._create_transformer_config_from_eagle(
+            sample_eagle3_config, "meta-llama/Llama-3.1-8B-Instruct"
+        )
+        assert llama_config.num_key_value_heads == 8
+
+    @pytest.mark.sanity
+    @patch(
+        "speculators.convert.eagle.eagle3_converter.PretrainedConfig.get_config_dict"
+    )
+    def test_config_num_key_value_heads_defaults_to_attention_heads(
+        self, mock_get_config, sample_eagle3_config
+    ):
+        """num_key_value_heads should fall back to num_attention_heads (MHA).
+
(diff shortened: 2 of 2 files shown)


orestis-z (commented):
LGTM. The hardcoded `8` was a real bug that silently corrupted KV projection shapes for models without explicit `num_key_value_heads` in their draft config. Fix is minimal and correct; regression tests cover both paths. Recommend approving.

rahul-tuli (approved):

SuperMarioYL:
Rebased onto the latest `main` (the branch had fallen 15 commits behind, which was tripping the `auto-rebase` rule). The rebase is clean — no conflicts, and our change (`eagle3_converter.py` + its test) replays unchanged on top. Force-pushed with `--force-with-lease` to keep the history linear (dropped the prior merge commit). Happy to address any fresh feedback.

SuperMarioYL:
Rebased onto latest `main` and force-pushed (same situation as #624 — the branch had fallen behind and the `auto-rebase` rule could not push the fork branch). Single commit, clean replay, no conflicts; fix unchanged. CI is re-running.

SuperMarioYL:
Rebased onto the latest `main` (the branch had fallen behind and the `auto-rebase` rule can't push to a fork). Clean replay — no conflicts, our change is unchanged; CI is re-running on the updated base.

SuperMarioYL:
Rebased onto the latest `main` (the branch had fallen behind and the auto-rebase rule couldn't rebase a fork branch). No conflicts — the `num_key_value_heads` default alignment is unchanged. Tests/ruff re-run clean.

SuperMarioYL:
Gentle ping — this one is approved and CI is green (5/5). Happy to rebase or address anything if there's a hold-up before merge. Thanks!

SuperMarioYL:
Rebased onto current `main` to clear the behind-main check. No logic changes — the converter fix and its test are unchanged (just replayed on top of the latest `main`).

SuperMarioYL:
Hi @rahul-tuli — gentle follow-up on this and the sibling #624.

Both are approved and `mergeable`, but each carries a failing Mergify "auto-rebase if approved, ready, and 5 commits behind main" check. That rule fails because the bot can't push to my fork branch, not because of a code issue — the branches are currently ~6 commits behind `main`. On #624 the buildkite run that failed was a GPU-infra flake (3 of 8 H100 jobs in the unit/integration step; `main` was green at the same time, and the touched unit tests pass on the branch).

I'm happy to rebase both onto current `main` (the diffs are small and still apply cleanly) to clear the auto-rebase gate and retrigger CI if that helps move them toward merge — just say the word. Otherwise, is there anything else blocking these from your side? Thanks for the review.

src/speculators/convert/eagle/eagle3_converter.py
```diff
@@ -163,7 +163,11 @@ def _create_transformer_config_from_eagle(
             intermediate_size=eagle_config.get("intermediate_size", 11008),
             num_hidden_layers=eagle_config.get("num_hidden_layers", 1),
             num_attention_heads=eagle_config.get("num_attention_heads", 32),
-            num_key_value_heads=eagle_config.get("num_key_value_heads", 8),
+            # Mirror the Eagle1 converter: when the draft config omits
+            # num_key_value_heads, leave it None so LlamaConfig falls back to
+            # num_attention_heads (standard MHA) instead of silently forcing
+            # GQA with a hardcoded group count.
```
fynnsu: Can we remove this comment.

tests/unit/convert/test_eagle3_converter.py
```diff
@@ -137,6 +137,52 @@ def test_config_num_hidden_layers_default(
         )
         assert llama_config.num_hidden_layers == 1
 
+    @pytest.mark.sanity
```
fynnsu, line 140: Please remove these tests.

fynnsu (changes requested):
Hi @SuperMarioYL, thanks for the pr. This is a good change to make but the pr is a little bit too verbose in the current state.

This is a pretty minor change in a not super active part of the codebase (Eagle3 conversion) and doesn't really justify a lengthy comment + tests for a small default change.

src/speculators/convert/eagle/eagle3_converter.py
SuperMarioYL: Done — removed the comment in a6bba04.

tests/unit/convert/test_eagle3_converter.py
SuperMarioYL: Done — removed both added tests in a6bba04.

SuperMarioYL:
Hi @fynnsu, thanks for the review. Trimmed the PR in a6bba04: removed the comment above the `num_key_value_heads` default and dropped the two added unit tests. The change is now just the one-line default fix, matching the Eagle1 converter. Happy to make further adjustments if needed.

SuperMarioYL:
Rebased onto latest `main` to clear the auto-rebase rule — the diff is unchanged (single-line `num_key_value_heads` default change per the earlier review).

SuperMarioYL:
Hi @fynnsu, gentle ping on this one — the requested trims (the inline comment and the extra tests) went in on 08-27 and the branch has since been re-rebased onto latest main, so the diff is now just the one-line `num_key_value_heads` default alignment. Happy to reshape it further if you'd prefer a different cut.

Issue #583, opened 2026-06-08
[RFC]: Prepare-data ingestion & normalization


imargulis:
### Motivation.

`scripts/prepare_data.py` and the data_generation loaders accept a handful of hard-coded dataset presets plus a single local `.json/.jsonl` path, with bespoke per-dataset normalizers and ad-hoc role mapping. 
That makes it hard to: train on a local *bundle* (a directory of split/parquet files), combine several sources, point at an arbitrary HF dataset/split, or rerun safely (the old `--overwrite` left stale files and
could silently produce an empty dataset). Conversation schemas (`messages` vs
`conversations`, ShareGPT `from/value` vs OpenAI `role/content`) are also normalized inconsistently.

### Proposed Change.

Generalize ingestion so local bundles, configured presets, and mixed prompt/message
schemas all reduce to one normalized conversation format before preprocessing.
- **Ingestion order** in `load_raw_dataset`: local file → local directory bundle → named
  preset (`DATASET_CONFIGS`) → HF hub path with optional `":split"` or `":[split_1|...|split_n]"` suffix. Directories are discovered recursively, grouped by `(format, source)` to preserve each split/file's
  provenance, and concatenated on their common columns (non-shared columns dropped with a
  per-column warning; mixed parquet schemas grouped and reconciled). An internal
  `__speculators_source` column tracks provenance and is dropped before save.
- **Registry**: add `nemotron` (+ per-split variants), `codealpaca`, `magpie`; extend
  `DatasetConfig` with split tuples, `prompt`/`messages` field hints, and a
  `preprocessing_supported` guard for response-only presets. Generalize the per-dataset
  normalizers into small factories.
- **Normalization**: factor `_normalize_turn` out behind a `ROLE_ALIASES` table and rename
  a `messages` column to `conversations` when none exists.
- **Proportional sampling**: under `--max-samples`, select a subset proportionally by
  source so a bundle/multi-source mix is preserved rather than truncated. Sampling is
  global across sources (documented, with a hook for alternative policies).
- **Safety/correctness**: `--data` takes multiple space-separated sources; `--output`
  defaults to `./output`; `--overwrite` refuses to delete a directory containing
  non-artifact files, then does a clean rebuild; truncate `input_ids`/`loss_mask` to
  `max_length`; raise on an all-empty result unless `--allow-empty-output`.

### Any Other Things.

These are separable and can be split into smaller PRs if preferred (ingestion/bundles,
[...]

PR #637, opened 2026-06-22
feat: Generalized data ingestion & normalization (RFC #583)

shotsan:
Implements the core ingestion and normalization improvements from RFC #583.

**Changes:**
- **Flexible Ingestion**: `load_raw_dataset` now naturally falls back to scanning local directories for json/parquet bundles, and finally attempts to load the argument as an arbitrary HuggingFace path if it's not a known preset. Users can pass split suffixes like `HuggingFaceH4/ultrachat_200k:train_sft`.
- **Role Mapping**: Factored out `_normalize_turn` behind a `ROLE_ALIASES` table to robustly handle `human`, `model`, `observation`, etc. across different datasets without custom normalization functions.
- **Schema Normalization**: Automatically renames the common `messages` column to `conversations` immediately after load, dramatically reducing the need to write custom Python normalization functions for simple schema differences.

### DIFF
--- a/src/speculators/data_generation/preprocessing.py
+++ b/src/speculators/data_generation/preprocessing.py
@@ -73,60 +73,60 @@ def _visualize_sample(preprocessed: HFDataset, processor: ProcessorLike, idx: in
 
     log.info(highlighted)
 
+ROLE_ALIASES = {
+    "human": "user",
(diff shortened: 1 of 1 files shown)


shotsan:
Closing in favor of already merged PRs

PR #644, opened 2026-06-23
Fix discarded shuffle when combining preprocessed datasets

guan404ming:
## Purpose

Assign the shuffle result back so the combined dataset is actually shuffled before truncating to max_samples. `Dataset.shuffle` returns a new dataset, so the discarded return left kept samples biased toward the first source dataset.

## Tests

python -m pytest "tests/integration/datagen/test_preprocessing.py::test_load_and_preprocess_dataset_shuffles_combined_datasets"

## Checklist

Added test fails before the fix (kept samples are all from the first dataset) and passes after.

I have filled in:


fynnsu (approved):
Nice catch! Thank you

guan404ming:
Thanks @fynnsu !

PR #645, opened 2026-06-23
Apply max_length truncation on HF assistant-mask path

guan404ming:
## Purpose

Truncate text-only inputs on the HF assistant-mask path to max_length, which it previously ignored, making `--seq-length` a no-op for text models using HF masks. Multimodal inputs stay untruncated since vLLM re-tokenizes them from messages.

## Tests

`test_preprocess_batch_truncation_hf_mask_path` → passed. Fails before the fix, passes after; `test_preprocess_batch_multimodal` still passes.

## Checklist

I have filled in:

### DIFF
--- a/src/speculators/data_generation/preprocessing.py
+++ b/src/speculators/data_generation/preprocessing.py
@@ -385,7 +385,14 @@ def _get_input_ids_loss_mask(
     hf_conv = _adapt_conv_for_hf(normalized_conv, processor)
 
     if assistant_pattern is None:
-        # HF assistant token mask
+        # HF assistant token mask. Multimodal input_ids are re-tokenized by vLLM
+        # from `messages`, so truncating them here would desync the two; only
+        # truncate text-only inputs, which are sent to vLLM as input_ids.
+        truncation_kwargs: dict = (
+            {}
+            if isinstance(processor, ProcessorMixin)
+            else {"max_length": max_length, "truncation": True}
+        )
         encoded_any = processor.apply_chat_template(
             hf_conv,
             tokenize=True,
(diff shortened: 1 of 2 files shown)


src/speculators/data_generation/preprocessing.py
fynnsu, line 435: > Multimodal inputs stay untruncated since vLLM re-tokenizes them from messages.

It seems like in the fallback case we are currently truncating the messages?

guan404ming: Thanks for catching that. Yes, the fallback path truncates MM input_ids via `processor_kwargs`, which desyncs from vLLM's re-tokenization of full messages (`the prompt_token_ids != token_ids` check in `extract_output` would reject it). 

I think that's a latent bug. Gate it the same way so only the text-only branch truncates is better.  Should I fold this fix into the current PR, or keep it scoped and open a separate one?

WindChimeRan:
https://github.com/vllm-project/speculators/issues/652

I think these are related. Will double check the behavior consistency.

orestis-z (commented):
LGTM. Clean fix for a real bug — the HF assistant-mask path was silently ignoring `max_length`. Agree with @coderabbitai’s suggestion to add the `_supports_assistant_mask` skip guard to the new test. Recommend approving once that’s addressed.

WindChimeRan (commented):
This PR is on the legacy HF path, which will be superseded by #794 . We want single-source of the truth from vllm.

speculatorsbot (commented):
LGTM. Correct fix — the `ProcessorMixin` gate is consistent with the fallback path, and the regression test validates both `input_ids` and `loss_mask` truncation. Outstanding items from earlier reviews (the `_supports_assistant_mask` skip guard and the #794 supersession question) are the right things to resolve before merging. Recommend approving once addressed.

shanjiaz:
Thanks for putting up this fix! We're deprecating this path way tho. Please take a look at https://github.com/vllm-project/speculators/pull/794 and let us know if you have other concerns/recommendations.

PR #653, opened 2026-06-24
Suppress uv and apt progress bars in logs

fynnsu:
## Purpose

Currently all of the buildkite logs are getting truncated because they exceed 2MB. The reason for this is the uv (and to a lesser degree apt) installs print and then delete progress updates as they go. This looks nice locally but results in massive log files in non-terminal environments. 

Edit: It looks like we're also currently downloading uv packages to a default cache dir, which then get copied for use in the environment. This seems to be happening because the default cache dir isn't on the same volume as the workplace where the env is created. I added a commit to set a new local cache dir. 

## Tests

Will run buildkite ci and confirm changes improve logs.

## Checklist

I have filled in:

### DIFF
--- a/.buildkite/gpu-tests/scripts/run-tests.sh
+++ b/.buildkite/gpu-tests/scripts/run-tests.sh
@@ -7,7 +7,7 @@ cat /etc/issue
 
 git fetch --tags --unshallow 2>/dev/null || git fetch --tags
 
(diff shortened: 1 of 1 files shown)


shanjiaz (approved):

deepak-kumar-neu (commented):
LGTM! thank you!

PR #655, opened 2026-06-24
Added claude skill to train speculators model

shanjiaz:
## Purpose
Despite our docs and examples, it might still be hard to spin up a training run for new models. The process could be confusing and manual. Added a claude skill to make this process easier. Would love to hear any suggestions.

## Tests
Tested locally.

## Checklist

I have filled in:


WindChimeRan:
I tried to train a DFlash drafter, using Claude Opus 4.8, without this skill. My prompt:

```
... use the new dataset, download model and data to /path ... you may  train a dflash drafter for google/gemma-4-31B-it, you are allowed to use GPU 4 5 6 7
```

And it just works. 

During the training process, it auto-fix a hang by `--no-async-scheduling`, not documented by the skill. 

I like the explanations on the speculator training knobs, but documentation might be a better place. 

Maybe opus 4.8 is already smart enough that we don't need a skill?

WindChimeRan:
Gap: I think response regeneration should be default for most training, not just MTP. Otherwise it's off-policy training.

orestis-z:
> I tried to train a DFlash drafter, using Claude Opus 4.8, without this skill. My prompt:
> 
> ```
> ... use the new dataset, download model and data to /path ... you may  train a dflash drafter for google/gemma-4-31B-it, you are allowed to use GPU 4 5 6 7
> ```
> 
> And it just works.
> 
> During the training process, it auto-fix a hang by `--no-async-scheduling`, not documented by the skill.
> 
> I like the explanations on the speculator training knobs, but documentation might be a better place.
> 
> Maybe opus 4.8 is already smart enough that we don't need a skill?

Thanks for your feedback @WindChimeRan and great that it just works for you!

We'll iterate on the documentation.

Regarding Opus 4.8 possibly being smart enough that we don't need this skill: I don't think we can assume everybody has access to Opus 4.8. Maybe a safer assumption to use an older/cheaper model like Opus 4.6. Wdyt?

Issue #659, opened 2026-06-25
DFlash train do not support MM!!!


chizhanyuefeng:
My train dataset is MM data

{
   "images": [],
   "system": "",
   "conversations": []
}

val_metrics.json

{"loss_epoch": 0.055430415494879114, "full_acc_epoch": 0.8921938772083632, "position_1_acc_epoch": 0.9333426497179864, "position_2_acc_epoch": 0.9084640020613753, "position_3_acc_epoch": 0.8972993950354843, "position_4_acc_epoch": 0.8900179479239685, "position_5_acc_epoch": 0.8816816581237188, "position_6_acc_epoch": 0.8718744566500382, "position_7_acc_epoch": 0.8612913956776911}

infer draft model: accpet only 8%!

shanjiaz:
@chizhanyuefeng Thanks for reaching out! Could you provide more details? Which model are you targeting and how much data is your model trained on? Which vLLM version did you test the model for? Would be happy to help!

Issue #652, opened 2026-06-24
[RFC]: Use vLLM Render Endpoint in preprocessing for Tokenization Alignment


shanjiaz:
### Motivation.
Speculators preprocesses training data by applying chat templates and tokenizing with HuggingFace's AutoProcessor/AutoTokenizer, then sends the resulting token IDs to vLLM's Completions API for hidden-state extraction. This assumes the HF tokenizer and vLLM's internal preprocessing produce identical token sequences for the same conversation. They sometimes don't, and the divergence occurs at two distinct layers. Tokenizer-level divergence. vLLM maintains a tokenizer registry (vllm/tokenizers/registry.py) with custom tokenizer backends that replace AutoTokenizer entirely when tokenizer_mode="auto". When vLLM auto-detects one of these models, it loads the custom tokenizer instead of HF's. Preprocessing-level divergence. Even when the underlying tokenizer is the same (e.g., gpt-oss uses HF's tokenizer for raw text encoding), vLLM applies additional conversation-level preprocessing that apply_chat_template() does not replicate. For gpt-oss specifically, vLLM uses Harmony's Rust encoder (openai_harmony.load_harmony_encoding) to render conversations into token sequences via `render_for_completion()`, which injects system preambles, formats tool descriptions, and structures messages in a way that is different from the Jinja2 chat template. vLLM also applies tool parsers and reasoning parsers that further transform the token sequence.

**Two code paths for text-only vs. multimodal.** Speculators currently maintains two separate data flows depending on content type. Text-only data is tokenized locally with HF and sent as pre-tokenized input_ids to the Completions API. Multimodal data sends raw messages to the Chat Completions API, letting vLLM handle tokenization, because the Completions API cannot carry image/video/audio references. This split creates its own problems: the text-only path is vulnerable to the tokenizer divergence described above, while the multimodal path relies on vLLM and HF inserting placeholder tokens at the same positions to keep the locally-built loss mask aligned.  A single path that delegates all tokenization to vLLM would eliminate this branching and make our preprocessing step more maintainable.

### Proposed Change.

[...]

PR #665, opened 2026-06-26
Add --render-endpoint to prepare_data.py for vLLM-delegated tokenization (#652)

WindChimeRan:
## Purpose

Adds a `--render-endpoint` flag to `prepare_data.py` that delegates tokenization to a running vLLM render server instead of doing it locally with HF `apply_chat_template`. 

The --render-endpoint should point at the same vllm serve instance used for hidden-state extraction. The render endpoint runs on the same server with zero GPU overhead. A separate vllm launch render instance is only needed for GPU-less offline preprocessing.

vLLM prerequisite PR has been merged, bun not released yet: https://github.com/vllm-project/vllm/pull/46846

When `--render-endpoint` is set, raw conversations are sent to `POST /v1/chat/completions/render` and the returned `token_ids` + `loss_mask` are used directly, eliminating the need for local chat-template rendering, assistant-pattern detection, and regex-based loss masking. 

Motivation: online training already requires a vLLM instance, so tokenization should happen once on the vLLM side rather than being duplicated between HF and vLLM. See RFC #652 for discussion.

## Tokenization 

* Render + server mask (template has `{% generation %}`): token_ids and mask both from /render → 1 tokenization, server-side, no local re-tokenization.
* Render + regex fallback (no gen-tags): server token_ids kept as input_ids; mask built from one local decode → re-tokenize for offsets only (row dropped if it doesn't round-trip) → 2 tokenizations (1 server + 1 local, offsets-only).
* Default HF path (no --render-endpoint, unchanged): single local apply_chat_template → 1 tokenization, no server.

## Corner cases

The current renderer endpoint + fallback may not fully guarantee the correctness of some corner cases, 

e.g., #659 , #645, #690 need to investigate further

## Tests

* passed end2end tests using vllm PR: https://github.com/vllm-project/vllm/pull/46846

## Limitation:
* **[Future work]** No concurrency: the render path is a sequential for loop of blocking HTTP calls.
* The headline benefit is niche. For the common models (Qwen2.5/3/VL — no {% generation %} tags), the render path falls to the same local regex the default path uses. 

## Checklist

I have filled in:


scripts/prepare_data.py
```diff
@@ -105,6 +105,18 @@ def parse_args():
             "per conversation for data augmentation."
         ),
     )
+    parser.add_argument(
+        "--render-endpoint",
```
rahul-tuli: nit: Name is acceptable and the help text makes it clear, but `--render-endpoint` might invite users to put in the complete endpoint `/v1/chat/completions/render` should we consider renaming to something like `--render-server-url` or `--vllm-server-url`?

src/speculators/data_generation/preprocessing.py
```diff
@@ -773,18 +883,29 @@ def load_and_preprocess_dataset(
...
-            dataset=raw_dataset,
-            processor=processor,
-            max_length=seq_length,
-            num_proc=build_dataset_num_proc,
-            assistant_pattern=assistant_pattern,
-            turn_dropout=turn_dropout,
-            minimum_valid_tokens=minimum_valid_tokens,
-        )
+        if render_endpoint is not None:
+            preprocessed_dataset = build_dataset_from_render(
```
rahul-tuli, line 1026: Could we warn/raise if both render-endpoint and assisstant pattern are set?

src/speculators/data_generation/preprocessing.py
```diff
@@ -709,32 +837,10 @@ def load_and_preprocess_dataset(
     turn_dropout: bool = False,
     minimum_valid_tokens: int | None = None,
     trust_remote_code: bool = False,
+    render_endpoint: str | None = None,
+    chat_template_kwargs: dict | None = None,
```
rahul-tuli, line 936: Shouldn't we also add `chat_template_kwargs` in the CLI?

src/speculators/data_generation/preprocessing.py
```diff
@@ -709,32 +837,10 @@ def load_and_preprocess_dataset(
...
-                          chat template.
-        turn_dropout: If True, randomly keeps first N consecutive turns per
-                     conversation
-        minimum_valid_tokens: Number of tokens to consider for a valid sample
-        trust_remote_code: If True, allows executing code from HF Hub.
-
-    Returns:
-        Tuple of (preprocessed_dataset, processor)
-    """
+    """Load, tokenize, and preprocess a dataset for EAGLE3 training."""
```
rahul-tuli: Since this is a public-ish function, could we update the docstring with new additions instead of truncating it?

rahul-tuli (commented):
Thank you for working on this @WindChimeRan, the diff looks good to me pending a few small comments, could you respond and update with latest main, and I will approve the workflows to run, great job on this!

tests/integration/datagen/test_render_multimodal.py
```diff
@@ -0,0 +1,130 @@
...
+
+from datasets import Dataset as HFDataset
+from PIL import Image
+
+from speculators.data_generation.preprocessing import (
+    build_dataset_from_render,
+    load_processor,
+)
+
+VLLM_PYTHON = "/home/ran/workspace/vllm/.venv/bin/python"
```
shanjiaz: probably needs to fix this.

src/speculators/data_generation/render_client.py
```diff
@@ -0,0 +1,53 @@
+"""Client for vLLM's /v1/chat/completions/render endpoint (RFC #652)."""
+
+from http import HTTPStatus
+
+import httpx
```
shanjiaz, line 5: Is httpx in our requirements, could you check.

rahul-tuli: I think it's installed by openai dependency

WindChimeRan: added the dependency explicitly.

scripts/prepare_data.py
WindChimeRan: use `--vllm-server-url` now

WindChimeRan:
Thanks for the review! All addressed @rahul-tuli @shanjiaz

tests/integration/datagen/test_render_multimodal.py
```diff
@@ -0,0 +1,130 @@
...
+
+from speculators.data_generation.preprocessing import (
+    build_dataset_from_render,
+    load_processor,
+)
+from tests.e2e.utils import VLLM_PYTHON
+
+MODEL = "Qwen/Qwen3-VL-2B-Instruct"
+MEDIA_DIR = "/tmp"
+IMG_PATH = "/tmp/speculators_render_mm_test.png"
```
rahul-tuli: Instead of hardcoding these temp paths, I believe it would be nicer if we can use pytest primitives or context managers for these?

You could use something like the following maybe?
```python
@pytest.fixture(scope="module")
def test_dir(tmp_path_factory) -> Path:
    """Provides a module-scoped isolated temporary directory."""
    return tmp_path_factory.mktemp("speculators_render")


@pytest.fixture(scope="module")
def image(test_dir: Path) -> str:
    img_path = test_dir / "render_mm_test.png"
    arr = np.zeros((128, 128, 3), dtype=np.uint8)
    arr[:64, :, 0] = 200
    arr[64:, :, 2] = 200
    arr[32:96, 32:96, 1] = 180
    Image.fromarray(arr).save(img_path)
    return str(img_path)


@pytest.fixture(scope="module")
def render_server(test_dir: Path) -> str:
    ...
```

rahul-tuli:
LGTM, pending CI!

rahul-tuli (approved):

WindChimeRan:
this has design conflicts with #774

WindChimeRan:
design conflict with #794

PR #610, opened 2026-06-16
feat(eagle3): add Eagle 3.1 training support

orestis-z:
## Purpose

Adds Eagle 3.1 training support with two config flags (matching vLLM's config fields):

- **`--norm-before-fc`**: Apply a single RMSNorm to the concatenated target hidden states before the FC projection layer — i.e. `Norm(concat(h₂, h₁₈, h₃₃))`. This was already implemented on main for gpt-oss checkpoint compatibility; this PR adds the training CLI flag and wires it through P-EAGLE. **Note:** the original Eagle 3.1 paper specifies per-layer normalization (`concat(Norm(h₂), Norm(h₁₈), Norm(h₃₃))`), which corresponds to vLLM's `fc_norm` field. Per-layer norm support (`--fc-norm`) will be added in a follow-up PR.
- **`--norm-output`**: Feed post-norm hidden states back across TTT steps to stabilize magnitude drift across speculation depths.

Both flags default to `False` (opt-in).

Also fixes P-EAGLE missing `input_norm` before FC and wires `norm_before_fc` / `norm_output` through the converter and P-EAGLE's `from_training_args`.

### vLLM compatibility

- `norm_before_fc` works out of the box with vLLM
- `norm_output` requires a one-line patch in vLLM's `algos.py` to propagate the config field (model code already supports it via `getattr`): https://github.com/vllm-project/vllm/pull/46488

## Tests

```bash
pytest tests/unit/test_config.py -k "eagle3_config" -v    # config roundtrip
pytest tests/integration/models/test_model_forward.py -k "NormOutput" -v  # forward pass
```

- `test_norm_output` — both flags enabled, verifies input_norm + forward + backward
- `test_norm_output_without_norm_before_fc` — norm_output alone (no input_norm)
- `test_peagle_norm_before_fc` — P-EAGLE with norm_before_fc
- 4 config roundtrip tests (model_dump, to_dict, save/load_pretrained, defaults)

## Checklist

- [x] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [x] The test plan/results, such as providing test command and pasting the results.
- [x] The necessary documentation update.
- [x] I (a human) have written or reviewed the code in this pr to the best of my ability.

PR #668, opened 2026-06-26
fix(train): keep norm weights in float32 under bfloat16 training

orestis-z:
## Summary

- **Single-GPU:** RMSNorm weights are initialized to 1.0, where bfloat16's representable precision gap (~8e-3) swallows typical Adam updates (~1e-4), so the weights never learn. After converting the model to the hidden-states dtype, upcast norm modules back to float32.
- **Distributed (FSDP):** FSDP2 requires uniform original parameter dtype within each shard group, so selectively upcasting norms to float32 crashes on the first forward pass. Instead, keep all master weights in float32 and let `MixedPrecisionPolicy` handle the bfloat16 compute cast — this preserves float32 optimizer updates for all params including norms.
- **Bug fix in `apply_fully_sharded`:** The root `fully_shard(model)` call was missing `mp_policy`, so top-level parameters (embeddings, fc, lm_head, norms) didn't get bf16 compute + fp32 gradient reduction like the layers did.

## Context

Bug introduced in #352 (`dd6deac`, "Support transformers v5"), first shipped in v0.4.0. Before that PR, the model stayed in float32 (only device placement was done). Published models (e.g. `inference-optimization/Qwen3-8B-eagle3-ckpt2`) train with float32 and have learned norms.

FSDP uniform-dtype issue identified by @imargulis.

## Test plan

- [x] `test_norm_weights_stay_float32_under_bfloat16` — verifies norm params are float32 while other params are bfloat16 after `setup_model` (single-GPU)
- [x] `test_norm_weights_learn_under_bfloat16` — verifies norm weights actually change after an AdamW step (single-GPU, fails without the fix)
- [x] `test_distributed_norm_weights_float32_after_fsdp` — full forward+backward+optimizer step under FSDP; verifies no dtype assertion crash and that norm weights learn (multi-GPU)
- [x] All existing unit tests pass


tests/unit/train/test_setup_model.py
orestis-z, line 1: note: both tests fail on `main` branch and pass here.

rahul-tuli (approved):

imargulis:
@orestis-z Thanks a lot for introducing the fix. Would you mind checking the actual training with FSDP.
I got the failure due to the lack of uniform precision of master weights.

My way to address this was:
src/speculators/train/trainer.py

```
def setup_model(self):
        # Verify model is compatible with training infrastructure
        SpeculatorModel.verify_training_compatible(self.model)

        load_checkpoint = (
            self.resume_from_checkpoint and self.checkpointer.previous_epoch != -1
        )

        if not self.is_distributed:
            # Single device: cast to the compute dtype, but keep norm weights in
            # float32 so Adam updates (~1e-4) aren't lost to bfloat16's precision
            # limit (~8e-3 at 1.0). No FSDP here, so mixed param dtypes are fine.
            self.model.to(self.config.hidden_states_dtype)  # type: ignore[arg-type]
            for module in self.model.modules():
                if isinstance(module, torch.nn.LayerNorm) or type(
                    module
                ).__name__.endswith("RMSNorm"):
                    module.float()
            self.model.to(self.local_rank)  # type: ignore[arg-type]
            if load_checkpoint:
            return

        # Distributed case.
        # FSDP2 (fully_shard) requires a uniform *original* parameter dtype within
        # each sharded group, so we cannot keep norms in float32 while other
        # weights are bfloat16. Instead keep all master weights in float32 and let
        # apply_fully_sharded's MixedPrecisionPolicy do the bfloat16 compute cast.
        # This preserves float32 norm (and all) master weights for stable optimizer
        # updates without mixing dtypes inside an FSDP group.
        self.model.float()

        # Capture full state dict on rank 0 before FSDP sharding
        full_state_dict = {}
        if not load_checkpoint and dist.get_rank() == 0:
            full_state_dict = self.model.state_dict()

        apply_fully_sharded(self.model)

        if load_checkpoint:
        else:
            # Broadcast full state dict from rank 0 to all ranks
            set_model_state_dict(
                self.model,
                full_state_dict,
                options=StateDictOptions(
                    full_state_dict=True,
                    broadcast_from_rank0=True,
                    strict=False,
                ),
            )
            del full_state_dict
            dist.barrier()
```

src/speculators/train/utils.py
```
def apply_fully_sharded(model: torch.nn.Module):
    """Applies torch FSDP fully_shard to the model, wrapping layers in FSDPModule.

    Assumes the model has a `layers` attribute containing the decoder layers.
    Model should be validated with SpeculatorModel.verify_training_compatible()
    before calling this function.
    """
    mp_policy = MixedPrecisionPolicy(
        param_dtype=torch.bfloat16,
        reduce_dtype=torch.float32,
    )

    # Apply the same policy to the root group (embeddings, fc, lm_head and the
    # final/hidden/verifier norms) so those top-level params use bf16 compute and
    # fp32 gradient reduction like the layers; without it they would compute in
    # their float32 master dtype.
    fully_shard(model, mp_policy=mp_policy)

    return model
```

Thanks.

orestis-z:
> @orestis-z Thanks a lot for introducing the fix. Would you mind checking the actual training with FSDP. I got the failure due to the lack of uniform precision of master weights.
> 
> My way to address this was: src/speculators/train/trainer.py
> 
> ```
> def setup_model(self):
>         # Verify model is compatible with training infrastructure
>         SpeculatorModel.verify_training_compatible(self.model)
> 
>         load_checkpoint = (
>             self.resume_from_checkpoint and self.checkpointer.previous_epoch != -1
>         )
> 
>         if not self.is_distributed:
>             # Single device: cast to the compute dtype, but keep norm weights in
>             # float32 so Adam updates (~1e-4) aren't lost to bfloat16's precision
>             # limit (~8e-3 at 1.0). No FSDP here, so mixed param dtypes are fine.
>             self.model.to(self.config.hidden_states_dtype)  # type: ignore[arg-type]
>             for module in self.model.modules():
>                 if isinstance(module, torch.nn.LayerNorm) or type(
>                     module
>                 ).__name__.endswith("RMSNorm"):
>                     module.float()
>             self.model.to(self.local_rank)  # type: ignore[arg-type]
>             if load_checkpoint:
>                 self.checkpointer.load_model_state_dict(self.model)
>             return
> 
>         # Distributed case.
>         # FSDP2 (fully_shard) requires a uniform *original* parameter dtype within
>         # each sharded group, so we cannot keep norms in float32 while other
>         # weights are bfloat16. Instead keep all master weights in float32 and let
>         # apply_fully_sharded's MixedPrecisionPolicy do the bfloat16 compute cast.
>         # This preserves float32 norm (and all) master weights for stable optimizer
>         # updates without mixing dtypes inside an FSDP group.
>         self.model.float()
> 
>         # Capture full state dict on rank 0 before FSDP sharding
>         full_state_dict = {}
>         if not load_checkpoint and dist.get_rank() == 0:
>             full_state_dict = self.model.state_dict()
> 
>         apply_fully_sharded(self.model)
> 
>         if load_checkpoint:
>             self.checkpointer.load_model_state_dict(self.model)
>         else:
>             # Broadcast full state dict from rank 0 to all ranks
>             set_model_state_dict(
>                 self.model,
>                 full_state_dict,
>                 options=StateDictOptions(
>                     full_state_dict=True,
>                     broadcast_from_rank0=True,
>                     strict=False,
>                 ),
>             )
>             del full_state_dict
>             dist.barrier()
> ```
> 
> src/speculators/train/utils.py
> 
> ```
> def apply_fully_sharded(model: torch.nn.Module):
>     """Applies torch FSDP fully_shard to the model, wrapping layers in FSDPModule.
> 
>     Assumes the model has a `layers` attribute containing the decoder layers.
>     Model should be validated with SpeculatorModel.verify_training_compatible()
>     before calling this function.
>     """
>     mp_policy = MixedPrecisionPolicy(
>         param_dtype=torch.bfloat16,
>         reduce_dtype=torch.float32,
>     )
> 
>     for layer in model.layers:  # type: ignore[union-attr]
>         fully_shard(layer, mp_policy=mp_policy)
> 
>     # Apply the same policy to the root group (embeddings, fc, lm_head and the
>     # final/hidden/verifier norms) so those top-level params use bf16 compute and
>     # fp32 gradient reduction like the layers; without it they would compute in
>     # their float32 master dtype.
>     fully_shard(model, mp_policy=mp_policy)
> 
>     return model
> ```
> 
> Thanks.

Thanks for this @imargulis, will check and report back

orestis-z:
@imargulis Thanks again for flagging the FSDP issue — you were right, the selective norm upcast crashed with `FSDP expects uniform original parameter dtype`. Pushed commit d49b992 which implements the approach you suggested: all master weights stay float32, `MixedPrecisionPolicy` handles the bf16 compute cast. Also added `mp_policy` to the root `fully_shard()` call (was missing before) and a multi-GPU regression test that runs a full forward+backward+optimizer step through FSDP.

imargulis:
> @imargulis Thanks again for flagging the FSDP issue — you were right, the selective norm upcast crashed with `FSDP expects uniform original parameter dtype`. Pushed commit [d49b992](https://github.com/vllm-project/speculators/commit/d49b9924f37ea704354bf15a8661395ba026e79f) which implements the approach you suggested: all master weights stay float32, `MixedPrecisionPolicy` handles the bf16 compute cast. Also added `mp_policy` to the root `fully_shard()` call (was missing before) and a multi-GPU regression test that runs a full forward+backward+optimizer step through FSDP.

@orestis-z I'm glad it worked out. BTW did your fix come from any real issue with training or just pure looking at the code?

orestis-z:
> > @imargulis Thanks again for flagging the FSDP issue — you were right, the selective norm upcast crashed with `FSDP expects uniform original parameter dtype`. Pushed commit [d49b992](https://github.com/vllm-project/speculators/commit/d49b9924f37ea704354bf15a8661395ba026e79f) which implements the approach you suggested: all master weights stay float32, `MixedPrecisionPolicy` handles the bf16 compute cast. Also added `mp_policy` to the root `fully_shard()` call (was missing before) and a multi-GPU regression test that runs a full forward+backward+optimizer step through FSDP.
> 
> @orestis-z I'm glad it worked out. BTW did your fix come from any real issue with training or just pure looking at the code?

Real issue with training: I was trying to debug an eagle3.1 checkpoint (something didn't look right) so I instructed the agent to check if the norm weights learned anything. There it stumbled upon 1.0 weights for the pre-FC norm. Then I inspected all the checkpoints I have linked in https://github.com/vllm-project/speculators/pull/610#issuecomment-4763251154 and they had all `1.0` as norm weight. Further debugging concluded that it was bfloat16 accuracy related and switching the norm weights to float32 made them move.

orestis-z:
bfloat16 has 7 mantissa bits, so its precision (ULP — unit in last place) at a given magnitude is:

  ```
  ┌─────────────────────┬───────────────────────────────────────────┬──────────────────────────────┐
  │  Weight magnitude   │ ULP (gap between consecutive bf16 values) │  Adam update (~1e-4) vs gap  │
  ├─────────────────────┼───────────────────────────────────────────┼──────────────────────────────┤
  │ ~0.01 (random init) │ 2^(-7-7) ≈ 6e-5                           │ 1e-4 > 6e-5 — update lands   │
  ├─────────────────────┼───────────────────────────────────────────┼──────────────────────────────┤
  │ ~1.0 (norm init)    │ 2^(0-7) ≈ 7.8e-3                          │ 1e-4 < 7.8e-3 — rounded away │
  └─────────────────────┴───────────────────────────────────────────┴──────────────────────────────┘
```

orestis-z:
@imargulis Would you mind checking the fix with FSDP?

imargulis:
> @imargulis Would you mind checking the fix with FSDP?


The training itself proceeds fine, I checked this right away. 
@orestis-z do you mean checking that RMSNorm reacts well to upscaling to fp32 during training?

orestis-z:
> > @imargulis Would you mind checking the fix with FSDP?
> 
> The training itself proceeds fine, I checked this right away. @orestis-z do you mean checking that RMSNorm reacts well to upscaling to fp32 during training?

Yes exactly — that the RMSNorm weights actually move away from 1.0 during training. If you've already seen training proceed fine, we're good. Thanks for confirming!

orestis-z:
As discussed with @fynnsu blocking this for a more general solution, https://github.com/vllm-project/speculators/tree/ddp_amp

orestis-z:
Closing in favor of #711 which takes a broader approach (DDP default + `torch.autocast` for AMP), which implicitly solves the norm precision issue without per-module dtype manipulation. The `mp_policy` one-liner from this PR has been flagged on #711.

Frozen7771:
Posted production-training evidence for this (A/B impact + a weight-decay angle not covered here) over in #711, since the AMP rework there seems to be the live thread.

PR #670, opened 2026-06-26
feat(eagle3): add per-layer FC normalization (--fc-norm)

orestis-z:
## Summary
- Add `--fc-norm` flag for per-layer RMSNorm on each auxiliary hidden state before concatenation and FC projection: `concat(Norm(h_a), Norm(h_b), Norm(h_c))`
- This matches the Eagle 3.1 paper specification and vLLM's `fc_norm` config field
- Differs from `--norm-before-fc` which applies a single norm to the full concatenation: `Norm(concat(h_a, h_b, h_c))`
- `norm_before_fc` and `fc_norm` are validated as mutually exclusive in the config (enabling both would cause double-norming)

## Test plan
```bash
pytest tests/unit/test_config.py -k "fc_norm" -v
pytest tests/integration/models/test_model_forward.py -k "fc_norm" -v
```

- `test_fc_norm` — Eagle3 with fc_norm + norm_output, verifies fc_norms module created and forward + backward pass
- `test_peagle_fc_norm` — P-Eagle with fc_norm
- `test_eagle3_config_fc_norm_roundtrip` — config serialization roundtrip
- `test_eagle3_config_rejects_both_norm_before_fc_and_fc_norm` — mutual exclusivity validation


src/speculators/convert/eagle/eagle3_converter.py
```diff
@@ -39,6 +39,9 @@ def convert(
         base_model: str,
         validate: bool = True,
         norm_before_residual: bool = False,
+        norm_before_fc: bool = False,
+        fc_norm: bool = False,
```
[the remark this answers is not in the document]
orestis-z: This is the existing pattern for norm_before_fc (line 142) and norm_output (line 144) too — they all use the same param or config.get(...) idiom. This is actually the intended behavior for a converter: the checkpoint's own config is the source of truth, and CLI flags only add flags that weren't in the original checkpoint. You wouldn't convert a checkpoint that was trained with fc_norm=True and want to strip that normalization — the weights depend on it.

src/speculators/models/eagle3/config.py
```diff
@@ -76,6 +86,15 @@ class Eagle3SpeculatorConfig(SpeculatorModelConfig):
         description="Whether embedding layer weights require gradients during training",
     )
 
+    @model_validator(mode="after")
+    def _check_norm_flags(self) -> "Eagle3SpeculatorConfig":
```
shanjiaz, line 90: Nice!

src/speculators/models/eagle3/core.py
```diff
@@ -102,6 +101,19 @@ def __init__(self, config: Eagle3SpeculatorConfig):
...
+                [
+                    self._model_definitions.norm_class(
+                        self.hidden_size,
+                        eps=config.transformer_layer_config.rms_norm_eps,
+                    )
+                    for _ in range(3)
+                ]
+            )
+        else:
+            self.fc_norm = None
```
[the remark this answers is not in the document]
shanjiaz: Ah I saw coderabbit caught this nice

src/speculators/models/eagle3/core.py
```diff
@@ -101,6 +101,19 @@ def __init__(self, config: Eagle3SpeculatorConfig):
...
             self.input_norm = None
 
+        if config.fc_norm:
+            self.fc_norm: torch.nn.ModuleList | None = torch.nn.ModuleList(
+                [
+                    self._model_definitions.norm_class(
+                        self.hidden_size,
+                        eps=config.transformer_layer_config.rms_norm_eps,
+                    )
+                    for _ in range(3)
```
shanjiaz: I know for eagle3 we just hard-coded to 3 and it's safe with the default. Not sure if it even matters but do you think we should try replacing with `len(self.target_layer_ids)` so if users want to do experiments they can? Will need to update the fc layer shape and norm before fc as well.

shanjiaz (approved):
Looks good just one small non-blocking concern

src/speculators/models/eagle3/core.py
orestis-z: That's a good suggestion I like making things more generic!

PR #671, opened 2026-06-26
feat(train): save launch command to checkpoint dir

### LINKED ISSUE #658: Save the training launch command to the checkpoint dir
## Summary

`scripts/train.py` has ~70 CLI flags, but a finished checkpoint records nothing about how it was produced. Once the shell scrollback is gone, the run is effectively unreproducible — there's no reliable way to know which verifier, loss, learning rate, seed, etc. produced a checkpoint sitting in `--save-path`. This is especially painful when running and comparing many experiments.

We should write the exact launch command into `--save-path` so any checkpoint can be reproduced by copy-paste.

## Proposed change

In `main()` (rank 0 only), write `sys.argv` (the `torchrun ... scripts/train.py ...` / `python scripts/train.py ...` invocation) to a file in `--save-path`, e.g. `train_command.txt`:

- reconstruct the command from `sys.argv`;
- write it atomically, rank-0 only, alongside the checkpoints;
- optionally prepend a short provenance header (timestamp, git commit SHA, `world_size`) as comments so a stale checkpoint can be traced back.

That's the whole feature: no schema, no namespace serialization, no new dependencies.

## Notes

- A few args are resolved/mutated at runtime (`mask_token_id`, `draft_vocab_size`, `num_speculative_steps`, default `target_layer_ids`). The saved command still reproduces them on re-run, so it is sufficient for reproducibility — it just isn't a snapshot of the *effective* resolved values.
- Good first issue. It touches the CLI, so per `CONTRIBUTING.md` it warrants maintainer sign-off before implementation.

orestis-z:
## Summary

- Write `sys.argv` and a provenance header to `train_command.txt` in `--save-path` so any checkpoint can be reproduced by copy-paste
- Header includes timestamp, git SHA, world size, and package versions (`speculators`, `vllm`, `transformers`, `torch`)
- Copy `train_command.txt` into each checkpoint subdirectory so it travels with the checkpoint to HuggingFace Hub
- Rank-0 only, atomic write via temp file, no new dependencies (stdlib only)

Closes #658

### Example output (`train_command.txt`)

```
# Timestamp: 2025-06-25T14:30:00.123456+00:00
# Git SHA: 3fdac63a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6e
# World size: 4
# speculators: 0.6.0.dev0
# vllm: not installed
# transformers: 4.52.4
# torch: 2.7.1
torchrun --nproc_per_node=4 scripts/train.py --speculator-type eagle3 --verifier-name-or-path meta-llama/Llama-3.1-8B-Instruct --save-path ./output/checkpoints --epochs 20 --lr 1e-4
```

## Test plan
- [ ] Run `python scripts/train.py --dry-run --save-path /tmp/test-cmd ...` and verify `train_command.txt` is created with correct content
- [ ] Verify the saved command is copy-pasteable and reproduces the run
- [ ] Verify non-rank-0 processes do not write the file
- [ ] Verify `train_command.txt` appears inside each epoch checkpoint subdirectory

## Summary by CodeRabbit

* **New Features**
  * Training runs now save a `train_command.txt` file with launch details, timestamp, environment info, and package versions.
  * Checkpoints now include this file in the main save folder and each epoch-specific checkpoint.

* **Tests**
  * Added coverage to verify the training command file is created and copied into checkpoint subfolders during finetuning and training-only flows.


scripts/train.py
```diff
@@ -466,6 +473,49 @@ def build_draft_model(
     )
 
 
+def _save_train_command(save_path: str) -> None:
```
fynnsu: This looks good but can we put it in one of the utils modules in speculators. `scirpts/train.py` is getting a bit bloated.

fynnsu (approved):

PR #673, opened 2026-06-27
Add hybrid LK loss (adaptive KL/TV blend) for speculative decoding

GIREESH7963:
## Summary
Adds the hybrid LK loss: an adaptive per-position blend of KL and TV,
L = lambda*KL + (1-lambda)*TV with lambda = exp(-eta * sg[alpha]), alpha = overlap.

## Motivation
TV optimizes acceptance directly but has a vanishing gradient from a cold start; KL has a
strong early gradient but optimizes a proxy. The hybrid leans on KL when overlap is low and
shifts to TV as the draft aligns, making TV's acceptance-optimal target trainable from
scratch. Source: Samarin et al., arXiv 2602.23881. Paper's best setting: eta=3.

## Changes
- `lk_hybrid_loss(logits, targets, eta=3.0)` in metrics.py, reusing `kl_div_loss` and the
  TV overlap term; alpha in the blend weight is detached.
- `resolve_loss_fn` gains `lk_loss_eta` (default 3.0), binding eta into the "lk_hybrid"
  entry via functools.partial (single binding point; call sites unchanged).
- `--lk-loss-eta` CLI flag (mirrors --dflash-decay-gamma), threaded through each
  get_trainer_kwargs into resolve_loss_fn; "lk_hybrid" added to --loss-fn choices.
- Tests: eta->0 == KL, eta->inf == TV, shape/finiteness, alpha-detach gradient check,
  resolve binds eta.

## Notes
- The adaptive KL/TV blend and the DFlash positional decay are orthogonal weightings.
- tv/nla are not in the --loss-fn choices list either; happy to add them here if desired.
- Loss key name "lk_hybrid" open to preference.

### DIFF
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -863,8 +863,8 @@ def parse_args():
         default="kl_div",
         help=(
             "Loss function specification. Pass a name for a single loss "
-            "(kl_div, ce, tv, nla) or a JSON dict for a weighted combination, "
-            'e.g. \'{"ce": 0.1, "tv": 0.9}\'.'
+            "(kl_div, ce, tv, nla, lk_hybrid) or a JSON dict for a weighted "
+            'combination, e.g. \'{"ce": 0.1, "tv": 0.9}\'.'
         ),
     )
     parser.add_argument(
--- a/src/speculators/models/metrics.py
+++ b/src/speculators/models/metrics.py
@@ -180,6 +180,48 @@ def neg_log_acceptance_loss(
     return elementwise_loss  # noqa: RET504
 
 
+def lk_hybrid_loss(
+    logits: torch.Tensor,  # shape: [1, seq_len, draft_vocab_size]
+    targets: torch.Tensor,  # shape: [1, seq_len, draft_vocab_size]
+    eta: float = 3.0,
+):
+    """Compute per-position hybrid LK loss (adaptive KL/TV blend).
+
+    Blends KL divergence and total variation per position:
+    ``L = lambda * KL(p||q) + (1 - lambda) * TV(p, q)`` with adaptive weight
+    ``lambda = exp(-eta * sg[alpha])``, where ``alpha = sum_v min(p_v, q_v)`` is the
+    acceptance rate (overlap) and ``sg`` is stop-gradient. When overlap is low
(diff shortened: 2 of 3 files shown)


GIREESH7963:
addressed in 03bc603. The `--loss-fn` help now lists `lk_hybrid` and points to `--lk-loss-eta` for its mixing parameter (`'lk_hybrid' = adaptive KL/TV blend (see --lk-loss-eta).`).

scripts/train.py
```diff
@@ -963,11 +963,12 @@ def parse_args():
         "--loss-fn",
         type=str,
         default="kl_div",
-        choices=["kl_div", "ce"],
+        choices=["kl_div", "ce", "lk_hybrid"],
```
fynnsu: We should also add the other loss functions here:
```
        "kl_div": kl_div_loss,
        "ce": ce_loss,
        "tv": tv_loss,
        "nla": neg_log_acceptance_loss,
```
("tv" and "nla").

fynnsu (approved):
Looks good but thing to fix below.

GIREESH7963:
Rebased onto `main` to clear the conflict (the loss refactor in #671/#677 is what conflicted). Summary of what changed since the approval, since it's a bit more than a plain rebase:

- `lk_hybrid` is now registered in the new module-level `_LOSS_FN_MAP`, so it works both as a standalone `--loss-fn lk_hybrid` and inside the JSON weighted-combination spec (e.g. `'{"lk_hybrid": 0.5, "ce": 0.5}'`).
- Re: the suggestion to add `tv`/`nla` to the `--loss-fn` `choices` list — that `choices=[...]` list was removed in the refactor (the arg is now a free-form name-or-JSON spec). `tv`, `nla`, and `lk_hybrid` are all listed in the updated free-form help text instead.
- Dropped the `--lk-loss-eta` flag. The refactored interface no longer has a per-loss hyperparameter channel (losses are bare `(logits, targets)` callables in `_LOSS_FN_MAP`), so `eta` is now baked at the paper's default of 3.0. Happy to restore tunability as a follow-up if you'd prefer.

Net functional diff is now just the `lk_hybrid_loss` function + one `_LOSS_FN_MAP` entry + the help-text mention, plus unit tests. Flagging in case you'd like another glance before merge given the `--lk-loss-eta` removal.

PR #675, opened 2026-06-28
Add flexible data source discovery to load_raw_dataset (#661)

### LINKED ISSUE #661: [RFC]: Flexible data source discovery
### Motivation.

Training a speculator on local data currently requires either passing individual `.json`/`.jsonl` files or using a named preset from `DATASET_CONFIGS`.  Separately, Magpie and Nemotron are popular open instruction-tuning datasets that would be useful presets in `DATASET_CONFIGS`. Magpie is already referenced in our scripts under a local config dict, so promoting it to the shared registry unifies the two code paths.

### Proposed Change.

Extend `load_raw_dataset()` in `src/speculators/data_generation/preprocessing.py` to support a resolution chain. Today the function handles two cases (local `.json`/`.jsonl` file, or named preset). We should extend to:
1. **Local file** (existing) — match paths ending in `.json`/`.jsonl`.
2. **Local directory** (new) — recursively glob `*.json`/`*.jsonl` from the directory and load as one dataset.
3. **Named preset** (existing) — look up `DATASET_CONFIGS`.
4. **Arbitrary HuggingFace dataset** (new) — a `hf:` prefix loads any HF dataset ID directly, with optional `:<subset>:<split>` suffix.

### Any Other Things.

- All source types compose with mixed `--data` usage: `--data sharegpt --data /my/shards/ --data hf:nvidia/HelpSteer2 --data extra.jsonl`.
- For the `hf:` path, datasets not in conversations format will fail at the preprocessing stage with an actionable error with no silent data corruption.
- A future follow-up could add auto-detection of common HF schemas (instruction/response, prompt/completion, messages) and normalize them automatically, reducing the need for per-dataset `normalize_fn` boilerplate.
- Test coverage: integration tests with a temp directory of JSONL shards and a small public HF dataset would validate both new paths.

KKothuri:
Resolves #661.

## Summary

Extends `load_raw_dataset()` in `src/speculators/data_generation/preprocessing.py` from two source types to the four-step resolution chain requested in RFC #661, and promotes Magpie/Nemotron to shared presets. Scope is intentionally limited to **data source discovery** — schema/role normalization is left to a separate effort.

## Changes

- `load_raw_dataset` now resolves a source in order:
  1. Local `.json`/`.jsonl` file (existing).
  2. **Local directory** — recursively glob `*.json`/`*.jsonl` (sorted for determinism) and load as one dataset; an empty directory raises an actionable error.
  3. Named preset from `DATASET_CONFIGS` (existing).
  4. **`hf:<id>[:<subset>:<split>]`** — load an arbitrary HuggingFace dataset.
- New `_load_hf_dataset` helper parses the `hf:` spec (split defaults to `train`). It **raises if the loaded dataset has no `conversations` column**, since `_preprocess_batch` otherwise only warns and silently yields zero samples — satisfying the RFC's "no silent data corruption" requirement.
- Adds `magpie` (already in ShareGPT `conversations` format, no `normalize_fn`) and `nemotron` (`nvidia/Llama-Nemotron-Post-Training-Dataset`, `SFT`/`chat`, with `_normalize_nemotron` building conversations from `input` + `output`) presets. Both schemas were verified against the live HF datasets-server before hardcoding.

All source types compose across mixed `--data` usage, e.g.:
```
--data sharegpt --data /my/shards/ --data hf:nvidia/HelpSteer2 --data extra.jsonl
```

### Note for the maintainer

Beyond the RFC's literal `hf:<id>:<subset>:<split>`, I also accept the 2-part form `hf:<id>:<split>` (subset omitted), since single-config datasets with a non-`train` split (e.g. `HuggingFaceH4/ultrachat_200k:train_sft`) are common. Happy to make it strict if you'd prefer.

## Why this is not a duplicate

#637 (RFC #583) touches the same function but bundles directory/HF discovery **together with** role/schema normalization. Per @shanjiaz's scoping on #583 and the assignment of #661, this PR delivers the **discovery slice only**, keeping it separate from normalization. The `hf:` handling here is also a different, more explicit design: an explicit `hf:` prefix (rather than treating any unmatched string as an HF path) plus a hard failure on non-conversations data.

## Tests

New tests in `tests/integration/datagen/test_preprocessing.py`: local file, recursive directory bundle (incl. nested `.json`), empty-directory error, preset dispatch, unsupported source, `hf:` spec parsing (parametrized), the conversations-format guard, malformed `hf:` specs, and the new presets. The `hf:`/preset cases mock `load_dataset` so they are deterministic and network-free.

Commands run (in the project venv):

```
python -m pytest tests/integration/datagen/test_preprocessing.py -q
# 51 passed, 2 skipped

ruff check  src/speculators/data_generation/{preprocessing,configs}.py tests/integration/datagen/test_preprocessing.py   # All checks passed
ruff format --check <same files>                                                                                          # already formatted
python -m mypy src/speculators/data_generation/{preprocessing,configs}.py                                                 # Success: no issues
```

## AI assistance disclosure

AI assistance (Claude Code) was used to help write this change. I (the submitter) have reviewed every changed line, understand the change end-to-end, and ran the tests above.


imargulis:
@KKothuri @shanjiaz 
Following up on RFC#583 I mostly had the same approach implemented with a slight difference below:

The local-directory path loads every discovered file in a single pass:

load_dataset("json", data_files=[all *.json/*.jsonl], split="train")
which forces one Arrow schema across all files, so a bundle whose files/splits don't share an identical schema fails with DatasetGenerationError. This isn't limited to exotic data — even two files that share conversations but differ in one top-level column raise:

```
# a.jsonl: {"conversations": [...], "idx": 1}
# b.jsonl: {"conversations": [...], "lang": "en"}
load_dataset("json", data_files=["a.jsonl", "b.jsonl"], split="train")
#-> DatasetGenerationError: Couldn't cast ...
```
Loading each file/source separately and then concatenate_datasets on the common columns (dropping non-shared ones with a warning, and raising if no shared data column remains) handles heterogeneous bundles cleanly, and extends to parquet by grouping on schema. For homogeneous bundles it's identical to the current single-pass.

I have this implemented and tested (directory-bundle slice of RFC #583). Happy to (a) push it as a follow-up on top of this PR, or (b) hand you the diff to fold in — your and @shanjiaz's call. The hf: handling and presets here are great as-is.

fynnsu (approved):

PR #680, opened 2026-06-29
minor fix(train): use global rank for dry-run checkpoint writes

orestis-z:
## Summary
- Use `rank == 0` instead of `local_rank == 0` for the dry-run checkpoint save guard
- In multi-node setups with a shared filesystem, `local_rank == 0` fires once per node, causing concurrent writes to the same `--save-path`

## Test plan
- [ ] Verify single-node dry-run still works (`python scripts/train.py --dry-run ...`)
- [ ] Verify multi-node dry-run writes only once

### DIFF
--- a/examples/train/dspark_qwen3_0_6b_sharegpt_online.sh
+++ b/examples/train/dspark_qwen3_0_6b_sharegpt_online.sh
@@ -1,0 +1,111 @@
+#!/bin/bash
+# Online DSpark Training Script
+#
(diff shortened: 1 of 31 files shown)


orestis-z:
Closing — the underlying bug (using `local_rank` instead of global rank for the dry-run checkpoint guard) was already fixed as a side effect of #656, which refactored the rank access to use `get_rank()` (returns `dist.get_rank()`, i.e. global rank). The PR diff is now empty after merging main.

PR #620, opened 2026-06-18
Change default drafter architecture from Llama to Qwen3

orestis-z:
## Summary
- Change the default drafter architecture from Llama to Qwen3 in the training script (`--draft-arch`) and Eagle3 config (`transformer_layer_config`)
- Qwen3 is now the recommended default for new speculator training

## Blocked by
- https://github.com/vllm-project/vllm/pull/43132

## Jira
- [INFERENG-8024](https://issues.redhat.com/browse/INFERENG-8024)

## Test plan
- [ ] Verify `--draft-arch` defaults to `qwen3` when not specified
- [ ] Verify `Eagle3SpeculatorConfig` uses `Qwen3Config` as default `transformer_layer_config`
- [ ] Run training with default args and confirm Qwen3 layers are used
- [ ] Run existing tests

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #681, opened 2026-06-29
Remove stale vLLM compatibility warning for draft arch

orestis-z:
## Summary
- Remove the stale vLLM compatibility warning in `create_transformer_layer_config` that was missed by #620
- Both llama and qwen3 are supported in vLLM, so the warning is no longer needed

## Test plan
- [ ] Run training with default args — no spurious warning emitted


orestis-z:
Closing — the upstream warning was already updated in a later commit to only warn for architectures other than llama/qwen3, which is valid. Nothing left to fix here.

PR #685, opened 2026-06-29
feat: Domino head for DFlash

### LINKED ISSUE #664: [RFC]: Add Domino support as a DFlash-based speculator
## Motivation

I would like to propose adding Domino support to `vllm-project/speculators`.

Domino has already been integrated in SpecForge, and the implementation can be used as a reference:

https://github.com/sgl-project/SpecForge/pull/571

## What is Domino?

Domino can be viewed as a DFlash extension rather than a completely new speculator family.

DFlash drafts a whole block of tokens in parallel, which gives low drafting overhead. However, because the tokens inside the block are generated in parallel, DFlash has weaker modeling of causal dependencies among draft tokens.

Domino keeps the DFlash-style parallel draft backbone, but adds a lightweight causal correction head. The head uses previous draft tokens to refine the base logits, improving the quality of the drafted tokens while keeping most of DFlash's parallel drafting efficiency.

In short:

- DFlash: parallel draft backbone → base logits → draft tokens
- Domino: parallel draft backbone → causal correction head → refined logits → draft tokens

## Difference from DFlash

The main difference is that Domino adds causal refinement on top of DFlash.

DFlash mainly optimizes drafting speed by generating the whole block in parallel. Domino keeps this advantage, but improves intra-block dependency modeling with a lightweight head. This should lead to higher acceptance length with only small extra overhead.

## Benefits

1. Reuses the existing DFlash infrastructure in speculators.
2. Improves draft quality and acceptance length compared with plain DFlash.
3. Keeps the main drafting computation parallel, so the additional cost should be small.
4. SpecForge already has a working implementation, which can be used as a reference.

## References

- SpecForge Domino PR: https://github.com/sgl-project/SpecForge/pull/571
- Domino paper: https://arxiv.org/abs/2605.29707
- Domino code: https://github.com/jianuo-huang/Domino

## Possible implementation direction

Domino could be added as a DFlash submode, for example:

- `projector_type = "domino"`
- reuse the existing DFlash backbone
- add a lightweight causal correction head
- train with both base DFlash logits and Domino-refined logits
- report both base acceptance length and final Domino acceptance length

### Any Other Things.

_No response_

arnabwithab:
## 1. Purpose

Closes #664

Add Domino causal correction head as a `projector_type` sub-mode of the existing DFlash draft model (issue #664). Domino extends the DFlash parallel draft backbone with a lightweight GRU-based correction head that refines base logits using causal state from previous draft tokens, with a scheduled dual-loss (base + final) that transitions from 100% base to 100% refined over training.

### 2.1 Domino Head
For `projector_type="domino"`, `DFlashDraftModel` adds:
- A GRU causal encoder (`prefix_gru`) over previous draft-token embeddings.
- A low-rank logit correction MLP (`embed_proj`) with bottleneck dimension `emb_dim`.
- A residual correction path that adds the Domino logit deltas to the parallel base logits at suffix positions `[suffix_start, block_size)`.
- 
### 2.2 Training Objective
Domino jointly supervises:
- `final_loss`: cross-entropy on Domino-refined logits.
- `base_loss`: cross-entropy on backbone-only base logits.
The loss uses a base-anchor curriculum:
loss = (1 - lambda_base) * final_loss + lambda_base * base_loss
`lambda_base` linearly decays from `lambda_base_start` to 0 over `lambda_base_decay_steps`, so early training keeps the parallel backbone strong and later training lets the Domino correction head take over.

### 2.3 Metrics
The training loop reports:
- `final_loss`, `final_full_acc` — Domino-refined logit quality.
- `base_loss`, `full_acc` — backbone-only logit quality.
- `position_X_acc` — per-position accuracy (from refined logits).
These follow the existing `*_sum` / `*_total` pattern consumed by `normalize_counted_metrics`.

## 3. Files Changed
| File | Action |
|------|--------|
| `src/speculators/models/dflash/domino.py` | **NEW** — `DominoHead` module (GRU + MLP correction head) |
| `src/speculators/models/dflash/config.py` | Added 8 Pydantic fields for Domino configuration |
| `src/speculators/models/dflash/__init__.py` | Added `DominoHead` export |
| `src/speculators/models/dflash/core.py` | Wired `DominoHead` in `__init__` and dual-loss in `forward` |
| `src/speculators/train/trainer.py` | Inject `global_step` into model forward calls |
| `scripts/train.py` | Added 7 CLI args: `--projector-type`, `--gru-hidden-dim`, `--emb-dim`, `--pure-draft-prefix-len`, `--shift-label`, `--domino-lambda-start`, `--domino-lambda-decay-steps` |
| `tests/unit/models/test_domino_head.py` | **NEW** — 14 unit tests |
## 4. Design Constraints
- Purely additive — existing `projector_type="dflash"` path is unchanged.
- No new `SpeculatorModel.register()` entry; Domino is a config option within `"dflash"`.
- `global_step` defaults to `0` for non-trainer callers.
- All new metrics follow `*_sum` / `*_total` pattern.
- `shift_label=True` → `suffix_start = pure_draft_prefix_len`; `shift_label=False` → `suffix_start = 1 + pure_draft_prefix_len`.
- `lambda_base = 0.0` when `lambda_base_decay_steps` is `None`.

## 5. Usage
```bash
torchrun --standalone --nproc_per_node=8 scripts/train.py \
    --verifier-name-or-path Qwen/Qwen3-8B \
    --speculator-type dflash \
    --projector-type domino \
    --block-size 16 \
    --num-layers 5 \
    --target-layer-ids 1 9 17 25 33 \
    --gru-hidden-dim 1024 \
    --emb-dim 256 \
    --pure-draft-prefix-len 1 \
    --domino-lambda-start 1.0 \
    --domino-lambda-decay-steps 15000 \
    --mask-token-id 151669 \
    --data-path ./output \
    --save-path ./output/checkpoints \
    --epochs 5 \
    --lr 1e-4 \
    --dflash-decay-gamma 7.0
```

## 6. Validation

```
make quality                                                # ruff + mypy + mdformat — all pass
uv run pytest tests/unit/models/test_domino_head.py -v      # 14/14 pass
uv run pytest tests/unit/models/test_dflash_metrics.py -v   # 19/19 pass (backward compat)
```
## 7. Benchmarks

I lack access to the GPU cluster required to run a full training run and obtain
loss/accuracy curves. 

The implementation has been validated via `make quality`
and unit tests; a maintainer with hardware access should verify training
converges by running the command in the Usage section and confirming loss
decreases and accuracy increases over the first ~500 steps.

## 8. Checklist

I have filled in:


src/speculators/models/dflash/core.py
orestis-z, line 1: what's the reason you removed the ` # shape: [1,total_seq_len,...]` comments?

docs/user_guide/algorithms/dflash.md
```diff
@@ -27,6 +27,19 @@ Pretrained DFlash speculator models are available on HuggingFace from the [RedHa
...
 
+## Domino Variant
+
+Domino is a lightweight causal correction head that can be added on top of the DFlash backbone. Instead of using the raw DFlash base logits at every block position, Domino refines the suffix positions with a single-layer GRU that encodes previous-token embeddings, concatenates the GRU output with the backbone hidden states, and projects through a bottleneck MLP to produce additive logit deltas.
+
+|                           | DFlash               | DFlash + Domino                                              |
+| ------------------------- | -------------------- | ------------------------------------------------------------ |
+| **Draft path**            | Base logits → argmax | Base logits → GRU correction → refined logits                |
+| **Per-position accuracy** | Good                 | Better (GRU conditions on prior tokens)                      |
+| **Training cost**         | ~1×                  | ~1.05× (negligible overhead)                                 |
```
orestis-z: where does this number come from? maybe worth referencing

docs/cli/train.md
```diff
@@ -156,6 +156,22 @@ torchrun --standalone --nproc_per_node=4 scripts/train.py \
...
+- **`--emb-dim`** (int, default: `256`) Bottleneck dimension for the Domino embed projection MLP that combines GRU output with backbone hidden states.
+
+- **`--pure-draft-prefix-len`** (int, default: `1`) Number of leading block positions (after the anchor) that use pure DFlash logits without Domino correction. The suffix positions are refined by the Domino head.
+
+- **`--shift-label`** (flag, default: `True`) Shift labels by 1 so the first predicted position is anchor+1. Enabled by default for Domino (aligns with DFlash's implicit target shift via `torch.roll`).
+
+- **`--domino-lambda-start`** (float, default: `1.0`) Initial weight of the base loss in the Domino dual-loss schedule. The combined loss is `(1-λ) * final_loss + λ * base_loss`.
+
+- **`--domino-lambda-decay-steps`** (int, default: `None`) Number of training steps over which lambda_base decays from `--domino-lambda-start` to 0. If not set, no decay is applied (lambda_base stays at `--domino-lambda-start`). Suggested value: ~15000 for typical runs.
+
```
orestis-z, line 194: There's an inconsistency of prefixing only 2 of 6 domino-specific args. I'd suggest prefixing them all with `--domino-*` since they're all domino specific

orestis-z (commented):
Thanks for your work @Eros483 !

The unit tests on DominoHead are solid, but none of the 14 tests exercise the integration points — so two bugs slip through:

1. `--domino-lambda-start` / `--domino-lambda-decay-steps` are silently ignored because `_build_base_config_kwargs` looks up `lambda_base_start` / `lambda_base_decay_steps` (name mismatch with argparse).
2. With defaults (`lambda_base_decay_steps=None`), `lambda_base` stays at `1.0`, so `loss = 0*final_loss + 1*base_loss` — the domino head gets zero gradient and never trains. (PR description says `lambda_base = 0.0` in this case, which contradicts the code.)

The existing `tests/integration/models/test_model_forward.py` pattern would catch these bugs with minimal additions. The framework is already set up — `make_dflash_model` in `conftest.py` just needs a `projector_type` `kwarg`, and a `DOMINO_SPEC` in `ALL_SPECS` gives you the shared `TestTraining` / `TestMultiBatch` coverage for free.

  Concretely:

  1. Extend the factory (`tests/integration/conftest.py`):

```python3
  def make_dflash_model(
      *,
      draft_vocab_size: int = 64,
      block_size: int = 4,
      max_anchors: int = 8,
      projector_type: str = "dflash",
      draft_attn_impl: str | None = None,
      device: str = "cuda:0",
      dtype: torch.dtype = torch.bfloat16,
      **domino_kwargs,
  ) -> DFlashDraftModel:
      config = DFlashSpeculatorConfig(
          ...,
          projector_type=projector_type,
          **domino_kwargs,
      )
      ...
```

  2. Add to shared specs (`test_model_forward.py`):

```python3
  DOMINO_SPEC = ModelSpec(
      name="domino",
      factory=partial(make_dflash_model, projector_type="domino"),
  )
```

  Adding this to `ALL_SPECS` gives forward+backward, multi-batch, and varying loss-mask coverage automatically — the same tests that cover DFlash, Eagle3, and PEagle.

  3. Add domino-specific param tests alongside `TestDFlashParams`:

```python3
  @requires_cuda
  class TestDominoParams:
      def test_domino_head_receives_gradient(self):
          """Domino head must get nonzero gradient from the training loss."""
          model = make_dflash_model(projector_type="domino")
          samples = _make_samples([128])
          batch = make_batch(max_len=MAX_LEN, samples=samples, hidden_size=HIDDEN_SIZE)
          _, loss, metrics = model(**batch)
          loss.backward()
          grad = model.domino_head.prefix_gru.weight_ih_l0.grad
          assert grad is not None and grad.abs().sum() > 0

      @pytest.mark.parametrize("lambda_base_start,lambda_base_decay_steps", [
          (1.0, 100),   # full decay schedule
          (0.5, None),   # constant lambda
          (0.0, None),   # pure final loss
      ])
      def test_lambda_schedule_variants(self, lambda_base_start, lambda_base_decay_steps):
          model = make_dflash_model(
              projector_type="domino",
              lambda_base_start=lambda_base_start,
              lambda_base_decay_steps=lambda_base_decay_steps,
          )
          samples = _make_samples([128])
          batch = make_batch(max_len=MAX_LEN, samples=samples, hidden_size=HIDDEN_SIZE)
          _, loss, metrics = model(**batch, global_step=50)
          assert loss.isfinite()
          loss.backward()

      def test_domino_reports_dual_metrics(self):
          model = make_dflash_model(projector_type="domino")
          samples = _make_samples([128])
          batch = make_batch(max_len=MAX_LEN, samples=samples, hidden_size=HIDDEN_SIZE)
          _, _, metrics = model(**batch)
          for key in ["base_loss_sum", "final_loss_sum", "final_full_acc_sum"]:
              assert key in metrics, f"Missing domino metric: {key}"

      @pytest.mark.parametrize("pure_draft_prefix_len", [0, 1, 3])
      def test_varying_prefix_len(self, pure_draft_prefix_len):
          model = make_dflash_model(
              projector_type="domino",
              pure_draft_prefix_len=pure_draft_prefix_len,
          )
          samples = _make_samples([128])
          batch = make_batch(max_len=MAX_LEN, samples=samples, hidden_size=HIDDEN_SIZE)
          _, loss, _ = model(**batch)
          assert loss.isfinite()
          loss.backward()
```

src/speculators/models/dflash/core.py
```diff
@@ -180,6 +189,13 @@ def _build_base_config_kwargs(
             "aux_hidden_state_layer_ids": target_layer_ids,
             "mask_token_id": kwargs.get("mask_token_id"),
             "sliding_window_non_causal": kwargs.get("sliding_window_non_causal", False),
+            "projector_type": kwargs.get("projector_type", "dflash"),
+            "shift_label": kwargs.get("shift_label", True),
+            "pure_draft_prefix_len": kwargs.get("pure_draft_prefix_len", 1),
+            "emb_dim": kwargs.get("emb_dim", 256),
+            "gru_hidden_dim": kwargs.get("gru_hidden_dim", 1024),
+            "lambda_base_start": kwargs.get("lambda_base_start", 1.0),
+            "lambda_base_decay_steps": kwargs.get("lambda_base_decay_steps"),
```
orestis-z: Argparse stores `--domino-lambda-start` as `domino_lambda_start`, but `_build_base_config_kwargs` looks up `lambda_base_start`. Same for `--domino-lambda-decay-steps` vs `lambda_base_decay_steps`. The user's CLI values are silently dropped and defaults are always used.

src/speculators/models/dflash/core.py
orestis-z, line 1: `compute_metrics` is called once before the `if self.projector_type == "domino"` branch, then called again identically inside both branches. The first call's results (including its autograd graph) are always discarded.

docs/user_guide/algorithms/dflash.md
```diff
@@ -27,6 +27,19 @@ Pretrained DFlash speculator models are available on HuggingFace from the [RedHa
...
+| **Inference**             | Same                 | Same (Domino is training-only; fused into base at inference) |
```
orestis-z: Is the GRU fused into the linear head?

orestis-z:
@Eros483 Once the bugs are fixed we can run a real training smoke test on a GPU cluster.

E2E (`tests/e2e/smoke/test_offline_training.py`) parametrized with eagle3, dflash, and peagle as separate entries. Adding domino is just another tuple:
```python3
          (TEXT_MODEL, "sharegpt", "dflash",
           ["--block-size", "8", "--max-anchors", "256", "--num-layers", "3",
            "--projector-type", "domino", "--domino-lambda-decay-steps", "100"],
           [1, 13, 25]),
```

arnabwithab:
> Thanks for your work @Eros483 !
> 
> The unit tests on DominoHead are solid, but none of the 14 tests exercise the integration points — so two bugs slip through:
> 
> 1. `--domino-lambda-start` / `--domino-lambda-decay-steps` are silently ignored because `_build_base_config_kwargs` looks up `lambda_base_start` / `lambda_base_decay_steps` (name mismatch with argparse).
> 2. With defaults (`lambda_base_decay_steps=None`), `lambda_base` stays at `1.0`, so `loss = 0*final_loss + 1*base_loss` — the domino head gets zero gradient and never trains. (PR description says `lambda_base = 0.0` in this case, which contradicts the code.)
> 
> The existing `tests/integration/models/test_model_forward.py` pattern would catch these bugs with minimal additions. The framework is already set up — `make_dflash_model` in `conftest.py` just needs a `projector_type` `kwarg`, and a `DOMINO_SPEC` in `ALL_SPECS` gives you the shared `TestTraining` / `TestMultiBatch` coverage for free.
> 
> Concretely:
> 
> 1. Extend the factory (`tests/integration/conftest.py`):
> 
> ```python
>   def make_dflash_model(
>       *,
>       draft_vocab_size: int = 64,
>       block_size: int = 4,
>       max_anchors: int = 8,
>       projector_type: str = "dflash",
>       draft_attn_impl: str | None = None,
>       device: str = "cuda:0",
>       dtype: torch.dtype = torch.bfloat16,
>       **domino_kwargs,
>   ) -> DFlashDraftModel:
>       config = DFlashSpeculatorConfig(
>           ...,
>           projector_type=projector_type,
>           **domino_kwargs,
>       )
>       ...
> ```
> 
> 2. Add to shared specs (`test_model_forward.py`):
> 
> ```python
>   DOMINO_SPEC = ModelSpec(
>       name="domino",
>       factory=partial(make_dflash_model, projector_type="domino"),
>   )
> ```
> 
> Adding this to `ALL_SPECS` gives forward+backward, multi-batch, and varying loss-mask coverage automatically — the same tests that cover DFlash, Eagle3, and PEagle.
> 
> 3. Add domino-specific param tests alongside `TestDFlashParams`:
> 
> ```python
>   @requires_cuda
>   class TestDominoParams:
>       def test_domino_head_receives_gradient(self):
>           """Domino head must get nonzero gradient from the training loss."""
>           model = make_dflash_model(projector_type="domino")
>           samples = _make_samples([128])
>           batch = make_batch(max_len=MAX_LEN, samples=samples, hidden_size=HIDDEN_SIZE)
>           _, loss, metrics = model(**batch)
>           loss.backward()
>           grad = model.domino_head.prefix_gru.weight_ih_l0.grad
>           assert grad is not None and grad.abs().sum() > 0
> 
>       @pytest.mark.parametrize("lambda_base_start,lambda_base_decay_steps", [
>           (1.0, 100),   # full decay schedule
>           (0.5, None),   # constant lambda
>           (0.0, None),   # pure final loss
>       ])
>       def test_lambda_schedule_variants(self, lambda_base_start, lambda_base_decay_steps):
>           model = make_dflash_model(
>               projector_type="domino",
>               lambda_base_start=lambda_base_start,
>               lambda_base_decay_steps=lambda_base_decay_steps,
>           )
>           samples = _make_samples([128])
>           batch = make_batch(max_len=MAX_LEN, samples=samples, hidden_size=HIDDEN_SIZE)
>           _, loss, metrics = model(**batch, global_step=50)
>           assert loss.isfinite()
>           loss.backward()
> 
>       def test_domino_reports_dual_metrics(self):
>           model = make_dflash_model(projector_type="domino")
>           samples = _make_samples([128])
>           batch = make_batch(max_len=MAX_LEN, samples=samples, hidden_size=HIDDEN_SIZE)
>           _, _, metrics = model(**batch)
>           for key in ["base_loss_sum", "final_loss_sum", "final_full_acc_sum"]:
>               assert key in metrics, f"Missing domino metric: {key}"
> 
>       @pytest.mark.parametrize("pure_draft_prefix_len", [0, 1, 3])
>       def test_varying_prefix_len(self, pure_draft_prefix_len):
>           model = make_dflash_model(
>               projector_type="domino",
>               pure_draft_prefix_len=pure_draft_prefix_len,
>           )
>           samples = _make_samples([128])
>           batch = make_batch(max_len=MAX_LEN, samples=samples, hidden_size=HIDDEN_SIZE)
>           _, loss, _ = model(**batch)
>           assert loss.isfinite()
>           loss.backward()
> ```

Thank you for the review, really appreciate it. 
With context to the specific bugs mentioned here, the fixes I made are:

1. Update all kwarg look up keys, to match the argeparse attribute names.
2. fixed the lambda_base never decaying -> zero gradient, by changing the default decay steps to 30,000 (referencing the SGLang implementation, where 6 epochs translated to that, we can change it, when we run a smoke test or something)
3. added new tests, and modified existing tests.

also added the tuple for the domino smoke test.

Will be updated with these changes in the next push.

src/speculators/models/dflash/core.py
Eros483: Restored the exact style of comments previously. Will be present in the next push.

docs/user_guide/algorithms/dflash.md
Eros483: Redacted. The intention was that it would be a negligible overhead, due to the additional GRU and MLP only taking up a tiny fraction of the FLOPs of the transformer. 

Will be present in the next push.

docs/cli/train.md
Eros483: Fixed with the domino prefix.

Will be present in the next push.

src/speculators/models/dflash/core.py
Eros483: fixed via the domino specific args being renamed with the domino prefix.

Will be added in the next push.

src/speculators/models/dflash/core.py
Eros483: The code had a stale metrics computation call (i.e orphaned call), that was being discarded as well. Removed it, makes no difference.

Will be added in the next push.

docs/user_guide/algorithms/dflash.md
Eros483: Updated the documentation to be more verbose about what actually happens.

Domino is a training time helper, i.e helps the lm head, learn better representations during training, and is discarded/not used at inference. 

Docs be modified in the next push.

orestis-z:
### E2E smoke test result

I ran the domino e2e smoke test (`test_offline_smoke[...-extra_train_args3-target_layer_ids3]`) on an A100-80GB. Data prep, hidden-state extraction, and training all pass, but **the vLLM inference validation step fails**:

```
KeyError: 'domino_head.embed_proj.0.weight'
```

The checkpoint includes `domino_head.*` weights (saved via `model.save_pretrained()` → `model.state_dict()`), but vLLM's `qwen3_dflash.py` weight loader doesn't recognize them. Since domino is training-only, these weights need to be stripped at save time.

### Fix

Add the 4 domino_head parameter keys to `_keys_to_ignore_on_save` and `_keys_to_ignore_on_load_missing` in `DFlashDraftModel` — this follows the existing pattern used for `verifier_lm_head.weight` / `verifier_norm.weight`:

```diff
--- a/src/speculators/models/dflash/core.py
+++ b/src/speculators/models/dflash/core.py
@@ -35,10 +35,20 @@
         "verifier_lm_head.weight",
         "t2d",
         "d2t",
+        # Domino head is training-only; weights are not saved to checkpoints.
+        "domino_head.prefix_gru.weight_ih_l0",
+        "domino_head.prefix_gru.weight_hh_l0",
+        "domino_head.embed_proj.0.weight",
+        "domino_head.embed_proj.2.weight",
     ]
     _keys_to_ignore_on_save: ClassVar[list[str]] = [  # type: ignore[misc,assignment]
         "verifier_lm_head.weight",
         "verifier_norm.weight",
+        # Domino head is training-only; not needed at inference.
+        "domino_head.prefix_gru.weight_ih_l0",
+        "domino_head.prefix_gru.weight_hh_l0",
+        "domino_head.embed_proj.0.weight",
+        "domino_head.embed_proj.2.weight",
     ]
```

With this fix applied, the full e2e smoke test passes end-to-end (including vLLM inference validation). The vanilla dflash test also passes — no regression.

orestis-z:
@Eros483 running an experiment baseline vs domino on our cluster and will report back

arnabwithab:
Ran the smoke test on my own hardware, cuz i looked at the test and realised i could rut it with a few minor tweaks to prevent the OOM.

Environment:
- GPU: NVIDIA RTX 4060 (8 GB, Compute Capability 8.9)
- Driver: 580.95.05, CUDA 13.0
- torch 2.12.1, transformers 5.12.1
- vLLM 0.24.0, flashinfer 0.6.12

Training metrics:
- Final val loss: 2.603
- Per-position accuracy: ~5% across 7 draft positions
- Acceptance length: 1.07 (expected for 1-epoch smoke test with 50 samples)

Full pipeline ran in just around 2 minutes.

The tweaks I used:
- VLLM_USE_FLASHINFER_SAMPLER=0 (flashinfer popped up incompatibility issues with CUDA 13.0)
- set gpu_memory_utilization=0.5

The concluding logs for your reference:
```
2026-07-02 20:41:53.738 | INFO     | tests.e2e.utils:run_vllm_engine:438 - outputs_token_ids: [[151667, 198, 32313, 11, 358, 1184, 311, 3270, 264, 7868, 2711, 729, 304, 13027, 13, 6771, 752, 1744, 911, 1246, 311, 5486, 419, 13, 17718, 2711, 374, 1483, 311, 1477, 458, 2392, 304, 264, 10615, 1140, 13, 576, 6770, 4522, 374, 311, 18822, 21749, 279, 2711, 9873, 304, 4279, 13], [151667, 198, 32313, 11, 773, 358, 1184, 311, 10339, 1246, 65736, 47116, 4278, 13, 88190, 11, 358, 6099, 429, 304, 7377, 7699, 8692, 11, 5310, 304, 7699, 47116, 11, 1052, 525, 2155, 5413, 13, 6771, 752, 1744, 13, 358, 1744, 65736, 47116, 374, 264, 14762, 1380, 279, 24551, 5711, 264], [151667, 198, 32313, 11, 358, 1184, 311, 1855, 264, 42578, 2504, 729, 13, 6771, 752, 1191, 553, 88646, 1246, 42578, 10010, 975, 13, 2379, 11136, 6685, 315, 5248, 13617, 11, 1817, 448, 656, 12, 53103, 23783, 13, 576, 6770, 5944, 5646, 264, 2309, 44439, 6529, 6193, 11, 264, 656, 12]]
2026-07-02 20:41:53.738 | INFO     | tests.e2e.utils:run_vllm_engine:439 - metrics_dict: {'total_num_output_tokens': 150, 'num_drafts': 138, 'num_draft_tokens': 966, 'num_accepted_tokens': 10, 'acceptance_length': 1.0724637681159421, 'acceptance_at_token_0': 0.07246376811594203, 'acceptance_at_token_1': 0.0, 'acceptance_at_token_2': 0.0, 'acceptance_at_token_3': 0.0, 'acceptance_at_token_4': 0.0, 'acceptance_at_token_5': 0.0, 'acceptance_at_token_6': 0.0}
PASSED
```

cc: @orestis-z

orestis-z:
## Experiment: DFlash vs Domino — 8x H100 Acceptance Length Comparison

I reproduced the paper's comparison from scratch (arXiv:2605.29707) to validate the Domino head implementation before merging. Both models trained identically except for `--projector-type`.

### Setup

| Parameter | Value |
|-----------|-------|
| Model | Qwen/Qwen3-8B |
| block_size | 16 |
| num_layers | 5 |
| target_layer_ids | 1 9 17 25 33 |
| draft_vocab_size | 151936 (full) |
| lr | 6e-4 |
| epochs | 3 |
| total_seq_len | 3072 |
| Dataset | ShareGPT (120K samples) |
| GPUs | 2 vLLM + 6 FSDP training (H100-80GB) |

**Domino extras:** `--domino-gru-hidden-dim 1024 --domino-emb-dim 256 --domino-pure-draft-prefix-len 1 --domino-lambda-start 1.0 --domino-lambda-decay-steps 5000`

### Acceptance Length Results

| Subset | DFlash | Domino | Delta |
|--------|--------|--------|-------|
| HumanEval | 2.21 | 1.57 | -29.1% |
| math_reasoning | 2.36 | 1.56 | -33.8% |
| qa | 1.92 | 1.42 | -26.2% |
| question | 2.05 | 1.47 | -28.5% |
| rag | 1.99 | 1.36 | -31.6% |
| summarization | 1.84 | 1.33 | -27.7% |
| tool_call | 2.09 | 1.48 | -29.1% |
| translation | 1.98 | 1.36 | -31.4% |
| writing | 2.05 | 1.46 | -29.0% |
| **MEAN** | **2.055** | **1.444** | **-29.7%** |

### Per-Position Acceptance (averaged across 9 subsets)

| Position | DFlash | Domino |
|----------|--------|--------|
| 0 | 57.3% | 39.9% |
| 1 | 26.3% | 4.3% |
| 2 | 11.6% | 0.3% |
| 3+ | >0% | ~0% |

### Training Metrics

During training, Domino's **final loss** (correction head active) looked excellent: val_loss=0.243, full_acc=78.2%. But the **base loss** (backbone only — what matters at inference) told a different story: base_full_acc=5.1% vs DFlash's 28.6%.

### Analysis

**The Domino correction head parasitizes the backbone.** During training, the GRU head handles the hard suffix positions so effectively that the backbone never learns to predict them on its own. At inference, the correction head is discarded, and the backbone alone performs 30% worse than vanilla DFlash.

This is a fundamental issue with the dual-loss schedule — the base loss lambda decays too fast (from 1.0→0 in 5K steps), giving the backbone insufficient standalone learning signal. The paper's much longer decay (30K steps over 266K total) may mitigate this, but with our 30K total steps, the backbone barely gets any standalone training pressure.

### Potential fixes to explore (not blocking merge, but should be investigated):

1. **Slower lambda decay** — proportionally match the paper's ratio (the paper decays over ~11% of training; our 5K/30K is ~17%, but the absolute step count may matter more)
2. **Gradient stopping** — prevent GRU gradients from flowing into the shared backbone
3. **Higher `--domino-pure-draft-prefix-len`** — force more positions to train without correction

### Checkpoint save fix

I also found and fixed a bug: `_keys_to_ignore_on_save` was missing the 4 `domino_head.*` weight keys, causing them to leak into saved checkpoints and break vLLM inference loading. The fix adds them to both `_keys_to_ignore_on_save` and `_keys_to_ignore_on_load_missing` in `core.py`.

---

**Bottom line:** The Domino implementation is mechanically correct (architecture matches the paper, training works, correction head does what it claims), but the trained backbone alone underperforms vanilla DFlash by ~30% on our setup. The checkpoint save bug also needs fixing. I'd suggest:
1. Merge with the checkpoint save fix applied ✅
2. Document that Domino requires careful tuning of `lambda_decay_steps` relative to total training steps
3. Follow up with experiments using longer training and slower decay to reproduce the paper's results

cc @Eros483

arnabwithab:
So, the checkpoint save fix was already fixed in  a previous commit, (commit: 2847082), so I'm not sure what's up with that. Seems ok to me, do let me know if there's still an issue.

Now, when it comes to the mismatch in performance, that's a real issue. I went back and referenced the SGLang implementation because that was able to reproduce the metrics from the paper.

So what I'm noticing, is that we have a target alignment bug in our implementation.

Current behavior
Per draft block of size block_size:

|Position |	Current target	|In loss? | 	Gets Domino help?|
| --- | --- | --- | --- |
| 0 (anchor a) | predicts token a (trivial) | No cuz masked  | No (suffix_start=1) |
|1..7 | predicts token a+1...a+7 | Yes | Yes |

Essentially the issue is that the backbone on its own alwayus have a trivial target, and is excluded from loss. Which is why domino is taking up all the training signals, and the backbone only makes predictions that domino handles anyways.

So the fix, that im gonna implement is, that position 0 now predicts position a+1, i.e the next token after the anchor. Position 0 is included in the loss, so that the backbone gets a gradient signal there, but gets no domino correction.

Will implement this in the next push.

cc: @orestis-z

orestis-z:
## Experiment v2: Domino with backbone alignment fix (`4b260af`)

Re-trained Domino on the latest PR code (commit `c73fa81`) which includes the `shift_targets` alignment fix. Same hyperparameters as v1. DFlash baseline kept from v1 (trained pre-fix, on code matching mainline behavior).

### Training Metrics (v2 vs v1)

| Metric | v1 (pre-fix) | v2 (with fix) |
|--------|-------------|---------------|
| val/base_full_acc (backbone only) | 5.1% | 12.2% |
| val/final_full_acc (with correction head) | 78.2% | 45.4% |
| val_loss | 0.243 | 0.521 |

The alignment fix improved backbone accuracy from 5.1% → 12.2% — confirming the targets were indeed misaligned in v1.

### Acceptance Length Results (v2)

| Subset | DFlash | Domino v1 | Domino v2 | v2 vs DFlash |
|--------|--------|-----------|-----------|-------------|
| HumanEval | 2.21 | 1.57 | 1.12 | -49.1% |
| math_reasoning | 2.36 | 1.56 | 1.13 | -52.2% |
| qa | 1.92 | 1.42 | 1.10 | -42.9% |
| question | 2.05 | 1.47 | 1.10 | -46.3% |
| rag | 1.99 | 1.36 | 1.09 | -45.2% |
| summarization | 1.84 | 1.33 | 1.10 | -40.4% |
| tool_call | 2.09 | 1.48 | 1.11 | -47.1% |
| translation | 1.98 | 1.36 | 1.12 | -43.6% |
| **MEAN** | **2.055** | **1.444** | **1.108** | **-46.1%** |

### Setup
- Code: commit `c73fa81` (latest PR head with alignment fix)
- Hardware: 8x H100-80GB (2 vLLM + 6 FSDP training)
- Hyperparameters identical to v1: Qwen3-8B, block_size=16, max_anchors=192, 3 epochs, lr=6e-4, ShareGPT 120K
- Domino: gru_hidden_dim=1024, emb_dim=256, lambda_decay_steps=5000

cc @Eros483

orestis-z:
## Investigation: Why Domino v2 performs worse than v1

### TL;DR

`shift_targets=True` (the "backbone alignment fix" in `4b260af`) creates a **train/inference off-by-one mismatch** that explains both the misleading training accuracy improvement and the worse inference results.

### The mismatch

In [`core.py` L378](https://github.com/vllm-project/speculators/blob/c73fa81/src/speculators/models/dflash/core.py#L378), when `shift_targets=True`:
```python
target_indices = anchored_block_indices + (1 if shift_targets else 0)
```

This makes the backbone at position `p` learn to predict the token at position `p+1`.

But in vLLM's DFlash inference ([`speculator.py` L449-471](https://github.com/vllm/vllm/blob/main/vllm/v1/worker/gpu/spec_decode/dflash/speculator.py)), mask tokens at `query_pos = last_valid_pos + 1 + query_off` are sampled directly — the output at position `p` is used as the draft token **for position `p`**, not `p+1`.

| | What position `p` output predicts |
|---|---|
| **v1 training** (`shift_targets=False`) | Token at position `p` |
| **v2 training** (`shift_targets=True`) | Token at position `p+1` |
| **vLLM inference** | Token at position `p` |

Every draft token proposed by the v2 backbone is actually its prediction for the *next* position — hence the very low acceptance rates.

### Why training metrics are misleading

v2's `val_full_acc=12.2%` (vs v1's `5.1%`) is computed against the *shifted* targets. Predicting token `p+1` from position `p` is standard next-token prediction — an easier task with more context. The metric improved, but it's measuring the wrong thing for DFlash inference.

### Experiment results

| Model | Acceptance Length | vs DFlash baseline |
|-------|------------------|--------------------|
| DFlash baseline | 2.055 | — |
| Domino v1 (no shift) | 1.444 | -29.7% |
| Domino v2 (shift_targets=True) | 1.108 | -46.1% |

### Recommended fix

`shift_targets` should be `False` (or removed). [`core.py` L441](https://github.com/vllm-project/speculators/blob/c73fa81/src/speculators/models/dflash/core.py#L441) currently passes `shift_targets=True` unconditionally — this is incorrect for both DFlash and Domino, since both use the same backbone at inference and vLLM expects unshifted position semantics.

If the intent was to include the anchor position in the Domino loss mask, that should be handled by adjusting the mask directly in the Domino branch, not by shifting the backbone's targets.

orestis-z:
Running a v3 experiment with the recommended fix applied:

1. `shift_targets=False` — backbone targets aligned with vLLM inference expectations
2. Domino loss mask always includes anchor positions (decoupled from `shift_targets`)
3. `suffix_start` uses the unshifted form

Same training setup and hyperparameters as v1/v2. Will post results once complete.

arnabwithab:
Been looking a bit more at the SGLang implementation.
One thing that stands out to me, is that their 6 epoch run, applies the lambda decay across all 30k steps, but in a decreasing fashion. So essentially decaying over the whole run. This is significantly different to our approach. Might be worth considering, to allow the decay over the whole run, instead of just a fraction of the steps.

additionally, i think this might end up being a issue on the inference side as well.
slot `i`'s output is supposed to be the draft for `anchor+i+1` not `anchor+i` whenever shift_label is True.

Ill go a bit deeper into the paper and the sgl code, and see if i can get some ideas that would work out well.

c: @orestis-z

orestis-z:
## v3 Experiment Results (with fix)

Applied the recommended fix:
1. `shift_targets=False` — backbone targets aligned with vLLM inference
2. Domino loss mask always includes anchor positions (decoupled from `shift_targets`)
3. `suffix_start` uses the unshifted form

### Results

| Subset | DFlash baseline | Domino v1 (no fix) | Domino v2 (shift_targets) | Domino v3 (with fix) |
|--------|:-:|:-:|:-:|:-:|
| HumanEval | 2.210 | 1.570 | 1.125 | 1.721 |
| math_reasoning | 2.357 | 1.560 | 1.126 | 1.765 |
| qa | 1.922 | 1.420 | 1.097 | 1.618 |
| question | 2.051 | 1.470 | 1.101 | 1.658 |
| rag | 1.993 | 1.360 | 1.092 | 1.619 |
| summarization | 1.841 | 1.330 | 1.097 | 1.559 |
| tool_call | 2.091 | 1.480 | 1.106 | 1.672 |
| translation | 1.978 | 1.360 | 1.116 | 1.592 |
| writing | 2.052 | 1.460 | — | 1.660 |
| **Mean** | **2.055** | **1.444 (-29.7%)** | **1.108 (-46.1%)** | **1.652 (-19.6%)** |

### Analysis

The fix confirms the off-by-one diagnosis: v3 (-19.6%) is substantially better than v2 (-46.1%) and v1 (-29.7%). However, the Domino correction head still degrades backbone quality by ~20% compared to vanilla DFlash.

### Training details

Same setup as v1/v2: Qwen3-8B verifier, 120K ShareGPT samples, 3 epochs, lr=6e-4, block_size=16, 6×H100 training. Best checkpoint: epoch 2, `val/full_acc=7.3%` (vs v1's 5.1%, v2's 12.2%).

orestis-z:
Running a v4 experiment with one Domino-specific change to better match the [Domino paper](https://arxiv.org/abs/2605.29707):

1. **Lambda decay over full run**: `--domino-lambda-decay-steps 30000` (paper linearly anneals λ from 1→0 across all training; v1-v3 used 5000 steps)

All other hyperparams (loss function, LR schedule, weight decay, gamma, warmup) kept identical to the DFlash baseline for an apples-to-apples comparison. Plus the v3 fix (`shift_targets=False` + decoupled anchor mask).

### Remaining discrepancies vs paper/SGLang

| Discrepancy | Paper/SGLang | Our setup | Source | Fair comparison? | Easy fix? |
|---|---|---|---|---|---|
| Target shift | Shifted (+1): all 16 block positions useful | Unshifted: position 0 wasted (~6% less signal) | Paper | No (Domino-specific) | No (requires vLLM inference changes) |
| Loss function | Cross-entropy on hard token IDs | KL divergence on soft verifier logits | Paper | Yes (same for both) | Yes (`--loss-fn ce`) |
| Loss normalization | Divides by sum of decay-weighted tokens | Divides by count of valid tokens | SGLang/SpecForge impl | Yes (same for both) | No (code change) |
| LR schedule | Cosine | Linear | Paper | Yes (same for both) | Yes (`--scheduler-type cosine`) |
| Weight decay | 0.0 | 0.01 | Paper | Yes (same for both) | Yes (`--weight-decay 0.0`) |
| Warmup | 4% | 1% | Paper | Yes (same for both) | Yes (`--scheduler-warmup-steps`) |
| Decay gamma | 7.0 | 4.0 | SGLang/SpecForge impl | Yes (same for both) | Yes (`--dflash-decay-gamma 7.0`) |
| FP32 master weights | Custom BF16Optimizer with FP32 copies | Standard AdamW in BF16 | SGLang/SpecForge impl | Yes (same for both) | No (code change) |
| Dataset | 1.42M samples | 120K samples | Paper | Yes (same for both) | Yes (use paper dataset) |

orestis-z:
### Domino v4 results — lambda decay over full run

v4 changes vs v3: `--domino-lambda-decay-steps 30000` (full-run decay instead of 5000 steps)

| Experiment | Mean acceptance length | vs DFlash baseline |
|---|---|---|
| DFlash baseline | 2.055 | — |
| Domino v1 (no fix, KL, λ decay 5k) | 1.444 | -29.7% |
| Domino v3 (fix, KL, λ decay 5k) | 1.653 | -19.6% |
| **Domino v4 (fix, KL, λ decay 30k)** | **2.059** | **+0.2%** |

**Per-category breakdown (v4 vs DFlash baseline):**

| Category | DFlash | Domino v4 | Delta |
|---|---|---|---|
| HumanEval | 2.210 | 2.206 | -0.2% |
| math_reasoning | 2.357 | 2.330 | -1.1% |
| qa | 1.922 | 1.998 | +3.9% |
| question | 2.051 | 2.044 | -0.3% |
| rag | 1.993 | 1.982 | -0.5% |
| summarization | 1.841 | 1.857 | +0.9% |
| tool_call | 2.091 | 2.096 | +0.3% |
| translation | 1.978 | 1.969 | -0.5% |
| writing | 2.052 | 2.047 | -0.2% |

**Conclusion:** Full-run lambda decay closes the gap entirely — Domino v4 matches DFlash baseline. The Domino head (discarded at inference) doesn't hurt, but doesn't help either with current setup. The remaining [discrepancies](https://github.com/vllm-project/speculators/pull/685#issuecomment-4881584788) (target shift, loss function, LR schedule, etc.) may be needed to see the improvement reported in the paper.

arnabwithab:
So based on all the work we've done so far, here's my proposal for the next steps.

1. Do a new training run with the following configurations
- `--loss-fn ce`
- `--scheduler-type cosine`
- `--dflash-decay-gamma 7.0`

2. ill raise a PR on vllm for checking if changes in how we're running inference, would actually help. can merge it if it does help, else we can dump it.

    1. Read shift_label from speculator checkpoint (DFlashSpeculatorConfig.shift_label). This field already exists in the speculator config.
    2. When shift_label=True, propose block_size tokens instead of block_size - 1:
    - Current (shift_label=False, 15 tokens): draft_tokens = [logits[anchor+1], logits[anchor+2], ..., logits[anchor+15]]
    - New (shift_label=True, 16 tokens): draft_tokens = [logits[anchor+0], logits[anchor+1], ..., logits[anchor+15]]
    4. Files to modify (rough file map):
    - vllm/model_executor/models/speculators/ — DFlash model loader (read shift_label from config)
    - vllm/spec_decode/ or the attention/engine code that samples draft tokens — extend the proposal loop from range(block_size - 1) to range(block_size) when shift_label=True
    - The speculator checkpoint proposal_methods[0].speculative_tokens should reflect the actual proposal count
    5. Backward compatibility: When shift_label is absent or False, keep existing behavior (15 tokens). No breaking changes.

then we can run a fresh experiment!
let me know your thoughts.

cc: @orestis-z

orestis-z:
Thanks for the proposal!

**v5 is already running** — I'm training both a new DFlash baseline and a Domino run with ALL easy-fix hyperparams applied (CE loss, cosine schedule, weight-decay 0.0, 4% warmup, gamma 7.0). Will post results once both are evaluated.

orestis-z:
@Eros483 Go ahead with your shift_label proposal on the vLLM side — the plan looks good.

orestis-z:
Opened https://github.com/Eros483/speculators/pull/2 targeting this branch — adds `--normalize-loss-by-decay` option (cherry-pick of #720). This normalizes the loss denominator by the sum of decay weights instead of token count, producing a proper weighted average. This follows SGLang impl.

orestis-z:
## v5 Experiment Results — Paper-matched hyperparameters

Retrained both DFlash baseline and Domino with hyperparameters closer to the paper/SGLang implementation:
- `--loss-fn ce` (cross-entropy, was KL)
- `--scheduler-type cosine` (was linear)
- `--weight-decay 0.0` (was 0.01)
- `--scheduler-warmup-steps 1185` (4% warmup, was 1%)
- `--dflash-decay-gamma 7.0` (was 4.0)
- Domino: `--domino-lambda-decay-steps 30000` (full-run decay, same as v4)

### v5 Per-Subset Results

| Subset | DFlash v5 | Domino v5 | Delta |
|---|---|---|---|
| HumanEval | 2.228 | 2.200 | -1.2% |
| math_reasoning | 2.307 | 2.287 | -0.9% |
| qa | 1.931 | 1.898 | -1.7% |
| question | 2.021 | 2.007 | -0.7% |
| rag | 1.943 | 1.933 | -0.5% |
| summarization | 1.816 | 1.799 | -0.9% |
| tool_call | 2.085 | 2.053 | -1.5% |
| translation | 1.943 | 1.928 | -0.8% |
| writing | 2.026 | 1.998 | -1.4% |
| **Mean** | **2.033** | **2.011** | **-1.1%** |

### Full Experiment Summary (v1–v5)

| Experiment | Config changes vs DFlash baseline | Mean acceptance length | vs DFlash baseline |
|---|---|---|---|
| **DFlash baseline** | KL, linear, wd=0.01, γ=4.0 | **2.055** | — |
| Domino v1 | + Domino (no fix), λ decay 5k | 1.444 | -29.7% |
| Domino v3 | + Domino (with fix), λ decay 5k | 1.653 | -19.6% |
| Domino v4 | + Domino (with fix), λ decay 30k | 2.059 | +0.2% |
| **DFlash v5** | CE, cosine, wd=0.0, warmup 4%, γ=7.0 | **2.033** | -1.1% |
| **Domino v5** | DFlash v5 + Domino, λ decay 30k | **2.011** | -2.1% |

### Analysis

1. **DFlash v5 (2.033) ≈ DFlash baseline (2.055):** The paper-matched hyperparams (CE, cosine, γ=7.0) produce comparable results to our original setup (KL, linear, γ=4.0) — a marginal -1.1% delta.

2. **Domino v5 (2.011) slightly worse than DFlash v5 (2.033):** The Domino head causes a consistent -1.1% degradation across all subsets. With full-run λ decay (30k steps), the head barely modifies the backbone — but the slight interference still hurts rather than helps.

3. **Pattern across v1–v5:** With short λ decay (v1, v3), the Domino head parasitizes the backbone (up to -30%). With full-run decay (v4, v5), the head is effectively a no-op — neither helping nor hurting significantly. We haven't found a regime where Domino improves acceptance length.

### Remaining discrepancies vs paper

| Discrepancy | Paper/SGLang | Our setup |
|---|---|---|
| Target shift | Shifted (+1): all 16 block positions useful | Unshifted: position 0 wasted | 
| Dataset size | 1.42M samples | 120K samples |
| FP32 master weights | Custom BF16Optimizer with FP32 copies | Standard AdamW in BF16 |
| Loss normalization | Divides by sum of decay weights | Divides by token count |

The target shift is likely the most impactful remaining difference, as it changes what the model learns at every position. This requires a coordinated vLLM-side change (inference must interpret shifted predictions correctly).

cc @Eros483

orestis-z:
## vLLM-side Domino implementation already exists

Heads up @Eros483 — there's already a working vLLM implementation of Domino inference by @swylyu001 on the Domino RFC issue:

- **RFC issue**: https://github.com/vllm-project/vllm/issues/46707
- **Commit**: https://github.com/vllm-project/vllm/commit/4d51bd958e6ff9bfa827fa9c8d69a23819ebf97a

It keeps `method="dflash"` and treats Domino as an optional projector selected by `dflash_config.projector_type == "domino"`. The GRU correction head runs during `_generate_draft()` — pure Python, no kernel changes. It also handles `shift_label=true` checkpoints by including the bonus token hidden state during draft generation.

swylyu001 posted acceptance-length numbers on Qwen3-8B using the original `Huang2020/Qwen3-8B-Domino-b16` checkpoint:

| Draft | Dataset | Mean AL |
|---|---|---|
| DFlash (z-lab) | GSM8K | 6.10 |
| Domino (Huang2020) | GSM8K | 9.06 |
| DFlash (z-lab) | HumanEval | 5.82 |
| Domino (Huang2020) | HumanEval | 6.65 |

Note: @benchislett deprioritized the RFC on July 1 "due to DSpark", so it's not merged yet — but the code works and can be cherry-picked for local eval.

arnabwithab:
Ok, so I think I have the full approach we need to be taking now.

1. We need to save the domino weights we build during training.
    - The domino head is essentially being trained for correcting the backbone weights, which is why discarding them at inference results in the same performance as dflash.

2. We use both Commit 4d51bd9 which handles the shift_label proposal I mentioned earlier.

This allows us to have a v6 experiment where we essentially
1. Train a new model, with `shift_labels=True` and save domino weights.
2. Use the PR on vllm for inference. The existing commit leverages the domino weights correctly, and the shift_label change, correctly applies logic for refining all tokens properly.

I'll create a vllm fork with the commit applied (credit the author), and open a PR, if it helps we merge it, else we dump it.

Will push the necessary changes on speculators in the next push.

cc: @orestis-z

orestis-z:
## Domino v6 Results — FP32 optimizer + vLLM inference bugfixes

We haven't been able to reproduce the SpecForge/paper results yet where the Domino GRU head improves acceptance length. Below is a summary of the current state.

### Known remaining discrepancies vs SpecForge/paper

| Factor | Paper/SpecForge | Our setup |
|---|---|---|
| Dataset size | 1.42M samples | ~120K samples (~12× smaller) |
| Loss normalization | Divides by sum of decay weights | Divides by token count |
| Teacher forcing | Not discussed in paper | Training is teacher-forced (ground-truth tokens), inference is autoregressive — gap may affect GRU quality |

### v6 Setup

Same data/hyperparams as DFlash v5: Qwen3-8B verifier, 5 transformer layers, block_size=16, CE loss, cosine scheduler, lr=1e-4, wd=0.0, warmup=1185, gamma=7.0, lambda_base_start=1.0, lambda_decay=30k steps, gru_hidden_dim=1024, emb_dim=256, 3 epochs, 6× GPU FSDP2.

Fixes vs v5:
- FP32 master weights (autocast instead of model-level bfloat16 cast)
- FSDP2: shard `domino_head` before root `fully_shard` (compound module wrapping)
- Domino GRU weights saved to checkpoint (removed from `_keys_to_ignore_on_save`)

### Results

| Subset | [DFlash v5](https://huggingface.co/orestis-z/Qwen3-8B-speculators.dflash-v5) | [Domino no-shift, plain DFlash eval](https://huggingface.co/orestis-z/Qwen3-8B-speculators.domino-noshift-v6-dflash) | [Domino no-shift + GRU head](https://huggingface.co/orestis-z/Qwen3-8B-speculators.domino-noshift-v6) | [Domino shift + GRU head](https://huggingface.co/orestis-z/Qwen3-8B-speculators.domino-shift-v6) |
|---|---|---|---|---|
| HumanEval | 2.228 | **2.416** | 1.697 | 1.575 |
| math_reasoning | 2.307 | **2.521** | 1.711 | 1.564 |
| qa | 1.931 | **2.153** | 1.606 | 1.393 |
| question | 2.021 | **2.181** | 1.641 | 1.463 |
| rag | 1.943 | **2.148** | 1.620 | 1.413 |
| summarization | 1.816 | **1.963** | 1.570 | 1.372 |
| tool_call | 2.085 | **2.229** | 1.654 | 1.464 |
| translation | 1.943 | **2.165** | 1.639 | 1.444 |
| writing | 2.026 | **2.183** | 1.639 | 1.456 |
| **Mean** | **2.033** | **2.218 (+9.1%)** | **1.642 (-19.2%)** | **1.460 (-28.2%)** |

**Config details:**

- **DFlash v5**: Trained on mainline speculators (PR branch). Evaluated with stock vLLM.
- **Domino no-shift, plain DFlash eval**: Trained with [`speculators@v6-normalize-experiment`](https://github.com/vllm-project/speculators/tree/v6-normalize-experiment) ([`df40fe9`](https://github.com/vllm-project/speculators/commit/df40fe9)), `shift_label=False`. Evaluated as plain DFlash (domino weights stripped, `projector_type: dflash`) — stock vLLM, no patches needed.
- **Domino no-shift + GRU head**: Same training as above. Evaluated with full domino checkpoint (`projector_type: domino`) using [`orestis-z/vllm@experiment/domino-shift-v6`](https://github.com/orestis-z/vllm/tree/experiment/domino-shift-v6) ([`21eefc952`](https://github.com/orestis-z/vllm/commit/21eefc952) — includes @swylyu001's Domino inference [`80f9acd4e`](https://github.com/orestis-z/vllm/commit/80f9acd4e) + config compat fixes + inference bugfixes).
- **Domino shift + GRU head**: Same training branch, `shift_label=True`. Same vLLM branch as above.

### Key Findings

1. **Backbone improvement is real (+9.1%)**: The Domino lambda schedule (start with base_loss only, gradually blend in GRU correction) produces a stronger backbone. Evaluating as plain DFlash (stripping domino weights) gives **2.218** avg acceptance — the best result in this experiment series.

2. **GRU correction head still hurts at inference**: Enabling the domino head drops acceptance length significantly. We don't yet know the definitive reason — the discrepancies listed above (dataset size, teacher forcing, loss normalization) are candidates but haven't been isolated.

3. **vLLM inference bugs found and fixed** (in [`21eefc952`](https://github.com/orestis-z/vllm/commit/21eefc952)):
   - **prefix_len off-by-one**: Training uses `suffix_start = 1 + pure_draft_prefix_len` but inference used `pure_draft_prefix_len` — correction started one position too early with shift
   - **vocab_size mismatch**: DominoHead used `config.vocab_size` (target model) instead of `draft_vocab_size`
   - **Falsy config handling**: Python `or` treated explicit `False`/`0` as falsy, falling through to wrong defaults
   
cc @Eros483

orestis-z:
@Eros483 I couldn't find any sources about domino helping with accuracy if training with it but dropping the head during inference. You mentioned that in the docs initially, could you point me to the source? My results above show that domino helps with training.

arnabwithab:
> @Eros483 I couldn't find any sources about domino helping with accuracy if training with it but dropping the head during inference. You mentioned that in the docs initially, could you point me to the source? My results above show that domino helps with training.

Yeah that's on me, wrong interpretation from my end, on the research. 

I had understood the domino head refines training, its actually learns how to refine raw logits.

Mentioned it in the previous commits, the commits on this PR and the vllm pr now actively use the domino head during inference. 

Found a few more bugs, will be mentioning them soon on this PR

arnabwithab:
## Experiment Update: End-to-End Validation of Training Fixes
Validated all three training fixes end-to-end with a patched vLLM inference backend on Qwen3-0.6B. Three models were trained identically (same data, epochs, architecture) to isolate the effect of each change.

### Model Configurations
| Parameter | All models |
|-----------|------------|
| Verifier | Qwen3-0.6B (hidden=1024, 28 layers) |
| Draft layers | 3 (hidden=1024, 16 attn heads, 8 KV heads) |
| block_size | 8 |
| draft_vocab_size | 32,000 |
| Auxiliary layers | [1, 13, 25] |
| Training | 5,000 ShareGPT, 2 epochs, lr=6e-4, linear schedule, 50 warmup |
| max_seq_len | 4,096 |

**Model A — DFlash shift=True**: `projector_type="dflash"`, `shift_label=True`. Position 0 predicts a+1 but receives zero gradient.

**Model B — DFlash shift=False**: `projector_type="dflash"`, `--no-domino-shift-label`. Position 0 is anchor (masked). Position 1 predicts a+1 with full training signal. This is the correct baseline — same token prediction, same architecture, same loss weighting as Domino position 0.

**Model C — Domino** (all three fixes): `projector_type="domino"`, `shift_label=True`. Position 0 predicts a+1 with full signal (base-anchored curriculum, `domino_loss_decay` weight 1.0, pos0 included in metrics). GRU hidden=1024, bottleneck=256, `pure_draft_prefix_len=0`.

| Model | First real prediction | Trained? | Correction? |
|-------|----------------------|----------|-------------|
| dflash_shift | pos0 (a+1) | No (masked, decay=0) | No |
| dflash_noshift | pos1 (a+1) | Yes | No |
| domino | pos0 (a+1) | Yes (curriculum) | No (pos0 uncorrected) |

The fair comparison for backbone quality is **Domino pos_0** vs **DFlash_noshift pos_1** — same token, same architecture, same training budget, same loss weighting. Only the base-anchored curriculum differs.

### Training Validation (ShareGPT held-out, epoch 2, greedy accuracy)
| Position | DFlash noshift pos1 | Domino pos0 | Delta |
|----------|---------------------|-------------|-------|
| Predicts a+1 | 21.5% | **41.0%** | +19.5 % |
| Overall | 11.8% | 27.2% (final) | +15.4 % |

### Inference Benchmark on 164 HumanEval prompts
greedy T=0, max 256 new tokens, single RTX 4060 8GB. Verifier: Qwen3-0.6B (HF `AutoModelForCausalLM`, bf16). 

Metric: unconditional acceptance (cumprod over draft tokens), wall-clock TPS.

| Metric | DFlash noshift | Domino | Delta |
|--------|---------------|--------|-------|
| Accept length | 1.079 | **1.378** | +27.7% |
| TPS | 45.1 | **53.1** | +17.8% |
| pos0 (a+1, backbone) | 7.6% | **31.9%** | +24.3% |
| pos1 | 0.3% | 4.9% | +4.6 % |
| pos2 | 0.0% | 0.8% | — |


Domino accept_len +27.7% with +17.8% TPS despite 15M extra parameters and GRU loop overhead. The higher acceptance saves more verifier forward passes than the correction costs.

| Position    | 0    | 1    | 2    | 3    | 4    | 5    | 6    | 7   |
|-------------|------|------|------|------|------|------|------|-----|
| Conditional | 31.9% | 15.2% | 16.5% | 23.4% | 29.8% | 41.2% | 71.4% | 80.0% |

Conditional acceptance rises with position — once the prefix chain survives, the GRU's growing causal context drives corrections to 80% at position 7.

### vLLM Inference Bugs Found and Fixed (patched vLLM, not in speculators)
Two bugs in the Domino inference backend, both caught by comparing against SGLang training code:

**Fix 1 — `correction_start` off-by-one:** Training leaves `1 + pure_draft_prefix_len` positions as pure base logits. vLLM started correction at `prefix_len`, missing the +1 for the anchor token. The Domino head corrected a position it was never trained on. 

**Source:** `specforge/core/domino.py:210-213` (`_suffix_start` property).

**Fix 2 — Domino head vocabulary mismatch:** Training outputs `draft_vocab_size` (32,000) residuals. vLLM used `config.vocab_size` (151,936), causing shape mismatches or corrections at wrong indices. Paper §4.1.2 describes this as a "low-rank residual update" restricted to the causal branch. 

**Source:** `specforge/modeling/draft/dflash.py:128-133`.

### Actual inference for validation
Use inference code via vllm from the [vLLM PR](https://github.com/vllm-project/vllm/pull/47627) or from hugging face inference code in commit [e97207d](https://github.com/vllm-project/vllm/commit/e97207d4d6a83afacd05de3e47840df31b517eba)

arnabwithab:
@orestis-z We finally have some positive results from our experiments :)

# v7 Experimental Results

| Key | Value |
|---------|-------|
| Verifier | Qwen3-0.6B |
| Training data | ShareGPT |
| Epochs | 2 |
| Hardware | RTX 4060 Laptop, 8GB VRAM |

Fixes are primarily in inference itself. 

Benchmarked only on **HumanEval** so far.

| Metric       | [dflash_shift](https://huggingface.co/Eros483/Qwen3-0.6B-DFlash-shift-b8) | [dflash_noshift](https://huggingface.co/Eros483/Qwen3-0.6B-DFlash-noshift-b8) | [domino](https://huggingface.co/Eros483/Qwen3-0.6B-Domino-b8) |
|--------------|-------------|---------------|--------|
| Accept len   | 1.070       | 1.079         | **1.378**  |
| TPS          | 47.3        | 45.1          | **53.1**   |
| Time (s)     | 886.9       | 930.2         | **790.6**  |

Domino delivers 
- **+27.7% acceptance length** 
- **+17.8% throughput**,
- **15% less wall-clock time** 

all while carrying 15M additional parameters and a GRU correction loop.

I understand that a much weaker model is being benchmarked, but everything is being compared on a equivalent base, so it is my hope that, a scaled up test with the same training paramters, and the vllm/hf inference code linked up above, should replicate similiar ratios!

orestis-z:
That's great, thanks so much @Eros483 !

arnabwithab:
@orestis-z PFA the arguments for the general training arguments (common to all models with parameter changes

```
python scripts/train.py \
    --verifier-name-or-path Qwen/Qwen3-0.6B \
    --data-path ./domino_data \
    --hidden-states-path ./domino_data/hidden_states \
    --speculator-type dflash \
    --target-layer-ids 1 13 25 \
    --block-size 8 \
    --max-anchors 1024 \
    --num-layers 3 \
    --total-seq-len 4096 \
    --draft-vocab-size 32000 \
    --epochs 2 \
    --lr 6e-4 \
    --scheduler-warmup-steps 50 \
    --on-missing raise

# Model A — dflash_shift
    --projector-type dflash \
    --save-path ./dflash_checkpoints

# Model B — dflash_noshift
    --projector-type dflash \
    --no-domino-shift-label \
    --save-path ./dflash_noshift_checkpoints

# Model C — domino
    --projector-type domino \
    --save-path ./domino_checkpoints
```

will see if i can run more evals

arnabwithab:
# v7 Experiment Results on further benchmarks

## Training Infra
| Key | Value |
|---------|-------|
| Verifier | Qwen3-0.6B |
| Training data | ShareGPT |
| Epochs | 2 |
| Hardware | RTX 4060 Laptop, 8GB VRAM |

## Benchmark Results

| subset       | [dflash_noshift](https://huggingface.co/Eros483/Qwen3-0.6B-DFlash-noshift-b8) | [domino](https://huggingface.co/Eros483/Qwen3-0.6B-Domino-b8) | Delta |
|--------------|-------------|---------------| ---------------|
| HumanEval | 1.079 | 1.378 | **+27.7%** |
| aime24 | 1.067 | 1.320 | **+23.7%** |
| aime25 | 1.077 | 1.324 | **+22.9%** |
| mt-bench | 1.082 | 1.321 | **+22.1%** |
| swe-bench | 1.090 | 1.335 | **+22.47%** |
| rag | 1.068 | 1.308 | **+22.47%** |

orestis-z:
## Qwen3-8B evaluation with `Huang2020/Qwen3-8B-Domino-b16`

Ran the paper authors' public checkpoint ([Huang2020/Qwen3-8B-Domino-b16](https://huggingface.co/Huang2020/Qwen3-8B-Domino-b16)) through our eval suite on 4xH100 using [vLLM PR #47627](https://github.com/vllm-project/vllm/pull/47627) with the `algos.py` config forwarding patch and the step-0 base-logit fix.

### Setup
- **Verifier**: Qwen/Qwen3-8B, TP=4
- **Draft**: `Huang2020/Qwen3-8B-Domino-b16` (`shift_label=True`, `pure_draft_prefix_len=1`, `gru_hidden_dim=1024`, `emb_dim=256`)
- **spec_tokens**: 15, `--enforce-eager`, `--max-num-seqs 256`
- **Baseline**: our DFlash v5 checkpoint (same verifier/TP/hardware, acceptance length 2.033 avg)

### Results

| Subset | DFlash v5 | Domino (Huang2020) | Delta |
|--------|-----------|-------------------|-------|
| HumanEval | 2.228 | 3.420 | **+53.5%** |
| math_reasoning | 2.307 | 3.873 | **+67.9%** |
| qa | 1.931 | 2.994 | **+55.0%** |
| question | 2.021 | 3.087 | **+52.7%** |
| rag | 1.943 | 3.270 | **+68.3%** |
| summarization | 1.816 | 2.579 | **+42.0%** |
| tool_call | 2.085 | 2.960 | **+42.0%** |
| translation | 1.943 | 3.029 | **+55.9%** |
| writing | 2.026 | 3.129 | **+54.4%** |
| **Average** | **2.033** | **3.149** | **+54.9%** |

**Caveats:** This is not an apples-to-apples comparison — the Huang2020 checkpoint was likely trained on different data, with a different training recipe, and possibly for longer. This confirms the inference path works and establishes an upper bound for Domino on Qwen3-8B b16, but the gains vs our DFlash v5 are not solely attributable to the Domino head.

@Eros483's results on Qwen3-0.6B/b8 showed +22-28% acceptance length with the same speculators training code. Our checkpoints trained with the same code on Qwen3-8B/b16 still produce ~1.0 acceptance, suggesting the current training recipe may not scale to larger models and block sizes without changes to hyperparameters, data, or training duration. The paper authors clearly have a recipe that works at this scale — we need to investigate what's different.

arnabwithab:
@orestis-z I do not think v5 used the same parameter configurations

Essentially The key changes we should do a bigger run with is,
- decay ratio 0.5
- Shift label True
- Verify if any changes in the default parameter
- and test it with the latest inference code.

I think its worth a try to replicate my experiments with the 0.6b model, with the parameters listed above, and in the training args i referenced in a previous comment to get a better idea. There might be also some default parameter changes that ive not mentioned explicitly, but are used by default in the training code.

orestis-z:
Hey moved to a clean state to https://github.com/vllm-project/speculators/pull/773 to run experiments on our clusters as discussed offline. I'm closing this one. The new PR is build on top of your commits so your contribution is preserved. Looking forward to your feedback in https://github.com/vllm-project/speculators/pull/773.

PR #687, opened 2026-06-29
feat(peagle): add random anchor subsampling (max_anchors)

### LINKED ISSUE #567: [RFC]: Add random anchor for p-eagle first position
### Motivation.

Peagle uses COD sampling, which predicts the next token for every possible starting token in a sample. Then also predicts the token after that for some percent of the tokens (e.g. 70%). And then predicts the 3rd token for 70% of the selected 70%, etc.

So our attention mask shape is: $(\text{sequence len} * 2.73)^2$, where $2.73 = \sum_{n=0}^4{0.7^n}$ and 4 is number of speculative tokens. What we're proposing is using `num_anchors` as starting points for peagle training because a good chunk of the original sequence length will have loss mask 0. That would give us attention masks of size: $(\text{num anchors} \times 2.73) \times (\text{num anchors} \times 1.73 + \text{sequence len})$ This would be smaller and potentially helps us train with longer sequence length.

### Proposed Change.

Add the `max_anchors` parameter to the P-Eagle config that caps the number of COD chain starting points. When set, `generate_cod_sample_indices` randomly selects `max_anchors` positions from the loss_mask=1 indices at depth 0 instead of using all positions. The attention mask construction reuses the existing
  `select_anchors` method to identify these starting points, then builds an asymmetric mask where queries are the COD-sampled tokens $(\text{max anchors} × 2.73)$ and keys/values include the full original sequence plus the deeper COD tokens $(\text{max anchors} × 1.73 + \text{sequence len})$. This shrinks the attention mask from
 $(\text{sequence len} × 2.73)^2$ to $(\text{num anchors} \times 2.73) \times (\text{num anchors} \times 1.73 + \text{sequence len})$, enabling training with longer sequence lengths under the same memory budget.

### Any Other Things.

_No response_

orestis-z:
Closes https://github.com/vllm-project/speculators/issues/567.

## Summary

- Add `max_anchors` config to P-EAGLE's COD sampling that randomly subsamples chain starting points at depth 1+, capping attention mask size for longer sequences while keeping depth 0 as the full sequence
- Change `--max-anchors` CLI default from 256 to None so P-EAGLE uses all positions by default (DFlash retains its 3072 hardcoded fallback)
- Add unit tests verifying depth-0 preservation, chain capping, sorted order, and backward compatibility

## Motivation

P-EAGLE's COD sampling starts a chain from every `loss_mask=1` position, producing attention masks that OOM at 8K+ sequences. The ablation study (https://github.com/vllm-project/speculators/issues/567#issuecomment-4835911248) showed that randomly subsampling with `max_anchors=1024` actually **improves** d0 accuracy by +3.1pp while dramatically reducing memory, enabling training at longer sequence lengths.

Supersedes #590 (contiguous window approach, which degraded quality) and #683.

## Test plan

- [x] 6 new unit tests in `tests/unit/models/test_peagle_data.py` — all pass
- [x] 38 existing P-EAGLE integration tests — all pass
- [x] `ruff check` clean on all changed files
- [ ] CI green


scripts/train.py
```diff
@@ -935,8 +935,9 @@ def parse_args():
     parser.add_argument(
         "--max-anchors",
         type=int,
-        default=256,
-        help="Maximum anchor positions for DFlash training (default: 256)",
+        default=None,
+        help="Maximum anchor positions. "
+        "DFlash default: 3072, P-EAGLE default: None (all).",
```
fynnsu: I think I'd prefer to keep a default here for both algorithms. Maybe something like 1024 or 2048 for both algs?

src/speculators/models/peagle/config.py
```diff
@@ -56,6 +56,17 @@ class PEagleSpeculatorConfig(Eagle3SpeculatorConfig):
...
+        default=None,
+        description=(
+            "Maximum number of COD chain starting points. When set, "
+            "randomly subsamples valid positions for depth-1+ chains. "
+            "Depth 0 always retains the full sequence. "
+            "None means use all valid positions."
+        ),
+        ge=1,
+    )
+
```
fynnsu: So we want to avoid putting training hyperparameters into the model config. I think we actually currently have this field for dflash, but that's kind of a mistake.

Instead we can use the `get_trainer_kwargs` method to get these hyperparameters into the models forward call.

src/speculators/models/peagle/data.py
```diff
@@ -36,12 +37,18 @@ def generate_cod_sample_indices(
     device = loss_mask.device
     all_valid_indices = torch.where(loss_mask == 1)[0]
 
+    if max_anchors is not None and all_valid_indices.shape[0] > max_anchors:
+        perm = torch.randperm(all_valid_indices.shape[0], device=device)[:max_anchors]
+        seed_indices = all_valid_indices[perm].sort()[0]
+    else:
+        seed_indices = all_valid_indices
```
fynnsu, line 44: There is maybe an argument for padding the indices here if we have less than max_anchors so that shapes can remain consistent for torch.compile. But if this seems to work reliably then perhaps it is okay as is.

fynnsu (changes requested):
Looks good, but please update the max_anchor piping to not use the model's config.

scripts/train.py
orestis-z: Sounds good. Default for `dflash` was 3072. Shall we stick to that? Happy to lower it also to 1024 or 2048 but I guess 3072 is battle-tested.

Regarding previous 256: https://github.com/vllm-project/speculators/pull/590#discussion_r3396705019

src/speculators/models/peagle/data.py
orestis-z: It seems to work fine as it is now, if it becomes a bottleneck I will address it as a follow-up.

src/speculators/models/peagle/config.py
orestis-z: Fixed.

I can move it from config to trainer kwargs for dflash/dspark too in a separate PR if you agree.

shanjiaz: Please do! Thanks.

shanjiaz (approved):
Thanks for the detailed ablation. Glad this works out!

src/speculators/models/peagle/core.py
fynnsu, line 65: nit: list `max_anchors` instead of extracting from kwargs below.

fynnsu (approved):
Looks good! One optional nit

PR #688, opened 2026-06-30
feat(data): declarative schema normalization for prepare-data (RFC #583)

imargulis:
Implements declarative conversation-schema normalization for prepare_data.py (part of RFC #583 per @shanjiaz scoping comments), replacing hand-written per-dataset normalizers with config-driven ones.

## Changes

- **Declarative role mapping**. Add a ROLE_ALIASES lookup table and a _normalize_turn helper; _normalize_conversation delegates to it instead of an inline if/elif chain.

- **Automatic messages → conversations**. The column is renamed during ingestion (_rename_messages_to_conversations), so OpenAI-style messages datasets need no per-dataset normalizer (e.g. _normalize_ultrachat is removed).

- **Config-generated normalizers**. DatasetConfig gains declarative hints — conversations_column, prompt_column/answer_column — and resolve_normalize_fn(config) builds the normalizer from them. ultrachat and gsm8k are converted to hints; normalize_fn remains an explicit escape hatch (e.g. for sharegpt4v_coco).

- **Correctness hardening of turn handling**: observation maps to the tool role (it's a tool result, not a user turn), thinking and reasoning_content are preserved independently (no or-overwrite), and empty conversations are guarded before turn-dropout sampling.

## Testing

- tests/integration/datagen/test_schema_normalization.py — offline unit tests for role aliases, resolve_normalize_fn generation (prompt/answer, conversations_column rename, explicit escape hatch), and the automatic messages→conversations rename.

- tests/integration/datagen/test_normalization_presets.py — a @pytest.mark.slow, self-skipping test that runs the real normalization path over small streamed slices of sharegpt, ultrachat, gsm8k, and nemotron, asserting well-formed conversations.

- Existing _normalize_conversation tests still pass. ruff, ruff format, and mypy --check-untyped-defs are clean.


src/speculators/data_generation/preprocessing.py
```diff
@@ -74,6 +74,53 @@ def _visualize_sample(preprocessed: HFDataset, processor: ProcessorLike, idx: in
...
+        normalized_turn["tool_calls"] = turn["tool_calls"]
+    if turn.get("tool_call_id"):
+        normalized_turn["tool_call_id"] = turn["tool_call_id"]
+
+    # Preserve thinking / reasoning_content independently so neither clobbers the
+    # other when both are present with different values.
+    if turn.get("thinking"):
+        normalized_turn["thinking"] = turn["thinking"]
+    if turn.get("reasoning_content"):
+        normalized_turn["reasoning_content"] = turn["reasoning_content"]
```
WindChimeRan: silently drops reasoning content for "thinking"

src/speculators/data_generation/preprocessing.py
```diff
@@ -701,6 +725,23 @@ def _load_hf_dataset(spec: str) -> tuple[HFDataset, None]:
     return raw_dataset, None
 
 
+def _rename_messages_to_conversations(dataset: HFDataset) -> HFDataset:
```
WindChimeRan, line 749: repro:
```
from speculators.data_generation.preprocessing import load_raw_dataset
load_raw_dataset("hf:HuggingFaceH4/ultrachat_200k:test_sft")
# ValueError: ... expected a 'conversations' column but found ['prompt', 'prompt_id', 'messages']
```

src/speculators/data_generation/configs.py
```diff
@@ -7,34 +7,95 @@
...
+                {"role": "user", "content": example[prompt_column]},
+                {"role": "assistant", "content": example[answer_column]},
+            ]
+        }
+
+    normalize.__name__ = f"normalize_{prompt_column}_{answer_column}"
+    return normalize
+
+
+def resolve_normalize_fn(config: DatasetConfig) -> NormalizeFn | None:
```
WindChimeRan: 1. It returns {"conversations": example[source_column]}, but Dataset.map() merges that into the row rather than replacing it, so source_column never gets dropped
2. If source_column is missing from a row (typo, schema drift), it silently returns the row unchanged

src/speculators/data_generation/preprocessing.py
WindChimeRan: Qwen3 template only reads the reasoning_content field. 

```
from transformers import AutoTokenizer
tok = AutoTokenizer.from_pretrained("Qwen/Qwen3-8B")

raw = {"role": "assistant", "content": "The answer is 4.", "thinking": "We are adding 2 and 2."}

t = raw.get("thinking") or raw.get("reasoning_content")           # OLD: mirror into both
old = {"role": "assistant", "content": raw["content"], "thinking": t, "reasoning_content": t}

new = {"role": "assistant", "content": raw["content"]}            # NEW (this PR): each key from its own source
if raw.get("thinking"): new["thinking"] = raw["thinking"]
if raw.get("reasoning_content"): new["reasoning_content"] = raw["reasoning_content"]

user = [{"role": "user", "content": "What is 2+2?"}]
r = lambda turn: tok.apply_chat_template(user+[turn], tokenize=False, enable_thinking=True).split("assistant\n",1)[1]
print("OLD:\n" + r(old))
print("NEW:\n" + r(new))   # <think> block is empty — reasoning dropped
```

WindChimeRan (commented):
Main concern: this is overcomplicated and the implementation is fragile. The config-driven column-layout normalization adds machinery no dataset uses yet, and that's where the bugs cluster.

Role names are uniform across datasets, so centralizing them is right. Column layout is per-dataset — generalizing it ahead of a real second case is what makes this fragile. Concretely: two mechanisms do the same `messages→conversations rename`, and it's wired into one downstream caller, so the hf: path skips it entirely. Collapse to one rename in one place (`load_raw_dataset`) that covers every source, and cut `conversations_column/_make_rename_normalizer.`

One blocker beyond that: the thinking change is a behavior regression, not normalization. 

Bigger "generalize now vs later" question is on RFC #583.

imargulis:
@WindChimeRan Thanks for the thorough review — I agree with the framing (role names are uniform,
column layout is per-dataset) and reworked the PR accordingly. 

**Thinking regression (blocker) — reverted.** You're right, this was a behavior change,
not normalization. Restored the mirror so `thinking` and `reasoning_content` are both
populated from whichever is present, so Qwen3's template (which reads
`reasoning_content`) still renders the `<think>` block. Your repro is covered by a test.

**`hf:` path skipped the rename — fixed.** The `messages→conversations` rename now runs
once in `load_raw_dataset`, covering every source (local file, directory, preset, and
`hf:`). `_load_hf_dataset` just loads; the conversations-format check happens after the
rename, so `load_raw_dataset("hf:HuggingFaceH4/ultrachat_200k:test_sft")` works now.
Removed the duplicate call in `load_and_preprocess_dataset`.

**Overcomplicated / premature generalization — trimmed.** Cut `conversations_column`,
`prompt_column`/`answer_column`, both builder factories, and `resolve_normalize_fn`
(no preset used `conversations_column`; only gsm8k used prompt/answer). gsm8k is back to
a bespoke `_normalize_fn`. What remains is the part you called out as correct: role
centralization (`ROLE_ALIASES` + `_normalize_turn`) plus the single automatic
`messages→conversations` rename, plus the turn-handling hardening (observation→tool,
empty-conversation guard). 

On the bigger "generalize now vs later" question — agreed it belongs on RFC #583; happy
to bring the prompt/answer or column-layout hints back there once there's a real second
case to justify them.

orestis-z (commented):
Core changes are correct, well-tested, and properly scoped after the trim. Two nits: falsy-content fallthrough in `_normalize_turn` when `value` is empty string, and missing test for dual `thinking`/`reasoning_content` priority. PR description is stale — still references cut machinery. See inline comments. Recommend approving.

src/speculators/data_generation/preprocessing.py
```diff
@@ -74,6 +74,53 @@ def _visualize_sample(preprocessed: HFDataset, processor: ProcessorLike, idx: in
...
+    """Normalize a single conversation turn to canonical role/content keys.
+
+    Returns ``None`` (signalling the caller to skip the turn) when the role is
+    not recognized.
+    """
+    raw_role = turn.get("from", turn.get("role", ""))
+    role = ROLE_ALIASES.get(raw_role)
+    if role is None:
+        log.warning(f"Unknown role '{raw_role}', skipping turn")
+        return None
```
orestis-z, line 100: nit: `or`-chain treats empty-string `value` as absent (falsy fallthrough to `content`). Now that this is the single canonical extraction point, consider:
```python
"content": turn["value"] if "value" in turn else (turn.get("content") or ""),
```
Not a regression (old inline code had the same pattern), but worth hardening since `_normalize_turn` is now the source of truth for all schemas.

tests/integration/datagen/test_schema_normalization.py
```diff
@@ -0,0 +1,141 @@
...
+def test_normalize_turn_mirrors_thinking_into_reasoning_content():
+    # Chat templates (e.g. Qwen3) read reasoning_content, so a turn carrying only
+    # 'thinking' must still populate reasoning_content (and vice versa).
+    from_thinking = _normalize_turn(
+        {"role": "assistant", "content": "a", "thinking": "T"}
+    )
+    from_reasoning = _normalize_turn(
+        {"role": "assistant", "content": "a", "reasoning_content": "R"}
+    )
+    assert from_thinking is not None
```
orestis-z: Consider adding a case for when both `thinking` and `reasoning_content` are present with **different** values, to document the `or`-short-circuit priority (`thinking` wins):
```python
both = _normalize_turn({"role": "assistant", "content": "a", "thinking": "T", "reasoning_content": "R"})
assert both["thinking"] == both["reasoning_content"] == "T"  # 'thinking' takes priority
```
This edge case briefly changed behavior in the intermediate revision, so a regression test for the priority would be good insurance.

imargulis:
Up to date PR description:

Implements declarative conversation-schema **role normalization** for prepare_data.py
(the schema-normalization slice of RFC #583), trimmed after
review to the parts that are uniform across datasets.

## Changes
- **Declarative role mapping.** Add a `ROLE_ALIASES` table and a `_normalize_turn`
  helper; `_normalize_conversation` delegates to it instead of an inline if/elif chain.
  `observation` maps to the `tool` role, `tool_calls`/`tool_call_id` are preserved, and
  `thinking`/`reasoning_content` are mirrored (both populated from whichever is present,
  so Qwen3's template — which reads `reasoning_content` — still renders the `<think>`
  block).
- **Automatic `messages` → `conversations`.** A single rename in `load_raw_dataset`
  covers every source (local file, directory, preset, and `hf:`), so OpenAI-style
  `messages` datasets need no per-dataset normalizer (`_normalize_ultrachat` removed).
- **Empty-conversation guard** before turn-dropout sampling.

`normalize_fn` remains the explicit per-dataset escape hatch (e.g. `gsm8k`,
`sharegpt4v_coco`). The config-driven column-layout machinery from the earlier revision
(`conversations_column`, `prompt_column`/`answer_column`, `resolve_normalize_fn`) was
dropped as premature generalization; the "generalize now vs later" discussion lives on
RFC #583.

## Testing
- `tests/integration/datagen/test_schema_normalization.py` — role aliases, turn
  normalization (tool fields, thinking/reasoning mirroring + priority, empty-value
  preference), the automatic `messages`→`conversations` rename, and preset normalizers.
- `tests/integration/datagen/test_normalization_presets.py` — network-gated, self-skipping
  validation over small streamed slices of sharegpt/ultrachat/gsm8k/nemotron.

orestis-z (commented):
LGTM. Both nits from my previous review (falsy-content fallthrough, missing thinking/reasoning priority test) are cleanly addressed in `9fafc86`. Recommend approving.

WindChimeRan (commented):
LGTM, recommend approve and merge.

WindChimeRan:
nit: `tests/integration/datagen/test_schema_normalization.py` this test is low value. Suggest delete. Others LGTM! @shanjiaz

speculatorsbot (commented):
LGTM. The refactoring is clean and well-scoped after the trim -- role centralization via `ROLE_ALIASES` + `_normalize_turn`, single `messages`→`conversations` rename covering all sources, and the `observation`→`tool` fix are all correct. Agree with @WindChimeRan and @orestis-z. Recommend approving.

PR #690, opened 2026-06-30
feat(data): prepare-data safety & correctness (RFC #583)

imargulis:
Hardens prepare_data.py reruns and preprocessing output (part of RFC #583).

## Changes

- **Safe --overwrite**. assert_safe_to_overwrite refuses to delete an output directory that contains files which aren't prepare_data.py artifacts (.arrow shards, dataset_info.json/state.json, and the token-frequency file). Previously --overwrite unconditionally rmtree'd the directory, so pointing --output at the wrong path could wipe unrelated files. token_freq_path is now resolved before the overwrite check so a custom path is recognized as an artifact.

- **Truncate to max_length**. _preprocess_batch now bounds input_ids/loss_mask to max_length, so a turn that runs past the sequence window keeps only its in-window tokens and the two stay aligned.

- **Fail loud on empty output**. load_and_preprocess_dataset raises when preprocessing yields zero samples (a common silent-failure mode from schema/mask/--minimum-valid-tokens issues). A new --allow-empty-output / allow_empty_output opts out, and sample visualization is skipped on an empty result.

## Testing

- tests/unit/train/test_prepare_data.py — assert_safe_to_overwrite accepts artifact-only directories (including a custom token-freq path) and rejects unexpected files/subdirectories; --allow-empty-output parsing (default False, set True).

### DIFF
--- a/scripts/prepare_data.py
+++ b/scripts/prepare_data.py
@@ -57,11 +57,12 @@ def assert_safe_to_overwrite(output: Path, token_freq_path: Path) -> None:
     metadata, and the token frequency file) may be deleted.
     """
     unexpected_paths = []
+    resolved_token_freq_path = token_freq_path.resolve()
     for path in output.iterdir():
         if path.is_file() and (
             path.suffix == ".arrow"
             or path.name in PREPARE_DATA_OVERWRITE_ALLOWED_FILES
-            or path == token_freq_path
+            or path.resolve() == resolved_token_freq_path
         ):
             continue
         unexpected_paths.append(path)
--- a/scripts/prepare_data.py
+++ b/scripts/prepare_data.py
@@ -40,6 +40,41 @@ logging.basicConfig(
 log = PipelineLogger(__name__)
 
 
+# Files prepare_data.py itself writes into --output; only these may be removed by
+# --overwrite.
+PREPARE_DATA_OVERWRITE_ALLOWED_FILES = {
+    "dataset_info.json",
+    "state.json",
+    "token_freq.pt",
+}
+
+
+def assert_safe_to_overwrite(output: Path, token_freq_path: Path) -> None:
+    """Refuse to ``--overwrite`` a directory holding non-artifact files.
+
+    Guards against pointing ``--output`` at a directory with unrelated user files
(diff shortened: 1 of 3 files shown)


scripts/prepare_data.py
```diff
@@ -143,6 +179,14 @@ def parse_args():
...
         ),
     )
+    parser.add_argument(
+        "--allow-empty-output",
+        action="store_true",
+        help=(
+            "Allow writing an empty preprocessed dataset. By default prepare_data.py "
+            "raises when normalization or filtering removes every sample."
+        ),
+    )
```
fynnsu, line 189: What's the use case for this?

src/speculators/data_generation/preprocessing.py
```diff
@@ -570,6 +570,11 @@ def _preprocess_batch(
             f"Shape mismatch: input_ids={len(input_ids)}, loss_mask={len(loss_mask)}"
         )
 
+        # Bound both to max_length: a turn running past the window keeps only its
+        # in-window tokens, and input_ids/loss_mask stay aligned and bounded.
```
fynnsu, line 574: Isn't this truncation already being handled in `_get_input_ids_loss_mask` / `apply_chat_template`?

scripts/prepare_data.py
imargulis: The real change is failing loud on an empty result (it used to silently write an empty dataset). The flag is just an escape hatch for cases where empty is intentional (e.g. dry-runs). If you'd rather not add the surface, I'm happy to drop the flag and always raise — an empty training set is essentially never intended

src/speculators/data_generation/preprocessing.py
imargulis: The regex path already truncates (passes truncation=True+max_length), but the HF assistant-mask path (assistant_pattern=None, used when the template supports {% generation %}) calls apply_chat_template without truncation. The slice bounds that path. Alternatively I can push truncation/max_length into that apply_chat_template call instead of post-slicing (slicing both arrays keeps the mask aligned and is simpler); or, if we only target templates without generation markers, drop it.

fynnsu (approved):

PR #691, opened 2026-06-30
fix(attention): convert boolean mask to float for eager/SDPA backends

### LINKED ISSUE #686: [Bug]: Investigate and resolve DFlash speculators-vLLM inference divergence for Multi-Modal training
### Your current environment

| Component | Version |
|-----------|---------|
| vLLM | Latest main (installed from source at `/home/cyrus/vllm/`) |
| Speculators | `ef5541ac` (DarkLight1337/speculators `main`, base of PR #1) |
| Python | 3.13 |
| PyTorch | >= 2.9.0, <= 2.12.0 (per `pyproject.toml`) |
| Transformers | >= 4.56.1, < 5.13.0 (code references rope changes in transformers 5.0) |
| CUDA | (not specified in PR) |
| Hardware | (not specified in PR — multi-GPU, at least 3 GPUs based on CUDA_VISIBLE_DEVICES 1,2 defaults in comparison script) |
| Model (verifier) | `google/gemma-4-26B-A4B-it` |
| Attention backends | FLEX_ATTENTION (main model), FlashAttention v2 (draft model in vLLM), SDPA (draft model in speculators) |

### 🐛 Describe the bug

The DFlash draft model is implemented in both speculators (for training) and vLLM (for inference). Given the same model weights and the same input, both implementations should produce identical intermediate tensors at every stage of the forward pass. [DarkLight1337/speculators#1](https://github.com/DarkLight1337/speculators/pull/1) built a comparison harness that runs a Gemma4-26B-A4B-it input through both implementations, saves every intermediate tensor (Q/K/V projections, norms, RoPE outputs, attention outputs, MLP outputs, logits) at each of the 5 DFlash transformer layers, and measures the cosine similarity between the speculators and vLLM versions of each tensor. The results show significant divergence:

| Layer | Attention Inputs (Q,K,V) | attn_output | o_proj | layer_output |
|-------|--------------------------|-------------|--------|--------------|
| 0     | ~1.000                   | 0.911       | 0.996  | 0.902        |
| 1     | ~0.925                   | 0.752       | 0.712  | 0.779        |
| 2     | ~0.930                   | 0.752       | 0.822  | 0.725        |
| 3     | ~0.950                   | -           | -      | ~0.7         |
| 4     | ~0.950                   | -           | -      | ~0.7         |

Layer 0 attention inputs match perfectly, confirming the weights and projections are identical. But attention output drops to 0.91, and error accumulates through subsequent layers until final logits cosine similarity reaches **0.783**. 

For the same data point, we get high acceptance rates in Speculators but very low acceptance rates in vLLM.

orestis-z:
## Summary

- **Root cause**: `torch.nn.attention.flex_attention.create_mask` returns a boolean tensor (True = attend), but `eager_attention_forward` from transformers adds the mask numerically (`scores + mask`), interpreting True as +1 and False as +0 — effectively ignoring the mask. This caused eager/SDPA backends to diverge from flex_attention and from vLLM inference.
- **Fix**: Added `create_float_mask` wrapper that converts the boolean mask to a float mask (0 for attend, -inf for masked), compatible with all non-flex backends. Applied to both DFlash and Eagle3.

## Details

After the fix, eager vs SDPA cosine similarity improved from **0.789 → 0.999953**, and eager vs flex_attention from **0.789 → 0.999953** (tested with `z-lab/gemma-4-26B-A4B-it-DFlash`).

**Reproduction script & output**: https://gist.github.com/orestis-z/3d51a66af49112dd6dba0327bee507e4

Note: the default attention backend is `simple_flex_attention`, so models trained with the default config are unaffected. This fix corrects behavior for the `eager` and `sdpa` backends only.

### Files changed
- `src/speculators/models/attention.py` — new `create_float_mask` function
- `src/speculators/models/dflash/core.py` — use `create_float_mask` instead of `create_mask`
- `src/speculators/models/eagle3/core.py` — use `create_float_mask` instead of `create_mask`
- `tests/integration/models/test_dflash_vllm_parity.py` — regression tests (backend parity, mask correctness, vLLM semantic equivalence)

## Test plan

- [x] All existing DFlash and Eagle3 integration tests pass
- [x] 3 new parity tests pass (backend agreement, mask correctness, vLLM mask semantics)
- [x] `ruff check` + `ruff format` pass
- [x] Reproduction script confirms fix with real model ([gist](https://gist.github.com/orestis-z/3d51a66af49112dd6dba0327bee507e4))
- [ ] Full CI

Closes #686

### DIFF
--- a/src/speculators/models/dflash/core.py
+++ b/src/speculators/models/dflash/core.py
@@ -2,7 +2,7 @@ from typing import ClassVar
 
 import torch
 from torch import nn
-from torch.nn.attention.flex_attention import create_block_mask
+from torch.nn.attention.flex_attention import create_block_mask, create_mask
 from transformers import PretrainedConfig
 from transformers.models.qwen3.modeling_qwen3 import (
     Qwen3RMSNorm,
@@ -58,6 +58,8 @@ class DFlashDraftModel(DraftVocabMixin, SpeculatorModel):
             create_block_mask
             if self._attn_impl == "simple_flex_attention"
             else create_float_mask
+            if self._attn_impl == "eager"
+            else create_mask
         )
         super().__init__(config=config)
         self._init_vocab(config)
--- a/src/speculators/models/dflash/core.py
+++ b/src/speculators/models/dflash/core.py
@@ -2,7 +2,7 @@ from typing import ClassVar
 
 import torch
 from torch import nn
-from torch.nn.attention.flex_attention import create_block_mask, create_mask
+from torch.nn.attention.flex_attention import create_block_mask
 from transformers import PretrainedConfig
 from transformers.models.qwen3.modeling_qwen3 import (
     Qwen3RMSNorm,
@@ -10,6 +10,7 @@ from transformers.models.qwen3.modeling_qwen3 import (
 )
 
 from speculators.model import DraftVocabMixin, SpeculatorModel
+from speculators.models.attention import create_float_mask
 from speculators.models.dflash import DFlashSpeculatorConfig
 from speculators.models.dflash.attention import create_anchor_block_mask_mod
 from speculators.models.dflash.metrics import compute_metrics
@@ -56,7 +57,7 @@ class DFlashDraftModel(DraftVocabMixin, SpeculatorModel):
         self._create_mask_fn = (
             create_block_mask
             if self._attn_impl == "simple_flex_attention"
-            else create_mask
+            else create_float_mask
         )
         super().__init__(config=config)
         self._init_vocab(config)
--- a/src/speculators/models/attention.py
+++ b/src/speculators/models/attention.py
@@ -77,8 +77,8 @@ def create_float_mask(
 ) -> torch.Tensor:
     """Wrap ``create_mask`` and convert the boolean result to a float mask.
 
-    ``eager_attention_forward`` adds the mask numerically (``scores + mask``),
-    so it needs 0 for attended positions and ``-inf`` for masked positions.
+    Non-flex attention backends (eager, SDPA) add the mask numerically
+    (``scores + mask``) and need 0 for attended and ``-inf`` for masked.
     """
     bool_mask = _create_mask(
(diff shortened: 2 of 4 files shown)


shanjiaz (commented):
This looks good to me pending the validation results. I'm a little curious why we didn't catch this in our CI. Perhaps we don't have tests that uses enforce-eager at the moment?

orestis-z:
> This looks good to me pending the validation results. I'm a little curious why we didn't catch this in our CI. Perhaps we don't have tests that uses enforce-eager at the moment?

Correct — CI only tests `simple_flex_attention` (the default), which uses `create_block_mask` and is unaffected. The new tests in this PR cover the eager/SDPA paths.

Validation with the real model (`z-lab/gemma-4-26B-A4B-it-DFlash`):

| | eager vs sdpa | eager vs flex |
|---|---|---|
| **Before** (boolean mask) | 0.789 | 0.789 |
| **After** (float mask) | 0.9999 | 0.9999 |

Reproduction script & output: https://gist.github.com/orestis-z/3d51a66af49112dd6dba0327bee507e4

To show improved vLLM acceptance rates end-to-end, we'd need to train a model with `--draft-attn-impl eager` (broken without fix), run it in vLLM, then retrain with the fix and compare — which is expensive.

src/speculators/models/dflash/core.py
```diff
@@ -56,7 +57,7 @@ def __init__(
         self._create_mask_fn = (
             create_block_mask
             if self._attn_impl == "simple_flex_attention"
-            else create_mask
+            else create_float_mask
```
fynnsu, line 60: It seems like this is only needed for the eager backend, sdpa should support bool masks, which is also supported by your comparison showing the same diff here:

>  	eager vs sdpa 	eager vs flex
Before (boolean mask) 	0.789 	0.789
After (float mask) 	0.9999 	0.9999

Probably not the biggest deal, but I'd prefer to keep the masks boolean for sdpa because I think there are some optimizations that happen with a bool mask. So maybe change this condition to check for "eager"

fynnsu (approved):
Looks good, but 1 small comment

PR #692, opened 2026-06-30
feat(train): default to Eagle 3.1 settings (llama arch + norms)

orestis-z:
## Summary

[PR #610](https://github.com/vllm-project/speculators/pull/610) benchmarks show that Eagle 3.1 with `--draft-arch llama --norm-before-fc --norm-output` gives **+7.2% average acceptance length** over Eagle 3 baselines (Qwen3-8B). This PR flips the CLI defaults so new training runs use these settings automatically.

**Changes:**
- `--draft-arch`: default `"qwen3"` → `"llama"`
- `--norm-before-fc`: default `False` → `True`, switched to `BooleanOptionalAction` (adds `--no-norm-before-fc`)
- `--norm-output`: default `False` → `True`, switched to `BooleanOptionalAction` (adds `--no-norm-output`)

**Backward compatibility:** Config class defaults (`Eagle3SpeculatorConfig`) remain `False` — old checkpoints deserialize correctly. Users can restore Eagle 3 + qwen3 behavior with `--draft-arch qwen3 --no-norm-before-fc --no-norm-output`.

## Test plan

- [x] `ruff check` passes on changed files
- [x] `pytest tests/unit/train/test_draft_config_init.py` — all 36 tests pass
- [x] `pytest tests/unit/test_config.py -k norm` — all 4 config roundtrip tests pass (config defaults unchanged)

### DIFF
--- a/docs/cli/train.md
+++ b/docs/cli/train.md
@@ -136,9 +136,9 @@ torchrun --standalone --nproc_per_node=4 scripts/train.py \
 
 - **`--embed-requires-grad` / `--no-embed-requires-grad`** (flag, default: `False`) Whether to train embedding layer weights.
 
-- **`--norm-before-fc` / `--no-norm-before-fc`** (flag, default: `True`) Apply RMSNorm before the FC layer in the draft path (Eagle 3.1). Use `--no-norm-before-fc` for the original Eagle 3 behavior.
+- **`--norm-before-fc` / `--no-norm-before-fc`** (flag, default: `True`) Apply RMSNorm before the FC layer in the draft path. Use `--no-norm-before-fc` to disable.
 
-- **`--norm-output` / `--no-norm-output`** (flag, default: `True`) Feed post-norm hidden states back across TTT steps to stabilize magnitude drift across speculation depths (Eagle 3.1). Use `--no-norm-output` for the original Eagle 3 behavior.
+- **`--norm-output` / `--no-norm-output`** (flag, default: `True`) Feed post-norm hidden states back across TTT steps to stabilize magnitude drift across speculation depths. Use `--no-norm-output` to disable.
 
 - **`--ttt-steps`** (int, default: `3`) Number of test-time training steps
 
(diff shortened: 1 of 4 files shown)


fynnsu (approved):
Looks good. Do we already have support for Eagle3.1 in vllm? Specifically loading from the speculators config?

orestis-z:
> Looks good. Do we already have support for Eagle3.1 in vllm? Specifically loading from the speculators config?

Yes, see https://github.com/vllm-project/vllm/blob/main/vllm/transformers_utils/configs/speculators/algos.py#L38-L40

orestis-z (commented):
## Defaults affect all speculator types, not just Eagle3

`--draft-arch`, `--norm-before-fc`, and `--norm-output` are global CLI args shared by all speculator types. Changing their defaults here silently changes behavior for DFlash/DSpark users who relied on the old defaults (e.g. `--draft-arch qwen3`).

**Suggested fix:** keep the argparse defaults as `None` and resolve per-speculator-type after parsing:

```python
# After parsing, fill in speculator-specific defaults
if args.draft_arch is None:
    args.draft_arch = "llama" if args.speculator_type == "eagle3" else "qwen3"
if args.norm_before_fc is None:
    args.norm_before_fc = args.speculator_type == "eagle3"
if args.norm_output is None:
    args.norm_output = args.speculator_type == "eagle3"
```

This way Eagle 3.1 gets the new defaults, DFlash/DSpark keep the old ones, and explicit flags still override everything.

---

@fynnsu what do you think? The global effect of this slipped through, my bad.

I remember we previously discussed not complicating the defaults.

I see 3 options:
- we use defaults per speculator type (my recommendation)
- we close this PR and leave the defaults as they are
- we extrapolate that  llama + norms is better for all spec types not just eagle3 (I believe this is risky and/or not scientific)

shanjiaz:
@orestis-z I think Fynn and I agree with you we should go with option 1, defaults for per speculator type.

orestis-z:
Pushed the per-speculator-type defaults fix in 09bdc78:

- Argparse defaults are now `None` for `--draft-arch`, `--norm-before-fc`, `--norm-output`
- After parsing, defaults resolve based on `--speculator-type`:
  - `eagle3` → `llama` arch, norms enabled
  - everything else → `qwen3` arch, norms disabled
- Explicit flags still override in all cases
- Added tests for dflash defaults alongside the existing eagle3 ones

PR #693, opened 2026-06-30
[On Policy Training] Multi-Turn Response Regeneration

WindChimeRan:
## Purpose

Response regeneration should support multi-turn (currently single-turn only). The driver is exposure bias: off-policy multi-turn data conditions each turn on the dataset's responses rather than the target's, so every turn past the first is misaligned. Regenerating turn-by-turn against the on-policy prefix fixes this — and yields longer, more realistic context as a bonus.

## Summary

- Multi-turn output. For any dataset with a messages/conversations field (ultrachat, magpie, open-perfectblend…), every user turn is regenerated against the freshly generated prefix, so a conversations list can now be many turns long instead of always exactly [human, gpt].
- System prompt preserved. A source system turn now appears in the output ({"from":"system",…}) and is used as context during regeneration. Previously system was always dropped.
- Per-turn reasoning_content. Reasoning is now attached to each assistant turn inside conversations. Previously it was a single field in top-level metadata (which the training pipeline drops).

## New behavior

- New <outfile>.errors.jsonl → partial multi-turn records were otherwise indistinguishable from clean ones downstream. (align with SpecForge)
- Empty/None assistant content → treated as a failed turn

## Out-of-scope

* tool use
* multi-modal
* concurrency
* doc: This PR indeed expands the behavior a bit. But I plan to make a larger refactoring PR to the doc, to make response regeneration explicitly the default. 

## Tests

* new unit tests, with diverse dataset format preset
* verified end2end: Qwen2.5-0.5B + `--dataset ultrachat --limit 2`

## Checklist

I have filled in:

### DIFF
--- a/scripts/response_regeneration/script.py
+++ b/scripts/response_regeneration/script.py
@@ -216,8 +216,12 @@ async def worker(
         # are built in lockstep as we walk the turns.
         prefix: list[dict[str, Any]] = []
         out_convs: list[dict[str, Any]] = []
-        finish_reason = None
-        usage = None
+        # Finish reasons are collected per turn: recording only the last would
+        # hide an earlier turn truncated at max_tokens (finish_reason "length").
+        finish_reasons: list[str | None] = []
+        # Usage is summed across turns: each user turn is a separate API call, so
+        # the conversation's cost is the sum, not just the final call's counts.
+        usage = {"prompt_tokens": 0, "completion_tokens": 0, "total_tokens": 0}
         start_time = time.time()
         try:
             for turn in turns:
@@ -235,6 +239,14 @@ async def worker(
                     "max_tokens": args.max_tokens,
                 }
                 async with sem, session.post(endpoint, json=payload) as response:
+                    # A non-2xx reply has no `choices` array; surface the status
+                    # and a short body so `.errors.jsonl` is diagnosable instead
+                    # of recording a bare KeyError('choices').
+                    if not response.ok:
+                        body = (await response.text())[:500]
+                        raise RuntimeError(
+                            f"HTTP {response.status} from {endpoint}: {body}"
+                        )
                     data = await response.json()
 
                 choice = data["choices"][0]
@@ -243,8 +255,10 @@ async def worker(
                 reasoning_content = message.get("reasoning_content")
                 if reasoning_content is None:
                     reasoning_content = message.get("reasoning")
-                finish_reason = choice.get("finish_reason")
-                usage = data.get("usage")
+                finish_reasons.append(choice.get("finish_reason"))
+                turn_usage = data.get("usage") or {}
+                for key in usage:
+                    usage[key] += turn_usage.get(key) or 0
 
                 # An empty/None content (e.g. a response truncated mid-reasoning)
                 # would corrupt the prefix of every following turn and emit a
@@ -262,7 +276,7 @@ async def worker(
 
             metadata = {
                 "idx": idx,
-                "finish_reason": finish_reason,
+                "finish_reasons": finish_reasons,
                 "latency_s": round(time.time() - start_time, 3),
                 "usage": usage,
                 "endpoint": endpoint,
@@ -398,7 +412,8 @@ async def main():
                     continue
 
                 turns = extract_turns(row, prompt_field)
-                if not any(turn["role"] == "user" for turn in turns):
+                # extract_turns returns [] when there is no usable user turn.
+                if not turns:
                     continue
 
                 uuid = row.get("uuid")
--- a/scripts/response_regeneration/script.py
+++ b/scripts/response_regeneration/script.py
@@ -104,6 +104,43 @@ def sanitize_filename(name: str) -> str:
     return name.strip("._")
 
 
+def extract_turns(row, prompt_field):
+    """Extract ordered system/user turns from a dataset row.
+
+    Multi-turn conversations are read from a ``messages`` or ``conversations``
+    field (either the role/content or from/value schema), preserving any system
+    prompt and dropping the original assistant turns so they can be regenerated.
+    Rows without a usable conversation fall back to a single user turn taken
+    from ``prompt_field``.
+    """
+    convs = row.get("messages")
+    if not (isinstance(convs, list) and convs):
+        convs = row.get("conversations")
+
+    if isinstance(convs, list) and convs:
+        turns = []
+        for m in convs:
+            if not isinstance(m, dict):
+                continue
+            role = m.get("role") or m.get("from")
+            content = m.get("content") or m.get("value")
+            if not content:
+                continue
+            if role == "system":
+                turns.append({"role": "system", "content": content})
+            elif role in ("user", "human"):
+                turns.append({"role": "user", "content": content})
+            # original assistant/gpt turns are dropped and regenerated
+        if any(turn["role"] == "user" for turn in turns):
+            return turns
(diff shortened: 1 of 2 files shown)


orestis-z (commented):
Taking this one to review

scripts/response_regeneration/script.py
```diff
@@ -154,74 +191,107 @@ async def worker(
...
+                    data = await response.json()
+
+                choice = data["choices"][0]
+                message = choice["message"]
+                generated_text = message.get("content")
+                reasoning_content = message.get("reasoning_content")
+                if reasoning_content is None:
+                    reasoning_content = message.get("reasoning")
+                finish_reason = choice.get("finish_reason")
+                usage = data.get("usage")
```
orestis-z: [minor] `usage` only captures the last turn. For multi-turn conversations with N user turns, N API calls are made, but `usage` in the output metadata only reflects the final call. Total prompt+completion tokens across the conversation would be more useful for cost tracking. Consider accumulating:

```python
total_usage = {"prompt_tokens": 0, "completion_tokens": 0, "total_tokens": 0}
```

and summing each turn's usage into it.

scripts/response_regeneration/script.py
```diff
@@ -154,74 +191,107 @@ async def worker(
...
+                async with sem, session.post(endpoint, json=payload) as response:
+                    data = await response.json()
+
+                choice = data["choices"][0]
+                message = choice["message"]
+                generated_text = message.get("content")
+                reasoning_content = message.get("reasoning_content")
+                if reasoning_content is None:
+                    reasoning_content = message.get("reasoning")
+                finish_reason = choice.get("finish_reason")
```
orestis-z: [minor] `finish_reason` only captures the last turn. If an earlier turn hit `length` but the last was `stop`, the metadata would hide the truncation. A `finish_reasons` list (or at least flagging if any turn was non-`stop`) would be more informative.

scripts/response_regeneration/script.py
```diff
@@ -320,8 +397,8 @@ async def main():
                 if args.language_filter and row.get("language") != args.language_filter:
                     continue
 
-                prompt = row.get(prompt_field)
-                if not prompt:
+                turns = extract_turns(row, prompt_field)
+                if not any(turn["role"] == "user" for turn in turns):
```
orestis-z: [nit] Redundant check — `extract_turns` already guarantees any non-empty return contains at least one user turn (line 134). So this is equivalent to `if not turns:` but scans the list.

scripts/response_regeneration/script.py
```diff
@@ -262,10 +332,15 @@ async def main():
...
+    error_outfile = f"{base}.errors{ext or '.jsonl'}"
+
     print(f"Using dataset: {dataset_id}")
     print(f"Split: {split}")
     print(f"Prompt field: {prompt_field}")
     print(f"Output file: {args.outfile}")
+    print(f"Error file: {error_outfile}")
     print()
 
     seen_ids = load_seen(args.outfile) if args.resume else set()
```
orestis-z, line 362: [minor] `--resume` will silently retry previously-failed rows since `load_seen` only reads the success file, not `*.errors.jsonl`. This is probably the right behavior (transient endpoint errors should be retried), but worth a one-line comment or `--help` note — a user seeing a row in the error file and then seeing it re-processed on resume might be confused.

scripts/response_regeneration/script.py
```diff
@@ -154,74 +191,107 @@ async def worker(
...
+
+                prefix.append({"role": "user", "content": turn["content"]})
+                out_convs.append({"from": "human", "value": turn["content"]})
+
+                payload = {
+                    "model": args.model,
+                    "messages": prefix,
+                    "max_tokens": args.max_tokens,
+                }
+                async with sem, session.post(endpoint, json=payload) as response:
```
orestis-z: [minor] Semaphore is acquired per-turn, not per-conversation. Each turn independently acquires the semaphore, so request concurrency stays bounded (good). But a single 5-turn conversation occupies a worker for 5 sequential acquisitions while other workers may idle waiting. With high-turn-count datasets this could reduce throughput vs. one-request-per-item. Not a problem for current datasets but worth noting.

orestis-z (commented):
Thanks for the contribution @WindChimeRan — this is a well-motivated and cleanly implemented extension. The on-policy multi-turn regeneration is a meaningful improvement, and the test coverage (especially the downstream normalizer contract test) is solid. A few notes below, nothing blocking.

orestis-z (approved):

orestis-z (commented):
Non-blocking: multi-turn e2e smoke test (even just 1 sample from ultrachat with a real or mocked server) in CI but not blocking given this lives in scripts/, not core library code.

scripts/response_regeneration/script.py
WindChimeRan: I think there are some pre-existing bugs before this PR. Will try to fix it in followup PR.

scripts/response_regeneration/script.py
WindChimeRan: will address semaphore in a dedicated followup PR. (pre-existing issue, code is not clean)

WindChimeRan:
E2E ablation study last week for multi-turn on-policy training: 

### DSpark (Qwen3-8B): training-data policy vs. validation set

Our speculator train & eval is default on **off-policy** (not response re-generation). This is not ideal

| Training responses              | Validation set | full_acc | pos1  | pos4  | pos7  |
|---------------------------------|----------------|----------|-------|-------|-------|
| off-policy (ShareGPT)           | on-policy      | 0.280    | 0.550 | 0.228 | 0.156 |
| on-policy, single-turn          | on-policy      | 0.374    | 0.609 | 0.334 | 0.250 |
| on-policy, multi-turn   #632        | on-policy      | 0.405    | 0.648 | 0.365 | 0.272 |

The bottom three rows share the same held-out on-policy validation set, so they are directly comparable.

WindChimeRan:
Followup PR soon
* semaphore cleanup
* investigate into `--resume`, could be a bug. 
* sample a mixed small set for CI testing
* doc update. multi-turn on policy should be the default.

shanjiaz:
@WindChimeRan @orestis-z  I think @imargulis  has mentioned the unreliable resume bug in this [RFC](https://github.com/vllm-project/speculators/issues/584) and is working on the refactor. Agree not blocking for this PR.

shanjiaz (approved):
I think the logic is sound, but would prefer we cleanup the comments a bit before merging. Thanks

benchislett:
@WindChimeRan can you compare against a _test_ set here? For example, our SPEED-Bench has several multi-turn conversations that are designed for "safe" reuse where collisions are not likely. The "turns" column specifically has multiple prompts you can send in sequence to measure AL over multiple turns without worrying about incoherent decay. PTAL

https://huggingface.co/datasets/nvidia/SPEED-Bench

WindChimeRan:
Thanks @benchislett 

I eval it on SPEED-Bench multi-turn conversations (120 convs with ≥2 turns), sending each turns entry in sequence and feeding the model's own reply back as context. (note that this is a out-of-distribution eval. The speculator is only trained on SharedGPT)

Training data | AL overall | turn 1 | turn 2 | turn 3+
-- | -- | -- | -- | --
on-policy single-turn | 2.031 | 2.019 | 2.066 | 1.628
on-policy multi-turn (this PR) | 2.169 | 2.153 | 2.199 | 1.848

PR #707, opened 2026-07-02
refactor: move training hyperparameters from model configs to trainer kwargs

orestis-z:
## Summary

Follow-up to #687 as discussed in [review comments](https://github.com/vllm-project/speculators/pull/687#discussion_r3500559679):

Move training-only hyperparameters out of model configs into `get_trainer_kwargs` → `forward()`. These fields don't affect model architecture or inference behavior and shouldn't be persisted in `config.json`.

**DFlash/DSpark:**
- `max_anchors` — anchor sampling count, training-only
- `sliding_window_non_causal` — training attention mask flag (vLLM uses its own attention path)

**PEagle:**
- `num_depths`, `down_sample_ratio`, `down_sample_ratio_min` — COD sampling params, training-only
- Inference-time speculation count remains in `SpeculatorsConfig.proposal_methods[0].speculative_tokens`

Not breaking: `SpeculatorModelConfig` uses `extra="allow"`, so old checkpoints with these fields load fine.

## Test plan

- [ ] `pytest tests/integration/models/test_model_forward.py -k dflash` — DFlash param tests
- [ ] `pytest tests/integration/models/test_model_forward.py -k peagle` — PEagle param tests
- [ ] `pytest tests/integration/models/test_model_forward.py` — all model forward tests (no regressions)
- [ ] Existing checkpoints load without error (removed fields silently accepted by `extra="allow"`)

### DIFF
--- a/src/speculators/models/dflash/config.py
+++ b/src/speculators/models/dflash/config.py
@@ -48,14 +48,6 @@ class DFlashSpeculatorConfig(SpeculatorModelConfig):
         ),
     )
 
-    max_anchors: int = Field(
-        default=256,
-        description=(
-            "Maximum number of anchor positions to sample during training "
-            "(controls memory usage and training efficiency)"
-        ),
-    )
-
     target_hidden_size: int | None = Field(
         default=None,
         description="Hidden size of the target model (if different from draft model)",
--- a/src/speculators/models/dflash/config.py
+++ b/src/speculators/models/dflash/config.py
@@ -63,6 +63,13 @@ class DFlashSpeculatorConfig(SpeculatorModelConfig):
         description="Token ID used for masking",
     )
 
+    sliding_window_non_causal: bool = Field(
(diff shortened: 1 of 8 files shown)


src/speculators/models/dflash/config.py
```diff
@@ -71,13 +63,6 @@ class DFlashSpeculatorConfig(SpeculatorModelConfig):
         description="Token ID used for masking",
     )
 
-    sliding_window_non_causal: bool = Field(
-        default=False,
-        description="Use non-causal (bidirectional) masking within draft blocks for "
-        "sliding window attention layers. Full attention layers are always "
-        "bidirectional.",
-    )
```
fynnsu, line 79: This is needed for inference (although possibly not currently used correctly in vllm)

fynnsu (approved):
Thanks for fixing this!

PR #709, opened 2026-07-02
(feat) train: enable train data ratio choice

dmasloff:
## Purpose

The purpose of this PR is to provided users an opportunity to determine the ratio of the training data that will be used for training. Now the proportion of 90% and 10% between train and validation respectively is fixed. In my opinion it is necessary to let user determine this proportion, especially if they use large datasets for training.

## Tests

The suggested fixes are minor and could be checked with fairly any train command. The fact that it works correctly follows from the change in the train length.

## Checklist

I have filled in:

- [+] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [-] The test plan/results, such as providing test command and pasting the results.
- [+] (Optional) The necessary documentation update.
- [+] I (a human) have written or reviewed the code in this pr to the best of my ability.

### DIFF
--- a/docs/cli/train.md
+++ b/docs/cli/train.md
@@ -98,6 +98,8 @@ torchrun --standalone --nproc_per_node=4 scripts/train.py \
 
 - **`--lr`** (float, default: `1e-4`) Learning rate.
 
+- **`--train-data-ratio`** (float, default: `0.9`) Ratio of data to use for training, the rest of the provided data will be used for validation.
+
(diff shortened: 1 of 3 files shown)


fynnsu (approved):
Looks good but could you add input validation in `create_train_val_loaders`? i.e. just check that the value is between 0 & 1?

fynnsu (approved):

PR #711, opened 2026-07-02
DDP + updated AMP strategy

fynnsu:
## Purpose

This PR introduces support for DistributedDataParallel (default distributed), while maintaining a `--fsdp-shard` option. We also maintain automatic mixed precision through `torch.auto_cast` which should automatically handle norm layers and other precision sensitive layers correctly.

We also no longer cast the full model to hidden states dtype before training. Instead we keep the model and optimizer state in F32, and just use autocast to reduce the precision during forward and backward passes. 

## Tests

Testing currently incomplete, will update. 

## Checklist

I have filled in:


orestis-z:
Small thing: the `--fsdp-shard` path still has a bug in `apply_fully_sharded` — the root `fully_shard(model)` call is missing `mp_policy`, so top-level parameters (embeddings, lm_head, norms) don't get bf16 compute + fp32 gradient reduction like the per-layer calls do. Already fixed in #668 (one-liner), worth pulling in:

```diff
-    fully_shard(model)
+    fully_shard(model, mp_policy=mp_policy)
```

fynnsu:
https://github.com/neuralmagic/llm-compressor-testing/actions/runs/28953917504

fynnsu:
Okay trained for 1 epoch on 1000 samples using DDP w/ 2 gpus:

```
┌───────────────────────────────┬───────┐
│            Metric             │ Value │
├───────────────────────────────┼───────┤
│ Acceptance length             │ 1.37  │
├───────────────────────────────┼───────┤
│ Position 0 acceptance         │ 30.3% │
├───────────────────────────────┼───────┤
│ Position 1 acceptance         │ 6.1%  │
├───────────────────────────────┼───────┤
│ Position 2 acceptance         │ 0.9%  │
├───────────────────────────────┼───────┤
│ Overall draft acceptance rate │ 11.2% │
└───────────────────────────────┴───────┘
```

Obviously these numbers aren't good but they make sense for the amount of data and we are getting training signal + no crashes

dsikka:
_Note: for internal purposes only. External contributors are safe to ignore this message._

Tracked in JIRA: https://issues.redhat.com/browse/INFERENG-8953

**JIRA Details:**
- Issue Type: Task
- Priority: Normal
- Activity Type: New Features
- Component: Speculators
- Fix Versions: Speculators 0.7.0
- Status: Backlog
- Assignee: Fynn Schmitt-Ulms

Frozen7771:
Adding some real-training evidence in support of this AMP restructuring — the "master weights are always kept in fp32" change fixes a real correctness bug we hit in production.

Context: on current main, `Trainer.setup_model` casts the model to `hidden_states_dtype` (bf16) **before** `apply_fully_sharded`, so the sharded tensors the optimizer steps on are bf16 and `MixedPrecisionPolicy(param_dtype=bf16)` is effectively a no-op — no fp32 master weights exist anywhere. We hit this training a DSpark draft for GLM-5.2 (753B MoE target, 8-rank FSDP2, torch 2.10, Ascend NPU). Two measurable consequences:

1. **All norm gains freeze at exactly 1.0** (per-step updates ~1e-7–2e-5 « bf16 ULP at 1.0 ≈ 2e-3). After 3 full epochs every `*norm*` tensor in the checkpoint is still bit-exact 1.0 — same symptom that motivated #668.
2. **AdamW decoupled weight decay is silently dropped for every parameter, every step** — this angle doesn't seem to have come up in #668: per-step relative decay lr·λ ≈ 6e-6 « 2⁻⁹ is quantized away on the bf16 in-place write-back. Forensic check: after ~6000 steps, decay alone should leave norm weights at exp(−Σ lr·λ) ≈ 0.982; we read exactly 1.000000, which also rules out fp32 copies hiding anywhere in memory.

**Impact of the fp32-master fix** (fp32 params reaching `fully_shard` + mp_policy on the root wrap; same data/hparams/seed, only the fix changed):

- 10k samples × 3 ep: val accept_len 1.225 → 1.578 (**+28.9%**)
- 50k samples × 3 ep: val accept_len 2.9596 → 3.3583 (**+13.5%**)

Cost we measured: ~+1.7 GB/rank resident, step time unchanged (compute/comm stay bf16).

One side effect to double-check while this PR touches the checkpointer: with fp32 in-memory params, `save_pretrained` writes `dtype: float32` into config.json while tensors are converted to bf16 on disk — downstream consumers (e.g. vLLM serving) then over-allocate or dtype-error. If the checkpointer changes here don't already normalize the config dtype after save, worth including.

Cheap regression canary we've been using — train 1 epoch on the distributed path, then:

```python
import glob
from safetensors import safe_open
worst = 0.0
for f in glob.glob("<ckpt_dir>/*.safetensors"):
    with safe_open(f, framework="pt", device="cpu") as sf:
        for k in sf.keys():
            if "norm" in k.lower():
                worst = max(worst, (sf.get_tensor(k).float() - 1).abs().max().item())
print("max |w-1| over norm tensors:", worst)  # buggy path = exactly 0.0
```

We hit this on NPU, but the mechanism is backend-agnostic bf16 rounding. Happy to contribute the canary as a regression test or the config-dtype normalization if useful — and thanks @orestis-z / @imargulis for the groundwork in #668.

orestis-z (commented):
## Review

### Should fix

**1. `config.json` dtype mismatch**

With fp32 in-memory params, `save_pretrained` writes `torch_dtype: "float32"` in config.json while tensors on disk are bf16 (via `convert_float_dtype`). Downstream consumers (vLLM, `from_pretrained`) may over-allocate or dtype-error. Should normalize config dtype after save, e.g.:

```python
raw_model.save_pretrained(self.path / str(epoch), state_dict=model_state_dict)
# Patch config.json to match the actual on-disk tensor dtype
config_path = self.path / str(epoch) / "config.json"
if config_path.exists():
    import json
    cfg = json.loads(config_path.read_text())
    cfg["torch_dtype"] = str(float_dtype).split(".")[-1]  # "bfloat16"
    config_path.write_text(json.dumps(cfg, indent=2))
```

This affects both `SingleGPUCheckpointer` and `DistributedCheckpointer` save paths.

**2. Validate `--fsdp-shard` outside distributed launch**

`--fsdp-shard` silently does nothing in single-GPU mode. Add an early error after `maybe_setup_distributed()` in `scripts/train.py`:

```python
if args.fsdp_shard and not is_distributed():
    raise ValueError("--fsdp-shard requires torchrun / distributed launch")
```

### Minor

**3. `timer.mark("fetch")` inside autocast** — cosmetic. The fetch timestamp logically belongs before the autocast context, though it doesn't affect correctness.

Frozen7771:
> Should normalize config dtype after save

+1 — one detail from hitting this in production: depending on the transformers version the config key is `dtype` (5.x) vs `torch_dtype` (older), so worth normalizing both if present.

orestis-z (commented):
Architecture is sound but the AMP strategy has gaps: no `GradScaler` for float16 risks gradient underflow, redundant `torch.autocast` layered on FSDP's `MixedPrecisionPolicy`, and hardcoded bf16 in `apply_fully_sharded` ignoring `--hidden-states-dtype`. DDP fresh-init path also lacks test coverage. See inline comments for details.

src/speculators/train/trainer.py
```diff
@@ -427,10 +448,13 @@ def train_epoch(self, epoch: int):
...
-            _draft_tokens, loss, metrics = self.model(
-                **gpu_batch, **(self.config.train_call_kwargs or {})
-            )
+            with torch.autocast(
+                self.device_type, dtype=self.config.hidden_states_dtype
+            ):
+                timer.mark("fetch")
+                _draft_tokens, loss, metrics = self.model(
+                    **gpu_batch, **(self.config.train_call_kwargs or {})
+                )
```
orestis-z, line 460: **AMP correctness: missing `GradScaler` for float16, and redundant autocast under FSDP**

Two concerns with the autocast approach:

**1. float16 without GradScaler.** `--hidden-states-dtype float16` is an explicitly supported option, but `torch.autocast(dtype=float16)` without a `torch.amp.GradScaler` risks gradient underflow (float16's limited exponent range). bf16 doesn't need a scaler because it shares fp32's exponent range, but float16 does. Either:
- Add a GradScaler when `hidden_states_dtype == float16`, or
- Document that `float16` is unsupported with the new AMP strategy and reject it early.

**2. Redundant autocast with FSDP.** When `--fsdp-shard` is active, `apply_fully_sharded` already applies `MixedPrecisionPolicy(param_dtype=bf16)`, which casts parameters to bf16 for forward/backward. Wrapping the forward call in `torch.autocast(dtype=bf16)` on top of that is redundant -- and if someone later changes the FSDP policy's `param_dtype` independently, the two mechanisms could conflict. Consider gating autocast on `not self.config.fsdp_shard`, or at least adding a comment explaining the intended interaction.

src/speculators/train/trainer.py
```diff
@@ -306,6 +315,18 @@ def setup_model(self):
...
+        self.model.to(self.local_rank)  # type: ignore[arg-type]
+
+        if load_checkpoint:
+            self.checkpointer.load_model_state_dict(self.model)
+        else:
+            for param in self.model.parameters():
+                dist.broadcast(param.data, src=0)
+            dist.barrier()
+
+        self.model = DistributedDataParallel(self.model)  # type: ignore[assignment]
```
orestis-z, line 331: **DDP setup: every rank loads checkpoint from disk, but only rank 0's weights survive**

When `load_checkpoint=True`, `SingleGPUCheckpointer.load_model_state_dict` reads the safetensors file on **every** rank independently. Then `DistributedDataParallel(self.model)` broadcasts rank 0's parameters to all other ranks, overwriting whatever they just loaded. The non-rank-0 disk I/O and memory churn are wasted work.

Consider loading only on rank 0 and letting DDP handle the broadcast:
```python
if load_checkpoint:
    if dist.get_rank() == 0:
# DDP constructor broadcasts rank 0's params to all ranks
self.model = DistributedDataParallel(self.model)
```

This also avoids a subtle failure mode on rank-local storage where non-rank-0 loads would fail needlessly.

Also: there is no `test_distributed_fresh_init` equivalent for the DDP path (only FSDP fresh init is tested). The parameter broadcast loop deserves its own test to verify all ranks converge to the same initial weights.

src/speculators/train/distributed.py
```diff
@@ -209,6 +209,6 @@ def apply_fully_sharded(model: torch.nn.Module):
     for layer in model.layers:  # type: ignore[union-attr]
         fully_shard(layer, mp_policy=mp_policy)
 
-    fully_shard(model)
+    fully_shard(model, mp_policy=mp_policy)
```
orestis-z, line 214: **FSDP `param_dtype` hardcoded to bf16, ignoring `--hidden-states-dtype`**

Good fix adding `mp_policy` to the root `fully_shard` call -- previously the root module lacked mixed precision, which was a real bug.

However, `param_dtype=torch.bfloat16` is hardcoded in `MixedPrecisionPolicy` (line 205), while `--hidden-states-dtype` allows `float16` and `float32`. If a user runs `--fsdp-shard --hidden-states-dtype float16`, FSDP will cast parameters to bf16 but `torch.autocast` in the trainer will request float16 -- a dtype mismatch between the two MP mechanisms. Consider threading `hidden_states_dtype` through to `apply_fully_sharded` so the policy is consistent with the trainer's autocast dtype.

eldarkurtic:
Could we perhaps re-evaluate the idea of introducing `torch.nn.parallel.DistributedDataParallel` to get the DDP behavior? It bloats the codebase and adds one more pathway (which is not future-proof) to maintain (especially since pytorch is slowly moving from "wrappers" style into "composable API"). We can get the same DDP training flow by leveraging FSDP pathways (like TorchSpec and DeepSpec are doing). For example:
1. DeepSpec wraps the model with FSDP but sets `sharding_strategy="no_shard"`
2. TorchSpec is using the "composable" API which pytorch is converging to for future development and releases 
```python
  if strategy == "REPLICATE":
      replicate(model)      # tag: "keep full copies, average gradients"
  else:
      fully_shard(model)    # tag: "split the weights across GPUs" 
```

I would lean more towards TorchSpec's solution to be more future-proof and better aligned with pytorch.

fynnsu:
> Could we perhaps re-evaluate the idea of introducing `torch.nn.parallel.DistributedDataParallel` to get the DDP behavior? It bloats the codebase and adds one more pathway (which is not future-proof) to maintain (especially since pytorch is slowly moving from "wrappers" style into "composable API"). We can get the same DDP training flow by leveraging FSDP pathways (like TorchSpec and DeepSpec are doing). For example:
> 


Hi @eldarkurtic, I spoke with a @thisisatharva-rh who's a Red Hatter working on pytorch distributed and it was through discussions with him that we landed on implementing ddp using `DistributedDataParallel`. 

The issue with:

> 1. DeepSpec wraps the model with FSDP but sets `sharding_strategy="no_shard"`

AFAICT this doesn't actually remove the collective operations sync step (even though no data is actually transferred). This is a big part of why we're switching to ddp is to reduce/remove some of these syncs. 

> 2. TorchSpec is using the "composable" API which pytorch is converging to for future development and releases
> 
> 
> ```python
>   if strategy == "REPLICATE":
>       replicate(model)      # tag: "keep full copies, average gradients"
>   else:
>       fully_shard(model)    # tag: "split the weights across GPUs" 
> ```

Although `replicate` seems like a good way to implement ddp, it's purpose (Atharva can maybe correct me if I'm misremembering here) is more for setting up DTensors and internal usages, and the way it implements DDP is slightly more complex and not as supported (yet). I'm probably not explaining that correctly, but from my conversation with Atharva, he recommended against using this for DDP.

> I would lean more towards TorchSpec's solution to be more future-proof and better aligned with pytorch.

The reasons to use DDP are:
1. This is a very well established, stable interface that has good support/composability with other pytorch features (including torch.compile, autocast, etc).
2. It seems like there are no plans to deprecate this, a least not anytime soon. [Related](https://github.com/pytorch/pytorch/issues/146909).
3. The implementation is actually fairly simple, primarily just wrapping the model with it. There is a small extra step when saving checkpoints but otherwise not much complexity.

eldarkurtic:
Alright, thanks a lot for the clarification. Let's then adopt DDP on the merit of its stability and see how it goes down the line. 

On a side note: could we please make sure to include fixes for FP32 master weights that we discussed offline?
1. optimizer state dict should preserve fp32 "bookkeeping": fp32 master weights, optimizer states 
2. checkpoint a model shouldn't destroy fp32 bookkeeping

orestis-z (commented):
All findings from my previous review have been addressed: float16 is now rejected (eliminating the GradScaler concern), `param_dtype` is passed through to FSDP's `MixedPrecisionPolicy`, only rank 0 loads checkpoint weights in the DDP resume path, and `patch_config_dtype` fixes the config.json mismatch. The `_rank0_only` decorator is a clean consolidation of the rank-guard + barrier pattern that previously had to be duplicated across `DistributedCheckpointer` overrides.

LGTM. Recommend approving.

docs/cli/train.md
```diff
@@ -271,7 +287,8 @@ CUDA_VISIBLE_DEVICES=0,1,2,3 torchrun \
   --scheduler-type cosine \
   --scheduler-warmup-steps 100 \
   --checkpoint-freq 2 \
-  --save-best
+  --save-best \
+  --fsdp-shard  # optional: shard parameters when model doesn't fit in one GPU
```
orestis-z: nit: I think the "optionality" is implied as in the other CLI args

scripts/train.py
```diff
@@ -522,6 +528,15 @@ def main(args: argparse.Namespace):  # noqa: C901
         )
     hidden_states_dtype = getattr(torch, args.hidden_states_dtype)
 
+    if hidden_states_dtype == torch.float16:
+        raise ValueError(
```
orestis-z: nit: could raise `NotImplementedError` instead

orestis-z (approved):
Cool!

PR #714, opened 2026-07-03
fix(test): disable torch.compile in MTP num_speculative_steps test

orestis-z:
## Summary

- Fix flaky `TestMTPParams::test_varying_num_speculative_steps` by disabling `torch.compile` on the model instance for this parametrized test
- Add `torch_compile` parameter to `make_mtp_model()` test helper to support unwrapping the dynamo-compiled forward

## Root cause

The MTP forward method is wrapped with `torch.compile` at class definition time. When `test_varying_num_speculative_steps` runs parametrized cases `[1, 2, 5]` in the same process, `torch.compile` first traces the forward loop with `effective_steps=1`. Subsequent cases trigger recompilation for different loop iteration counts, and the inductor-generated kernels produce NaN losses in bfloat16 on the tiny test model.

Test ordering across the full CI suite determines the initial compile cache state (e.g. whether `TestTraining` MTP tests with `effective_steps=3` ran first), making the failure flaky rather than consistent.

Evidence:
- Each parametrized case passes **in isolation**
- Running in reversed order `[5, 2, 1]` passes (larger loop compiles first, smaller recompilations are stable)
- `TORCH_COMPILE_DISABLE=1` → all pass
- `float32` instead of `bfloat16` → 0/60 failures across 20 seeds

## Validation

| | Failures | Runs |
|---|---|---|
| **Without fix** | 16/30 | 8 of 10 runs had failures |
| **With fix** | 0/30 | 10 of 10 runs clean |

## Test plan

- [x] `test_varying_num_speculative_steps[1]` passes
- [x] `test_varying_num_speculative_steps[2]` passes
- [x] `test_varying_num_speculative_steps[5]` passes
- [x] Full `test_model_forward.py` suite passes (134 tests)
- [x] `make quality` (ruff) passes

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -214,6 +214,7 @@ def make_mtp_model(
     num_speculative_steps: int = 3,
     device: str = "cuda:0",
     dtype: torch.dtype = torch.bfloat16,
+    torch_compile: bool = True,
 ) -> MTPDraftModel:
     """Create a tiny MTP model mirroring Qwen3.5-0.8B architecture."""
     from transformers.models.qwen3_5.configuration_qwen3_5 import (  # noqa: PLC0415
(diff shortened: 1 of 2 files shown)


rahul-tuli (approved):
Thanks for looking into this!

PR #716, opened 2026-07-03
fix(regen): stable --resume identity + request retries

imargulis:
Follow-up to #693 (multi-turn response regeneration). This is the resume/robustness
hardening from the RFC #584 discussion.

cc @shanjiaz — flagging you here since this is the follow-up you signed off on.

## Why

`--resume` currently never skips anything. `load_seen` keys on top-level `uuid`/`idx`,
but the writer emits `id` + `metadata.idx` (no top-level `uuid`/`idx`), so `seen_ids`
is always empty and a resumed run reprocesses every row. Even a corrected index-based
key would be unstable: the streaming enumeration index shifts whenever
`--limit`, `--language-filter`, or the input order changes, so it can't safely
identify a row across runs.

Separately, a single transient HTTP/network hiccup sends the whole conversation to
the error file, which is costly on long runs against a busy vLLM server.

## What

- **Stable resume identity.** Resolve a per-row primary id (`id`/`uuid` →
  nested `metadata` id → deterministic content hash), persist it as
  `metadata.primary_id`, and resume against it. Legacy output files still resume
  for rows that carried a real id.
- **Request retries.** Retry transient failures with exponential backoff
  (`--max-retries`, default 3; `--retry-backoff`, default 2.0s), releasing the
  concurrency slot while backing off, instead of failing the conversation on the
  first error.

Both changes preserve #693's behavior.

### DIFF
--- a/scripts/response_regeneration/script.py
+++ b/scripts/response_regeneration/script.py
@@ -116,7 +116,12 @@ def parse_args():
             f"(default: {DEFAULT_RETRY_BACKOFF})"
         ),
     )
-    return parser.parse_args()
+    args = parser.parse_args()
+    if args.max_retries < 0:
+        parser.error("--max-retries must be >= 0")
+    if args.retry_backoff < 0:
+        parser.error("--retry-backoff must be >= 0")
+    return args
 
 
 def sanitize_filename(name: str) -> str:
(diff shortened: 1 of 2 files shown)


shanjiaz (changes requested):
Let me know if you disagree, instead of the multi-source fallback chain (id → uuid → metadata.* → content hash) computed at runtime, could we stamp a `primary_id` on every row once during data prep and persist it? Then both the script and `--resume` just read a single field with no recomputation, fallback logic, and we can get rid of `primary_id_source` as well.  A one-time preprocessing pass to assign ids is a small upfront cost that simplifies this code significantly.

imargulis:
Thanks @shanjiaz — agreed, and I've simplified accordingly. The updated implementation
uses a single stable `metadata.primary_id`: it's stamped onto each sample once, at
regeneration time (right before the record is written), and `--resume` just reads that
one field back. I dropped `primary_id_source` and the multi-source runtime fallback
chain, and removed the legacy back-compat branch entirely — resume never actually worked
before this PR, so there are no old output files to stay compatible with.

One nuance on "stamp during data prep": the script reads the source datasets as an
online HF stream (`load_dataset(..., streaming=True)`), so there's no pre-pass we can run
to materialize and stamp them first — and they don't share a natural id (`magpie` has
`uuid`, `ultrachat` has `prompt_id`, `gsm8k` has none). So instead of a separate prep
pass, the id is derived deterministically from the row (explicit `id`/`uuid`, else a
content hash) and stamped at regeneration time. On resume the same deterministic function
is applied to the streamed rows to match them against the stored `primary_id`, which is
what makes resume correct without a preprocessing step. 

If I misuderstood your intent, will be happy to follow-up on this.

scripts/response_regeneration/script.py
```diff
@@ -185,22 +242,64 @@ async def detect_model(endpoint: str) -> str:
...
+
+class _NonRetryableChatError(RuntimeError):
+    """A non-2xx reply that should fail fast instead of being retried."""
+
+
+def _is_retryable_http_status(status: int) -> bool:
+    return status >= SERVER_ERROR_STATUS or status in RETRYABLE_HTTP_STATUSES
+
+
 async def _post_chat(
```
WindChimeRan, line 245: `_post_chat` reimplements the existing `with_retries` async decorator. `src/speculators/data_generation/vllm_client.py`

scripts/response_regeneration/script.py
WindChimeRan: `metadata.primary_id` is now canonical. Here is some legacy id keys?

scripts/response_regeneration/script.py
imargulis: Thanks — good catch, fixed in the latest push.

**`with_retries` reuse.** Dropped the hand-rolled retry loop and now decorate `_post_chat`
with the existing `with_retries` from `data_generation/vllm_client`. Mapped the
transient-vs-permanent distinction onto its contract: transient failures (network errors
+ HTTP 408/409/425/429/5xx) raise `RuntimeError` so the decorator retries with backoff,
while permanent non-2xx (400/404/…) raise `InvalidResponseError`, which `with_retries`
never retries, so they fail fast. Removed `_NonRetryableChatError` and the `--retry-backoff`
flag (the decorator owns backoff); `--max-retries` passes straight through.

scripts/response_regeneration/script.py
imargulis: Thanks — good catch, fixed in the latest push.

**Legacy id key.** You're right — with `metadata.primary_id` canonical, I missed after reworking the implementation that the top-level `"id": uuid or sample_{idx}` was inconsistent (and the `sample_{idx}` fallback isn't
stable). Set the output `id` to `primary_id` in both the success and error records so they
agree, and dropped the now-unused `uuid` from the queue item. I kept a top-level `id`
(equal to `primary_id`) to preserve #693's output shape rather than drop it — happy to
remove it entirely if you'd prefer, since it's redundant with `metadata.primary_id`.

scripts/response_regeneration/script.py
```diff
@@ -185,21 +234,39 @@ async def detect_model(endpoint: str) -> str:
...
 
 
+# Transient statuses worth retrying: request timeout, conflict, too-early, and
+# rate limiting, plus all 5xx. Other non-2xx replies (e.g. 400/401/404) are
+# permanent config/client errors and fail fast.
+SERVER_ERROR_STATUS = 500
+RETRYABLE_HTTP_STATUSES = {408, 409, 425, 429}
+
+
+def _is_retryable_http_status(status: int) -> bool:
```
WindChimeRan: can be simplified.

scripts/response_regeneration/script.py
```diff
@@ -141,9 +160,39 @@ def extract_turns(row, prompt_field):
     return []
 
 
-def load_seen(path: str):
-    """Load previously processed record IDs from output file."""
-    seen = set()
+def _is_present(value: Any) -> bool:
+    """Return True for a usable identifier (not None / not empty string)."""
+    return value is not None and str(value) != ""
```
WindChimeRan: value not in (None, "")

scripts/response_regeneration/script.py
```diff
@@ -296,9 +369,10 @@ async def worker(
             # Failures go to a separate error file, not the training output; an
             # in-band marker would be invisible (the pipeline drops metadata).
             error_output = {
-                "id": item.get("uuid") or f"sample_{idx}",
+                "id": item["primary_id"],
                 "conversations": out_convs,
                 "metadata": {
+                    "primary_id": item["primary_id"],
```
WindChimeRan: do we need to store `primary_id` three times?

tests/unit/scripts/test_response_regeneration.py
```diff
@@ -234,3 +246,218 @@ def test_emitted_conversation_survives_downstream_normalization():
...
+    """A response whose context-manager entry raises (network-error path)."""
+
+    async def __aenter__(self):
+        raise ConnectionError("boom")
+
+    async def __aexit__(self, *exc):
+        return False
+
+
+def test_post_chat_retries_on_network_error_then_succeeds():
```
WindChimeRan: redundant with `test_post_chat_retries_transient_failure_then_succeeds`

WindChimeRan:
@imargulis only some nits. Then it's ready to go! Thanks.

imargulis:
Thanks @WindChimeRan — applied all the nits: inlined the retryable-status check, simplified `_is_present` to `value not in (None, "")`, dropped the redundant network-error test, and de-duplicated the id by keeping a single top-level `id` (= the row's primary id) that `--resume` keys on (dropping `metadata.primary_id`).

WindChimeRan:
LGTM @shanjiaz

shanjiaz (approved):
Much simpler now thanks for cleaning it up!

shanjiaz:
@imargulis Thanks for the clarification! As long as the hashing is deterministic I think this should work, no need to pre-stamp.

PR #719, opened 2026-07-04
Add mistral training support

shanjiaz:
## Purpose
Added support for Mistral styled checkpoints. Three changes:
- mrope fix: Mistral uses partial rotary embeddings, and it's in the verifier's rope config. When creating the draft model's transformer layer config, we copy the verifier's rope parameters. But the draft model doesn't support partial_rotary_factor. Pop it so we get the correct rope config.
- hidden states generation: Adds continue_final_message: True to the chat completions path used for multimodal hidden state extraction. Without it, Mistral's chat template appends a spurious EOS token, causing a token ID mismatch with the preprocessed input_ids. This is now the default. Text-only models are unaffected, they go through the Completions API with pre-tokenized IDs and bypass chat templates entirely.
- weight loading: adds weight name aliases (tok_embeddings.weight → embed_tokens.weight, etc.) so Mistral's non-standard naming resolves correctly. If we end up supporting more models with non-traditional names we might need a model-free pathway like llm-compressor does.

## Tests
see trained model [evaluation](https://docs.google.com/spreadsheets/d/15NqjQrkOJxQbqKZZZRKli8A2sJHWU6QEmlrmOt6xnS8/edit?gid=0#gid=0)

## Checklist

I have filled in:

### DIFF
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -218,7 +218,8 @@ def create_transformer_layer_config(  # noqa: C901
                 )
                 rope_params.pop("type", None)
                 rope_params.pop("mrope_interleaved", None)
-                rope_params.pop("partial_rotary_factor", None)
+                if "mrope_section" not in rope_params:
+                    rope_params.pop("partial_rotary_factor", None)
             config.rope_parameters = rope_params
     else:
         if hasattr(verifier_config, "rope_scaling"):
@@ -229,7 +230,8 @@ def create_transformer_layer_config(  # noqa: C901
                 )
                 rope_scaling.pop("type", None)
                 rope_scaling.pop("mrope_interleaved", None)
-                rope_scaling.pop("partial_rotary_factor", None)
+                if "mrope_section" not in rope_scaling:
+                    rope_scaling.pop("partial_rotary_factor", None)
             config.rope_scaling = rope_scaling
         config.rope_theta = getattr(verifier_config, "rope_theta", 10000.0)
 
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -216,11 +216,9 @@ def create_transformer_layer_config(  # noqa: C901
                 _maybe_apply_mrope_full_head_hack(
                     rope_params, resolved_head_dim, mrope_full_head_hack
                 )
-                # ``type`` is a legacy alias (only "mrope" on VL models) that
-                # transformers strips during validation and that breaks vLLM's
-                # config checks; drop it while keeping the real MRoPE fields.
                 rope_params.pop("type", None)
                 rope_params.pop("mrope_interleaved", None)
+                rope_params.pop("partial_rotary_factor", None)
             config.rope_parameters = rope_params
     else:
         if hasattr(verifier_config, "rope_scaling"):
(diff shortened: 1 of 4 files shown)


src/speculators/data_generation/vllm_client.py
```diff
@@ -200,7 +200,11 @@ async def generate_hidden_states_async(
             model=model,
             messages=messages,
             max_tokens=1,
-            extra_body={"add_generation_prompt": False, "return_token_ids": True},
+            extra_body={
+                "add_generation_prompt": False,
+                "continue_final_message": True,
+                "return_token_ids": True,
+            },
```
orestis-z, line 207: **`continue_final_message=True` is applied unconditionally — was this tested with non-Mistral models?**

The PR description frames this as a Mistral-specific fix ("avoid spurious EOS tokens from Mistral's chat template"), but it's applied to all models using the chat completions path. For Llama 3-style templates this strips the trailing `<|eot_id|>`, which changes the extracted hidden states (one fewer position — no EOS-position hidden state).

This is arguably *more correct* for hidden state extraction in general (you want the model's representation of the content, not the end-of-turn marker), but it's a silent behavior change for existing data generation pipelines. If you regenerate data for a non-Mistral model with this change, the new hidden states won't match the old ones.

A couple of options:
1. If this is intentionally global (i.e., you believe it's better for all models), document that in the PR description and commit message.
2. If it's Mistral-only, make it conditional — e.g., a flag in the data generation config, or detect from the tokenizer's chat template.

Also applies to lines 249–253 (sync variant).

orestis-z (commented):
Clean design — three well-scoped changes for Mistral support. Weight aliasing approach is pragmatic and well-tested. I agree with @coderabbitai's points on guarding `rope_scaling`/`rope_parameters` against `None`. One design question below.

src/speculators/data_generation/vllm_client.py
shanjiaz: You're right! I'll update the PR description for this. Good thing is this only affects multi-modal models. We haven't done much experiments around these, but I believe this is the more correct behavior.

orestis-z (commented):
Re-reviewed after the `mrope_section` guard fix (9631565). The `partial_rotary_factor` is now correctly preserved for MRoPE models and stripped for non-MRoPE models like Mistral — good fix. Previous concern about `continue_final_message` was addressed in the inline discussion. No new issues. Recommend approving once the merge conflict is resolved.

scripts/train.py
```diff
@@ -221,11 +221,10 @@ def create_transformer_layer_config(  # noqa: C901
                 _maybe_apply_mrope_full_head_hack(
                     rope_params, resolved_head_dim, mrope_full_head_hack
                 )
-                # ``type`` is a legacy alias (only "mrope" on VL models) that
-                # transformers strips during validation and that breaks vLLM's
-                # config checks; drop it while keeping the real MRoPE fields.
                 rope_params.pop("type", None)
                 rope_params.pop("mrope_interleaved", None)
+                if "mrope_section" not in rope_params:
+                    rope_params.pop("partial_rotary_factor", None)
```
rahul-tuli, line 228: nit: could we add a small comment here describing why we are popping this field, the PR desription explains this well, but it'll be nice to have it here

scripts/train.py
```diff
@@ -234,9 +233,10 @@ def create_transformer_layer_config(  # noqa: C901
                 _maybe_apply_mrope_full_head_hack(
                     rope_scaling, resolved_head_dim, mrope_full_head_hack
                 )
-                # Strip legacy fields for consistency with rope_parameters path
                 rope_scaling.pop("type", None)
                 rope_scaling.pop("mrope_interleaved", None)
+                if "mrope_section" not in rope_scaling:
+                    rope_scaling.pop("partial_rotary_factor", None)
```
rahul-tuli, line 242: nit: Same as above

rahul-tuli (approved):
LGTM!

Issue #694, opened 2026-06-30
[Roadmap] On Policy Training


WindChimeRan:
## Motivation

The driver is exposure bias: off-policy multi-turn data conditions each turn on the dataset's responses rather than the target's, so every turn past the first is misaligned. Regenerating turn-by-turn against the on-policy prefix fixes this — and yields longer, more realistic context as a bonus.

## Topic

Ensure no cross-instance attention leakage

- [x] #627

You Only Tokenize Once 

- [ ] #652 #665
- [x] vllm renderer endpoint https://github.com/vllm-project/vllm/pull/46846
- [ ] concurrent preprocessing

Response Regeneration as default

- [ ] [Data] Response regeneration should support multi-turn #693. 
- [ ] [Doc] Make on-policy the documented default for training. Keep off-policy as a cheaper fallback (it skips a full target-model pass over the data) with its acceptance penalty called out, so the tradeoff is explicit.
- [ ] Compatible with multi-modal 
- [ ] Compatible with tool use
- [ ] Support customized sampling config
- [ ] Concurrent preprocessing

Leslie360:
A controlled data point for this roadmap, from a Markov-conditioned (DSpark) drafter on a multimodal OCR target. We trained two drafters for 300 steps with identical data, warm-start, and budget — the only variable being the prev-token source:

- **Arm A (teacher forcing, current default):** train loss 0.697 / token acc 0.643 → served aggregate acceptance **46.3%** (6.94 tokens per round)
- **Arm B (in-loop draft-sampled prev, unrolled Markov argmax as next position's context):** train loss 0.604 / token acc 0.722 → served **40.3%** (6.04 per round)

Served on stock vLLM 0.27.1, temp=0, 100 held-out business samples, same engine and protocol for both arms (a draft-layout footgun — `sample_from_anchor` default mismatch — was corrected on both arms before measuring, so the comparison is layout-clean).

In-loop on-policy sampling did align the training objective — lower loss, higher accuracy — but it served *worse* by ~6pp aggregate acceptance. For this head family at this scale, the exposure-bias intuition did not transfer through to acceptance. This matches the roadmap's emphasis on data regeneration over in-loop sampling, and we wanted to share the counterexample so others don't have to run it.

Happy to share the eval harness details if useful.

PR #724, opened 2026-07-06
 Add --sampling-params to response regeneration script

guan404ming:
## Purpose

Response regeneration requests only set max_tokens, so sampling behavior (temperature, top_p, seed) was stuck on server defaults with no way to customize. Adds a --sampling-params JSON flag merged into each chat-completion request, records it in output metadata for reproducibility, and documents it in the CLI reference. Part of #694.

## Tests

- Sanity-checked arg parsing: --sampling-params '{"temperature": 0.6, "top_p": 0.95}' parses to a dict, omitted flag yields {}, non-object JSON errors out

## Checklist

I have filled in:

### DIFF
--- a/docs/cli/response_regeneration.md
+++ b/docs/cli/response_regeneration.md
@@ -89,6 +89,8 @@ python scripts/response_regeneration/script.py --dataset magpie
 
 - **`--max-tokens`** (int, default: `8192`) Max tokens for generation.
 
+- **`--sampling-params`** (str, default: `None`) JSON object merged into each chat-completion request, e.g. `'{"temperature": 0.6, "top_p": 0.95, "seed": 0}'`. Unset keys use the server defaults.
+
 #### Output Arguments
 
 - **`--outfile`** (str, default: auto-generated) Output JSONL path. If not specified, auto-generated as `{dataset}_{model}.jsonl`.
@@ -133,7 +135,8 @@ Rows are pre-tokenized and ready for training: one row per assistant turn, holdi
     "idx": 0,
     "finish_reason": "stop",
     "usage": {...},
-    "endpoint": "http://127.0.0.1:8000/v1/chat/completions"
+    "endpoint": "http://127.0.0.1:8000/v1/chat/completions",
+    "sampling_params": {...}
   }
 }
 ```
--- a/scripts/response_regeneration/script.py
+++ b/scripts/response_regeneration/script.py
@@ -86,6 +86,14 @@ def parse_args():
         help="max_tokens for generation",
     )
     parser.add_argument(
+        "--sampling-params",
+        default=None,
+        help=(
+            "JSON object merged into each chat-completion request, "
+            'e.g. \'{"temperature": 0.6, "top_p": 0.95, "seed": 0}\''
+        ),
+    )
+    parser.add_argument(
         "--outfile",
         default=None,
         help="Output JSONL path (auto-generated if not specified)",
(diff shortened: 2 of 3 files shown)


scripts/response_regeneration/script.py
```diff
@@ -94,7 +102,13 @@ def parse_args():
...
         help="Only process rows where language==this (e.g., EN)",
     )
-    return parser.parse_args()
+    args = parser.parse_args()
+    args.sampling_params = (
+        json.loads(args.sampling_params) if args.sampling_params else {}
+    )
+    if not isinstance(args.sampling_params, dict):
+        parser.error("--sampling-params must be a JSON object")
+    return args
```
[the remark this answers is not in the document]
guan404ming: Fixed!

docs/cli/response_regeneration.md
WindChimeRan: line 138 and 144 are stale. Now it's no longer meta-data. We store them per turn

WindChimeRan:
LGTM. Please fix the nit then we are ready to go!

docs/cli/response_regeneration.md
guan404ming: Thanks for the good catch, updated the doc: `reasoning_content` is now shown on the gpt turn and metadata uses finish_reasons.

scripts/response_regeneration/script.py
```diff
@@ -317,6 +333,7 @@ async def worker(
                     "model": args.model,
                     "messages": prefix,
                     "max_tokens": args.max_tokens,
+                    **args.sampling_params,
```
orestis-z, line 345: **`**args.sampling_params` can silently override `model`, `messages`, or `max_tokens`.**

Because the unpacking comes *after* the explicit keys, a user passing `--sampling-params '{"max_tokens": 100}'` alongside `--max-tokens 8192` would silently get 100 tokens. Worse, `{"messages": [...]}` would replace the conversation context entirely.

Swapping the order so explicit keys win:

```suggestion
                    **args.sampling_params,
                    "model": args.model,
                    "messages": prefix,
                    "max_tokens": args.max_tokens,
```

Alternatively, you could pop the reserved keys from `args.sampling_params` at parse time with a warning — but reordering is the minimal fix.

orestis-z (commented):
LGTM overall — clean, well-scoped change. One correctness nit on dict unpacking order below.

scripts/response_regeneration/script.py
guan404ming: Thanks, that makes sense!

shanjiaz (approved):
Thanks for adding this!

orestis-z (commented):
**Parsing & validation** -- solid. The `json.loads` call is guarded with `try/except JSONDecodeError`, and the `isinstance(..., dict)` check rejects JSON arrays or scalars. Both error paths route through `parser.error` for clean CLI output.

**Payload construction** -- the earlier ordering issue (my inline comment) is now fixed: `**args.sampling_params` is unpacked first so the explicit keys (`model`, `messages`, `max_tokens`) always take precedence and cannot be silently overridden.

**Metadata recording** -- `sampling_params` is recorded in each output record's metadata, which is good for reproducibility. Minor note: error records (written to the separate error file) don't include `sampling_params`, so you lose that context on failures. Since the params are constant across a run this is not a real problem, but adding it to the error metadata too would keep the schemas symmetric.

**Doc updates** -- the `--sampling-params` entry is clear. The bundled fixes (moving `reasoning_content` to the gpt turn, `finish_reason` -> `finish_reasons` array) correct stale docs from a prior refactor and are appropriate to include here.

Looks good to merge once CI is green.

guan404ming:
Thanks, just pushed to fix the ci issue. Please help enable merge again, thanks!

guan404ming:
Thanks!

PR #728, opened 2026-07-06
fix: resolve hidden state layer count mismatches in DFlash (Fixes #705)

### LINKED ISSUE #705: [Bug]: Hidden state layer count mismatch in GLM-5.2 DFlash training pipeline
### Your current environment

Please provide the following information about your environment if applicable:

- vLLM 0.23
- Speculators 0.6
- CUDA 12.9
- PyTorch 2.11
- Transformers 5.10.4
- Hardware H200 && H100
- Model zai-org/GLM-5.2-FP8

### 🐛 Describe the bug

  There are two interconnected bugs in the DFlash training pipeline that cause shape mismatches when training with `--include-last-layer` / `--no-include-last-layer`.

  **Bug 1: Off-by-one in `launch_vllm.py` (line 67)**

  When `--include-last-layer` is used (the default), `launch_vllm.py` appends `num_hidden_layers` (the *count* of layers, e.g. 78) to `target_layer_ids`, but the last valid layer index is `num_hidden_layers - 1` 
  (e.g. 77) since layers are 0-indexed. This causes vLLM to try extracting hidden states from a non-existent layer, resulting in an expand error.

  Error message:
  RuntimeError: The expanded size of the tensor (6) must match the existing size (5) at non-singleton dimension 1.
  Target sizes: [2443, 6, 6144]. Tensor sizes: [2443, 5, 6144]

  **Bug 2: `ArrowDataset._get_raw_data()` unconditionally strips the last layer (line 381)**

  `ArrowDataset._get_raw_data()` always applies `[:, :-1]` to the loaded hidden states tensor, assuming the last layer is always the verifier's final layer. When data is generated with `--no-include-last-layer`
  (used as a workaround for Bug 1), the hidden states tensor contains only target layers. The unconditional `[:, :-1]` strips one target layer, causing a mismatch with the model's fully-connected layer.

  Error message:
  RuntimeError: mat1 and mat2 shapes cannot be multiplied (8192x24576 and 30720x6144)

  ### To Reproduce

  1. Launch vLLM with 5 target layer IDs and default `--include-last-layer`:
     ```bash
     python scripts/launch_vllm.py /path/to/model \
       --target-layer-ids 8 23 39 55 70 \
       --hidden-states-path /path/to/hidden_states \
       -- --tensor-parallel-size 8
     → Results in Bug 1 (layer index out of bounds)

  2. As a workaround, add --no-include-last-layer:
  python scripts/launch_vllm.py /path/to/model \
    --target-layer-ids 8 23 39 55 70 \
    --no-include-last-layer \
    --hidden-states-path /path/to/hidden_states \
    -- --tensor-parallel-size 8
  2. → Then run training → Results in Bug 2 (shape mismatch)

  Root Cause

  Bug 1 (launch_vllm.py:67):
```
  # num_hidden_layers is the COUNT (e.g. 78), not the last index (e.g. 77)
  if args.include_last_layer and num_hidden_layers not in target_layer_ids:
      target_layer_ids.append(num_hidden_layers)  # wrong: appends 78, should be 77
```

  Bug 2 (data.py:381):
```
  # Always strips the last layer, even when there's no final verifier layer
  return {
      "hidden_states": loaded_hs["hidden_states"][:, :-1].flatten(1),
      ...
  }
```

shotsan:
Closes #705.

Signed-off-by: santosh-agebold <12378904+shotsan@users.noreply.github.com>

### DIFF
--- a/scripts/launch_vllm.py
+++ b/scripts/launch_vllm.py
@@ -63,8 +63,8 @@ def main():
 
     if args.target_layer_ids:
         target_layer_ids = args.target_layer_ids
-        if args.include_last_layer and num_hidden_layers not in target_layer_ids:
-            target_layer_ids.append(num_hidden_layers)
+        if args.include_last_layer and (num_hidden_layers - 1) not in target_layer_ids:
+            target_layer_ids.append(num_hidden_layers - 1)
         warnings.warn(
             f"Using custom target layer ids {target_layer_ids}. These "
             "must also be explicitly passed into the training script.",
@@ -75,7 +75,7 @@ def main():
             2,
             num_hidden_layers // 2,
             num_hidden_layers - 3,
-            num_hidden_layers,
+            num_hidden_layers - 1,
         ]
 
     speculative_config = {
--- a/src/speculators/train/data.py
+++ b/src/speculators/train/data.py
@@ -93,7 +93,7 @@ def create_empty_sample(
     }
 
 
-def standardize_data_v1(data: dict[str, Any]) -> dict[str, Any]:
+def standardize_data_v1(data: dict[str, Any], include_last_layer: bool = True) -> dict[str, Any]:
     # v1 data format:
     # {
     #  "input_ids": [seq_len],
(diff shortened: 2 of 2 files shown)


WindChimeRan:
Tend to reject this PR because of the upstream fix: https://github.com/vllm-project/speculators/issues/705#issuecomment-4877271483

shotsan:
Closing based on upstream fix clarification. Apologies for the noise.

PR #729, opened 2026-07-06
Refactor(regen): emit generation boundary as loss mask for on-policy data, eliminate redundant re-tokenize roundtrip

WindChimeRan:
# Purpose

The current on-policy response regeneration has re-tokenization roundtrip redundancy. This PR helps eliminate the roundtrip:

```
BEFORE                               AFTER
regen -> text                        regen -> return_token_ids
write conversations                  write input_ids + loss_mask (boundary)
re-tokenize + REGEX/HF mask          _preprocess_batch passthrough (no masking)
vLLM -> hidden states                vLLM -> hidden states
```

The masking machinery still exists (off-policy needs it until PR2); it is just no longer on the on-policy path.

## Concrete example — a 2-turn conversation

Before (one row of text, masked later by regex over the whole rendering):

```json
{"id": "sample_7", "conversations": [{"from":"human","value":"2+2?"}, {"from":"gpt","value":"4"}, {"from":"human","value":"3+3?"}, {"from":"gpt","value":"6"}]}
```

After (one row per assistant turn, mask already applied):

```json
{"id":"sample_7_turn0", "input_ids":[<prompt>,<"4">], "loss_mask":[0,0,0,1,1], "conversations":[...]}
{"id":"sample_7_turn1", "input_ids":[<prompt+history>,<"6">], "loss_mask":[0,...,0,1,1], "conversations":[...]}
```

`loss_mask` 0 is the prompt the target conditioned on; 1 is what it generated. No `{% generation %}` markers, no regex — just the boundary the serving engine reports.

## Why

On-policy data never needed regex masking: we generated the response, so we already know where the assistant tokens are, and capturing that boundary is exact.

It is also correct for multi-turn reasoning models, where each turn supervises the `<think>` it actually generated while history keeps only parsed content — matching inference. The regex path cannot reproduce this, because the same turn tokenizes differently as "generated" vs as "history".

## What changes (observable)

| | Before | After |
|---|---|---|
| Output schema | `conversations` (text) | `input_ids` + `loss_mask` (+ `conversations`, review-only) |
| Rows per conversation | 1 | 1 per assistant turn |
| Mask made | downstream regex / HF | generation boundary |
| Endpoint | text out | must support `return_token_ids` |

`conversations` is retained as an inert human-review twin of `input_ids`: `_preprocess_batch` routes on `input_ids`/`loss_mask` (checked first) and drops `conversations` before training, so it is never re-masked.

## Tests

`build_boundary_sample` produces the boundary mask; pre-tokenized rows pass through `_preprocess_batch` without a processor, and a review `conversations` field is ignored. Off-policy suite unchanged (56 passed / 2 skipped).

## Scope — step 1 of 5 ("renderer is the single source of truth")

1. (this PR) on-policy → generation boundary; stop feeding the regex path. (Please note that this PR is a concrete standalone PR. No speculative scaffolding. )
2. To unify the chat template generation tag. we should vendor generation-tagged template from trl https://github.com/huggingface/trl/issues/4879. 
3. on-policy → we have to diverge a bit from pure on-policy regen: real inference has a thinking compression mechanism - newer models, e.g., GLM5.2, Qwen3.6 only keep the latest turn's thinking, and drop 0:n-1  thinking. But we need all thinking because that's where speculative decoding can helps. PR1 is correct (completely on policy, no divergence) but inefficient because it fan out multiturn to multi-row jsonl. PR3 can linearize them back to single-row, with some tradeoff. This is also the same solution with trl, llama-factory for multi-turn SFT. We can discuss further on the tradeoff when I have a draft PR. 
4. off-policy → vLLM `/render` this path can be decoupled from the regex & hf path
5. delete the regex/HF masking once both producers are redirected.

small followup PRs:

`--turn-dropout` not compatible with this PR. need to run some tests on real data to see if we should improve it or remove it. 

## Checklist

I have filled in:

### DIFF
--- a/scripts/response_regeneration/script.py
+++ b/scripts/response_regeneration/script.py
@@ -6,7 +6,6 @@ import json
 import os
 import re
 import sys
-import time
 from typing import Any
 
 import aiohttp
@@ -187,11 +186,16 @@ def _primary_identifier(row: dict[str, Any]) -> str:
 
 
 def load_seen(path: str) -> set[str]:
-    """Load previously written output ids from the output file.
+    """Load previously completed conversation ids from the output file.
 
-    Each record stores its stable id under the top-level ``id`` (equal to the
-    row's :func:`_primary_identifier`), so a resumed run skips a row when its
-    recomputed id is already present.
+    A conversation fans out to one row per assistant turn, whose ``id`` carries a
+    ``_turn<N>`` suffix; the conversation's own :func:`_primary_identifier` is
+    kept alongside it as ``primary_id``. Resume keys on that, since the suffixed
+    ids never match a recomputed one. Rows are written only after every turn
+    succeeds, so one row is enough to mark the conversation done.
+
+    ``id`` is the fallback for output files written before the fan-out, where the
+    top-level ``id`` *was* the primary identifier.
     """
     seen: set[str] = set()
     if not os.path.isfile(path):
@@ -203,8 +207,11 @@ def load_seen(path: str) -> set[str]:
                 obj = json.loads(line)
             except json.JSONDecodeError:
                 continue
-            if _is_present(obj.get("id")):
-                seen.add(str(obj["id"]))
+            key = obj.get("primary_id")
+            if not _is_present(key):
+                key = obj.get("id")
+            if _is_present(key):
+                seen.add(str(key))
     return seen
 
 
@@ -270,6 +277,19 @@ async def _post_chat(
         return await response.json()
 
 
+def build_boundary_sample(
+    prompt_token_ids: list[int],
+    completion_token_ids: list[int],
+) -> tuple[list[int], list[int]]:
+    """Build one training sample: prompt (loss_mask 0) + generated tokens (1).
+
+    The generation boundary is the mask -- no ``{% generation %}`` markers, no regex.
+    """
+    input_ids = [*prompt_token_ids, *completion_token_ids]
+    loss_mask = [0] * len(prompt_token_ids) + [1] * len(completion_token_ids)
+    return input_ids, loss_mask
+
+
 async def worker(
     session: aiohttp.ClientSession,
     queue: "asyncio.Queue[dict[str, Any]]",
@@ -280,12 +300,10 @@ async def worker(
     progress,
     stats: dict[str, int],
 ):
-    """Worker that pulls conversations from the queue and regenerates them.
+    """Regenerate each queued conversation into pre-tokenized training samples.
 
-    Each user turn is sent to the endpoint with the freshly generated prefix
-    (system + regenerated history); the original assistant turns are discarded
-    and replaced by the model's responses. Single-turn rows are the degenerate
-    case of one user turn.
+    One sample per assistant turn: the prompt the target conditioned on
+    (loss_mask 0) followed by the tokens it generated (1).
     """
     while True:
         item = await queue.get()
@@ -295,28 +313,23 @@ async def worker(
 
         idx = item["idx"]
         turns = item["turns"]
+        conv_id = item["primary_id"]
 
-        # The API prefix (role/content) and the output conversation (from/value)
-        # are built in lockstep as we walk the turns.
         prefix: list[dict[str, Any]] = []
-        out_convs: list[dict[str, Any]] = []
-        finish_reasons: list[str | None] = []
-        usage = {"prompt_tokens": 0, "completion_tokens": 0, "total_tokens": 0}
-        start_time = time.time()
+        samples: list[dict[str, Any]] = []
         try:
             for turn in turns:
                 if turn["role"] == "system":
                     prefix.append({"role": "system", "content": turn["content"]})
-                    out_convs.append({"from": "system", "value": turn["content"]})
                     continue
 
                 prefix.append({"role": "user", "content": turn["content"]})
-                out_convs.append({"from": "human", "value": turn["content"]})
 
                 payload = {
                     "model": args.model,
                     "messages": prefix,
                     "max_tokens": args.max_tokens,
+                    "return_token_ids": True,  # prompt_token_ids + completion token_ids
                 }
                 data = await _post_chat(
                     session,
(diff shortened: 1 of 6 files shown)


scripts/response_regeneration/script.py
```diff
@@ -228,79 +238,76 @@ async def worker(
 
         idx = item["idx"]
         turns = item["turns"]
+        conv_id = item.get("uuid") or f"sample_{idx}"
```
orestis-z: **Bug: `--resume` is broken — wrong key + id mismatch.**

`item.get("uuid")` will always be `None` because `main()` puts the stable id under the key `"primary_id"` ([line 496](https://github.com/vllm-project/speculators/blob/a6fe4e8f55291dcfb81e66858d5f0566e04790f7/scripts/response_regeneration/script.py#L496)), not `"uuid"`. So `conv_id` falls back to `f"sample_{idx}"` for every row.

Second issue: even after fixing the key to `"primary_id"`, `--resume` still won't work. The output rows now have ids like `{primary_id}_turn0`, `{primary_id}_turn1` (line 284), but `main()` checks `if primary_id in seen_ids` (line 490) — exact string match. `load_seen()` collects the `_turnN`-suffixed ids, which never match the bare `primary_id`, so every row gets re-processed on resume.

Suggested fix: use `item.get("primary_id", f"sample_{idx}")` here, and also update `load_seen()` to strip the `_turnN` suffix (or store the bare `primary_id` as an additional field in each output row for resume lookup).

scripts/response_regeneration/script.py
```diff
@@ -228,79 +238,76 @@ async def worker(
...
                 prefix.append({"role": "user", "content": turn["content"]})
-                out_convs.append({"from": "human", "value": turn["content"]})
 
                 payload = {
                     "model": args.model,
                     "messages": prefix,
                     "max_tokens": args.max_tokens,
+                    "return_token_ids": True,  # prompt_token_ids + completion token_ids
                 }
                 data = await _post_chat(session, endpoint, payload)
```
orestis-z: **Bug: `max_retries=args.max_retries` dropped — `--max-retries` CLI arg silently ignored.**

The original call passed `max_retries=args.max_retries` to `_post_chat` (which is decorated with `@with_retries`). This line drops it, so the decorator falls back to `DEFAULT_MAX_RETRIES` regardless of what the user passes via `--max-retries`.

```suggestion
                data = await _post_chat(session, endpoint, payload, max_retries=args.max_retries)
```

orestis-z (commented):
Design is sound — capturing the generation boundary directly as the loss mask instead of round-tripping through text+regex is the right approach, and the implementation is clean. Two bugs in the worker that break existing CLI behavior:

WindChimeRan:
Rebased onto `main` (now includes #716 fix).

- `--max-retries`: no change needed — the rebase restores `max_retries=args.max_retries` on its own.
- `--resume`: fixed — `conv_id = item["primary_id"]`, every row now carries `primary_id`, and `load_seen()` keys on it.
- Added a `worker()` test over a fake endpoint that fails on both of the original bugs.
- `_passthrough_pretokenized` now raises on an `input_ids`/`loss_mask` length skew instead of silently shifting the mask.
- Dropped a stray test fixture,  updated the regen docs to the new output schema.

Both comments describe the tree after rebasing onto #716; the reviewed head predates it, which is why the first one needs no code change. Force-pushed, so the inline comments are marked outdated.

WindChimeRan:
scope updated. 

multi-turn multi-row fan out is a bit annoying to me. will address it in PR3
image

orestis-z (commented):
Well-motivated refactor, correctly implemented. Two minor inconsistencies in the passthrough path: `minimum_valid_tokens` truthiness check differs from the normal path, and `turn_dropout` is silently ignored for pre-tokenized data. See inline comments. Recommend approving.

src/speculators/data_generation/preprocessing.py
```diff
@@ -507,6 +507,34 @@ def _parse_conv_tools(conv_tools: object, idx: int) -> list | None:
...
+        # the collator packs each key independently, silently shifting the mask
+        # against the ids for every sample packed after this one.
+        if len(ids) != len(mask):
+            raise ValueError(
+                f"Pre-tokenized row shape mismatch: "
+                f"input_ids={len(ids)}, loss_mask={len(mask)}"
+            )
+        trimmed_ids = ids[:max_length]
+        trimmed_mask = mask[:max_length]
+        if minimum_valid_tokens and sum(trimmed_mask) < minimum_valid_tokens:
```
orestis-z: Nit: this uses truthiness (`if minimum_valid_tokens`), while the normal path below uses `if minimum_valid_tokens is not None:`. With `minimum_valid_tokens=0`, the passthrough skips the filter while the normal path runs it. The practical impact is zero (threshold 0 filters nothing), but aligning the check avoids a future foot-gun if the semantics of 0 ever matter:

```suggestion
        if minimum_valid_tokens is not None and sum(trimmed_mask) < minimum_valid_tokens:
```

src/speculators/data_generation/preprocessing.py
```diff
@@ -517,6 +545,11 @@ def _preprocess_batch(
 ) -> dict[str, list]:
     """Process a batch of conversations into tokenized format with loss masks."""
 
+    # On-policy regeneration rows are already masked (boundary); pass them through
+    # instead of re-tokenizing and re-masking.
+    if "input_ids" in examples and "loss_mask" in examples:
+        return _passthrough_pretokenized(examples, max_length, minimum_valid_tokens)
```
orestis-z, line 554: `turn_dropout` is accepted by `_preprocess_batch` but silently ignored here because the early-return bypasses it. If someone passes `--turn-dropout` when training on a pre-tokenized regen dataset, nothing will warn them. Consider logging when the passthrough skips a non-default parameter:

```python
    if "input_ids" in examples and "loss_mask" in examples:
        if turn_dropout:
            log.warning("turn_dropout is not supported for pre-tokenized rows; ignoring")
        return _passthrough_pretokenized(examples, max_length, minimum_valid_tokens)
```

rahul-tuli (commented):
LGTM, pending CI

orestis-z (commented):
LGTM. Both prior findings addressed in the latest commit. Recommend approving.

orestis-z (approved):

rahul-tuli (approved):

Issue #613, opened 2026-06-16
[Bug]: RedHatAI/Qwen3.6-35B-A3B-NVFP4 - Dflash - Very Low Acceptance Despite Good Validation Metrics


huaxuan250:
### Your current environment

Please provide the following information about your environment if applicable:

- vLLM: 0.22.1 for Hidden State Extraction + Inference Benchmark 
- Speculators: Mainline `b6bccf38525971b7b9721a47d69e434fbce31391` as of 06/13/2025
- CUDA: 13.2
- PyTorch: 2.11.0
- Transformers: 5.10.2
- Hardware: p4de for benchmark, p5en for training
- Model: RedHatAI/Qwen3.6-35B-A3B-NVFP4

### 🐛 Describe the bug

# Overview:
Benchmarking with a subset of training data yields very low acceptance rate. 
Training data is 130K worth of at most 10240 context size of internal domain-specific prompts. 
Even when I invoke with a singular prompt (still in training distribution), the acceptance rate is very low.

Training's validation metrics seem fine and no anomalies. 

### Low Acceptance Rate:
```
(APIServer pid=1874323) INFO:     Started server process [1874323]
(APIServer pid=1874323) INFO:     Waiting for application startup.
(APIServer pid=1874323) INFO:     Application startup complete.
(Worker_TP0 pid=1875538) WARNING 06-16 19:27:42 [jit_monitor.py:103] Triton kernel JIT compilation during inference: _zero_kv_blocks_kernel. This causes a latency spike; consider extending warmup to cover this shape/config.
(Worker_TP0 pid=1875538) WARNING 06-16 19:27:42 [jit_monitor.py:103] Triton kernel JIT compilation during inference: _compute_slot_mapping_kernel. This causes a latency spike; consider extending warmup to cover this shape/config.
(Worker_TP0 pid=1875538) WARNING 06-16 19:27:42 [jit_monitor.py:103] Triton kernel JIT compilation during inference: eagle_prepare_next_token_padded_kernel. This causes a latency spike; consider extending warmup to cover this shape/config.
(Worker_TP0 pid=1875538) WARNING 06-16 19:27:42 [jit_monitor.py:103] Triton kernel JIT compilation during inference: copy_and_expand_dflash_inputs_kernel. This causes a latency spike; consider extending warmup to cover this shape/config.
[...]

PR #568, opened 2026-06-03
[Features] Qwen3.6 PartialRoPE supports

PatchouliTIS:
# PR Description: Qwen3.6 and Multimodal Verifiers Support

## Purpose
Add end-to-end training support for Qwen3.6 (Hybrid Attention, MoE) and other multimodal verifiers (Qwen3-Omni, Qwen3.5-MoE) to the speculators training pipeline. This includes proper handling of MRoPE position encoding, partial rotary factors, multimodal data preprocessing with sidecar tensors, and a fix for the token-id mismatch that caused 100% sample rejection during online hidden-states generation.

### Qwen3.6 Hybrid Attention & RoPE Training (`eagle3/core.py`, `scripts/train.py`)
- Add `_select_rotary_emb_class` and `PartialMRoPE` to support `partial_rotary_factor < 1` in MRoPE-aware draft models.
- Add `--draft-mrope-full-head-hack` (default on): rescales `mrope_section` by `1/partial_rotary_factor` and pins `partial_rotary_factor=1.0` to ensure bit-equivalent rotation between HF's `rotate_half` (training) and vLLM's neox-partial rotation (inference).
- Add `--draft-rope-scaling`, `--draft-rope-theta`, `--draft-max-position-embeddings` for explicit RoPE override.
- Add `--draft-intermediate-size`, `--draft-num-attention-heads`, `--draft-num-key-value-heads`, `--draft-head-dim` for MoE drafter geometry override.
- Support `unwrap_verifier_text_config` for nested multimodal configs (`thinker_config.text_config`).

### Multimodal Data Pipeline (`preprocessing.py`, `train/data.py`)
- Add multimodal sidecar persistence: `_save_multimodal_sidecar` writes `image_grid_thw`, `video_grid_thw`, etc. to `.safetensors` files.
- Add `_build_multimodal_loss_mask`: zeroes loss at placeholder token positions (image/video/audio pad tokens).
- Add `_loss_mask_from_assistant_token_spans` and placeholder-aware `_loss_mask_from_ids_fallback` for accurate multimodal assistant mask alignment.
- `ArrowDataset` now calls `.with_format(None)` after `load_from_disk` to expose multimodal metadata columns hidden by `prepare_data.py`'s persisted `set_format`.
- Add `_make_rope_index_fn` supporting both Qwen3-Omni (nested `thinker_config`) and Qwen3.5/3.6 MoE (flat `vision_config` + `text_config`) config layouts for 3D MRoPE `position_ids` generation.
[...]

PR #733, opened 2026-07-07
[Features] Partial MRope Support in Draft Model Training

PatchouliTIS:
## Purpose
Fix the RoPE/MRoPE training-vs-inference mismatch that can cause 0% acceptance for DFlash on partial-MRoPE verifier configs such as Qwen3.5/3.6 and Gemma4. This addresses the issue discussed in #613.

This PR keeps the scope intentionally small and only includes the partial-MRoPE fix split out from the broader Qwen3.6/multimodal work:

- Add a partial-neox RoPE helper in `src/speculators/models/eagle3/rotary_partial.py` that:
  - preserves HF behavior when `rotary_dim == head_dim`;
  - rotates only the leading `rotary_dim` channels when `rotary_dim < head_dim`;
  - keeps pass-through channels unchanged;
  - matches vLLM partial-neox channel layout.
- Update `scripts/train.py` so draft configs preserve real MRoPE fields:
  - `mrope_section`
  - `mrope_interleaved`
  - `partial_rotary_factor`
- Drop only the legacy `type` alias from RoPE config because it breaks downstream config validation.
- Add `mrope_full_head_hack` handling:
  - by default, partial MRoPE configs are rescaled to full-head semantics for HF training / vLLM inference compatibility;
  - users can disable it with `--no-draft-mrope-full-head-hack`;
  - invalid non-integer rescaling fails fast with a clear error.
- Preserve existing nested RoPE flattening behavior for configs with per-layer-type RoPE settings such as `full_attention` / `sliding_attention`.

## Checklist

I have filled in:


PatchouliTIS:
I have already extracted the partial mrope section and submitted it as a separate PR. Could you please check it out ? @shanjiaz @orestis-z

orestis-z (commented):
Nice split from #568 — this is exactly the RoPE fix needed for #613.

The current tests verify the new behavior works, but neither would have caught the original bug on main — they depend on the new code (`rotary_partial` module / `mrope_full_head_hack` parameter) and fail with `ImportError`/`TypeError` rather than a meaningful assertion.

Would be great to add a lightweight regression test. Using the `patch_verifier` fixture already in the PR, it's pure config math — no downloads, no GPU, runs in milliseconds:

```python
def test_partial_rotary_factor_not_leaked(patch_verifier):
    """Regression for #613: partial_rotary_factor must not cause
    a train/inference rotary dimension mismatch."""
    vc = _make_verifier_config(
        rope_parameters={
            "rope_type": "default",
            "mrope_section": [8, 12, 12],
            "partial_rotary_factor": 0.25,
            "rope_theta": 1000000.0,
        },
        head_dim=256,
    )
    patch_verifier(vc, "5.0.0")

    config = _build()

    # Simulate what vLLM's get_rope does at inference:
    partial = config.rope_parameters.get("partial_rotary_factor", 1.0)
    inference_rotary_dim = int(config.head_dim * partial)

    assert inference_rotary_dim == config.head_dim, (
        f"Train/inference rotary dim mismatch: training uses {config.head_dim}, "
        f"inference would use {inference_rotary_dim} (partial_rotary_factor={partial})"
    )
```

This test fails on main (256 vs 64) and passes with this PR.

scripts/train.py
```diff
@@ -160,18 +201,30 @@ def create_transformer_layer_config(  # noqa: C901
...
+                rope_params.pop("type", None)
             config.rope_parameters = rope_params
-
-            _MROPE_KEYS = ("mrope_section", "mrope_interleaved", "type")  # noqa: N806
-            for key in _MROPE_KEYS:
-                config.rope_parameters.pop(key, None)
     else:
         if hasattr(verifier_config, "rope_scaling"):
-            config.rope_scaling = deepcopy(verifier_config.rope_scaling)
+            rope_scaling = deepcopy(verifier_config.rope_scaling)
```
orestis-z, line 232: `resolved_head_dim` is not defined in this function — on `main` the variable is called `head_dim` (line 106). This will `NameError` at runtime. Looks like a leftover from the #568 refactor.

Also note `head_dim` can be `None` (from `getattr(..., None)`), but this function's type hint expects `int`. You may need:
```python
resolved_head_dim = head_dim or (verifier_config.hidden_size // num_attention_heads)
```

scripts/train.py
```diff
@@ -160,18 +201,30 @@ def create_transformer_layer_config(  # noqa: C901
...
+            if isinstance(rope_scaling, dict):
+                _maybe_apply_mrope_full_head_hack(
+                    rope_scaling, resolved_head_dim, mrope_full_head_hack
```
orestis-z, line 235: The old code also stripped `mrope_interleaved`:
```python
_MROPE_KEYS = ("mrope_section", "mrope_interleaved", "type")
```
This PR only strips `type`. Is keeping `mrope_interleaved` in the draft config intentional?

src/speculators/models/eagle3/rotary_partial.py
```diff
@@ -0,0 +1,109 @@
+"""Patch HF rotary helper to match vLLM partial-neox behavior.
```
orestis-z, line 1: This module is added and tested but `install_partial_neox_rotary()` is never called from the training pipeline. Is this intended for a follow-up (e.g. the `--no-draft-mrope-full-head-hack` path), or should it be wired up in this PR?

PatchouliTIS:
> Nice split from #568 — this is exactly the RoPE fix needed for #613.
> 
> The current tests verify the new behavior works, but neither would have caught the original bug on main — they depend on the new code (`rotary_partial` module / `mrope_full_head_hack` parameter) and fail with `ImportError`/`TypeError` rather than a meaningful assertion.
> 
> Would be great to add a lightweight regression test. Using the `patch_verifier` fixture already in the PR, it's pure config math — no downloads, no GPU, runs in milliseconds:
> 
> ```python
> def test_partial_rotary_factor_not_leaked(patch_verifier):
>     """Regression for #613: partial_rotary_factor must not cause
>     a train/inference rotary dimension mismatch."""
>     vc = _make_verifier_config(
>         rope_parameters={
>             "rope_type": "default",
>             "mrope_section": [8, 12, 12],
>             "partial_rotary_factor": 0.25,
>             "rope_theta": 1000000.0,
>         },
>         head_dim=256,
>     )
>     patch_verifier(vc, "5.0.0")
> 
>     config = _build()
> 
>     # Simulate what vLLM's get_rope does at inference:
>     partial = config.rope_parameters.get("partial_rotary_factor", 1.0)
>     inference_rotary_dim = int(config.head_dim * partial)
> 
>     assert inference_rotary_dim == config.head_dim, (
>         f"Train/inference rotary dim mismatch: training uses {config.head_dim}, "
>         f"inference would use {inference_rotary_dim} (partial_rotary_factor={partial})"
>     )
> ```
> 
> This test fails on main (256 vs 64) and passes with this PR.

New test method added and the params missing error fixed.

orestis-z (approved):
LGTM — `resolved_head_dim` fix and regression test look good. Thanks for the quick turnaround.

Minor nit: `rotary_partial.py` is added and tested but `install_partial_neox_rotary()` is never called from the training pipeline — dead code for now.

tests/unit/train/test_rope_config.py
```diff
@@ -0,0 +1,215 @@
...
+
+def _build(**kwargs):
+    defaults = {
+        "verifier_name_or_path": "dummy",
+        "num_layers": 1,
+        "draft_arch": "llama",
+        "hidden_act": "silu",
+        "sliding_window": 0,
+        "sliding_window_indices": [],
+    }
```
orestis-z, line 61: The `_build` helper passes `sliding_window_indices=[]`, but [`create_transformer_layer_config`](https://github.com/vllm-project/speculators/blob/5497822367a307c8fd6d38961135c088685d50e2/scripts/train.py#L120) expects `full_attention_indices`. This will `TypeError` at runtime — `full_attention_indices` is required (no default), and `sliding_window_indices` is not a recognized parameter.

```suggestion
    defaults = {
        "verifier_name_or_path": "dummy",
        "num_layers": 1,
        "draft_arch": "llama",
        "hidden_act": "silu",
        "sliding_window": 0,
        "full_attention_indices": [],
    }
```

orestis-z (commented):
All three items from the prior review are resolved (ABC fix, mrope_interleaved stripping, install wiring). One test bug below from the latest push.

tests/unit/train/test_rope_config.py
PatchouliTIS: params error fixed.

orestis-z (commented):
LGTM. Clean approach to a real train/inference mismatch (#613) -- the two-path design (full-head rescaling by default, partial-neox patch as opt-out) is well-motivated and both paths correctly align HF training with vLLM's `get_rope` semantics. All prior review items resolved. Recommend approving once merge conflicts and quality checks are addressed.

PatchouliTIS:
Hi @orestis-z , I have updated my codes and all checks been passed. Would you kindly check it out ?

PR #735, opened 2026-07-07
Refactor hidden states transfer w/ abstract backend

fynnsu:
## Purpose

This is a precursor to landing other hidden states connectors (e.g. Mooncake connector: #605 #710). It refactors out logic for setting up the connector, adding argparse entries, etc. Currently this pr only includes the default "file" transfer system with follow up prs adding mooncake and similar. 

This pr:
- Adds a new subpackage to supply hidden states connectors (separate package so that it can be installed in both vllm and speculators environments)
- Vendors a simple version of the "file" connector so that existing behavior will be maintained even if the user doesn't install hs_connectors
- Updates train/launch_vllm scripts to setup argparse based on available connectors
- Updates dataloading to defer to the connector implementations. 

## Tests

Nightly run: https://github.com/neuralmagic/llm-compressor-testing/actions/runs/28890598247/job/85702003140

## Checklist

I have filled in:


orestis-z (commented):
Thanks for the refactor — the `HiddenStatesTransfer` abstraction is the right direction for pluggable backends. One architectural question and a few mechanical issues:

### Does `hs_connectors` need to be a subpackage in this repo?

The motivation is that it needs to be installable in vLLM environments — #605 confirms this, vLLM loads the Mooncake connector at runtime via `kv_connector_module_path: "hs_connectors.mooncake_hidden_states_connector"`. But nesting it inside the speculators repo means you can't `pip install hs-connectors` independently in a vLLM deployment that doesn't have a speculators checkout. The vendored `_InlineFileBackend` fallback already acknowledges this fragility. If it's a shared dependency between two projects, it should be its own repo/PyPI package. If it's not ready for that, the abstraction could live in `speculators.data_generation.transfer` for now — `launch_vllm.py` already has the fallback, and vLLM loads the Mooncake connector by module path, not through the registry.

### Mechanical

- **`transfer.setup()` only runs on the generate path** — it's called inside `_setup_client()`, which only executes when `on_missing="generate"` and the client hasn't been created. Fine for `FileTransfer` (no-op) but Mooncake will need initialization for cached reads too.
- **`HiddenStatesBackend` uses `@abstractmethod` without inheriting `ABC`** — `HiddenStatesTransfer` inherits `ABC` but `HiddenStatesBackend` doesn't, so the decorators on `from_train_args` and `build_kv_transfer_config` are inert.
- **All backends' args registered unconditionally** — the `for backend_cls in registry.values(): add_*_args(parser)` loops will cause argparse conflicts once a second backend adds overlapping flags.

fynnsu:
> Does hs_connectors need to be a subpackage in this repo?

> The motivation is that it needs to be installable in vLLM environments — https://github.com/vllm-project/speculators/pull/605 confirms this, vLLM loads the Mooncake connector at runtime via kv_connector_module_path: "hs_connectors.mooncake_hidden_states_connector". But nesting it inside the speculators repo means you can't pip install hs-connectors independently in a vLLM deployment that doesn't have a speculators checkout. The vendored _InlineFileBackend fallback already acknowledges this fragility. If it's a shared dependency between two projects, it should be its own repo/PyPI package. If it's not ready for that, the abstraction could live in speculators.data_generation.transfer for now — launch_vllm.py already has the fallback, and vLLM loads the Mooncake connector by module path, not through the registry.

We discusses this a bit offline [Internal slack link](https://redhat-internal.slack.com/archives/C093RTWQ5T8/p1781729522341489), and this seems like the best approach. The plan is to publish this folder to pypi so that it will be easy to install, but avoid the overhead of a full new repo. 

> transfer.setup() only runs on the generate path — it's called inside _setup_client(), which only executes when on_missing="generate" and the client hasn't been created. Fine for FileTransfer (no-op) but Mooncake will need initialization for cached reads too.

Yes this is largely intentional. We make the setup lazy, so that it isn't required if we aren't doing generation. The Mooncake connector isn't included here yet, but we might try to get Mooncake to read from disk for it's cached reads too. That way users can still do a mix of offline and online. Although we can re-evaluate when adding Mooncake. I think the alternative is we could call setup_client() on the cached read path too.

> HiddenStatesBackend uses @abstractmethod without inheriting ABC — HiddenStatesTransfer inherits ABC but HiddenStatesBackend doesn't, so the decorators on from_train_args and build_kv_transfer_config are inert.

Yeah, makes sense. I'll fix that.

> All backends' args registered unconditionally — the for backend_cls in registry.values(): add_*_args(parser) loops will cause argparse conflicts once a second backend adds overlapping flags.

Yeah I agree this is an issue but because all backends currently live in the repo, this just means added backends need distinct arg names. This seems okay for now, especially because we're working on refactoring how we handle the args in general.

tests/unit/train/test_data.py
```diff
@@ -491,4 +491,5 @@ def test_arrow_dataset_on_generate_cache_creates_hidden_states_dir(tmp_path: Pat
         on_generate="cache",
     )
 
-    assert arrow_ds.hidden_states_path.is_dir()
+    assert hasattr(arrow_ds.transfer, "hidden_states_path")
+    assert arrow_ds.transfer.hidden_states_path.is_dir()
```
orestis-z, line 507: This assertion will fail. The test's purpose (per its docstring) is to verify that `on_generate="cache"` creates the hidden-states directory eagerly — preventing silent `FileNotFoundError` on `shutil.move`. The old code did `self.hidden_states_path.mkdir(parents=True, exist_ok=True)` in `ArrowDataset.__init__`, but the PR removes that and defers the `mkdir` to [`FileTransfer.cache()`](https://github.com/vllm-project/speculators/blob/6fcef87f6ceb7af482c08e55032542462abfb02b/hs_connectors/src/hs_connectors/transfer.py#L146-L149), which is only called when a sample is actually generated and cached — not during construction.

Since `FileTransfer.__init__` just stores the path without creating it, `arrow_ds.transfer.hidden_states_path.is_dir()` will return `False` here.

The lazy mkdir in `cache()` is arguably better (the directory is guaranteed to exist right before the move), so the fix is probably to update this test to reflect the new contract — e.g. assert that `cache()` creates the directory, or remove the eager-creation assertion.

orestis-z (commented):
The ABC inheritance fix looks good. One test regression below from the deferred-mkdir change.

tests/unit/train/test_data.py
fynnsu: Fixed

orestis-z (commented):
LGTM. Test fix from the last round looks correct — lazy `mkdir` is now properly exercised. Previous feedback (ABC inheritance, architectural questions) has been addressed. Recommend approving.

hs_connectors/src/hs_connectors/transfer.py
```diff
@@ -0,0 +1,200 @@
...
+            type=str,
+            default="/tmp/hidden_states",  # noqa: S108
+            help="The directory to save hidden states to. Default '/tmp/hidden_states'",
+        )
+
+    @staticmethod
+    def from_train_args(
+        args: argparse.Namespace,
+        data_path: str,
+        **_kwargs: Any,
```
rahul-tuli: nit: are these kwargs used anywhere?

scripts/train.py
```diff
@@ -530,14 +531,17 @@ def main(args: argparse.Namespace):  # noqa: C901
     }
     preprocess = preprocess_fns.get(args.speculator_type)
 
+    backend_registry = HiddenStatesBackend.registry or {}
```
rahul-tuli: nit:
```suggestion
    backend_registry = HiddenStatesBackend.registry
```

Since HiddenStatesBackend.registry will never be None

scripts/train.py
```diff
@@ -724,15 +729,19 @@ def parse_args():
             "(token_freq.pt), and hidden states (default: ./output)"
         ),
     )
+    backend_registry = HiddenStatesBackend.registry or {}
```
rahul-tuli:
```suggestion
    backend_registry = HiddenStatesBackend.registry
```

Same as above, HiddenStatesBackend.registry will never be `None` and if its an empty dict then we are just replacing it with another empty dict

src/speculators/train/data.py
```diff
@@ -242,14 +229,6 @@ def __init__(
         request_timeout: float | None = DEFAULT_REQUEST_TIMEOUT,
         max_retries: int = DEFAULT_MAX_RETRIES,
     ):
-        """Initialize the ArrowDataset.
```
rahul-tuli, line 245: nit: an updated docstring would be nice

rahul-tuli (approved):
LGTM! minor nits, nothing major

pyproject.toml
```diff
@@ -258,6 +265,14 @@ select = [
     "INP001", # allow implicit namespace packages in examples
 ]
 
+"hs_connectors/**/*.py" = [
+    "INP001", # separate package
+    "S101", # asserts acceptable
+    "BLE001", # catching Exception acceptable
+    "SIM105", # contextlib.suppress not needed
+    "PTH", # os.path acceptable
```
rahul-tuli, line 270: nit: As of now I don't think this diff violates `S101`/`SIM105`/`BLE001`

speculatorsbot (commented):
LGTM. Agree with @rahul-tuli's open nits (`**_kwargs` unused, `registry or {}` redundant, unnecessary ruff suppressions for `S101`/`SIM105`/`BLE001`). Recommend approving once those are tidied up.

PR #736, opened 2026-07-07
Add D-PACE loss implementation for D-Flash training

weifanjiang:
## Purpose

This PR implements [D-PACE: Dynamic Position-Aware Cross-Entropy for Parallel Speculative Drafting](https://arxiv.org/abs/2605.18810). D-PACE proposes an alternative per-position loss weighting for training block-parallel draft models. Instead of applying a fixed exponential decay over block positions, D-PACE dynamically adjusts the weights as training progresses. This is based on the intuition that although earlier positions are naturally more important (i.e., an early rejection forces all subsequent draft positions to be discarded), when they are sufficiently stabilized, the training priority should shift to later positions to improve overall acceptance rate.

Two arguments are added to `train.py` to support D-PACE loss:

- `--per-position-loss-weight ["fixed-exp-decay", "dpace"]` (default `fixed-exp-decay`).  This flag enables D-PACE loss; while missing, it defaults to the original fixed exponential decay.
- `--dpace-alpha` (default `0.5`). This flag is only effective when D-PACE is enabled. It smoothes per-position confidence according to equation 7 of the paper. The paper uses 0.5 as the default value for their evaluation and justifies this choice in sensitivity analysis.

Note that D-PACE requires cross-entropy as the training loss (`--loss-fn ce`), allowing per-position confidence score computation to be piggybacked on the loss computation. Applying D-PACE to other losses is not supported and will result in `ValueError`.

D-PACE is currently enabled for D-Flash and D-Spark draft models. Theoretically, it is compatible with all block-parallel drafters.

## Tests

**D-Flash + D-PACE evaluation.** Multiple D-Flash checkpoints are trained with different loss function configurations:

- `--loss-fn kld`: KL Divergence (current default)
- `--loss-fn ce`: Cross Entropy with fixed exponential decay
- `--per-position-loss-weight dpace --dpace-alpha ALPHA`: D-PACE with `ALPHA` = 0.3, 0.5, 0.7.

The training set is 100K random samples from Magpie+Ultrachat. The verifier model is Qwen3-8B. Each checkpoint uses 1 A100 for hidden state generation and 7 A100s for training. Training configurations are identical for all checkpoints:

```
--target-layer-ids 1 9 17 25 34  --num-layers 5 --draft-vocab-size 32000 --max-anchors 3072 --block-size 8 --sliding-window 2048 --sliding-window-indices 0 1 2 3 4 --scheduler-type cosine --lr 0.0006 --epochs 1 --noise-std 0.0
```

Metric: since D-PACE is a pure training-time optimization and does not change draft model's forward pass complexity, its gains are captured entirely by acceptance rates.

Evaluation results: Mean-Accepted-Length (MAL) on [Speculator Benchmarks](https://huggingface.co/datasets/RedHatAI/speculator_benchmarks):

| Dataset | KL Divergence | CE | D-PACE $\alpha=0.3$ | D-PACE $\alpha=0.5$| D-PACE $\alpha=0.7$ |
|---|---:|---:|---:|---:|---:|
| HumanEval | 2.977 | 2.956 | 3.047 | **3.169** | 3.139 |
| math_reasoning | 3.275 | 3.322 | **3.541** | 3.532 | 3.531 |
| qa | 2.279 | 2.343 | 2.455 | **2.495** | 2.431 |
| question | 2.585 | 2.567 | 2.685 | 2.680 | **2.708** |
| rag | 2.302 | 2.269 | **2.393** | 2.372 | 2.336 |
| summarization | 1.906 | 1.918 | 1.981 | 1.980 | **1.991** |
| tool_call | 2.372 | 2.327 | **2.480** | 2.454 | 2.467 |
| translation | 2.096 | 2.072 | 2.176 | 2.183 | **2.191** |
| writing | 2.565 | 2.586 | 2.683 | 2.684 | **2.718** |
| Mean | 2.484 | 2.485 | 2.605 | **2.617** | 2.612 |

**D-Spark + D-PACE evaluation.** D-Spark checkpoints use Markov rank 256 and vanilla Markov head type. The confidence head is disabled. Remaining training configurations match above D-Flash experiments.

- `--loss-fn ce`: Vanilla fixed-decay cross-entropy
- `--position-loss-weight dpace --dpace-alpha 0.5`: D-PACE with $\alpha=0.5$

MALs:

| Dataset | CE | D-PACE $\alpha=0.5$ |
|---|---:|---:|
| HumanEval | 3.145 | **3.302** |
| math_reasoning | 3.457 | **3.615** |
| qa | 2.352 | **2.492** |
| question | 2.671 | **2.770** |
| rag | 2.276 | **2.384** |
| summarization | 1.916 | **1.996** |
| tool_call | 2.451 | **2.538** |
| translation | 2.082 | **2.202** |
| writing | 2.686 | **2.776** |
| Mean | 2.560 | **2.675** |

D-PACE improves upon fixed-decay loss for both D-Flash and D-Spark drafters.

## Checklist

I have filled in:

### DIFF
--- a/src/speculators/models/dflash/core.py
+++ b/src/speculators/models/dflash/core.py
@@ -105,6 +105,8 @@ class DFlashDraftModel(DraftVocabMixin, SpeculatorModel):
         )
         self.verifier_norm.weight.requires_grad = False
         self.block_size = config.block_size
+        self.per_position_loss_weight = config.per_position_loss_weight
+        self.dpace_alpha = config.dpace_alpha
         self.post_init()
 
     @property
@@ -181,6 +183,8 @@ class DFlashDraftModel(DraftVocabMixin, SpeculatorModel):
             "aux_hidden_state_layer_ids": target_layer_ids,
             "mask_token_id": kwargs.get("mask_token_id"),
             "sliding_window_non_causal": kwargs.get("sliding_window_non_causal", False),
+            "per_position_loss_weight": kwargs.get("per_position_loss_weight", "fixed-exp-decay"),
+            "dpace_alpha": kwargs.get("dpace_alpha", 0.5),
             "speculators_config": SpeculatorsConfig(
                 algorithm=algorithm,
                 proposal_methods=[
@@ -207,10 +211,14 @@ class DFlashDraftModel(DraftVocabMixin, SpeculatorModel):
         loss_config = resolve_loss_config(kwargs["loss_fn"])
         gamma = kwargs.get("dflash_decay_gamma", 4.0)
         max_anchors = kwargs.get("max_anchors", 3072)
+        per_position_loss_weight = kwargs.get("per_position_loss_weight", "fixed-exp-decay")
+        dpace_alpha = kwargs.get("dpace_alpha", 0.5)
         shared = {
             "loss_config": loss_config,
             "gamma": gamma,
             "max_anchors": max_anchors,
+            "per_position_loss_weight": per_position_loss_weight,
+            "dpace_alpha": dpace_alpha,
         }
         return dict(shared), dict(shared)
 
@@ -412,6 +420,8 @@ class DFlashDraftModel(DraftVocabMixin, SpeculatorModel):
             self.block_size,
             gamma=gamma,
             loss_config=loss_config,
+            per_position_loss_weight=self.per_position_loss_weight,
+            dpace_alpha=self.dpace_alpha,
         )
         draft_tokens = torch.argmax(logits, dim=-1)
 
--- a/src/speculators/models/dflash/core.py
+++ b/src/speculators/models/dflash/core.py
@@ -105,8 +105,6 @@ class DFlashDraftModel(DraftVocabMixin, SpeculatorModel):
         )
         self.verifier_norm.weight.requires_grad = False
         self.block_size = config.block_size
-        self.per_position_loss_weight = config.per_position_loss_weight
-        self.dpace_alpha = config.dpace_alpha
         self.post_init()
 
     @property
@@ -183,8 +181,6 @@ class DFlashDraftModel(DraftVocabMixin, SpeculatorModel):
             "aux_hidden_state_layer_ids": target_layer_ids,
             "mask_token_id": kwargs.get("mask_token_id"),
             "sliding_window_non_causal": kwargs.get("sliding_window_non_causal", False),
-            "per_position_loss_weight": kwargs.get("per_position_loss_weight", "fixed-exp-decay"),
-            "dpace_alpha": kwargs.get("dpace_alpha", 0.5),
             "speculators_config": SpeculatorsConfig(
                 algorithm=algorithm,
                 proposal_methods=[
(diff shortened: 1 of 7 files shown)


src/speculators/models/dflash/core.py
```diff
@@ -412,6 +420,8 @@ def forward(
             self.block_size,
             gamma=gamma,
             loss_config=loss_config,
+            per_position_loss_weight=per_position_loss_weight,
```
fynnsu, line 423: Can we update both dflash and dspark to support this?

src/speculators/models/dflash/metrics.py
```diff
@@ -46,13 +52,28 @@ def compute_metrics(
...
+
     loss, term_losses = compound_loss(
         logits,
         targets,
         loss_mask,
         pos_idx,
         loss_config=loss_config,
-        decay_fn=partial(dflash_loss_decay, gamma=gamma),
+        decay_fn=decay_fn,
+        dpace_precomputed_ce=elementwise_ce,
```
fynnsu: This completely circumvents the point of having single generalized `compound_loss` function if we start passing precomputed loss values into the function.

src/speculators/models/dflash/metrics.py
```diff
@@ -46,13 +52,28 @@ def compute_metrics(
...
+    if per_position_loss_weight == "dpace":
+        elementwise_ce = ce_loss(logits, targets)
+        decay_mult = dpace_loss_weight(
+            elementwise_ce, loss_mask, block_size, alpha=dpace_alpha
+        )
+
+        def decay_fn(
+            _pos_idx: torch.Tensor, _weight: torch.Tensor = decay_mult
+        ) -> torch.Tensor:
+            return _weight
```
fynnsu: If we need elementwise loss as an input to dpace weighting then we may need to update the `decay_fn` signature.

src/speculators/models/metrics.py
```diff
@@ -361,6 +408,7 @@ def compound_loss(
     pos_idx: torch.Tensor,
     loss_config: LossConfig,
     decay_fn: Callable[[torch.Tensor], torch.Tensor] | None = None,
+    dpace_precomputed_ce: torch.Tensor | None = None,
```
fynnsu: Remove

src/speculators/models/metrics.py
```diff
@@ -406,7 +461,10 @@ def loss_function(
     Returns:
         Scalar mean loss across the batch.
     """
-    elementwise_loss = loss_fn(logits, targets)  # shape: [1, seq_len]
+    if dpace_precomputed_ce is not None:
+        elementwise_loss = dpace_precomputed_ce
+    else:
+        elementwise_loss = loss_fn(logits, targets)  # shape: [1, seq_len]
```
fynnsu: We can just have the user input cross entropy as the loss function here, and then `elementwise_loss = loss_fn(...)` will just compute ce. 

If dpace only works with ce loss, then we should have a check for that when we parse the args.

src/speculators/models/metrics.py
fynnsu: If this usage is too restrictive to support dpace, then we can change the signature of `decay_fn` (for example we can have it take in `elementwise_loss` as an input.

fynnsu (changes requested):
Hi @weifanjiang thanks for the pr! I think we need to change how we're piping some things through speculators so that it fits within the existing abstractions.

fynnsu (approved):

PR #737, opened 2026-07-07
feat(convert): add DSpark checkpoint converter

kashif:
## What

Adds a converter for DSpark draft checkpoints so a trained DSpark draft can be turned into the speculators format and loaded with `DSparkDraftModel.from_pretrained`.

The DSpark speculator model already landed (#677), but there was no way to convert an external checkpoint into it. This fills that gap, mirroring the DFlash converter.

## How

DSpark is DFlash plus a Markov head and a confidence head, and the parameter names already match `DSparkDraftModel`, so the weights are copied as-is with no renaming. The converter's real work is config: it reads DSpark's flat top-level fields (`block_size`, `mask_token_id`, `target_layer_ids`, `markov_rank`, `markov_head_type`, and the two confidence flags), applies the `+1` hidden-state offset (with `-1` mapping to the embedding output), and builds a `DSparkSpeculatorConfig` with `algorithm="dspark"`. `embed_tokens` and `lm_head` are filled from the verifier at save time, same as DFlash.

It also registers `"dspark"` in `convert_model` and in the auto-detection path.

## Tests

Added `tests/unit/convert/test_dspark_converter.py` (config building, the layer-id offset including the embedding case, error paths, and the missing-weight guard). All 59 tests in `tests/unit/convert/` pass locally.

Not covered here: an end to end vLLM smoke test, which needs a GPU and a real checkpoint. Happy to add that as a follow up.


orestis-z (commented):
LGTM. Agree with @coderabbitai's suggestion to add tests for auto-detection and `_save` happy path. The codex bot's `block_size - 1` concern is a false positive. Recommend approving once test gaps are addressed.

speculatorsbot (commented):
LGTM. Clean implementation that mirrors the DFlash converter with appropriate DSpark-specific additions (Markov head, confidence head, flat config layout). Recommend approving once the outstanding test gaps noted by @coderabbitai are addressed.

src/speculators/convert/dspark/converter.py
```diff
@@ -0,0 +1,214 @@
...
+
+        mask_token_id = source_config.get("mask_token_id")
+        if mask_token_id is None:
+            raise ValueError("Checkpoint config has no top-level `mask_token_id`")
+
+        speculators_config = SpeculatorsConfig(
+            algorithm="dspark",
+            proposal_methods=[
+                GreedyTokenProposalConfig(
+                    speculative_tokens=block_size - 1,
```
speculatorsbot, line 145: After the merge, `DSparkSpeculatorConfig.sample_from_anchor` defaults to `True`, which means speculative tokens = `block_size` (not `block_size - 1`). See [`dflash/core.py` L205](https://github.com/vllm-project/speculators/blob/d5b2812b576c78c51bf281cd8f6e8b800984d51d/src/speculators/models/dflash/core.py#L205):

```python
speculative_tokens = block_size if sample_from_anchor else block_size - 1
```

The DFlash converter is fine because DFlash defaults to `sample_from_anchor=False`, but DSpark defaults to `True`. This line should read:

```python
speculative_tokens=block_size,
```

The test assertion on line 84 of `test_dspark_converter.py` (`== 15` for `block_size == 16`) will also need updating to `== 16`.

If there are external DSpark checkpoints that use the non-anchor-sampling convention, consider reading `sample_from_anchor` from the source config with a fallback to `True`.

speculatorsbot (commented):
The merge of main introduced `sample_from_anchor` (default `True` for DSpark, `False` for DFlash). The converter now needs to account for this — see inline comment.

PR #741, opened 2026-07-08
Add EAL metric for speculator training

eldarkurtic:
Motivation: 
- loss independent metric (so we can ablate on various loss functions and have apples-to-apples comparison) which also incorporates accuracy across all positions (i.e. a proxy for the true vllm-based AL metric)
- confirmed that it correlates well (better than loss) with ranking from benchmarking on speculator-benchmarks (humaneval, math, qa, question, etc)

EAL (Expected Accepted Length) estimates how many draft tokens you expect the target model to accept per speculation block, on average.

The probability that the token at position k survives is the product of the per-position accuracies up to and including k:
```
  P(accept position k) = acc_1 · acc_2 · … · acc_k   =   ∏_{i≤k} acc_i
```
The expected number of accepted tokens is the sum of those survival probabilities over all drafted positions:
```
  EAL = Σ_k ∏_{i≤k} acc_i        (k = 1 … block_size-1)
```

Note position 0 (the anchor) is intentionally excluded — EAL only counts the drafted tokens.

### DIFF
--- a/src/speculators/models/dflash/metrics.py
+++ b/src/speculators/models/dflash/metrics.py
@@ -74,12 +74,7 @@ def compute_metrics(
     metrics["full_acc_sum"] = correct_per_pos[1:].sum()
     metrics["full_acc_total"] = total_per_pos[1:].sum()
 
-    # Also accumulate Expected Accepted Length: EAL = sum_k prod_{i<=k} acc_i over
-    # drafted positions 1..block_size-1 (a token is accepted only if all earlier
-    # drafted tokens were correct). It is nonlinear in the per-position accuracies,
-    # so it is computed from this step's correct/total ratios and logged via the
-    # generic _sum/_total pair (eal_total = 1), giving a per-step average in training
-    # and "eal_epoch" in validation.
+    # EAL = sum_k prod_{i<=k} acc_i over drafted positions
     eal = torch.zeros((), device=logits.device)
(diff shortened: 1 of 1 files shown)


src/speculators/models/dflash/metrics.py
```diff
@@ -73,7 +74,20 @@ def compute_metrics(
     metrics["full_acc_sum"] = correct_per_pos[1:].sum()
     metrics["full_acc_total"] = total_per_pos[1:].sum()
 
+    # Also accumulate Expected Accepted Length: EAL = sum_k prod_{i<=k} acc_i over
+    # drafted positions 1..block_size-1 (a token is accepted only if all earlier
+    # drafted tokens were correct). It is nonlinear in the per-position accuracies,
+    # so it is computed from this step's correct/total ratios and logged via the
+    # generic _sum/_total pair (eal_total = 1), giving a per-step average in training
+    # and "eal_epoch" in validation.
```
orestis-z: minor nit (non-blocking): the 6-line comment is a bit verbose for what the code already makes clear

orestis-z (approved):

src/speculators/models/dflash/metrics.py
eldarkurtic: good point, it is "vibe-commented". will shorten it manually

rahul-tuli (approved):

PR #743, opened 2026-07-08
Fix on_generate="cache" silently failing on a fresh data directory

jessiewei7:
## Bug

`ArrowDataset` never creates `hidden_states_path`. On a data directory that has never held a hidden-states cache (the normal first-run case for `--on-generate cache`), every `shutil.move` into it raises `FileNotFoundError`, which `_maybe_generate_hs` catches and downgrades to a warning — so caching silently fails for **every** sample and the run still exits 0. The next epoch/run regenerates everything from scratch, defeating the flag's purpose.

Reproduced live: fresh `--data-path`, 128/128 samples warn `Failed to load/cache hidden states ... No such file or directory`; after a manual `mkdir -p`, the same command caches 128/128. The path was never exercised by tests — e2e online training hardcodes `--on-generate delete`.

## Fix

Create the directory once at dataset construction when `on_generate="cache"`. Regression test included (fails without the fix).

### DIFF
--- a/src/speculators/train/data.py
+++ b/src/speculators/train/data.py
@@ -273,6 +273,8 @@ class ArrowDataset(BaseDataset):
         self.vllm_endpoint = vllm_endpoint
         self.on_missing = on_missing
         self.on_generate = on_generate
+        if self.on_generate == "cache":
+            self.hidden_states_path.mkdir(parents=True, exist_ok=True)
         self.client: openai.OpenAI | None = None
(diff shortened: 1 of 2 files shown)


fynnsu (approved):

fynnsu:
Looks good but please fix DCO @jessiewei7

jessiewei7:
Fixed @fynnsu

PR #749, opened 2026-07-08
feat(train): default sliding window attention for dflash/dspark

rahul-tuli:
## Purpose

Invert the sliding window attention API for `dflash`/`dspark` draft training so
that **sliding window is the default on every draft layer**.

- **Before:** all draft layers used full attention unless you enumerated every
  layer index via `--sliding-window-indices 0 1 2 3 ...`.
- **After:** all draft layers use sliding window by default. Use
  `--full-attention-indices 0 2` to opt specific layers back into full attention.

**Motivation:** our internal DFlash experiments show sliding window attention is
almost always at least as good as full attention — comparable or higher
per-position acceptance / average accepted length — while using substantially
less KV cache on long-context sequences. Full attention on every draft layer is
therefore the wrong default: it costs more memory for no consistent quality gain,
and the old API forced users to enumerate every layer index to get the *better*
configuration. This PR flips that so the recommended setup is the zero-config
default.

No model code changes are needed: `dflash`/`dspark` derive their per-layer
attention from `transformer_layer_config.layer_types`, which this PR still builds
correctly, and vLLM reads the same serialized field at inference time.

<details>
<summary>What changed</summary>

- **Removed** `--sliding-window-indices` (opt in to sliding window per layer).
- **Added** `--full-attention-indices` (opt specific layers out of sliding window).
- **Kept** `--sliding-window` (window size, default 2048) and
  `--sliding-window-non-causal`.
- Inverted `layer_types` construction in `create_transformer_layer_config`;
  `use_sliding_window` is now derived from the resulting `layer_types` (single
  source of truth, robust to duplicate indices).
- `build_draft_model` routes non-sliding speculator types (`eagle3`, `peagle`) to
  all-full-attention, preserving their prior behavior exactly. `dflash`/`dspark`
  forward `--full-attention-indices` (empty by default → all sliding window).
- The `dflash`/`dspark`-only guard for `--full-attention-indices` is a plain
  `ValueError` in `main()`, next to where the draft is synthesized.
- Docs/help polish: fixed the malformed `--sliding-window-non-causal` help string,
  corrected the stale `--speculator-type` options list, and grouped the
  sliding-window flags under their own docs section.
</details>

<details>
<summary>Breaking change &amp; migration</summary>

`--sliding-window-indices` is removed. Existing scripts that pass it will fail
with a clear argparse error. Migration:

- To get **all** layers on sliding window (the new default): just drop
  `--sliding-window-indices`.
- To keep **specific** layers on full attention: use `--full-attention-indices`
  with the complement of your old sliding-window set.

`eagle3`/`peagle` behavior is unchanged — they remain full-attention-only, and
passing `--full-attention-indices` with a non-`dflash`/`dspark` type raises a
clear error.
</details>

## Tests

```
pytest tests/unit/train/test_draft_config_init.py -q
```

```
48 passed
```

<details>
<summary>Coverage added</summary>

New/updated tests are parametrized:

- `test_full_attention_indices_parsing` — default parses to `[]` (all sliding
  window); explicit values parse to `[0, 2]`.
- `test_create_layer_config_layer_types` — empty → all `sliding_attention`
  (`use_sliding_window=True`); mixed → correct per-layer `layer_types` with the
  flag still `True`; all-full → `use_sliding_window=False`.
- `test_create_layer_config_rejects_out_of_range_indices` — `[-1]`, `[3]`,
  `[0, 3]` all raise.
- `test_build_draft_model_routing` — `dflash`/`dspark` default to all-sliding
  (`[]`) and forward an explicit non-empty list unchanged; other types put every
  layer on full attention (`[0, 1, 2]`).
- Existing `--from-pretrained` / `--draft-config` mutual-exclusion guards still
  hold with the renamed flag.
</details>

## Checklist

I have filled in:


orestis-z (approved):
Clean!

PR #750, opened 2026-07-08
[Regen][tool call] Support tool calls in on-policy response regeneration

WindChimeRan:
## Purpose

Before this PR speculator doesn't support tool use regen. This PR fixes it. 

Design:

* **Semi-on-policy** tool call regen: the target regenerates the **tool-call tokens on-policy**, but tools are **not executed** — the *i*-th cached tool result from the source data is spliced positionally after the target's *i*-th regenerated call.
* Fail loud on malform input tool. 

> Q: Why semi-on-policy (and not full agentic replay)? 
> A: Full agentic replay requires sandbox env, which is expensive and extremely slow. 

**SpecForge drops tool call same as our main. This PR will be a differentiator.**

## Requirements

- **Server:** must run with `--enable-auto-tool-choice --tool-call-parser <hermes|llama3_json|...>`,   otherwise tool calls come back as raw text in `content` rather than structured   `tool_calls`.

## Limitation

* Parallel tool call is out of scope. Currently truncate to the first call. 
* need to cleanup name normalization after  #688 landed

## Tests

Tool-free datasets are byte-for-byte unchanged: `extract_tools` returns `None`, `extract_tool_results` returns `[]`, and no `tools` field is sent.

## Checklist

I have filled in:

### DIFF
--- a/scripts/response_regeneration/script.py
+++ b/scripts/response_regeneration/script.py
@@ -570,6 +570,8 @@ async def regenerate_conversation(
             # To continue past a tool call we need exactly one call and a cached
             # result to pair with it; otherwise keep this (committed) call row
             # and truncate.
+            # TODO(regen): support parallel/multi tool calls -- a turn with >1
+            # call truncates here (dropped ~16/50 rows in a Hermes validation run).
             if len(tool_calls) != 1 or not tool_results:
                 truncated = True
                 break
--- a/scripts/response_regeneration/script.py
+++ b/scripts/response_regeneration/script.py
@@ -6,6 +6,7 @@ import json
 import os
 import re
 import sys
+from collections import deque
 from typing import Any
 
 import aiohttp
@@ -33,6 +34,11 @@ def _dataset_choice(name: str) -> str:
             "images yet. Use it off-policy with `prepare-data`."
         )
     return name
+
+
+# ---------------------------------------------------------------------------
+# CLI & run configuration
+# ---------------------------------------------------------------------------
 
 
 def parse_args():
@@ -136,6 +142,19 @@ def sanitize_filename(name: str) -> str:
     return name.strip("._")
 
 
+# ---------------------------------------------------------------------------
+# Row ingestion: user/system turns, tool schema, cached tool results
+# ---------------------------------------------------------------------------
+
+
+def _conversation_messages(row: dict[str, Any]) -> list:
+    """The ``messages`` or ``conversations`` list from a row, else []."""
+    convs = row.get("messages")
+    if not (isinstance(convs, list) and convs):
+        convs = row.get("conversations")
+    return convs if isinstance(convs, list) else []
+
+
 def extract_turns(
     row: dict[str, Any], prompt_field: str | None
 ) -> list[dict[str, Any]]:
@@ -147,28 +166,23 @@ def extract_turns(
     Rows without a usable conversation fall back to a single user turn taken
     from ``prompt_field``.
     """
-    convs = row.get("messages")
-    if not (isinstance(convs, list) and convs):
-        convs = row.get("conversations")
+    turns = []
+    for m in _conversation_messages(row):
+        if not isinstance(m, dict):
+            continue
+        role = m.get("role") or m.get("from")
+        content = m.get("content") or m.get("value")
+        if not content:
+            continue
+        if role == "system":
+            turns.append({"role": "system", "content": content})
+        elif role in ("user", "human"):
+            turns.append({"role": "user", "content": content})
+        # original assistant/gpt turns are dropped and regenerated
+    if any(turn["role"] == "user" for turn in turns):
+        return turns
 
-    if isinstance(convs, list) and convs:
-        turns = []
-        for m in convs:
-            if not isinstance(m, dict):
-                continue
-            role = m.get("role") or m.get("from")
-            content = m.get("content") or m.get("value")
-            if not content:
-                continue
-            if role == "system":
-                turns.append({"role": "system", "content": content})
-            elif role in ("user", "human"):
-                turns.append({"role": "user", "content": content})
-            # original assistant/gpt turns are dropped and regenerated
-        if any(turn["role"] == "user" for turn in turns):
-            return turns
-        # no usable user turn: fall through to the prompt_field fallback
-
+    # no usable user turn: fall back to the prompt_field
     prompt = row.get(prompt_field)
     if prompt:
         return [{"role": "user", "content": prompt}]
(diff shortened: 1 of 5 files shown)


WindChimeRan:
@coderabbitai  fixed.

orestis-z (commented):
LGTM. The semi-on-policy design (regenerate tool-call tokens, positionally splice cached results) is clean and well-justified. The refactoring from inline worker code into `regenerate_conversation` and `_sample_from_response` improves testability without disrupting the tool-free path. Tests are comprehensive — extraction, normalization, validation guards, the splice/truncate loop, and the unchanged plain path are all covered. Agree with @coderabbitai's earlier suggestion about graceful `extract_tools` error handling in the main loop, and it's been addressed. Recommend approving.

scripts/response_regeneration/script.py
```diff
@@ -531,19 +786,48 @@ async def main():
...
+                        "Skipping row %s: input tool schema is broken (%s)",
+                        primary_id,
+                        exc,
+                    )
+                    error_output = {
+                        "id": primary_id,
+                        "metadata": {
+                            "idx": index,
+                            "error": repr(exc),
+                            "turns_completed": 0,
```
orestis-z: **Doc/code field-name mismatch.** The [error output format in the docs](https://github.com/vllm-project/speculators/blob/46f2f6a25efb3572e7887ab030407c41fd373964/docs/cli/response_regeneration.md#L171) was renamed from `turns_completed` to `generations_completed` (consistent with the turn-to-generation rename throughout the PR), but the code still emits `turns_completed` here and in the worker error handler (line 680). The test at line 535 also asserts on `turns_completed`.

Since the PR renames the semantic unit from "turn" to "generation" everywhere else (row IDs `_turn<N>` -> `_gen<N>`, docstrings, comments), this field should probably follow suit:

```suggestion
                            "generations_completed": 0,
```

(and the same rename at line 680 + the test assertion at line 535).

orestis-z (commented):
LGTM, one minor non-blocking note below. The two new commits since my last review (graceful error handling for broken tool schemas, and extracting tools from the normalized row instead of the raw row) both look correct and address real issues. Recommend approving.

scripts/response_regeneration/script.py
```diff
@@ -147,45 +169,144 @@ def extract_turns(
...
+        role = m.get("role") or m.get("from")
+        if role not in _TOOL_RESULT_ROLES:
+            continue
+        content = m.get("content")
+        if content is None:
+            content = m.get("value")
+        if content is not None:
+            results.append(content)
+    return results
+
```
Roderick-Wu: `extract_turns()` and this function both produce lists of outputs but don't map parts to each other. Later in `regenerate_conversation()` the tool_results are put in a deque:

```python
    turns = item["turns"]
    tools = item.get("tools")
    tool_results = deque(item.get("tool_results") or [])
    conv_id = item["primary_id"]
```

And during the tool-call loop we extract the next tool-call if we identify a toolcall in the assistant message:
```python
            prefix.append(_tool_result_message(tool_calls[0], tool_results.popleft()))
```

I'm not completely sure that this is problematic but since the two lists are decoupled, if the model doesn't call tools in the same turn-by-turn pattern, the tool call position has potential to be incorrect? I think it might be better to modify the `extract_turns` directly instead of building on top. But would be good for someone more experienced to take a look in case I missed something ehre. Otherwise this pr lgtm

WindChimeRan: update: 
* merged to single `extract_conversation `
* Only reuse a result if the call is for the same tool. 

```python
content, result_names = tool_results.popleft()
call_name = (tool_calls[0].get("function") or {}).get("name")
# Splice only if the regenerated call is for the tool this cached
# result answers; a different tool cannot be paired coherently, so
# keep the committed call row and truncate.
if result_names and call_name not in result_names:
    truncated = True
    break

prefix.append(_tool_result_message(tool_calls[0], content))
```

scripts/response_regeneration/script.py
```diff
@@ -136,56 +145,157 @@ def sanitize_filename(name: str) -> str:
...
     if prompt:
-        return [{"role": "user", "content": prompt}]
-    return []
+        return [{"role": "user", "content": prompt}], []
+    return [], []
 
 
-def prepare_row(row: dict[str, Any], config: DatasetConfig) -> list[dict[str, Any]]:
-    """Extract regeneration turns from a raw dataset row, ``[]`` to skip it.
+def normalize_row(row: dict[str, Any], config: DatasetConfig) -> dict[str, Any] | None:
```
shanjiaz: nit: This function is only called by `prepare_row`, should probably be moved inline there.

scripts/response_regeneration/script.py
```diff
@@ -136,56 +145,157 @@ def sanitize_filename(name: str) -> str:
...
+
+def extract_tools(row) -> list | None:
+    """Return the OpenAI-style ``tools`` schema for a row, or ``None``.
+
+    Reads the ``tools`` column -- a list, or a JSON-string encoding one, as the
+    Hermes function-calling dataset stores it. A row that declares a ``tools``
+    field we cannot read as a list raises ``ValueError`` rather than silently
+    regenerating tool-free; a tool-free row returns ``None``.
+    """
+    tools = _list_field(row, "tools")
```
shanjiaz, line 253: It's kinda awful this field could be either a string or a list in hf.

scripts/response_regeneration/script.py
```diff
@@ -319,6 +435,182 @@ def build_boundary_sample(
...
+            samples.append(sample)
+            prefix.append(assistant_msg)
+
+            if not tool_calls:
+                break  # final answer: this user turn is done
+
+            # To continue past a tool call we need exactly one call and a cached
+            # result to pair with it; otherwise keep this (committed) call row
+            # and truncate.
+            if len(tool_calls) != 1 or not tool_results:
```
shanjiaz, line 575: I guess multi-tool call is tricky, since we only have a single response saved in the queue and having multiple tool calls would cause a turn mismatch. Just want to point out when I was running a validation, 16/50 samples are dropped because of this. Maybe let's add a TODO here to revisit?

shanjiaz:
Ran tool call regen and here's an example output [tool_call_regen_output.json](https://github.com/user-attachments/files/30064567/tool_call_regen_output.json). Verified the regenerated data could be directly consumed by our training pipeline. Great job!

shanjiaz (approved):

scripts/response_regeneration/script.py
WindChimeRan: I was trying to test on many tool use datasets, and this could be somewhat complicated. Hard to imagine without concrete datapoints. 

Suggest to revisit it when we have more tool call datasets

shanjiaz (approved):

PR #752, opened 2026-07-09
[Long Context] TV Loss triton fusion

WindChimeRan:
## Purpose

Long-context draft training is bottlenecked by the **loss region**. The eager acceptance-rate losses (`tv`, `nla`) materialize *both* full softmax distributions to GPU memory:

```python
# src/speculators/models/metrics.py — eager tv_loss
def tv_loss(logits, targets):              # both [1, T, V_draft]
    draft_p  = F.softmax(logits,  dim=-1)  # [T, V] written to HBM
    target_p = F.softmax(targets, dim=-1)  # [T, V] written to HBM
    overlap  = torch.minimum(draft_p, target_p).sum(dim=-1)   # α, [1, T]
    return 1.0 - overlap                                      # 1 − d_TV
```

`draft_p` and `target_p` are **saved for the backward** and kept alive across all `S` TTT steps, so the loss-region peak is `~S · T · V_draft · bytes` and grows *unboundedly* with context length while everything else in the step (draft weights, optimizer, verifier) is fixed. Since #788 the softmax runs in fp32 — 4 bytes/element.

This PR routes `tv`/`nla` through a **fused online-softmax Triton kernel** that computes the same overlap **without ever materializing `draft_p`/`target_p`**: it streams the softmax in registers, saves only five per-row scalars, and recomputes `p, q` in the backward. Peak loss-region memory drops ~40–60% and, unlike the eager path, no longer OOMs at long context (measurements below).

**Why a kernel and not `torch.compile`:** the two softmaxes are **autograd-saved outputs**, not fusable temporaries — `torch.compile` still writes them to HBM to feed the backward. Eliminating them requires the online-softmax reformulation (forward stats → backward recompute), which has to be a hand-written kernel. `tv`/`nla` route straight to the kernels (Triton required — no eager fallback); the eager losses remain only as the numeric reference the kernels are tested against.

Derived from SpecForge's fused loss (Apache-2.0, Liger lineage) — see [SpecForge#185](https://github.com/sgl-project/SpecForge/pull/185). This port keeps a separate gradient buffer.

## Tests

Verified on 8×A100-80GB (kernel imported standalone). Reference = the repo's fp32 eager `tv_loss` (current `main`).

**Correctness** — forward matches the eager loss to **1.2e-7** (bit-exact) and the gradient to cosine **0.99999988**; also verified at DeepSeek-V3 vocab (`V=129280`). `test_fused_tv_matches_eager_on_cuda` now runs on bf16 inputs and asserts the loss stays fp32.

**Peak memory** — loss region, `torch.cuda.max_memory_allocated`, `V=32000`, `S=3`, bf16 logits, joint backward (GiB):

| Context `T` | eager fp32 (`main`) | fused (this PR) |
|---|---|---|
| 4K   | 5.13  | **1.95**  |
| 8K   | 10.25 | **3.91**  |
| 16K  | 20.51 | **7.81**  |
| 32K  | 41.02 | **15.63** |
| 64K  | **OOM** | **31.25** |
| 128K | **OOM** | **62.51** |

The fused peak is a constant **0.38×** of eager fp32 across context length → the loss region is exactly `S · T · V · bytes`, so the saving scales without bound and is *larger* at DeepSeek vocab (**25.6 GiB** saved at `V=129280, T=8K`). Eager OOMs at 64K on an 80 GB card; the fused kernel trains **128K in 62.5 GiB**.

**Scope:** isolated loss-region measurement — it lowers the end-to-end training peak where the loss region is the peak (long context, this PR's target; at ≤8K the backbone/optimizer dominate). Gradients are unchanged.

## Follow-up

- **Shared kernel for the divergence-loss family** — this PR fuses `tv`/`nla` only; the same online-softmax fusion applies unchanged to `kl_div`/`rkl`/`jsd`/`lk_hybrid`, so one generic kernel would cover them without per-loss duplication.
- **In-place gradient write** — that shared kernel then unlocks writing the gradient into the logits buffer instead of a separate one ([SpecForge#185](https://github.com/sgl-project/SpecForge/pull/185)), a further `[T, V]`/step saving.

## Reproduce

The 128K row above as a runnable check — single 80 GB A100, `V_draft=32000`, `S=3`, bf16:

```python
import torch
from speculators.models.fused_tv_loss import fused_tv_loss   # ← swap to metrics.tv_loss for eager

T, V, S = 131072, 32000, 3          # 128K context, 32K draft vocab, 3 TTT steps
dev = torch.device("cuda", 0)
torch.cuda.reset_peak_memory_stats()
total = None
for _ in range(S):                  # steps retained for one joint backward
    logits  = torch.randn(1, T, V, device=dev, dtype=torch.bfloat16, requires_grad=True)
    targets = torch.randn(1, T, V, device=dev, dtype=torch.bfloat16)
    step = fused_tv_loss(logits, targets).float().mean()
    total = step if total is None else total + step
total.backward()
print(f"peak {torch.cuda.max_memory_allocated() / 1024**3:.1f} GiB")
```

```
fused_tv_loss   →  peak 62.5 GiB
metrics.tv_loss →  torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 15.62 GiB
```

Same input, same 80 GB card: the fused kernel fits; the eager loss OOMs on the first fp32 `[T, V]` softmax buffer it materializes (15.62 GiB).

## Checklist

I have filled in:

### DIFF
--- a/src/speculators/models/metrics.py
+++ b/src/speculators/models/metrics.py
@@ -1,6 +1,7 @@
 import json
 import math
 from collections.abc import Callable
+from functools import cache
 
 import torch
 
@@ -386,38 +387,32 @@ def dpace_loss_decay(
 # Ascend NPU where mainline Triton has no backend. ``logits.is_cuda`` gates
 # CUDA/ROCm; the import is lazy so this module imports without Triton installed.
 
-_FUSED_KERNELS = None  # None = not yet tried; False = unavailable; else (tv, nla)
 
-
-def _fused_kernels():
-    """Lazily import and cache the fused TV/NLA kernels; ``None`` if unavailable."""
-    global _FUSED_KERNELS  # noqa: PLW0603
-    if _FUSED_KERNELS is None:
-        try:
-            from speculators.models.fused_tv_loss import (  # noqa: PLC0415
-                fused_nla_loss,
-                fused_tv_loss,
-            )
-
-            _FUSED_KERNELS = (fused_tv_loss, fused_nla_loss)
-        except ImportError:
-            _FUSED_KERNELS = False
-    return _FUSED_KERNELS or None
+@cache
+def _fused_kernel(name: str):
+    """Import and cache a fused kernel by name; ``None`` if Triton is unavailable."""
+    try:
+        from speculators.models import fused_tv_loss as mod  # noqa: PLC0415
+    except ImportError:
+        return None
+    return getattr(mod, name)
 
 
 def tv_loss_fused_or_eager(logits: torch.Tensor, targets: torch.Tensor):
     """TV loss: fused Triton on CUDA/ROCm (fp32), eager ``tv_loss`` on CPU/NPU."""
-    kernels = _fused_kernels()
-    if kernels is not None and logits.is_cuda:
-        return kernels[0](logits, targets)
+    if logits.is_cuda:
+        kernel = _fused_kernel("fused_tv_loss")
+        if kernel is not None:
+            return kernel(logits, targets)
     return tv_loss(logits, targets)
 
 
 def nla_loss_fused_or_eager(logits: torch.Tensor, targets: torch.Tensor):
     """NLA loss: fused Triton on CUDA/ROCm (fp32), eager on CPU/NPU."""
-    kernels = _fused_kernels()
-    if kernels is not None and logits.is_cuda:
-        return kernels[1](logits, targets)
+    if logits.is_cuda:
+        kernel = _fused_kernel("fused_nla_loss")
+        if kernel is not None:
+            return kernel(logits, targets)
     return neg_log_acceptance_loss(logits, targets)
 
 
--- a/src/speculators/models/metrics.py
+++ b/src/speculators/models/metrics.py
@@ -4,8 +4,6 @@ from collections.abc import Callable
 
 import torch
 
-from speculators.models.fused_tv_loss import fused_nla_loss, fused_tv_loss
-
 _EPS = 1e-5
 
 LossConfig = dict[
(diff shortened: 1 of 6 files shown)


rahul-tuli:
@ElizaWszola

tests/unit/models/test_metrics.py
```diff
@@ -197,8 +199,32 @@ def test_matches_l1_form_shape_and_range(self):
...
+        Uses bf16 inputs (the real training regime): the loss must be returned in
+        fp32 like the eager path, not downcast to the bf16 input dtype.
+        """
+        torch.manual_seed(0)
+        base = torch.randn(1, 64, 512)
+        targets = torch.randn(1, 64, 512, device="cuda", dtype=torch.bfloat16)
+        le = base.clone().to("cuda", torch.bfloat16).requires_grad_(True)
+        lf = base.clone().to("cuda", torch.bfloat16).requires_grad_(True)
+        out_e = tv_loss(le, targets)
+        out_f = resolve_loss_fn("tv")(lf, targets)
```
ElizaWszola: nit: I think it would be good for readability to explicitly call the fused kernel here, without running it through the resolve function

src/speculators/models/metrics.py
```diff
@@ -367,13 +367,52 @@ def dpace_loss_decay(
...
+        except ImportError:
+            _FUSED_KERNELS = False
+    return _FUSED_KERNELS or None
+
+
+def tv_loss_fused_or_eager(logits: torch.Tensor, targets: torch.Tensor):
+    """Fused Triton TV loss on CUDA (fp32, like eager ``tv_loss``); eager fallback."""
+    kernels = _fused_kernels()
+    if kernels is not None and logits.is_cuda:
+        return kernels[0](logits, targets)
```
ElizaWszola: nit: why not just try-import-cache only the needed kernel?

WindChimeRan:
@ElizaWszola Thanks for the review! I removed the eager or fused kernel dispatch. Now we only use fused kernel (because eager will OOM and it should be just used as numeric reference)

ElizaWszola (commented):
The changes look good to me now :)

speculatorsbot (commented):
LGTM. The online-softmax fusion is mathematically correct — the backward gradient `grad = -go * p * (I(p<=q) - s_s)` is the right subgradient of TV distance through softmax, and the five-scalar stat buffer avoids materializing the full `[T, V]` distributions. Agree with @coderabbitai's suggestion to add a fused-vs-eager NLA test on CUDA for completeness. Recommend approving.

WindChimeRan:
Updated for CI fix: explicitly move  metric tests to GPU (fused tv loss is CUDA-only)

fynnsu (changes requested):
I'm concerned that this will break things for certain accelerator backends (e.g. NPU). Can we have some check (e.g. checking for cuda/amd availability), and fallback to the torch implementation if not?

WindChimeRan:
@fynnsu Updates: revert the dispatch. It should fallback to eager path when cuda is not available. For the dispatch, added try-import-cache only the needed kernel as suggested by Eliza.

speculatorsbot (commented):
Re-reviewed after the dispatch restructure (commits 281eca1..bbc178b). The `@cache`-based per-kernel lazy import cleanly addresses both @fynnsu's non-CUDA fallback concern and @ElizaWszola's import-only-what-you-need suggestion. The `logits.is_cuda` gate runs before any Triton import attempt, so NPU/CPU backends never touch Triton. Kernel math is unchanged and still correct. CI is green. Recommend approving.

fynnsu (approved):
Quality needs a fix, but otherwise looks good

PR #755, opened 2026-07-09
fix: apply --draft-attn-impl when loading with --from-pretrained

### LINKED ISSUE #754: --draft-attn-impl is silently dropped when loading with --from-pretrained
### Your current environment

- speculators: main @ 87d47ff (0.7.0.dev67), where we hit the bug; confirmed still present on current main by code inspection (the `model_class.from_pretrained(...)` call in scripts/train.py passes only t2d / d2t / verifier)
- torch / torch_npu: 2.10.0
- transformers: 5.5.4
- Python: 3.11
- Hardware: Huawei Ascend NPU (Atlas A2 / 910B), CANN 8.5.1, Linux

Note: this is a code-path issue on the `--from-pretrained` branch and is backend-independent; NPU is just where it becomes fatal, because FlexAttention is unavailable there.

### 🐛 Describe the bug

`--draft-attn-impl` is only consumed on the fresh-model construction branch of `scripts/train.py`. When loading via `--from-pretrained`, the CLI choice is silently ignored: HF `PretrainedConfig` never serializes `_attn_implementation`, so the checkpoint's config.json has no such field and loading resolves the model default (`simple_flex_attention`). The `model_class.from_pretrained(...)` call passes only `t2d` / `d2t` / `verifier` — no attention implementation.

This looks like the same disease class as #601 (target_layer_ids lost on from_pretrained) — config-carried choices that don't survive the load path.

On CUDA this is invisible because flex happens to work. On backends without FlexAttention it fails in two stages:

1. kernel lookup routes to flex → `ValueError` in attention dispatch;
2. even after rewriting the configs post-load, the mask builder has already produced a flex `BlockMask` object, which then reaches non-flex kernels expecting a dense bool tensor:

```
AttributeError: 'BlockMask' object has no attribute 'dim'
```

### Reproduction

1. Train briefly with a non-default `--draft-attn-impl`, save a checkpoint.
2. Reload with `--from-pretrained <ckpt> --draft-attn-impl <same>` (e.g. for offline eval with `--val-only`) → the flag is ignored; on non-flex backends this crashes as above.

### Proposed fix (PR ready)

Apply the CLI selection on the from-pretrained path the same way the fresh-construction branch does (set it on the loaded config before/at `from_pretrained`), so HF's plumbing propagates it to all submodule configs and the mask path follows. Optionally, as defense in depth, coerce `BlockMask` → dense at the non-flex kernel entrances; we tested the coercion bit-exact against `block_mask_to_dense_attention_mask` on CPU. Happy to open a PR either way — hit this on Ascend NPU while running offline evaluation of a trained DSpark draft.

Frozen7771:
### Problem

`--draft-attn-impl` is silently dropped on the `--from-pretrained` load path: HF configs never serialize `_attn_implementation`, and the load call doesn't re-apply the CLI choice, so the loaded draft always resolves the model default (`simple_flex_attention`). On backends without FlexAttention this crashes during offline eval of trained checkpoints. Details and reproduction in #754.

### Fix

Re-apply the CLI selection on the loaded config before model construction, mirroring the `from_training_args` mechanism (`transformer_layer_config._attn_implementation`), on both the weights-loading path and the config-only path. Applying the default value is behavior-preserving, so this is a no-op for anyone not using the flag.

### Validation

- Loading a DSpark checkpoint with `--from-pretrained --val-only --draft-attn-impl <non-default>` now routes to the selected backend (previously: flex dispatch error).
- Offline eval of a GLM-5.2 draft over six public benchmarks ran end-to-end on this path; scoring reproducibility across independent runs Δ=0.14%.

Fixes #754

### DIFF
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -370,17 +370,28 @@ def build_draft_model(
                 t2d=t2d,
                 d2t=d2t,
                 verifier_name_or_path=args.verifier_name_or_path,
-                draft_attn_impl=args.draft_attn_impl,
+                draft_attn_impl=(
+                    args.draft_attn_impl if args.speculator_type != "mtp" else None
+                ),
             )
-        config = model_class.config_class.from_pretrained(args.from_pretrained)
-        # _attn_implementation is never serialized by HF configs, so re-apply the
-        # CLI selection before construction -- mirroring from_training_args.
-        config.transformer_layer_config._attn_implementation = (  # noqa: SLF001
-            args.draft_attn_impl
-        )
+        if args.speculator_type != "mtp":
+            # _attn_implementation is never serialized by HF configs, so re-apply
+            # the CLI selection before construction -- mirroring from_training_args.
+            # MTP is skipped: its from_training_args never sets the field and its
+            # __init__ resolves its own default ("eager") when it is absent.
+            config = model_class.config_class.from_pretrained(args.from_pretrained)
+            config.transformer_layer_config._attn_implementation = (  # noqa: SLF001
+                args.draft_attn_impl
+            )
+            return model_class.from_pretrained(
+                args.from_pretrained,
+                config=config,
+                t2d=t2d,
+                d2t=d2t,
+                verifier=args.verifier_name_or_path,
+            )
         return model_class.from_pretrained(
             args.from_pretrained,
-            config=config,
             t2d=t2d,
             d2t=d2t,
             verifier=args.verifier_name_or_path,
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -315,9 +315,7 @@ def _build_from_config_only(
     """
     config = model_class.config_class.from_pretrained(path)
     if draft_attn_impl is not None:
-        config.transformer_layer_config._attn_implementation = (  # noqa: SLF001
-            draft_attn_impl
-        )
+        config.transformer_layer_config._attn_implementation = draft_attn_impl
     speculators_config = getattr(config, "speculators_config", None)
     # Fall back to the CLI --verifier-name-or-path only when the saved config has
     # no verifier path -- either null or blanked to "". A real path in the config
@@ -380,9 +378,7 @@ def build_draft_model(
             # MTP is skipped: its from_training_args never sets the field and its
             # __init__ resolves its own default ("eager") when it is absent.
             config = model_class.config_class.from_pretrained(args.from_pretrained)
-            config.transformer_layer_config._attn_implementation = (  # noqa: SLF001
-                args.draft_attn_impl
-            )
+            config.transformer_layer_config._attn_implementation = args.draft_attn_impl
             return model_class.from_pretrained(
                 args.from_pretrained,
                 config=config,
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -305,6 +305,7 @@ def _build_from_config_only(
     t2d: torch.Tensor | None,
     d2t: torch.Tensor | None,
     verifier_name_or_path: str | None = None,
+    draft_attn_impl: str | None = None,
 ) -> SpeculatorModel:
     """Initialize a fresh draft from a saved speculator *config* (no weights).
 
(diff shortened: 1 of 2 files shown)


scripts/train.py
```diff
@@ -365,9 +370,17 @@ def build_draft_model(
...
                 d2t=d2t,
                 verifier_name_or_path=args.verifier_name_or_path,
+                draft_attn_impl=args.draft_attn_impl,
             )
+        config = model_class.config_class.from_pretrained(args.from_pretrained)
+        # _attn_implementation is never serialized by HF configs, so re-apply the
+        # CLI selection before construction -- mirroring from_training_args.
+        config.transformer_layer_config._attn_implementation = (  # noqa: SLF001
+            args.draft_attn_impl
+        )
```
orestis-z: **MTP behavior change**: This unconditional assignment applies `args.draft_attn_impl` (always `"simple_flex_attention"` for MTP, enforced at [line 456](https://github.com/vllm-project/speculators/blob/9464cdb9f36bd1ed258547db68e776352557fc34/scripts/train.py#L456)) to MTP checkpoints loaded via `--from-pretrained`.

But MTP's `__init__` [defaults to `"eager"`](https://github.com/vllm-project/speculators/blob/9464cdb9f36bd1ed258547db68e776352557fc34/src/speculators/models/mtp/core.py#L72-L73) when `_attn_implementation is None`, and MTP's [`from_training_args`](https://github.com/vllm-project/speculators/blob/9464cdb9f36bd1ed258547db68e776352557fc34/src/speculators/models/mtp/core.py#L230) doesn't set `_attn_implementation` at all (unlike eagle3/dflash/peagle which all [read `draft_attn_impl` from kwargs](https://github.com/vllm-project/speculators/blob/9464cdb9f36bd1ed258547db68e776352557fc34/src/speculators/models/eagle3/core.py#L351-L352)). So on the fresh-construction path, MTP always gets `"eager"`.

After this change, MTP loaded via `--from-pretrained` would get `"simple_flex_attention"` instead of `"eager"`. Since MTP uses standard HF decoder layers (not speculators' custom mask builders), `"simple_flex_attention"` is not a recognized HF attention type and could cause a dispatch error — the same class of bug this PR fixes for eagle3/dflash, but in reverse.

Suggestion: guard the assignment so it skips MTP:

```python
if args.speculator_type != "mtp":
    config = model_class.config_class.from_pretrained(args.from_pretrained)
    config.transformer_layer_config._attn_implementation = (
        args.draft_attn_impl
    )
    return model_class.from_pretrained(
        args.from_pretrained,
        config=config,
        t2d=t2d,
        d2t=d2t,
        verifier=args.verifier_name_or_path,
    )
return model_class.from_pretrained(
    args.from_pretrained,
    t2d=t2d,
    d2t=d2t,
    verifier=args.verifier_name_or_path,
)
```

The same issue applies to the `_build_from_config_only` path (line 373 passes `draft_attn_impl=args.draft_attn_impl` for all speculator types, including MTP).

orestis-z (commented):
**Design & approach**: The fix correctly identifies the root cause — `_attn_implementation` is not serialized by HF configs, so loading via `--from-pretrained` loses the CLI selection. The approach of re-applying it on the loaded config before construction mirrors `from_training_args` and is the right pattern.

Two items to address:

1. **MTP correctness** — the unconditional assignment on the weights-loading path changes MTP behavior (see inline comment).
2. **Regression test** — this is a bug fix but includes no test. A regression test that asserts `_attn_implementation` survives the `from_pretrained` round-trip (and that MTP keeps its own default) would prevent this from regressing again. The existing `test_build_from_config_only` tests in `tests/unit/train/test_draft_config_init.py` would be a natural place to add coverage.

scripts/train.py
Frozen7771: Good catch on MTP — you're right: its `from_training_args` never sets the field and `__init__` resolves `"eager"`, so the unconditional re-application would have introduced for MTP exactly the class of bug this PR fixes elsewhere. Pushed:

1. the guard as suggested, on both the weights path and the `_build_from_config_only` call site;
2. regression tests in `test_draft_config_init.py`: a round-trip test documenting that `_attn_implementation` is dropped by serialization, and a test asserting `_build_from_config_only` re-applies the CLI selection (fails on the pre-fix signature).

For MTP end-to-end from-pretrained coverage: with the guard in place the MTP path is byte-identical to the previous behavior, and the existing MTP parse/routing tests cover that surface — happy to add a heavier fixture if you'd prefer explicit coverage.

orestis-z (commented):
MTP concern from my prior review is addressed — both the weights path and the config-only path now skip the override for MTP. LGTM. Recommend approving.

fynnsu (approved):

PR #760, opened 2026-07-09
Add support sample_from_anchor logic to DFlash/DSpark 

fynnsu:
## Purpose

Currently, DFlash and DSpark block slot `k` predicts token `anchor + k` (e.g. the query token itself). This is based on a convention from DFlash. However, DSpark implementation sometimes predict token `anchor + k + 1` instead. 

This pr adds a configuration option to both alg configs indicating the prediction mode. It also adds support for `anchor + k + 1` style training to speculators. 

Config change:

Both DFlash and DSpark configs now include a
```
sample_from_anchor: bool
```
field. The default for DFlash is `False` and the default for DSpark is `True`. 

## Tests

Working on creating a test checkpoint. 

## Checklist

I have filled in:

### DIFF
--- a/src/speculators/models/dflash/core.py
+++ b/src/speculators/models/dflash/core.py
@@ -1,3 +1,4 @@
+import logging
 from typing import ClassVar
 
 import torch
@@ -21,6 +22,8 @@ from speculators.models.dflash.utils import (
 )
 from speculators.models.metrics import LossConfig, resolve_loss_config
 from speculators.models.utils import conditional_torch_compile, resolve_target_layer_ids
+
+logger = logging.getLogger(__name__)
 
 
 @SpeculatorModel.register("dflash")
@@ -105,6 +108,15 @@ class DFlashDraftModel(DraftVocabMixin, SpeculatorModel):
         )
         self.verifier_norm.weight.requires_grad = False
         self.block_size = config.block_size
+
+        # Warn if using DFlash with sample_from_anchor=True (may not be supported)
+        if type(self).__name__ == "DFlashDraftModel" and config.sample_from_anchor:
+            logger.warning(
+                "DFlash with sample_from_anchor=True may not be supported in "
+                "all inference engines (e.g., vLLM). Verify compatibility with your "
+                "deployment target."
+            )
+
         self.post_init()
 
     @property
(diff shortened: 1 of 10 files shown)


fynnsu:
@coderabbitai review

shanjiaz (approved):
Looks good! Thanks for adding this change. Is it already supported in vLLM main?

src/speculators/models/dflash/core.py
```diff
@@ -174,18 +186,29 @@ def _build_base_config_kwargs(
             "draft_attn_impl", "simple_flex_attention"
         )
         block_size = kwargs.get("block_size", 8)
+
+        default_sample_from_anchor = algorithm == "dspark"
+        sample_from_anchor = kwargs.get(
+            "sample_from_anchor", default_sample_from_anchor
+        )
```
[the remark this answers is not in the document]
shanjiaz: `sample-from-anchor` is only None if we don't pass in `sample-from-anchor` right? Then `kwargs.get("sample_from_anchor", default_sample_from_anchor)` should be equal to `default_sample_from_anchor`?

src/speculators/models/dflash/core.py
```diff
@@ -383,7 +408,9 @@ def _backbone_forward(
             .to(aligned_loss_mask.dtype)
         )  # shape: [1, num_anchors*block_size]
 
-        aligned_loss_mask[:, :: self.block_size] = 0
+        # For sample_from_anchor=False, mask slot 0 (anchor) since it's not trained
+        if not self.config.sample_from_anchor:
```
orestis-z, line 415: The loss mask change here is correct — when `sample_from_anchor=True`, slot 0 is trained and its gradient flows properly. However, the downstream metrics computation in both [`dflash/metrics.py:88-101`](https://github.com/vllm-project/speculators/blob/fbf6e48d0094314ccc462c3ecb619aee8ce28eab/src/speculators/models/dflash/metrics.py#L88-L101) and [`dspark/metrics.py:144-154`](https://github.com/vllm-project/speculators/blob/fbf6e48d0094314ccc462c3ecb619aee8ce28eab/src/speculators/models/dspark/metrics.py#L144-L154) hardcodes skipping position 0:

```python
# Position 0 is the anchor — intentionally excluded from accuracy
metrics["full_acc_sum"] = correct_per_pos[1:].sum()
metrics["full_acc_total"] = total_per_pos[1:].sum()
for pos in range(1, block_size):
```

When `sample_from_anchor=True`, position 0 is a valid speculative position, but:
- `full_acc` excludes it → accuracy underreported
- EAL/`accept_len` starts from position 1 → expected accepted length underestimated
- No `position_0_acc` metric is ever emitted

The DSpark metrics additionally have `accept_blocks[:, 1:]` and `conf_prefix` slicing that skips position 0 in the confidence calibration path.

Suggestion: pass `sample_from_anchor` (or derive the starting slot from the loss mask) into both `compute_metrics` functions and use `start_pos = 0 if sample_from_anchor else 1` for the accuracy/EAL loops.

orestis-z (commented):
Design looks solid — adding `sample_from_anchor` as an algorithm-specific config field with the right defaults per algorithm is the correct approach.

One issues: the metrics computation in both `dflash/metrics.py` and `dspark/metrics.py` hardcodes skipping position 0, so reported accuracy and EAL will undercount for `sample_from_anchor=True` models even though training loss is correct.

src/speculators/models/dflash/core.py
fynnsu: Fixed

PR #765, opened 2026-07-10
Fix numerical issues by holding params in fp32

eldarkurtic:
## Human written explanation (see below for more verbose Clauded version)
All models trained with `speculators` end up with **untrained normalization layers** (`.weight == 1.0`). For example, in all RedHatAI's DFlash speculators for Qwen3-8B, gemma-4-31B-it, Laguna-XS.2 
  - layers.N.input_layernorm.weight 
  - layers.N.post_attention_layernorm.weight 
  - layers.N.self_attn.q_norm.weight
  - layers.N.self_attn.k_norm.weight 
  - hidden_norm.weight
  - norm.weight

are exactly equal to `1.0`. This is happening because our training is done fully in `bf16`. This is a problem for weights which start training from a "large value", in this case normalization layers start from `.weight = 1.0`. In bf16, the step to move away from 1.0 is quite large and therefore the weights are never updated as the gradient update delta is smaller than the step size to move away from 1.0 in bf16 format. This didn't affect linear layers that much because they start from much smaller random init values and the step size is smaller, therefore the gradient update can "move them".  

I already ran some ablations with a manual hacky fix for this and saw ~10-15% improvement:
image

Don't merge this PR until I confirm that this clean integration also works well and doesn't break something else. 

## TL;DR

Optimizer updates now accumulate in **fp32 master weights** instead of pure bf16. Forward/backward compute is unchanged (still bf16, bit-identical activations) — only the weights the optimizer steps on are fp32. Always on, no flag, matching SpecForge / TorchSpec / DeepSpec.

## Why

In pure-bf16 training, any weight update smaller than half a bf16 ulp (~0.2% relative) is rounded away at `optimizer.step()`. At typical LRs (`1e-4`–`1e-5`) a large fraction of per-step updates is sub-ulp and only survives when momentum happens to push it over the rounding threshold.

From the new test suite — 300 AdamW steps of `1e-5` updates on a weight of `1.0`:

| | final weight |
|---|---|
| pure bf16 (before) | `1.0` — every update lost |
| fp32 masters (after) | `bf16(0.997)` — all updates accumulated |

Every comparable trainer (SpecForge, TorchSpec, DeepSpec) hard-codes fp32 masters for this reason.

## How it works

Two mechanisms, one internal switch (`Trainer._fp32_master_mode` — on for bf16/fp16, skipped when already training in fp32):

| Path | Mechanism |
|---|---|
| **FSDP2 (torchrun)** | Model kept in fp32 before sharding; `apply_fully_sharded` gains a `param_dtype` (compute dtype, default bf16). FSDP2 casts params to bf16 at all-gather → compute unchanged, but sharded params, gradients (`reduce_dtype` was already fp32), grad clipping, and Muon/AdamW moments are all fp32. No wrapper; composes with Muon-on-DTensor and `get_optimizer_state_dict` as-is. |
| **Single GPU** | New `FP32MasterOptimizer` wrapper (`train/optimizers.py`): bf16 model params + fp32 masters of trainable params. Each step: copy bf16 grads → fp32, step inner Muon/AdamW on masters, write downcast result back. Shares `param_groups` with the inner optimizer, so existing LR schedulers work unchanged. Masters only for `requires_grad` params — no fp32 copies of frozen embeddings / verifier lm_head. |

### Checkpointing (`train/checkpointer.py`)

Required so resume doesn't silently defeat the masters:

- Optimizer state is saved in **native dtypes** (no more bf16 downcast). No-op for old pure-bf16 checkpoints.
- Distributed checkpoints embed the fp32 trainable params in the optimizer payload → resume restores exact masters, while **`model.safetensors` stays bf16 and vLLM-deployable**.
- `FP32MasterOptimizer.state_dict()` embeds its masters; round-trip is bit-exact.
- Missing `optimizer_state_dict.pt` now warns + resumes with fresh optimizer state (matches scheduler-loader behavior).

## ⚠️ Compatibility notes

- **Optimizer checkpoints are ~2× larger** (fp32 masters + moments). Old checkpoints still load; bf16 state is upcast on load.
- **Legacy single-GPU `--optimizer adamw` checkpoints** have a different param grouping and can't load into the wrapper — the error message says to delete `optimizer_state_dict.pt` and resume from model weights. Default `muon` checkpoints resume cleanly.
- Optimizer-side GPU memory roughly doubles (sharded across ranks under FSDP2). Activation memory unchanged.

## Testing

- ✅ New `tests/unit/train/test_fp32_master_optimizer.py` — 15 tests: sub-ulp accumulation, scheduler integration, bit-exact checkpoint round-trip, legacy loads (compatible + mismatched), Muon/AdamW group wrapping, frozen-param filtering, dtype gating, missing-file resume.
- ✅ Full `tests/unit`: **437 passed**, 7 skipped (CUDA/multi-GPU).
- ✅ ruff check/format clean.

### DIFF
--- a/docs/cli/train.md
+++ b/docs/cli/train.md
@@ -130,6 +130,9 @@ torchrun --standalone --nproc_per_node=4 scripts/train.py \
 
 - **`--muon-adjust-lr-fn`** (str, default: `"match_rms_adamw"`) Muon LR adjustment strategy. Options: `original`, `match_rms_adamw`. Only used with `--optimizer muon`.
 
+!!! note "fp32 master weights"
+    Optimizer updates always accumulate in fp32 master weights; forward/backward compute stays in `--hidden-states-dtype` (bf16 by default). Under FSDP2 (torchrun) the sharded parameters and optimizer state are held in fp32 and cast to the compute dtype at all-gather; on a single GPU the optimizer keeps an fp32 master copy of the trainable weights. Without this, updates smaller than a bf16 ulp would round away every step.
+
 ### Eagle3-Specific Arguments
 
 - **`--use-off-policy-tokens`** (flag) Use off-policy tokens during training (required for [regenerated data](response_regeneration.md)).
--- a/src/speculators/train/checkpointer.py
+++ b/src/speculators/train/checkpointer.py
@@ -270,21 +270,30 @@ class SingleGPUCheckpointer(BaseCheckpointer):
 
     def load_optimizer_state_dict(
         self,
-        model: PreTrainedModel,
+        model: PreTrainedModel,  # noqa: ARG002 (kept for the base-class interface)
         optimizer: OptimizerOrList,
-        float_dtype: torch.dtype | None = None,
+        float_dtype: torch.dtype | None = None,  # noqa: ARG002
     ):
+        optimizer_path = self.optimizer_path(self.previous_epoch)
+        if not optimizer_path.exists():
+            logger.warning(
+                f"No optimizer state found at {optimizer_path}; "
+                "resuming with a fresh optimizer state."
+            )
+            return
         device = get_current_device()
         loaded = torch.load(
-            self.optimizer_path(self.previous_epoch),
+            optimizer_path,
             weights_only=True,
             map_location=device,
         )
         optimizers = _as_list(optimizer)
         loaded_list = loaded if isinstance(loaded, list) else [loaded]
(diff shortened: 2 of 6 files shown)


rahul-tuli:
@eldarkurtic it seems like @fynnsu 's diff already resolves this: https://github.com/vllm-project/speculators/pull/711 could you verify?

fynnsu:
Closing since #711 has landed

PR #758, opened 2026-07-09
feat(data): extract verifier KV caches for Gemma4 MTP

shotsan:
### Purpose
Resolves the offline data ingestion requirements for Gemma 4 MTP training by properly extracting and batching the verifier's KV caches.

The Gemma 4 MTP architecture uses a query-only attention mechanism that borrows the KV cache from the verifier model. To train this offline, the dataset must provide the extracted KV caches for every token. This PR updates the data loading and batching pipeline to handle these new 5-D KV tensors.

**Code Changes:**
1. **Dynamic Tensor Stacking**: Updated `ArrowDataset._get_raw_data` and `standardize_data_v1` to dynamically detect `kv_last_local_k` and `_v` keys. It uses `torch.stack` to combine them into the unified `verifier_kv_last_local` tensor that the downstream model expects.
2. **Type Casting**: In `BaseDataset.__getitem__`, the KV caches are correctly cast to the `hidden_states_dtype` (bfloat16) to prevent downstream mixed-precision crashes.
3. **MTP Batch Shifting**: Updated `shift_batch_mtp` in `src/speculators/models/mtp/data.py` to correctly forward the 5-D KV cache tensors through the training loop without sequence misalignment.

### Tests
- **Unit Tests**: Updated `tests/unit/models/test_mtp_data.py`. The unit tests pass locally and verify that the 5-D KV tensors `(batch_size, seq_len, 2, num_kv_heads, head_dim)` are correctly preserved and aligned.
- **End-to-End Test**: Roderick-Wu has successfully tested this branch (alongside the two subsequent PRs) to train a full Gemma 4 MTP speculator end-to-end on vLLM.

PR #767, opened 2026-07-10
feat(mtp): add multi-level LM head for Gemma4 MTP (2/3)

### LINKED ISSUE #586: [RFC]: Gemma4 MTP support
### Motivation.

We want to support **training Gemma4 MTP draft models** in speculators.

MTP finetuning already works for the Qwen-style architecture, as of: 1. the MTP algorithm, converter, and stitcher are merged (#452), and 2. MTP finetuning is wired into the shared training infrastructure with online/offline paths, and e2e
tests (#366). 

This RFC covers the Gemma4-specific delta on top of that MTP support. Gemma4's MTP drafter differs from the Qwen-style architecture in two ways the current stack cannot express:

1. The drafter shares the verifier's KV cache. Gemma4's MTP attention is
   query-only: it has no K/V projections and no KV cache of its own. At inference it
   borrows the verifier's last local and last global attention KV caches and cross-attends into them. 
2. A sparse LM head. The drafter does not use a flat draft-vocab
   LM head; it uses a centroid-masked (multi-level) head for efficient decoding.

Consequence for training:  Gemma4's drafter is not self-contained: its query-only attention has nothing to attend to
unless the verifier's KV tensors are supplied as training inputs, the same way hidden states are supplied today. There is currently no way to get those KV tensors out of the verifier and into the training process.

So Gemma4 MTP finetuning needs work on both sides of:
- vLLM: export the relevant verifier KV caches alongside hidden states.
- speculators: a Gemma-specific model definition, a data path that loads and aligns those KV tensors, and LM-head support.

This RFC proposes the speculators-side design and the shared data contract; the vLLM
KV-export mechanism is covered by the companion vLLM RFC.

### Proposed Change.

High-level: extend the offline-extraction data contract with verifier KV caches, add a Gemma4 MTP model definition that consumes them, and add multi-level LM-head support.

1. Shared data contract.
For Gemma4 we extend it with the two verifier KV caches the drafter shares:

```
{
  hidden_states:            [seq, n_layers, hidden]   
  token_ids:                [seq]                      
  verifier_kv_last_local:   [seq, kv_heads, head_dim] # new — K and V of the last sliding layer
  verifier_kv_last_global:  [seq, kv_heads, head_dim] # new — K and V of the last full layer
}
```

2. Gemma4 MTP variant. The Gemma4 variant provides a query-only attention module that consumes the injected verifier K/V instead of projecting/attending over its own sequence; the pre/post projection that fuses verifier hidden states with the drafter's token embedding.

3. Training data path. Extend the dataset/collation layer to load the verifier KV caches from the per-sample files, position-align them to the token sequence (the existing alignment check between stored tokens and training inputs must cover the KV tensors too), and route them into the Gemma draft model during the forward pass. This should reuse the existing per-algorithm argument-passing mechanism, so no changes to the core training loop are expected.

4. Multi-level LM head support. Represent the centroid-masked head in the Gemma variant and its training loss. Note the MTP code currently uses a flat LM head and explicitly disallows vocab reduction, so the multi-level head is a new capability.

### Any Other Things.

_No response_

shotsan:
This is **Part 2 of 3** in the rollout to introduce Gemma4 MTP support (Resolves #586).
Builds on top of the offline data pipeline from Part 1 (#758).

### Overview
This PR implements the `MultiLevelLMHead` for Gemma4 MTP, which replicates the centroid-masked sparse argmax approach used by vLLM (`Gemma4MTPMaskedEmbedder`). 

Instead of duplicating the standard flat cross-entropy logic, the `MultiLevelLMHead` acts as an auxiliary module that only computes the centroid classification loss using the `token_ordering` mapping buffer. The unreduced centroid loss is then added to the standard token loss inside `core.py`.

### Changes
- Added `num_centroids` and `top_k_centroids` to `MTPSpeculatorConfig`.
- Created a compact `MultiLevelLMHead` (`head.py`) containing the centroid linear projection and `token_ordering` buffer. Note: Parameter names match vLLM's implementation exactly (`masked_embedding.centroids.weight` and `masked_embedding.token_ordering`) for seamless checkpoint loading.
- Conditionally integrated `MultiLevelLMHead` into `MTPDraftModel`'s forward pass.
- Added comprehensive unit tests in `tests/unit/models/test_mtp_head.py`.

### Context
By keeping the standard cross-entropy computation intact and making the centroid projection an additive loss term, we ensure that the logic is highly robust and compact. No existing code was deleted.

### DIFF
--- a/src/speculators/models/mtp/config.py
+++ b/src/speculators/models/mtp/config.py
@@ -2,6 +2,7 @@
 
 from typing import Any, Literal
 
+import pydantic
 from pydantic import Field, field_serializer, field_validator
 from transformers import AutoConfig, PretrainedConfig
 from transformers.models.qwen2.configuration_qwen2 import Qwen2Config
@@ -52,6 +53,29 @@ class MTPSpeculatorConfig(SpeculatorModelConfig):
         ),
     )
 
+    num_centroids: int | None = Field(
+        default=None,
+        description=(
+            "Number of centroids for a centroid-masked multi-level LM head. "
+            "If None, standard flat LM head is used."
+        ),
+    )
+
+    centroid_intermediate_top_k: int = Field(
+        default=32,
+        description="Number of top centroids to select during multi-level LM head decoding.",
+    )
+
+    @pydantic.model_validator(mode="after")
+    def validate_centroids(self) -> "MTPSpeculatorConfig":
+        if self.num_centroids is not None:
+            if self.num_centroids < 1:
+                raise ValueError(f"num_centroids must be >= 1, got {self.num_centroids}")
+            if self.vocab_size % self.num_centroids != 0:
+                raise ValueError(f"vocab_size ({self.vocab_size}) must be divisible by num_centroids ({self.num_centroids})")
+            if self.centroid_intermediate_top_k > self.num_centroids:
+                raise ValueError(f"centroid_intermediate_top_k ({self.centroid_intermediate_top_k}) cannot exceed num_centroids ({self.num_centroids})")
+        return self
     @property
     def hidden_size(self) -> int:
         return self.transformer_layer_config.hidden_size  # type: ignore[return-value]
--- a/src/speculators/models/mtp/config.py
+++ b/src/speculators/models/mtp/config.py
@@ -52,18 +52,6 @@ class MTPSpeculatorConfig(SpeculatorModelConfig):
         ),
     )
 
-    num_centroids: int | None = Field(
-        default=None,
-        description=(
-            "Number of centroids for a centroid-masked multi-level LM head. "
-            "If None, standard flat LM head is used."
-        ),
-    )
-
-    top_k_centroids: int = Field(
-        default=32,
-        description="Number of top centroids to select during multi-level LM head decoding.",
-    )
 
     @property
     def hidden_size(self) -> int:
--- a/src/speculators/models/mtp/config.py
+++ b/src/speculators/models/mtp/config.py
@@ -52,7 +52,6 @@ class MTPSpeculatorConfig(SpeculatorModelConfig):
         ),
     )
 
-
     @property
     def hidden_size(self) -> int:
         return self.transformer_layer_config.hidden_size  # type: ignore[return-value]
--- a/src/speculators/models/mtp/config.py
+++ b/src/speculators/models/mtp/config.py
@@ -52,6 +52,19 @@ class MTPSpeculatorConfig(SpeculatorModelConfig):
         ),
     )
 
+    num_centroids: int | None = Field(
+        default=None,
+        description=(
+            "Number of centroids for a centroid-masked multi-level LM head. "
+            "If None, standard flat LM head is used."
+        ),
+    )
+
+    top_k_centroids: int = Field(
+        default=32,
+        description="Number of top centroids to select during multi-level LM head decoding.",
+    )
+
     @property
     def hidden_size(self) -> int:
         return self.transformer_layer_config.hidden_size  # type: ignore[return-value]
--- a/src/speculators/models/mtp/core.py
+++ b/src/speculators/models/mtp/core.py
@@ -63,6 +63,8 @@ class MTPDraftModel(DraftVocabMixin, SpeculatorModel):
         "verifier_lm_head.weight",
         "t2d",
         "d2t",
+        "masked_embedding.centroids.weight",
+        "masked_embedding.token_ordering",
     ]
 
     t2d: torch.Tensor | None
@@ -86,6 +88,17 @@ class MTPDraftModel(DraftVocabMixin, SpeculatorModel):
         self.rotary_emb = self._model_definitions.rotary_emb_class(tc)
 
         self.post_init()
+
+        if getattr(self.config, "num_centroids", None) is not None:
+            from speculators.models.mtp.head import MultiLevelLMHead
+
+            self.masked_embedding = MultiLevelLMHead(
+                hidden_size=tc.hidden_size,
+                vocab_size=tc.vocab_size,
+                num_centroids=self.config.num_centroids,
+            )
+        else:
+            self.masked_embedding = None
 
     @property
     def layers(self) -> nn.ModuleList:
@@ -208,12 +221,20 @@ class MTPDraftModel(DraftVocabMixin, SpeculatorModel):
                 step_targets = step_targets.clone()
                 step_targets[step_mask == 0] = _IGNORE_INDEX
             weight = step_weights[step] if step_weights is not None else 1.0
+
             unreduced = nn.functional.cross_entropy(
                 logits.permute(0, 2, 1),
                 step_targets,
                 ignore_index=_IGNORE_INDEX,
                 reduction="none",
             )
+
+            if getattr(self, "masked_embedding", None) is not None:
+                unreduced += self.masked_embedding.compute_centroid_loss(
+                    hidden_states=mtp_output,
+                    targets=step_targets,
+                    ignore_index=_IGNORE_INDEX,
+                )
             valid_count = (step_targets != _IGNORE_INDEX).sum()
             step_loss = weight * unreduced.sum() / valid_count.clamp(min=1)
             total_loss = total_loss + step_loss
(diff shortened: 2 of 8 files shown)


src/speculators/models/mtp/config.py
```diff
@@ -52,6 +52,19 @@ class MTPSpeculatorConfig(SpeculatorModelConfig):
...
+        description=(
+            "Number of centroids for a centroid-masked multi-level LM head. "
+            "If None, standard flat LM head is used."
+        ),
+    )
+
+    top_k_centroids: int = Field(
+        default=32,
+        description="Number of top centroids to select during multi-level LM head decoding.",
+    )
```
orestis-z: `top_k_centroids` is defined here but never referenced anywhere in this PR's code -- neither `MultiLevelLMHead` nor `MTPDraftModel` reads it.

More importantly, the field name does not match what vLLM expects. vLLM's `Gemma4MTPModel` reads the top-K centroid count as:

```python
# vllm/model_executor/models/gemma4_mtp.py:538
top_k = getattr(config, "centroid_intermediate_top_k", 32)
```

So a speculators-trained checkpoint that serializes `top_k_centroids` in its `config.json` will silently miss vLLM's lookup, and vLLM will fall back to its hardcoded default of 32. With the current default of 32 this is invisible, but it breaks interop if someone changes the value.

Suggestion: either rename to `centroid_intermediate_top_k` to match the vLLM contract, or defer adding the field until Part 3 when it is wired up (YAGNI).

orestis-z (commented):
Design is clean -- the centroid classification loss as an additive auxiliary term keeps the change compact and the existing flat CE path untouched. Centroid inverse-mapping and shape handling in `compute_centroid_loss` look correct.

Two notes:

1. **`top_k_centroids` naming / usage** -- see inline comment.
2. **Undocumented data.py scope** -- the `verifier_kv_last_local` / `verifier_kv_last_global` pipeline changes are part of RFC #586 point 3 ("Training data path") but are not mentioned in this PR's description. They are orthogonal to the multi-level head and could be split out or at least called out in the overview so reviewers know they are intentional.

Agree with @coderabbitai's suggestion to add input validation for `num_centroids` in both config and `MultiLevelLMHead.__init__`. Recommend approving once the naming concern below is resolved.

shotsan:
Thanks for the detailed review @coderabbitai and @orestis-z! I've pushed a new commit addressing all of the feedback:

1. **Config Naming & Validation:** Renamed `top_k_centroids` to `centroid_intermediate_top_k` to match vLLM's contract. Added strict Pydantic validators enforcing `num_centroids >= 1` and ensuring `vocab_size` is evenly divisible by `num_centroids`.
2. **Head Inverse Mapping:** Precomputed the token-to-centroid inverse mapping in `MultiLevelLMHead.__init__` as a registered buffer (`token_ordering_inv`) and exposed a `_rebuild_inverse()` helper so we aren't recomputing it on every forward pass. 
3. **`torch.compile` Compatibility:** Removed in-place additions during the loss computation, utilizing `masked_scatter` and out-of-place assignment (`unreduced = unreduced + ...`). Added the missing `masked_embedding` class-level annotation.
4. **Data Type Casting:** Replaced the fuzzy substring float casting in `data.py` with explicit checks for `verifier_last_hidden_states` and `.startswith("verifier_kv_last_")`.
5. **Testing:** Added all the requested unit tests, including validation checks for non-divisible vocabulary sizes, integration checks verifying the centroid loss is included in the forward pass, and robust dataset collator tests handling missing verifier KV caches across mixed batches.

Let me know if there's anything else needed here!

shotsan:
@coderabbitai Thanks for catching that gap in the collator tests! You were spot on.

I just pushed a new commit to harden `create_collate_fn` directly. Instead of strictly relying on `batch[0]`'s keys, the collator now unions all keys across the entire batch. If an optional field like `verifier_kv_last_local` is missing for a particular sample, we now dynamically infer its required shape using the matching sequence length from its `input_ids`, and insert a zeroed-out tensor placeholder. 

This ensures that mixed batches degrade gracefully and collate perfectly without crashing or silently dropping data for the rest of the batch! I also updated `test_verifier_kvs_mixed_batch` to explicitly assert the exact zero-padded shapes for both forward and reverse batch orderings.

Let me know if you need anything else to sign off!

src/speculators/models/mtp/core.py
```diff
@@ -63,10 +63,13 @@ class MTPDraftModel(DraftVocabMixin, SpeculatorModel):
         "verifier_lm_head.weight",
         "t2d",
         "d2t",
+        "masked_embedding.centroids.weight",
+        "masked_embedding.token_ordering",
```
orestis-z, line 67: `masked_embedding.token_ordering_inv` (the precomputed inverse buffer added in `head.py`) is missing from this list. vLLM's `Gemma4MTPMaskedEmbedder` only stores `token_ordering`, not the inverse, so loading from a vLLM-style checkpoint will emit a spurious missing-key warning for this buffer.

Since `_rebuild_inverse()` always recomputes it in `__init__`, there is no need to load it from a checkpoint -- it just needs to be silenced here:

```suggestion
        "masked_embedding.token_ordering",
        "masked_embedding.token_ordering_inv",
```

orestis-z (commented):
Previous feedback has been thoroughly addressed -- naming, validation, inverse precomputation, `torch.compile` safety, explicit dtype casting, and mixed-batch collation all look correct. One minor gap below.

shotsan:
Good catch! Added `masked_embedding.token_ordering_inv` to `_keys_to_ignore_on_load_missing` in `core.py` to silence the warning when loading older vLLM checkpoints.

orestis-z (commented):
Previous finding (`token_ordering_inv` missing from `_keys_to_ignore_on_load_missing`) is addressed. LGTM. Recommend approving.

speculatorsbot (commented):
LGTM. Weight names and config fields align with vLLM's `Gemma4MTPMaskedEmbedder`, centroid loss math and shape handling are correct, inverse mapping is properly precomputed with `masked_scatter` for `torch.compile` safety, and the hardened collator gracefully handles mixed KV/non-KV batches. Agree with the previous review -- recommend approving.

tests/unit/train/test_data.py
```diff
@@ -492,6 +492,115 @@ def test_arrow_dataset_default_split_ratio_does_not_crash(tmp_path: Path):
...
+        hidden_states_dtype=torch.bfloat16,
+    )
+    
+    item = ds[0]
+    assert item["verifier_kv_last_local"].dtype == torch.bfloat16
+    assert item["hidden_states"].dtype == torch.bfloat16
+    assert item["verifier_last_hidden_states"].dtype == torch.bfloat16
+
+
+>>>>>>> 40e5d06 (fix(mtp): address PR 767 CodeRabbit and collaborator comments)
```
speculatorsbot, line 603: This file contains unresolved merge conflict markers (`<<<<<<< HEAD`, `=======`, `>>>>>>> 40e5d06`), which will cause a `SyntaxError` and prevent all tests in this module from running.

Additionally, this creates two issues:
1. `test_verifier_kvs_survive_pipeline` is defined twice -- once inside the conflict block (line 497, older version using pre-stacked KV tensors) and once below (line 640, newer version using separate `kv_last_local_k`/`kv_last_local_v` keys). The newer version is the correct one matching the `standardize_data_v1` changes.
2. `test_verifier_kvs_mixed_batch` and `test_verifier_kvs_dtype_cast` only exist inside the conflict block. They need to be preserved after resolving the conflict.

To fix: remove lines 495-496 (`<<<<<<< HEAD` and `=======`) and line 603 (`>>>>>>> 40e5d06 ...`), then remove the older `test_verifier_kvs_survive_pipeline` (lines 497-537) since the updated version at line 640 supersedes it. Keep `test_verifier_kvs_mixed_batch` and `test_verifier_kvs_dtype_cast`.

speculatorsbot (commented):
Unresolved merge conflict markers in `tests/unit/train/test_data.py` need to be fixed before merge. The source files (`head.py`, `core.py`, `config.py`, `data.py`) and test coverage in `test_mtp_head.py` all look correct -- previous feedback has been thoroughly addressed. One blocking item below.

PR #768, opened 2026-07-10
feat(mtp): implement QueryOnlyGemma2Attention for Gemma4 MTP draft models

shotsan:
## Context
This PR introduces the final architecture modifications required for **Gemma4 MTP draft models**, completing the 3-part rollout strategy outlined in Issue #586. 

Unlike Qwen-style MTP which maintains its own full self-attention layers and distinct KV caches, **Gemma4 MTP borrows the verifier's local and global KV caches** directly, stripping K/V projections out of its own layers.

**Dependencies:**
- Depends on **PR #758** (Data pipeline extraction of verifier KV caches - Merged).
- Depends on **PR #767** (Multi-Level LM Head - Under Review). 
*(Note: Commits from #767 currently appear in this PR as this was branched from it for stacking. This will resolve once #767 is merged).*

## Implementation Details
1. **Registered Gemma2 Components**: Added `Gemma2DecoderLayer`, `Gemma2RMSNorm`, and `Gemma2RotaryEmbedding` to the `base_components.py` registry.
2. **`QueryOnlyGemma2Attention`**: Built a specialized customized self-attention module that overrides the base `Gemma2Attention`. It zeroes out K/V linear projections and instead cross-attends into `verifier_kv_last_local` and `verifier_kv_last_global`.
3. **Sliding Window Support**: Dynamic routing of local versus global caches depending on the target layer's `sliding_window` size.
4. **RoPE Handling**: Isolated Rotary Positional Embeddings to only apply to query states since the vLLM extracted verifier KV caches have already been rotated.

## Testing
- Unit tests added in `tests/unit/models/test_mtp_gemma2.py` explicitly mock out the KV cache tensor routing, verifying that local vs. global tensor routing succeeds based on window configuration and shape propagation works as expected.

### DIFF
--- a/src/speculators/models/mtp/core.py
+++ b/src/speculators/models/mtp/core.py
@@ -212,6 +212,8 @@ class MTPDraftModel(DraftVocabMixin, SpeculatorModel):
                 attention_mask=causal_mask,
                 position_ids=step_pos_ids,
                 position_embeddings=step_pos_emb,
+                verifier_kv_last_local=kwargs.get("verifier_kv_last_local"),
+                verifier_kv_last_global=kwargs.get("verifier_kv_last_global"),
             )
 
             logits = self.lm_head(mtp_output)
--- a/src/speculators/models/mtp/core.py
+++ b/src/speculators/models/mtp/core.py
@@ -63,6 +63,8 @@ class MTPDraftModel(DraftVocabMixin, SpeculatorModel):
         "verifier_lm_head.weight",
         "t2d",
         "d2t",
+        "masked_embedding.centroids.weight",
+        "masked_embedding.token_ordering",
     ]
 
     t2d: torch.Tensor | None
@@ -86,6 +88,17 @@ class MTPDraftModel(DraftVocabMixin, SpeculatorModel):
         self.rotary_emb = self._model_definitions.rotary_emb_class(tc)
 
         self.post_init()
+
+        if getattr(self.config, "num_centroids", None) is not None:
+            from speculators.models.mtp.head import MultiLevelLMHead
+
+            self.masked_embedding = MultiLevelLMHead(
+                hidden_size=tc.hidden_size,
+                vocab_size=tc.vocab_size,
+                num_centroids=self.config.num_centroids,
+            )
+        else:
+            self.masked_embedding = None
 
     @property
     def layers(self) -> nn.ModuleList:
@@ -208,12 +221,20 @@ class MTPDraftModel(DraftVocabMixin, SpeculatorModel):
                 step_targets = step_targets.clone()
                 step_targets[step_mask == 0] = _IGNORE_INDEX
             weight = step_weights[step] if step_weights is not None else 1.0
+
             unreduced = nn.functional.cross_entropy(
                 logits.permute(0, 2, 1),
                 step_targets,
                 ignore_index=_IGNORE_INDEX,
                 reduction="none",
             )
+
+            if getattr(self, "masked_embedding", None) is not None:
+                unreduced += self.masked_embedding.compute_centroid_loss(
+                    hidden_states=mtp_output,
+                    targets=step_targets,
+                    ignore_index=_IGNORE_INDEX,
+                )
             valid_count = (step_targets != _IGNORE_INDEX).sum()
             step_loss = weight * unreduced.sum() / valid_count.clamp(min=1)
             total_loss = total_loss + step_loss
--- a/src/speculators/models/mtp/core.py
+++ b/src/speculators/models/mtp/core.py
@@ -65,6 +65,7 @@ class MTPDraftModel(DraftVocabMixin, SpeculatorModel):
         "d2t",
         "masked_embedding.centroids.weight",
         "masked_embedding.token_ordering",
+        "masked_embedding.token_ordering_inv",
     ]
 
     t2d: torch.Tensor | None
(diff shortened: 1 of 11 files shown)


orestis-z (commented):
Design looks solid. The query-only cross-attention into verifier KV caches is the right approach for Gemma4's shared-cache architecture, and the kwargs propagation path through `Gemma2DecoderLayer.forward` → `self.self_attn` is confirmed working (the installed transformers 5.12.1 passes `**kwargs` through). Agree with @coderabbitai on the sliding-window mask gap (the eager attention path silently drops the `sliding_window` kwarg, so local layers see the full causal prefix during training) and on strengthening the attention tests to assert cache routing, not just output shape. Recommend approving once those are addressed.

shotsan:
Thanks for the detailed review @coderabbitai and @orestis-z! I've pushed a new commit addressing all the remaining edge cases and test gaps:

1. **Safer Gemma2 Imports:** Replaced the fragile `NameError` block in `base_components.py` with an explicit `HAS_GEMMA2` boolean flag to cleanly and safely guard the model registry.
2. **Sliding Window Mask:** `eager_attention_forward` drops the `sliding_window` parameter, so I've manually constructed a boolean sliding window mask using `torch.triu/tril` and applied it directly to `attention_mask` inside `QueryOnlyGemma2Attention.forward`.
3. **Cache Routing Unit Tests:** Rewrote the test cases in `test_mtp_gemma2.py` to mathematically prove the KV cache routing works. The tests now run three forward passes: a baseline pass, a pass where we scramble the inactive cache (asserting `torch.equal` true), and a pass where we scramble the active cache (asserting `torch.equal` false).
4. **Empty Sample Shape:** Updated `create_empty_sample` in `data.py` to accept `kv_feature_dim` and construct dummy KV tensors with shape `[0, feature_dim]`, passing the true feature dimension down directly from the collator fallback to prevent downstream shape crashes.

shotsan:
Thanks for the incredibly thorough review @coderabbitai! 

I have addressed all three of your non-blocking items:

1. **`test_verifier_kvs_mixed_batch` assertions:** That test is actually introduced in the Multi-Level LM Head PR (PR #767), and we already pushed a fix over in that PR to explicitly assert the exact zero-padded tensor shapes rather than relying on the `try/except KeyError`! Once PR #767 merges, those fixes will automatically propagate here.
2. **`token_ordering_inv` persistence:** Great call. I updated `src/speculators/models/mtp/head.py` to explicitly set `persistent=False` when registering the `token_ordering_inv` buffer, which should completely prevent any missing-key issues when loading older checkpoints.
3. **Sliding Window Mask Test:** I added a brand new test (`test_query_only_gemma2_attention_sliding_window_mask`) in `test_mtp_gemma2.py` that explicitly passes a full causal `attention_mask` into the forward pass. The test mathematically proves that when `sliding_window` is set (e.g. 2), queries only attend to their window boundary and completely ignore perturbations in the KV cache outside of that boundary.

I just pushed these updates to this branch. We should be totally good to go!

orestis-z (commented):
Both outstanding items from my previous review are addressed:

- **Sliding-window mask**: The eager attention path now explicitly enforces the window constraint on the causal mask, so local layers no longer silently see the full causal prefix.
- **Cache-routing tests**: Tests now verify that changing the active cache changes output while changing the inactive cache does not, and a dedicated sliding-window mask test confirms tokens outside the window are correctly masked.

The coderabbitai feedback (NameError guard, centroid validation, inverse caching, empty-sample KV shape) has also been addressed. LGTM. Recommend approving.

src/speculators/models/mtp/core.py
```diff
@@ -197,6 +211,8 @@ def forward(
                 attention_mask=causal_mask,
                 position_ids=step_pos_ids,
                 position_embeddings=step_pos_emb,
+                verifier_kv_last_local=kwargs.get("verifier_kv_last_local"),
+                verifier_kv_last_global=kwargs.get("verifier_kv_last_global"),
```
speculatorsbot, line 216: **Bug: verifier KV caches are not sliced to `valid_len`, causing a shape mismatch during attention.**

All other sequence-dimension tensors in the loop are sliced to `valid_len` (e.g., `step_hidden = current_hidden[:, :valid_len]` at L202, and the causal mask is created for `valid_len` at L191–198). The KV caches are passed through at full `seq_len`.

Inside `eager_attention_forward` the attention scores have shape `[batch, heads, valid_len, seq_len]` but the causal mask has shape `[batch, 1, valid_len, valid_len]`. The element-wise add crashes:

```
RuntimeError: The size of tensor a (seq_len) must match the size of
tensor b (valid_len) at non-singleton dimension 3
```

(I reproduced this locally with `seq_len=10, valid_len=6`.)

Fix: slice once before the loop and pass the sliced tensors:

```python
# before the for-loop
step_kv_local = kwargs.get("verifier_kv_last_local")
step_kv_global = kwargs.get("verifier_kv_last_global")
if step_kv_local is not None:
    step_kv_local = step_kv_local[:, :valid_len]
if step_kv_global is not None:
    step_kv_global = step_kv_global[:, :valid_len]
```

then pass `step_kv_local` / `step_kv_global` at L214–215 instead of re-reading from `kwargs`.

The unit tests in `test_mtp_gemma2.py` do not catch this because they feed KV tensors with the same `seq_len` as the query and never go through `MTPDraftModel.forward()` (which is where the slicing to `valid_len` happens).

speculatorsbot (commented):
The query-only cross-attention design for Gemma4 MTP is clean, and the kwargs propagation path through `Gemma2DecoderLayer.forward(**kwargs)` into `QueryOnlyGemma2Attention` is correct. One shape-mismatch bug in `core.py` will crash Gemma2 MTP training at runtime -- details in the inline comment.

PR #195, opened 2025-11-20
refactor loss mask calculation

shanjiaz:
Refactored loss mask calculation.

Added:
- view masked examples 
- for irregular text passages, give users options to randomly drop the last few turns and only keep the thinking tokens for last turn
- give users option to pass in a regex for loss mask calculation
- updated readmes
Example trainable tokens visualization:
Screenshot 2025-12-01 at 9 45 46 AM

PR #772, opened 2026-07-11
refactor(data): remove --turn-dropout

WindChimeRan:
## Purpose

`--turn-dropout` was added in #195 as a data-augmentation option: during preprocessing it randomly truncates each conversation to keep only the first N turns (N picked at random), so the model would train on conversations of varying lengths and "learn to generate continuations from various conversation contexts." It also had a reasoning-model angle: models like Qwen3 keep only the last turn's thinking tokens at inference (earlier turns' thinking is dropped), so randomly varying which turn is last let each turn's thinking get trained at least some of the time.

This PR removes it. That goal isn't served on either data path today, and after #729 the flag is dead code on the on-policy path.

**On-policy (regen).** #729 made regen emit pre-tokenized rows (`input_ids` + a generation-boundary `loss_mask`) that skip normalization via the passthrough in `_preprocess_batch`, so turn-dropout never runs — `build_eagle3_dataset` already logs `turn_dropout does not apply to pre-tokenized rows; ignoring`.

**Off-policy (chat template + mask).** On this path every assistant turn is already trained on. Turn-dropout just cuts a conversation short, keeping the first few turns and dropping the rest. But because the model only looks at earlier tokens, a kept turn is trained the same way whether or not later turns follow it — so the full conversation already covers everything the kept turns would, and dropping the rest only removes training signal, never adds any.

Net: removing it changes no real training data. Deletes the flag, the `turn_dropout` parameter and its now-dead code across preprocessing, the turn-dropout test, and two doc references.

**Follow-up.** The thinking-token case deserves a real fix rather than turn-dropout's random workaround: keep (uncompress) thinking tokens in every turn — as trl and llama-factory do for multi-turn SFT — instead of dropping them for turns `:n-1` the way inference does. We plan a separate PR for this, which also lets a multi-turn conversation train as a single row rather than fanning out to one row per assistant turn.

## Tests

- Unit/integration: `pytest -k normalize_conversation` (9) and the pre-tokenized passthrough tests (3) pass; `ruff` clean.
- Full offline e2e on `Qwen/Qwen3-0.6B` (prepare_data → vLLM hidden states → train → acceptance) passes — training consumes the prepared `input_ids`/`loss_mask` and produces a checkpoint.

## Checklist

I have filled in:

---

This is a followup PR of the sig-spec-decode meeting. @fynnsu 

cc code owner: @shanjiaz

### DIFF
--- a/docs/cli/prepare_data.md
+++ b/docs/cli/prepare_data.md
@@ -44,8 +44,6 @@ python scripts/prepare_data.py \
 
 - **`--assistant-pattern`** (str, default: `None`) Custom regex pattern for matching assistant responses. If not provided, auto-detected from chat template.
 
-- **`--turn-dropout`** (flag) Enable turn dropout: randomly keeps first N consecutive turns per conversation for data augmentation.
-
 - **`--minimum-valid-tokens`** (int, default: `None`) Drop samples whose loss mask contains fewer than this many trainable tokens.
 
 ### Output Arguments
@@ -70,6 +68,5 @@ python scripts/prepare_data.py \
   --output ./prepared_data \
   --seq-length 4096 \
   --max-samples 10000 \
-  --turn-dropout \
   --num-preprocessing-workers 16
 ```
--- a/scripts/prepare_data.py
+++ b/scripts/prepare_data.py
@@ -133,14 +133,6 @@ def parse_args():
             "If not provided, auto-detected from chat template."
         ),
     )
-    parser.add_argument(
-        "--turn-dropout",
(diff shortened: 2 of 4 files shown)


orestis-z (commented):
LGTM. Clean removal of dead code with thorough cleanup across CLI, core logic, docs, and tests. Removal is complete — no stale references remain. Recommend approving.

Roderick-Wu (approved):

rahul-tuli (approved):

PR #773, opened 2026-07-11
feat(dflash): add Domino causal correction head (projector_type=domino)

### LINKED ISSUE #664: [RFC]: Add Domino support as a DFlash-based speculator

orestis-z:
## Purpose

Adds the **Domino** causal correction head as a `projector_type` sub-mode of the existing DFlash draft model (closes #664, based on training-side work in PR #685).

Domino (arXiv 2605.29707) extends DFlash's parallel draft backbone with a lightweight GRU + low-rank MLP head (~56M params) that refines base logits using causal state from previous draft tokens — entirely in logit space, without extra forward passes. Architecture verified against the paper and the [SpecForge reference implementation](https://github.com/sgl-project/SpecForge/pull/571).

Corresponding PR https://github.com/vllm-project/vllm/pull/48241

### Architecture

- `DominoHead` module (`prefix_gru` + `embed_proj`) wired into `DFlashDraftModel`
- Dual-branch training objective: `loss = (1 - λ) * final_loss + λ * base_loss` with linear λ decay schedule
- **Accept-Until-Fail (AUF)** loss truncation (`--auf`, arXiv 2607.01893): zeros the loss support after the first prediction error per block, aligning training with the verifier's prefix-acceptance semantics. Works with both plain DFlash and Domino (for Domino, only the base branch is truncated; the final branch keeps the full mask).
- **Decay-weighted loss normalization** (`--normalize-loss-by-decay`): divides the loss by the sum of decay weights instead of the token count, producing a proper weighted average that stabilizes loss magnitude across different gamma values. Matches SGLang's DFlash loss normalization.

### New CLI flags

| Flag | Section | Default | Description |
|------|---------|---------|-------------|
| `--auf` | DFlash | off | Enable AUF truncation (arXiv 2607.01893) — works with both DFlash and Domino |
| `--normalize-loss-by-decay` | DFlash | off | Normalize loss by sum of decay weights instead of token count |
| `--projector-type` | Domino | `dflash` | `domino` to enable the correction head |
| `--domino-gru-hidden-dim` | Domino | `1024` | GRU hidden dimension |
| `--domino-emb-dim` | Domino | `256` | MLP bottleneck dimension |
| `--domino-pure-draft-prefix-len` | Domino | `1` | Positions using pure base logits before GRU correction |
| `--domino-lambda-start` | Domino | `1.0` | Initial base-loss weight λ |
| `--domino-lambda-decay-ratio` | Domino | `1.0` | Fraction of total steps over which λ decays to 0 |

## Tests

```bash
make quality
uv run pytest tests/unit/models/test_domino_head.py -v
uv run pytest tests/unit/models/test_dflash_metrics.py -v
uv run pytest tests/integration/models/test_model_forward.py -k domino -v
```

## Checklist

I have filled in:


orestis-z (commented):
Design looks solid — the Domino head architecture (GRU + bottleneck MLP with additive logit deltas) and the dual-branch loss schedule match the paper (arXiv 2605.29707). The target-shift logic for enabling position-0 loss, OOB clamping, FSDP wrapping, and the AUF mask are all correctly implemented.

The main issues are in the test suite: several parameter names appear to be leftovers from a prior refactor (`lambda_base_decay_steps` → `lambda_base_decay_ratio`, `shift_targets`/`shift_label` → removed). These cause tests to either fail or silently test no-ops.

tests/e2e/smoke/test_offline_training.py
```diff
@@ -41,6 +41,24 @@
...
+            [
+                "--block-size",
+                "8",
+                "--max-anchors",
+                "256",
+                "--num-layers",
+                "3",
+                "--projector-type",
+                "domino",
+                "--domino-lambda-decay-steps",
```
orestis-z: **Bug: non-existent CLI arg will fail argparse.**

`--domino-lambda-decay-steps` doesn't exist — the CLI defines `--domino-lambda-decay-ratio` (a float 0–1 representing the fraction of total steps). This will cause an `unrecognized arguments` error.

```suggestion
                "--domino-lambda-decay-ratio",
```

The value on the next line (`"100"`) would also need to change to a valid ratio, e.g. `"1.0"`.

tests/integration/models/test_model_forward.py
```diff
@@ -309,6 +315,60 @@ def test_attention_backends_match(self, seq_lengths):
...
+            (1.0, 100),
+            (0.5, 0),
+            (0.0, 0),
+        ],
+    )
+    def test_lambda_schedule_variants(self, lambda_base_start, lambda_base_decay_steps):
+        model = make_dflash_model(
+            projector_type="domino",
+            lambda_base_start=lambda_base_start,
+            lambda_base_decay_steps=lambda_base_decay_steps,
```
orestis-z: **Bug: `lambda_base_decay_steps` is not a config field — test is a no-op.**

`DFlashSpeculatorConfig` has `lambda_base_decay_ratio` (float 0–1), not `lambda_base_decay_steps`. Since the config has `extra="allow"` (inherited from `SpeculatorModelConfig`), the unknown field is silently accepted and stored, but `lambda_base_decay_ratio` retains its default value of `1.0` for every parametrize variant. The test thinks it's exercising different decay schedules but it's always running with `decay_ratio=1.0`.

Rename to `lambda_base_decay_ratio` and adjust the parametrize values to valid ratios (floats in [0, 1]).

tests/unit/models/test_domino_head.py
```diff
@@ -0,0 +1,545 @@
...
+        data = _make_synthetic_data(seq_len=32, loss_mask_all=False)
+        loss_mask = data["loss_mask"]
+
+        _, _, shifted_targets, _, shifted_anchored_idx = model._backbone_forward(
+            data["hidden_states"],
+            data["input_ids"],
+            loss_mask,
+            data["verifier_last_hidden_states"],
+            data["document_ids"],
+            shift_targets=True,
```
orestis-z: **Bug: `shift_targets` parameter doesn't exist — `TestBackboneTargetAlignment` tests are no-ops.**

`_backbone_forward` has no `shift_targets` parameter; it goes into `**kwargs` and is never read. Both `shift_targets=True` and `shift_targets=False` produce identical results, so `test_shift_targets_true_shifts_by_one`, `test_shift_targets_false_no_shift`, and `test_shift_targets_true_oob_clamped_and_masked` all pass vacuously.

The target shift is unconditionally applied when `self.config.projector_type == "domino"` ([`core.py:424`](https://github.com/vllm-project/speculators/blob/5f1069ea0c76c33e014ef7133177b5a143959027/src/speculators/models/dflash/core.py#L424)), so there's no runtime toggle to test. Consider restructuring these tests to verify the shift behavior directly (e.g., assert that Domino targets differ from DFlash targets by exactly one position), rather than passing a phantom parameter.

tests/unit/models/test_domino_head.py
```diff
@@ -0,0 +1,545 @@
...
+# ---------------------------------------------------------------------------
+# Domino loss mask tests
+# ---------------------------------------------------------------------------
+
+
+@requires_cuda
+class TestDominoLossMask:
+    def test_position_zero_in_shift_label_true(self):
+        torch.manual_seed(42)
+        model = _make_tiny_domino_model(block_size=4, max_anchors=4, shift_label=True)
```
orestis-z: **Bug: `shift_label` is not a `DFlashSpeculatorConfig` field — `TestDominoLossMask` tests are no-ops.**

Same issue as `shift_targets`: `shift_label` is silently accepted as an extra config attribute but never read by any code path. Both `test_position_zero_in_shift_label_true` and `test_position_zero_excluded_when_shift_label_false` produce identical results regardless of the `shift_label` value passed.

The position-0 loss mask behavior is unconditionally determined by `projector_type == "domino"` — there's no separate `shift_label` toggle. If these tests are meant to verify the Domino vs DFlash masking difference, compare a `projector_type="domino"` model against a `projector_type="dflash"` model instead.

src/speculators/models/dflash/core.py
```diff
@@ -215,11 +238,79 @@ def get_trainer_kwargs(**kwargs) -> tuple[dict, dict]:
...
             "gamma": gamma,
             "max_anchors": max_anchors,
+            "normalize_by_decay": normalize_by_decay,
             "per_position_loss_weight": per_position_loss_weight,
             "dpace_alpha": dpace_alpha,
         }
         return dict(shared), dict(shared)
 
+    @staticmethod
+    def _auf_mask(
```
orestis-z, line 278: **Missing test coverage for `_auf_mask` and `--domino-auf`.**

The `_auf_mask` method has non-trivial logic (cumsum-based prefix acceptance, breaker-token inclusion, `j_star` computation) and there are zero tests exercising it — no unit test, no integration test, no e2e test with `domino_auf=True`. `grep -rn auf tests/` returns nothing.

Suggested coverage:
- A unit test with known `logits` and `targets` where errors occur at specific positions, verifying the mask keeps the accepted prefix + breaker and zeros the suffix.
- A case where all predictions are correct (j* = block_size, full mask retained).
- A case where the first position is wrong (j* = 0, only position 0 retained).
- An integration test passing `domino_auf=True` through the full `forward()` and verifying `j_star_sum` appears in metrics.

orestis-z (commented):
Previous review findings (stale `shift_targets` / `lambda_base_decay_steps` params in tests) have been addressed — the test suite now exercises the actual config fields.

Design and implementation are solid. The DominoHead architecture (GRU + low-rank MLP additive logit deltas) matches the paper (arXiv 2605.29707, §3). The dual-branch loss schedule `(1-λ)·L_final + λ·L_base`, the target-shift alignment that makes position-0 a real prediction, OOB clamping/masking, and the `domino_loss_decay` (position-0 gets weight 1.0 instead of DFlash’s 0) are all correctly implemented. The FSDP wrapping of the Domino head is correctly placed before the root `fully_shard`.

One test coverage gap below.

**Minor notes (non-blocking):**

- `fully_shard(model, mp_policy=mp_policy)` in `distributed.py:215` — the original code had `fully_shard(model)` without `mp_policy` on the root shard. Adding `mp_policy` is arguably more correct (root-level params like embeddings/norms should get the same mixed-precision treatment), but this is a behavioral change for **all** model types, not just Domino. Worth a sanity check that existing Eagle3/DFlash/DSpark training isn’t affected.

- Several doc files (CONTRIBUTING.md, branding.md, decision_guide.md, etc.) have multi-line markdown tables collapsed to single lines by mdformat, which will break rendering in both GitHub and mkdocs. The admonition syntax (`!!! warning ...`) is also collapsed. This appears to be a formatter config issue — might want to exclude `docs/` from mdformat or use a table-aware plugin.

docs/user_guide/tutorials/train_dflash_online.md
```diff
@@ -167,6 +167,43 @@ CUDA_VISIBLE_DEVICES=2,3 torchrun \
...
+  --epochs 5 \
+  --lr 1e-4
+```
+
+**Additional Domino-specific parameters:**
+
+- `--projector-type domino` — Enable the Domino correction head (default: `dflash`)
+- `--domino-gru-hidden-dim 1024` — Hidden dimension for the GRU
+- `--domino-emb-dim 256` — Bottleneck dimension for the projection MLP
+- `--domino-pure-draft-prefix-len 1` — Number of suffix positions left uncorrected
```
orestis-z: **Inverted description.** This says "Number of suffix positions left uncorrected" but `pure_draft_prefix_len` is the number of **prefix** (leading) positions that use pure DFlash logits *without* correction. The suffix positions *are* corrected.

The [CLI docs](https://github.com/vllm-project/speculators/blob/89be94b0c601dd60683ef052e4571ef32e3bedf2/docs/cli/train.md#L168) correctly say "Number of leading block positions (after the anchor) that use pure DFlash logits without Domino correction."

Suggested fix:
```suggestion
- `--domino-pure-draft-prefix-len 1` — Number of leading positions using pure DFlash without correction
```

docs/cli/train.md
```diff
@@ -160,6 +160,20 @@ torchrun --standalone --nproc_per_node=4 scripts/train.py \
...
+
+- **`--domino-gru-hidden-dim`** (int, default: `1024`) Hidden dimension for the Domino GRU correction head.
+
+- **`--domino-emb-dim`** (int, default: `256`) Bottleneck dimension for the Domino embed projection MLP that combines GRU output with backbone hidden states.
+
+- **`--domino-pure-draft-prefix-len`** (int, default: `1`) Number of leading block positions (after the anchor) that use pure DFlash logits without Domino correction. The suffix positions are refined by the Domino head.
+
+- **`--domino-lambda-start`** (float, default: `1.0`) Initial weight of the base loss in the Domino dual-loss schedule. The combined loss is `(1-λ) * final_loss + λ * base_loss`.
+
+- **`--domino-lambda-decay-ratio`** (float, default: `1.0`) Fraction of total training steps over which λ decays from `--domino-lambda-start` to 0. E.g. `1.0` means a full linear decay over the entire run. Set to `0` to disable decay.
```
orestis-z, line 195: **Missing `--domino-auf` doc entry.** The `--domino-auf` flag is defined in [`scripts/train.py`](https://github.com/vllm-project/speculators/blob/89be94b0c601dd60683ef052e4571ef32e3bedf2/scripts/train.py#L1056-L1063) but isn't listed in this section. Worth adding a line here for completeness, e.g.:

```
- **`--domino-auf`** (flag, default: off) Enable Accept-Until-Fail (B-AUF+D) truncation on the Domino base branch loss (arXiv 2607.01893). Zeros positions after the first greedy prediction error in each block.
```

orestis-z (commented):
Previous findings (stale test params, missing AUF coverage, mdformat docs) are all addressed by the three new commits.

Design and implementation remain solid. The DominoHead architecture, dual-branch loss schedule, target-shift alignment, OOB masking, AUF truncation, and FSDP wrapping are all correctly implemented. The new `_auf_mask` tests and `domino_auf` integration tests provide thorough coverage of the AUF feature.

Two doc issues below.

**Minor non-blocking observations (not posted inline):**

- `--normalize-loss-by-decay` and its plumbing through DFlash + DSpark are included in this PR but not mentioned in the description. If this is an intentional companion feature, a brief note in the PR body would help reviewers understand the scope.
- Domino metrics don't include `eal_sum`/`eal_total` (they're filtered out by the `startswith("position_")` comprehension in `_compute_domino_metrics`). Per-position accuracy is available so EAL can be derived, but if you want it logged directly, the fix is a one-liner.

orestis-z (commented):
Architecture and core logic look solid. The Domino head, dual-loss schedule, target shift, AUF mask, and FSDP wrapping are all correctly implemented and align with arXiv 2605.29707. The earlier self-review items (phantom test params, doc fixes) appear addressed in subsequent commits.

Two non-blocking observations:

1. **Scope**: The PR bundles `--normalize-loss-by-decay` (threaded through DFlash, DSpark, and Domino metrics, ~80 lines + 3 tests) without mentioning it in the title or description. Consider splitting it into its own PR or at least documenting it here before marking ready for review.

2. **`distributed.py`**: The root `fully_shard(model)` call gained `mp_policy=mp_policy` — this adds `reduce_dtype=float32` to root-level parameters (fc, norms, lm_head) for all FSDP-wrapped models, not just Domino. Probably an improvement (consistent with the per-layer policy), but worth noting as a behavioral change for existing DFlash/DSpark distributed training.

Recommend approving once the `quality-failed` label is resolved and the normalize-loss-by-decay scope is clarified.

orestis-z (commented):
Two new commits since last review — both clean.

**`8c62174` (EAL metric fix):** The `eal_sum`/`eal_total` metrics are now correctly propagated from `final_metrics` into the Domino metrics dict. The `@torch.compiler.disable` on `_compute_domino_metrics` is appropriate — the method's Python control flow (AUF conditional, lambda schedule) can't compile, and the DFlash path in `forward` compiles fine since its AUF branch is all tensor ops with config-based guards.

**`68287e3` (AUF generalization):** The `--domino-auf` → `--auf` rename and DFlash-path AUF are correctly implemented. In the DFlash path, `aligned_loss_mask` already zeros position 0, so AUF error detection starts at position 1 (the first draft position) — matching the paper's semantics. The `j_star` observability metrics are logged identically in both branches. Tests cover both Domino and DFlash AUF forward/backward and loss-difference assertions.

Previous outstanding items (normalize-loss-by-decay scope, `fully_shard` `mp_policy` behavioral change) are unchanged and already noted — not re-raising.

Recommend approving once `quality-failed` is resolved.

orestis-z:
## Microcycle Experiment Results (Qwen3-0.6B, 10k samples, 2 epochs)

Small-scale ablation comparing DFlash baseline vs Domino variants. Eval: vLLM offline inference (20 prompts, greedy, 256 max tokens).

### Results

| Config | Checkpoint | Acc Len | Δ |
|--------|-----------|---------|---|
| DFlash | [dflash](https://huggingface.co/orestis-z/dflash-qwen3-0.6b-microcycle-dflash) | 1.738 | — |
| DFlash + AUF | [dflash-auf](https://huggingface.co/orestis-z/dflash-qwen3-0.6b-microcycle-dflash-auf) | 1.695 | -0.042 |
| Domino + AUF | [domino-auf](https://huggingface.co/orestis-z/dflash-qwen3-0.6b-microcycle-domino-auf) | 1.812 | +0.074 |
| **Domino** | [domino](https://huggingface.co/orestis-z/dflash-qwen3-0.6b-microcycle-domino) | **1.864** | **+0.126** |
| Domino, prefix=0 | [domino-prefix0](https://huggingface.co/orestis-z/dflash-qwen3-0.6b-microcycle-domino-prefix0) | 1.858 | +0.121 |
| Domino, decay=0.5 | [domino-decay05](https://huggingface.co/orestis-z/dflash-qwen3-0.6b-microcycle-domino-decay05) | 1.718 | -0.019 |
| Domino+AUF, prefix=0 | [domino-auf-prefix0](https://huggingface.co/orestis-z/dflash-qwen3-0.6b-microcycle-domino-auf-prefix0) | 1.797 | +0.059 |

### DFlash vs Domino

DFlash vs Domino eval

### Training Plots

[TrackIO dashboard](https://huggingface.co/spaces/orestis-z/speculators-microcycle-0.6b)

### Key Observations

- **Domino is the best config** with +0.126 acceptance length over DFlash baseline (+7.3%)
- DFlash has higher acceptance at early positions (pos 0-1), but Domino surpasses it from pos 2 onward with increasingly larger gains at later positions — this tail advantage drives the overall acceptance length improvement
- AUF consistently hurts — both DFlash+AUF and Domino+AUF underperform their non-AUF counterparts
- `prefix_len=0` performs similarly to default (`prefix_len=1`)
- `decay_ratio=0.5` hurts significantly — faster base-loss decay degrades quality

cc @Eros483

jianuo-huang:
Thanks for the implementation. I would like to clarify the intended design and point out an issue in the current `shift_label=False` path.

1. **Why shifted labels are used**

   Shifted labels avoid wasting the anchor position. With unshifted labels, block position 0 contains the known anchor and must be excluded from the loss, so a block of size (B) produces only (B-1) predictions. With shifted labels, position 0 predicts the first draft token, allowing all (B) positions to be used.

2. **Meaning of `pure_draft_prefix_len`**

   `pure_draft_prefix_len` counts draft predictions, not raw block positions. With `pure_draft_prefix_len=1`, the first draft token uses pure DFlash logits because the anchor has already been consumed by the backbone. Domino correction starts from the second draft token, when the first realized draft token provides additional causal information.

3. **Issue in the unshifted path**

   When `shift_label=False`, prediction position (j) targets token (x_{a+j}), but `gru_out[j]` has already consumed that same token. Therefore, the correction head sees the token it is trained to predict, causing label leakage and a train–inference mismatch.

4. **Suggested fix**

   The unshifted path should use:

```python
suffix_start = 1 + pure_draft_prefix_len
domino_loss_mask[:, ::block_size] = 0
correction_state[:, j] = gru_out[:, j - 1]
```

I suggest either requiring `shift_label=True` for Domino, which matches the intended compute-efficient design, or implementing these separate offsets for the unshifted path.

A regression test should also verify that, in unshifted mode, changing token (j) cannot affect the correction logits used to predict token (j).

cc @orestis-z @Eros483

orestis-z:
## Full-Scale Experiment Results (Qwen3-8B, 120k ShareGPT, 6 epochs)

Full-scale training comparing DFlash baseline vs Domino on Qwen3-8B target model.

### Training Setup

| Parameter | Value |
|-----------|-------|
| Target model | Qwen/Qwen3-8B |
| Dataset | 120k ShareGPT samples |
| Epochs | 6 |
| Block size | 16 |
| Num layers | 3 |
| Target layer IDs | 2, 18, 33 |
| LR | 6e-4 (cosine, 4% warmup) |
| Loss | CE |
| Decay gamma | 7.0 |
| Max anchors | 256 |
| Total seq len | 6144 |

### Results

| Config | Checkpoint | EAL | Δ |
|--------|-----------|-----|---|
| DFlash | [dflash](https://huggingface.co/orestis-z/dflash-qwen3-8b-120k-dflash) | 1.900 | — |
| **Domino** | [domino](https://huggingface.co/orestis-z/dflash-qwen3-8b-120k-domino) | **2.292** | **+0.392 (+20.6%)** |

### Eval Setup

- **Method**: vLLM offline inference (`vllm.LLM`) with `method="dflash"`, `num_speculative_tokens=16`
- **Hardware**: Single A100-80GB (w-domino node)
- **Prompts**: 20 diverse prompts (general knowledge, coding, science, systems), greedy decoding, 256 max tokens
- **Draft rounds**: ~1,400 per config — delta CI [+0.28, +0.51], statistically significant but narrow domain coverage (no math/translation/tool-call prompts). Will follow up with guidellm multi-subset benchmark for broader validation.

### DFlash vs Domino

DFlash vs Domino eval

### Per-Position Acceptance Rates

| Pos | DFlash | Domino |
|-----|--------|--------|
| 0 | 100.0% | 100.0% |
| 1 | 51.3% | 56.5% |
| 2 | 22.1% | 30.6% |
| 3 | 9.7% | 18.0% |
| 4 | 4.4% | 10.0% |
| 5 | 1.2% | 5.9% |
| 6 | 0.7% | 3.5% |
| 7 | 0.3% | 1.7% |
| 8 | 0.2% | 1.2% |
| 9 | 0.1% | 0.8% |

### Key Observations

- **Domino delivers +20.6% EAL improvement over DFlash baseline** at full scale (Qwen3-8B, 120k samples) — substantially larger than the +7.3% seen in the 0.6B microcycle
- The improvement scales with model size: from +0.126 EAL on 0.6B to +0.392 on 8B
- Same pattern as microcycle: DFlash has comparable acceptance at early positions, but Domino's GRU correction increasingly dominates from pos 2 onward (2-5x better at pos 3-9)
- DFlash trained on H100s (~4.5h), Domino on A100s (~9h) — same hyperparameters, only `--projector-type domino` differs

orestis-z:
@jianuo-huang thanks for the feedback. The shift label option was set to check dflash inference with a domino trained checkpoint. Because I had observed an improvement using that previously. Since no benefit is observed with no shift + domino in this branch, I will revert to defaults per spec method (false for dflash + true for domino)

PR #774, opened 2026-07-11
refactor(data): derive off-policy loss masks from chat-template render boundaries (roadmap step 2 of #729)

WindChimeRan:
# Purpose

Step 2 of the "renderer is the single source of truth" roadmap (#729): off-policy loss masks now come from the **chat template's render boundary** — the same prompt/completion boundary the serving engine reports for on-policy data, reconstructed from the model's own template. The regex span-detection machinery and the `{% generation %}`-tag probing are deleted.

```
BEFORE (off-policy)                        AFTER (off-policy)
render whole conversation                  render prompt & full per assistant turn
auto-detect regex from template            boundary = where full extends prompt
char-offset span matching over text        mask = [0]*boundary + [1]*tail
(silent all-zero/EOS bugs)                 append-only template -> packed single row
                                           history-rewriting  -> per-turn fan-out
                                           unstable           -> loud error, no row
```

One mechanism, two boundary sources: on-policy rows get the boundary from the engine (#729), off-policy rows get the identical `(input_ids, loss_mask)` shape from the template. `_preprocess_batch` keeps a single consumption route.

## How the boundary is derived

For each assistant turn `j`: render `conv[:j]` with `add_generation_prompt=True` (context) and `conv[:j+1]` (full). The tokens the full render adds past the prompt are supervised. Two guarded relaxations, both discovered by real templates:

- **Scaffold fallback**: Qwen3.5's generation prompt pre-fills an empty `<think>` scaffold; with recorded reasoning the strict prefix check fails. The boundary falls back to the longest common prefix — exactly where generation starts on auto-opened thinking templates — guarded by the history render staying a token-prefix (otherwise: loud `BoundaryUnstableError`).
- **Emission routing**: if every turn's full render is a token-prefix of the next turn's prompt render (append-only template: Llama3, Qwen2.5), all spans are valid in one sequence → **packed single row**, same row count and cost as before. Thinking templates that rewrite history (Qwen3 strips past `<think>`) break the chain → **one row per turn**, each supervising its own thinking against the stripped-history context inference actually sees. The check is the router; no per-model registry anywhere.

## Why

- **The regex tier was the bug source.** On the ProcessorMixin path it regexes text decoded from *already-truncated* ids and requires a literal `<|im_end|>`: measured on dolphin-r1 × Qwen3.5-0.8B at `--seq-length 1024`, **94/94 truncating conversations produced all-zero masks** (long-reasoning samples silently lost, biasing against exactly the data speculative decoding needs). The boundary path recovers **93/94** (the 94th starts past the window — dropped loudly).
- **The turn terminator is finally supervised.** The regex captured content *between* markers, so `<|im_end|>`/`<|eot_id|>` was never trained — a draft that can't predict turn endings loses acceptance at every turn boundary. Boundary masks include the terminator by construction.
- **Reasoning models are supervised faithfully.** Per-turn fan-out supervises each turn's `<think>` against stripped history — matching inference and matching what on-policy regen (#729) produces. A single-row render structurally cannot do this: historical thinking isn't even in the render.
- **No `{% generation %}` tags needed.** TRL's tag approach requires a curated per-model template table (byte-identity matched, hard-fail otherwise) and deliberately diverges from inference on reasoning models. The boundary needs nothing from the template but prefix-stability, which is asserted per conversation, not assumed.

## Old vs new masks on real data (measured)

| model × dataset | route | old ⊂ new? | added | removed |
|---|---|---|---|---|
| Llama-3.1-8B × ultrachat (150) | packed ×150 | ✅ all 150 | +486 EOT | 0 |
| Qwen2.5-0.5B × ultrachat (150) | packed ×150 | ✅ all 150 | +486 EOT, +486 ws | 0 |
| Llama-3.1-8B × sharegpt (150) | packed ×150 | ✅ | +437 EOT | only turn-0-assistant* |
| Qwen3-0.6B × sharegpt (100) | packed 42 / fan-out 58 | ✅ | +EOT/ws | only turn-0-assistant* |
| Qwen3.5-0.8B × dolphin-r1 (60) | packed ×60 | ✅ all 60 | +EOT/ws | 0 |

\* ShareGPT dump artifacts: conversations *starting* with a gpt turn. A context-free assistant turn has no prompt to render a boundary against; it stays context for later turns, unsupervised (SpecForge drops these turns entirely).

Every supervised position the old path produced is preserved except those turn-0 artifacts; every addition is the turn terminator (+ ChatML's trailing newline). Token-frequency/vocab-mapping counts only supervised tokens, so the d2t/t2d mapping is unaffected on non-reasoning data.

## What changes (observable)

| | Before | After |
|---|---|---|
| Mask source | HF tag probe → regex auto-detect → char offsets | render boundary per assistant turn |
| `<|im_end|>` / `<|eot_id|>` | never supervised | supervised every turn |
| Truncated long responses | all-zero mask (ProcessorMixin: sample lost) | in-window prefix supervised; zero-supervised rows dropped with a warning |
| Reasoning multi-turn | single row, history thinking unsupervisable | one row per turn, faithful stripped-history context |
| Non-reasoning multi-turn | single row | single row (packed) — same cost |
| Unknown/unstable template | silent wrong or empty mask | `BoundaryUnstableError`, conversation skipped loudly |
| `--assistant-pattern` / `assistant_pattern=` | regex override | removed |
| Rows ending in user/tool turns | trailing messages kept (masked, wasted) | trailing non-assistant messages dropped |
| Pre-tokenized (regen) rows truncated to zero supervision | kept (zero gradient) | dropped with warning, same policy as rendered rows |
| Conversations yielding no rows | silent | counted and warned per batch |

## Tests

- `test_boundary_masks.py` (18 tests): coverage matrix over Qwen2/2.5/3/3.5 (ungated), packed route (span == `content<|im_end|>` per turn), fan-out route (history thinking stripped from context, current thinking supervised), Qwen3.5 scaffold LCP fallback (both reasoning and plain), assistant-first skip, trailing-message drop, truncation partial supervision, `BoundaryUnstableError` fail-loud + batch skip, frozen real-data fixtures (sharegpt/ultrachat/hermes-fc tool-calls/dolphin-r1 reasoning with PROVENANCE).
- Existing suites adapted: 115 passed (datagen integration + regen passthrough + prepare_data + offline datagen). `ruff check`, `ruff format`, `mypy --check-untyped-defs` (158 files) all clean.
- e2e: `scripts/prepare_data.py` smoke over mixed fixtures — visualization shows turn-2 rows conditioning on think-stripped history with `<think>…</think>answer<|im_end|>` supervised; saved dataset schema unchanged (`input_ids`, `loss_mask`, `seq_len`).

## Scope — step 2 of 5 ("renderer is the single source of truth")

1. #729 (merged): on-policy → generation boundary from the engine.
2. **(this PR)** off-policy → the same boundary from the template; regex/tag machinery deleted.
3. next: linearize per-turn fan-out back to single rows where lossless (non-reasoning), reusing the packed-emission helper.

Costs called out: per-turn fan-out applies only to history-rewriting (thinking) templates — non-reasoning datasets keep one row per conversation via packing. Preprocessing tokenizes 2 renders per assistant turn (3 when the scaffold divergence engages); turns whose context already fills `--seq-length` are never rendered (measured ~3× saving on long-tailed multi-turn data, where up to 71% of render time otherwise lands past the window). CPU-side, parallelized by `datasets.map`, negligible next to hidden-state generation. Possible follow-up: cache the constant generation-prompt suffix on scaffold templates to skip the third render (~33% CPU on those runs).


WindChimeRan:
close because I will further simplify the design.

PR #775, opened 2026-07-12
Enable Ascend NPU training: opt-in transfer_to_npu + robust device_id

Sawyer117:
- scripts/train.py: when CUDA is unavailable, best-effort import torch_npu.contrib.transfer_to_npu so the trainer's torch.cuda.* perf timer + seeding run unmodified on Ascend. Guarded by torch.cuda.is_available() -> a strict no-op on CUDA/CPU.
- distributed.py: pass device_id as an explicit torch.device(acc.type, local_rank) to init_process_group instead of a bare int. torch>=2.x reads device_id.type; the explicit device is correct on CUDA and fixes backends that rewrite the int (e.g. NPU).

## Checklist

I have filled in:

### DIFF
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -1,5 +1,6 @@
 import argparse
 import gc
+import importlib.util
 import logging
 import random
 import warnings
@@ -8,6 +9,19 @@ from pathlib import Path
 
 import numpy as np
 import torch
+
+# Ascend NPU: torch.cuda.is_available() is False on Ascend, so import
+# transfer_to_npu to monkeypatch torch.cuda.* -> torch.npu.* (dist backend ->
+# hccl) and let the trainer's CUDA-assuming code run unchanged. Gated on torch_npu
+# actually being installed, so a CUDA-misconfigured or CPU-only host does NOT enter
+# the NPU path (no spurious NPU message) -- it falls through to normal accelerator
+# resolution. CUDA hosts skip this via the outer check.
+if not torch.cuda.is_available() and importlib.util.find_spec("torch_npu"):
+    try:
+        from torch_npu.contrib import transfer_to_npu  # noqa: F401
+    except ImportError as e:
+        logging.getLogger(__name__).warning("torch_npu present but failed to import: %s", e)
+
 import transformers
 from packaging import version
 from transformers import LlamaConfig, PretrainedConfig
--- a/src/speculators/train/distributed.py
+++ b/src/speculators/train/distributed.py
@@ -153,7 +153,9 @@ def maybe_setup_distributed(sp_size: int = 1) -> None:
     if acc is None:
         raise ValueError("No accelerator found")
     backend = torch.distributed.get_default_backend_for_device(acc)
-    dist.init_process_group(backend, device_id=local_rank)
+    # Explicit torch.device, not a bare int: torch>=2.x reads device_id.type, and
+    # some backends (Ascend transfer_to_npu) rewrite the int.
+    dist.init_process_group(backend, device_id=torch.device(acc.type, local_rank))
 
     _rank = dist.get_rank()
     _world_size = dist.get_world_size()


orestis-z (commented):
LGTM. Small, well-scoped changes that are backward-compatible on CUDA. The `transfer_to_npu` import is needed for `torch.cuda.manual_seed_all` and `torch.cuda.empty_cache` to work on NPU; the explicit `torch.device` in `init_process_group` is a good defensive improvement (note: PyTorch 2.11+ already auto-completes bare ints via `torch.device(acc.type, device_id)`, so this is more about clarity than fixing a current bug). Recommend approving.

Sawyer117:
Pushed a small follow-up (b30242d) that tightens the NPU shim import.
  
  Previously the transfer_to_npu import was gated only on not torch.cuda.is_available(). But that condition is also true on a CUDA-misconfigured GPU host or a CPU-only host — not just on
  Ascend — so those cases would enter the NPU path (and could surface an NPU-flavored message), which is misleading.

  It now also checks that torch_npu is actually installed (importlib.util.find_spec("torch_npu")):
 
  - Ascend host (torch_npu installed): shim loads as before.
  - CUDA-misconfigured / CPU-only host (no torch_npu): skips the NPU path entirely and falls through to the normal accelerator resolution / "no accelerator" error — no spurious NPU output.
  - torch_npu installed but failing to import (broken install): surfaced as a warning.

  The CUDA path remains a strict no-op.

fynnsu:
Is there a reason we need `transfer_to_npu`? We've generally tried to make the codebase accelerator agnostic and avoid direct calls to `torch.cuda`. 

It seems like a few have slipped through but these can be replaced by calls to:
- [torch.accelerator.synchronize](https://docs.pytorch.org/docs/2.13/generated/torch.accelerator.synchronize.html)
- [torch.accelerator.memory.empty_cache](https://docs.pytorch.org/docs/2.13/generated/torch.accelerator.memory.empty_cache.html)
- [torch.accelerator.is_available](https://docs.pytorch.org/docs/2.13/generated/torch.accelerator.is_available.html)

To fix: 
- 2 lines in `scripts/`: https://github.com/search?q=repo%3Avllm-project%2Fspeculators+cuda+path%3Ascripts%2F&type=code

- 2 lines in `src/`: https://github.com/search?q=repo%3Avllm-project%2Fspeculators+cuda+path%3Asrc%2F&type=code



I think this is a cleaner approach to solving this issue. It also avoids a bug with this line: 
https://github.com/vllm-project/speculators/blob/6aeb8864ae13728e11d61b431683940ea27a43d1/src/speculators/models/utils.py#L11

This line was specifically added to fix an issue where flex attention + compile was run on NPUs which didn't support this combination https://github.com/vllm-project/speculators/pull/232. If I understand correctly, `transfer_to_npu` would make this condition true, breaking that fix.

Sawyer117:
> Is there a reason we need `transfer_to_npu`? We've generally tried to make the codebase accelerator agnostic and avoid direct calls to `torch.cuda`.
> 
> It seems like a few have slipped through but these can be replaced by calls to:
> 
> * [torch.accelerator.synchronize](https://docs.pytorch.org/docs/2.13/generated/torch.accelerator.synchronize.html)
> * [torch.accelerator.memory.empty_cache](https://docs.pytorch.org/docs/2.13/generated/torch.accelerator.memory.empty_cache.html)
> * [torch.accelerator.is_available](https://docs.pytorch.org/docs/2.13/generated/torch.accelerator.is_available.html)
> 
> To fix:
> 
> * 2 lines in `scripts/`: https://github.com/search?q=repo%3Avllm-project%2Fspeculators+cuda+path%3Ascripts%2F&type=code
> * 2 lines in `src/`: https://github.com/search?q=repo%3Avllm-project%2Fspeculators+cuda+path%3Asrc%2F&type=code
> 
> I think this is a cleaner approach to solving this issue. It also avoids a bug with this line:
> 
> https://github.com/vllm-project/speculators/blob/6aeb8864ae13728e11d61b431683940ea27a43d1/src/speculators/models/utils.py#L11
> 
> This line was specifically added to fix an issue where flex attention + compile was run on NPUs which didn't support this combination #232. If I understand correctly, `transfer_to_npu` would make this condition true, breaking that fix.

Hi dear @fynnsu, I totally agree, this would be a very much cleaner approach. I will close this PR for now.

PR #776, opened 2026-07-12
train: --init-on-meta for large draft model construction

Sawyer117:
Adds build_on_meta(): a context manager that builds nn.Module parameters on the meta device (buffers stay real), so a large draft is not replicated in host RAM on every rank during construction. Wired into scripts/train.py via --init-on-meta: non-rank0 ranks build on meta and receive real weights via set_model_state_dict(broadcast_from_rank0=True) in FSDP setup; only rank 0 builds real weights. This is the torchtitan/torchtune large-model init pattern.

Opt-in and off by default. The base `load_verifier_weights` now no-ops on meta params, so the built-in speculators work under it through the shared load path (verified end-to-end for Eagle3 — see Tests). The general contract still holds for any *custom* draft: data-dependent init (weight load, or anything calling `.item()`/`.isnan()` on a param) must be guarded with `param.is_meta`.

  

  ## Purpose

  For large draft models, having every rank build the full model in host RAM during construction is wasteful — only rank 0's weights survive (the rest are overwritten by the rank-0 broadcast in FSDP setup). `--init-on-meta` lets non-rank0 ranks construct the draft on the `meta` device (≈0 resident memory); real weights then arrive via the trainer's existing `set_model_state_dict(broadcast_from_rank0=True)`. Pure memory optimization — the resulting weights are identical to a normal build.

  Changes:
  - `build_on_meta()` context manager (`speculators/train/distributed.py`): overrides `register_parameter` so params land on `meta`, buffers stay real, and restores the original on exit.
  - `--init-on-meta` flag (`scripts/train.py`, opt-in): wraps draft construction in `build_on_meta()` on non-rank0 ranks.
  - Base `SpeculatorModel.load_verifier_weights()` early-returns when params are on `meta`. Without this, `from_training_args` crashes calling `.isnan()` on a meta tensor (`Tensor.item()
  cannot be called on meta tensors`). Covers eagle3/dflash/dspark/peagle (which don't override the method); harmless no-op for mtp.

## Tests

  **Unit (CPU, runs in CI)** — `tests/unit/train/test_build_on_meta.py`:
  - params land on `meta` while buffers keep real storage/values;
  - the `register_parameter` monkeypatch is restored on exit (no global leak);
  - a real `Eagle3DraftModel.from_training_args(...)` builds under `build_on_meta` (all params meta / buffers real), exercising the `is_meta` guard so the verifier load no-ops instead of
  crashing.

  ```
  $ pytest tests/unit/train/test_build_on_meta.py -v
  3 passed
  ```

  **Multi-GPU (e2e, `@requires_multi_gpu` — auto-skips with <2 GPUs)** — `tests/e2e/smoke/test_init_on_meta.py`:
  - launches the trainer's real build + broadcast-materialize under `torchrun` and asserts non-rank0 ranks build on `meta`, then after `set_model_state_dict(broadcast_from_rank0=True)`
  every rank holds real, finite weights **bit-equal to rank 0** — i.e. `--init-on-meta` does not change the result. Validated on 2 and 8 GPUs.

  ```
  $ pytest tests/e2e/smoke/test_init_on_meta.py -v
  PASS (2 ranks): non-rank0 built on meta, all ranks materialized real+finite weights, and the gathered state matches rank0 exactly.
  1 passed
  ```
## Checklist

  I have filled in:

### DIFF
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -1,4 +1,5 @@
 import argparse
+import contextlib
 import gc
 import logging
 import random
@@ -28,6 +29,7 @@ from speculators.models.utils import (
 )
 from speculators.train.dataloader import create_train_val_loaders
 from speculators.train.distributed import (
+    build_on_meta,
     get_rank,
     maybe_destroy_distributed,
     maybe_setup_distributed,
@@ -422,12 +424,20 @@ def build_draft_model(
         )
 
     args.draft_vocab_size = draft_vocab_size
-    return model_class.from_training_args(
-        verifier_config=transformer_layer_config,
-        t2d=t2d,
-        d2t=d2t,
-        **vars(args),
+    # --init-on-meta: non-rank0 build params on meta (no N-way host-RAM copy); real
+    # weights arrive via broadcast_from_rank0 in FSDP setup. Only rank 0 builds real.
+    meta_ctx = (
+        build_on_meta()
+        if args.init_on_meta and get_rank() != 0
+        else contextlib.nullcontext()
     )
+    with meta_ctx:
+        return model_class.from_training_args(
+            verifier_config=transformer_layer_config,
+            t2d=t2d,
+            d2t=d2t,
+            **vars(args),
+        )
 
 
 def main(args: argparse.Namespace):  # noqa: C901
@@ -822,6 +832,15 @@ def parse_args():
     parser.add_argument("--log-dir", type=str, default="./logs")
     parser.add_argument("--run-name", type=str, default=None)
     parser.add_argument("--num-layers", type=int, default=1)
+    parser.add_argument(
+        "--init-on-meta",
+        action="store_true",
+        help=(
+            "Build the draft on meta on all ranks except rank 0 (no N-way host-RAM "
+            "copy); real weights broadcast from rank 0 in FSDP setup. Requires the "
+            "model to no-op data-init on meta params (guard param.is_meta)."
+        ),
+    )
     parser.add_argument(
         "--draft-arch",
         type=str,
--- a/src/speculators/model.py
+++ b/src/speculators/model.py
@@ -128,6 +128,13 @@ class DraftVocabMixin(nn.Module):
         is True. Subclasses can override to load additional weights (e.g. norms,
         tokenizer) by calling super().load_verifier_weights() first.
         """
+        # Meta-device init (--init-on-meta): params carry no data, so skip the
+        # verifier load; real weights arrive via set_model_state_dict(
+        # broadcast_from_rank0=True) in the trainer's FSDP setup. (isnan() below
+        # would raise on a meta tensor.)
+        if self.embed_tokens.weight.is_meta:
+            return
+
         import warnings  # noqa: PLC0415
 
         from speculators.utils.loading import load_model_layers  # noqa: PLC0415
--- a/src/speculators/model.py
+++ b/src/speculators/model.py
@@ -128,11 +128,16 @@ class DraftVocabMixin(nn.Module):
         is True. Subclasses can override to load additional weights (e.g. norms,
         tokenizer) by calling super().load_verifier_weights() first.
         """
-        # Meta-device init (--init-on-meta): params carry no data, so skip the
-        # verifier load; real weights arrive via set_model_state_dict(
-        # broadcast_from_rank0=True) in the trainer's FSDP setup. (isnan() below
-        # would raise on a meta tensor.)
+        # --init-on-meta: params carry no data, so skip the verifier load (isnan()
+        # below would raise on meta); real weights arrive via broadcast in FSDP setup.
+        # Still freeze here -- requires_grad_() works on meta, and the trainable-param
+        # set must match across ranks or FSDP2's grad reduce_scatter hangs.
         if self.embed_tokens.weight.is_meta:
+            self.embed_tokens.weight.requires_grad_(False)
+            self.lm_head.weight.requires_grad_(False)
+            self.verifier_lm_head.weight.requires_grad_(False)
+            if hasattr(self, "verifier_norm"):
+                self.verifier_norm.weight.requires_grad_(False)  # type: ignore[union-attr]
             return
 
         import warnings  # noqa: PLC0415
--- a/tests/unit/train/test_build_on_meta.py
+++ b/tests/unit/train/test_build_on_meta.py
@@ -44,16 +44,10 @@ def test_build_on_meta_restores_register_parameter():
 
 
 def test_from_training_args_under_build_on_meta(tmp_path):
-    """A real speculator builds fully under build_on_meta: all params land on meta
-    and buffers stay real. Exercises the base-class ``is_meta`` guard -- with meta
-    params, ``load_verifier_weights`` must early-return instead of calling
-    ``isnan()`` on a meta param (that raises "Tensor.item() cannot be called on
-    meta tensors").
-
-    ``from_training_args`` resolves the verifier's config.json up front (for
-    architectures via ``VerifierConfig.from_pretrained``), so point it at a local
-    dir to keep the build offline. The verifier *weights* are never read: the meta
-    guard returns first.
+    """A real speculator builds under build_on_meta: params land on meta, buffers stay
+    real, and the base-class is_meta guard makes load_verifier_weights early-return
+    (instead of calling isnan() on a meta param). A local verifier dir keeps config
+    resolution offline; the verifier weights themselves are never read.
     """
     copy.deepcopy(_TINY).save_pretrained(tmp_path)  # writes config.json only
 
--- a/tests/unit/train/test_build_on_meta.py
+++ b/tests/unit/train/test_build_on_meta.py
@@ -1,0 +1,63 @@
+"""Tests for build_on_meta (meta-device construction for --init-on-meta)."""
+
+import copy
+
+import torch
+from torch import nn
+from transformers.models.llama.configuration_llama import LlamaConfig
+
+from speculators.models.eagle3 import Eagle3DraftModel
+from speculators.train.distributed import build_on_meta
+
+_TINY = LlamaConfig(
+    vocab_size=64,
+    hidden_size=32,
+    intermediate_size=128,
+    num_hidden_layers=2,
+    num_attention_heads=4,
+    num_key_value_heads=4,
+    head_dim=8,
+    max_position_embeddings=32,
+    rms_norm_eps=1e-6,  # type: ignore[arg-type]
+    tie_word_embeddings=False,
+    _attn_implementation="eager",  # type: ignore[call-arg]
+)
+
+
+def test_build_on_meta_params_meta_buffers_real():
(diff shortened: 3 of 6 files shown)


scripts/train.py
```diff
@@ -422,12 +424,20 @@ def build_draft_model(
...
+        if args.init_on_meta and get_rank() != 0
+        else contextlib.nullcontext()
+    )
+    with meta_ctx:
+        return model_class.from_training_args(
+            verifier_config=transformer_layer_config,
+            t2d=t2d,
+            d2t=d2t,
+            **vars(args),
+        )
```
orestis-z, line 440: `from_training_args` internally calls `load_verifier_weights()`, which inspects parameter data -- e.g., [`self.embed_tokens.weight.isnan().any()`](https://github.com/vllm-project/speculators/blob/b41cd3c8b6131d868aea31447bbe37d7c3a118ab/src/speculators/model.py#L157-L159) in the base `SpeculatorModel`. On meta tensors, `isnan().any()` returns a meta scalar and the `if` statement raises:

```
RuntimeError: Tensor.item() cannot be called on meta tensors
```

(Verified experimentally on this environment's PyTorch.)

Every existing model class (Eagle3, DFlash, DSpark, MTP) triggers this crash because all their `from_training_args` implementations call `load_verifier_weights()`. MTP's override additionally calls [`self.embed_tokens.weight.fill_(torch.nan)`](https://github.com/vllm-project/speculators/blob/b41cd3c8b6131d868aea31447bbe37d7c3a118ab/src/speculators/models/mtp/core.py#L104-L106), which is silently a no-op on meta tensors (doesn't crash, but defeats the NaN-sentinel purpose).

The documented contract ("models must no-op data-init on meta params") would fix this, but no model class implements those guards yet, so `--init-on-meta` is unusable as-is.

Simplest fix: guard `load_verifier_weights` in the base class early-return when params are meta, since the broadcast provides all weights anyway:

```python
def load_verifier_weights(self):
    # Skip on meta-device init; real weights arrive via broadcast.
    if self.embed_tokens.weight.is_meta:
        return
    ...
```

And similarly in MTP's override. A unit test that builds a model under `build_on_meta()` and verifies parameters are on meta / buffers are real would also help lock this down.

orestis-z (commented):
Clean application of the torchtitan/torchtune meta-device init pattern. The `build_on_meta` context manager design is sound -- monkey-patching `register_parameter` correctly redirects parameters to meta while keeping buffers real, and the existing `set_model_state_dict(broadcast_from_rank0=True)` call in `Trainer.setup_model` already handles materialization on all ranks.

However, the feature will crash at runtime with every existing model class due to `load_verifier_weights()` being called on meta parameters inside the `build_on_meta` context -- see inline comment. The PR also has no tests for the new context manager or CLI flag, and the Purpose/Tests sections of the PR description are empty.

tests/unit/train/test_build_on_meta.py
```diff
@@ -0,0 +1,63 @@
...
+    with build_on_meta():
+        model = Eagle3DraftModel.from_training_args(
+            verifier_config=copy.deepcopy(_TINY),
+            t2d=None,
+            d2t=None,
+            draft_vocab_size=64,
+            norm_before_residual=False,
+            ttt_steps=1,
+            # never fetched: the meta guard returns before the verifier load
+            verifier_name_or_path="does-not-exist/verifier",
```
orestis-z: This test will crash with `OSError` before the `is_meta` guard is ever reached. `Eagle3DraftModel.from_training_args` calls [`resolve_target_layer_ids(None, "does-not-exist/verifier")`](https://github.com/vllm-project/speculators/blob/9c98fadda14d6610bff066f317ebc3244d25f01e/src/speculators/models/eagle3/core.py#L347-L349) first (since `target_layer_ids` is not passed), which runs `AutoConfig.from_pretrained("does-not-exist/verifier")` and raises:

```
OSError: does-not-exist/verifier is not a local folder and is not a valid model identifier
```

Even with `target_layer_ids` supplied, the same path is hit by [`VerifierConfig.from_pretrained`](https://github.com/vllm-project/speculators/blob/9c98fadda14d6610bff066f317ebc3244d25f01e/src/speculators/models/eagle3/core.py#L372-L373) during config construction, and again by [`AutoConfig.from_pretrained` in Eagle3's `load_verifier_weights` override](https://github.com/vllm-project/speculators/blob/9c98fadda14d6610bff066f317ebc3244d25f01e/src/speculators/models/eagle3/core.py#L149) (which runs after `super()` returns early).

Quickest fix: mock `AutoConfig.from_pretrained` and `PretrainedConfig.get_config_dict` to return the `_TINY` config, or construct the `Eagle3DraftModel` directly (bypassing `from_training_args`) to isolate just the meta-device + `load_verifier_weights` behavior.

orestis-z (commented):
The crash fix (`is_meta` guard in the base class) and the two unit tests (`test_build_on_meta_params_meta_buffers_real`, `test_build_on_meta_restores_register_parameter`) are solid. One inline comment on the integration test below.

orestis-z (commented):
Both prior findings addressed: the `is_meta` guard in the base class prevents the `isnan()` crash, and the unit test now uses a local verifier dir. The new e2e test (`init_on_meta_worker.py` under torchrun) thoroughly verifies the full build + FSDP-shard + broadcast-materialize path. LGTM. Recommend approving.

src/speculators/model.py
```diff
@@ -128,6 +128,13 @@ def load_verifier_weights(self):  # noqa: C901
         is True. Subclasses can override to load additional weights (e.g. norms,
         tokenizer) by calling super().load_verifier_weights() first.
         """
+        # Meta-device init (--init-on-meta): params carry no data, so skip the
+        # verifier load; real weights arrive via set_model_state_dict(
+        # broadcast_from_rank0=True) in the trainer's FSDP setup. (isnan() below
+        # would raise on a meta tensor.)
+        if self.embed_tokens.weight.is_meta:
```
orestis-z: The early return skips the `requires_grad_(False)` calls at the end of this method (lines 195-199 on main). On meta tensors, `requires_grad_()` works fine (it just sets a flag), so the freeze should still be applied. Without it, non-rank0 ranks keep `requires_grad=True` on `lm_head`, `verifier_lm_head`, and `verifier_norm`, while rank 0 has them frozen. In FSDP2, this produces a gradient-reduction collective mismatch: `post_backward` ([source](https://github.com/pytorch/pytorch/blob/main/torch/distributed/fsdp/_fully_shard/_fsdp_param_group.py)) gathers `fsdp_params_with_grad` per rank, and the resulting `reduce_scatter` buffer sizes differ across ranks.

Verified empirically:
```python
m = nn.Linear(4, 4, device="meta")
m.weight.requires_grad_(False)  # works on meta
assert not m.weight.requires_grad  # True
```

Suggested fix -- freeze verifier weights before the early return:
```python
if self.embed_tokens.weight.is_meta:
    self.embed_tokens.weight.requires_grad_(False)
    self.lm_head.weight.requires_grad_(False)
    self.verifier_lm_head.weight.requires_grad_(False)
    if hasattr(self, "verifier_norm"):
        self.verifier_norm.weight.requires_grad_(False)
    return
```

(Or refactor the freeze into a separate helper called unconditionally, so it is not duplicated.)

orestis-z (commented):
The `is_meta` early-return in the base class skips the `requires_grad_(False)` calls for verifier weights (`lm_head`, `verifier_lm_head`, `verifier_norm`). After `set_model_state_dict(broadcast_from_rank0=True)`, rank 0 has those weights frozen (`requires_grad=False`) while non-rank0 ranks have them trainable (`requires_grad=True`). This creates an FSDP2 gradient-reduction mismatch during training -- `post_backward` collects a different set of params into `fsdp_params_with_grad` on each rank, producing different-sized `reduce_scatter` buffers, which will hang. See inline comment for the concrete fix.

Sawyer117:
Good catch, and apologies — this was a genuine gap in our test coverage. Our e2e check only went as far as build + broadcast + weight comparison, so it never ran a backward pass and missed this hang entirely. It should have exercised real training from the start.

To close the gap: the e2e worker now also asserts `requires_grad` is consistent across ranks after the broadcast, and I validated end-to-end with a real 2-GPU `--init-on-meta` training run (Qwen3-0.6B, eagle3) — trains to completion, no hang, loss decreasing.

Thanks for the careful review.

orestis-z (commented):
LGTM. The `requires_grad_(False)` freeze in the meta early-return path matches the normal-path freeze exactly, and the new `all_gather`-based `requires_grad` assertion in the e2e worker locks this down. Recommend approving.

src/speculators/model.py
```diff
@@ -128,6 +128,18 @@ def load_verifier_weights(self):  # noqa: C901
...
+        # Still freeze here -- requires_grad_() works on meta, and the trainable-param
+        # set must match across ranks or FSDP2's grad reduce_scatter hangs.
+        if self.embed_tokens.weight.is_meta:
+            self.embed_tokens.weight.requires_grad_(False)
+            self.lm_head.weight.requires_grad_(False)
+            self.verifier_lm_head.weight.requires_grad_(False)
+            if hasattr(self, "verifier_norm"):
+                self.verifier_norm.weight.requires_grad_(False)  # type: ignore[union-attr]
+            return
+
```
Roderick-Wu, line 142: For some inherited classes like `Eagle3DraftModel` (under `speculators/models/eagle3/core.py`) more verifier code is run after returning early:
```python
    def load_verifier_weights(self):
        super().load_verifier_weights()

        self.embed_tokens.weight.requires_grad_(self.config.embed_requires_grad)

### more code ###
        verifier_config = self.config.speculators_config.verifier
        verifier_model_config = AutoConfig.from_pretrained(verifier_config.name_or_path)  # type: ignore[arg-type] 

        # For multimodal models (Qwen3VL, etc.), extract text_config
        if hasattr(verifier_model_config, "text_config"):
            verifier_model_config = verifier_model_config.text_config

        if verifier_model_config.hidden_size != self.hidden_size:
            raise ValueError(
                f"Verifier hidden size {verifier_model_config.hidden_size} does not"
                f" match draft hidden size {self.hidden_size}."
            )
```

we should add a guard after `super().load_verifier_weights()`? If we skip loading weights, none of the code running at the moment is dangerous, just some wasteful disk i/o. But I think there might be some future edits to `load_verifier_weights()` that could cause issues in the future?

speculatorsbot (commented):
Clean application of the torchtitan/torchtune meta-device init pattern. The `build_on_meta()` context manager correctly redirects `register_parameter` to meta while preserving buffers, the `is_meta` guard in the base `load_verifier_weights` prevents the `isnan()` crash and freezes verifier weights to avoid FSDP2 gradient-reduction mismatches, and the test coverage (unit + multi-GPU e2e) is thorough.

One open item before merging: @Roderick-Wu's [unresolved question](https://github.com/vllm-project/speculators/pull/776#discussion_r3579945757) about whether Eagle3's `load_verifier_weights` override should also guard against running post-`super()` code on the meta path (currently does unnecessary `AutoConfig.from_pretrained` disk I/O on non-rank0 ranks). LGTM once that discussion is resolved. Recommend approving.

PR #788, opened 2026-07-14
fix(loss): compute divergence losses in float32 under bf16 training

WindChimeRan:
## Purpose

The divergence losses (`kl_div`, `rkl`, `tv`, `nla`, `lk_hybrid`) take their softmax in the logits' dtype, which under our bf16 training (FSDP2 `param_dtype=bfloat16`, no autocast) means the whole loss runs in bf16.

Their gradient w.r.t. the draft logits reduces to **`p_draft - p_target`**. A well-trained draft has `p ≈ q`, so in bf16 (8-bit mantissa) that subtraction is a **catastrophic cancellation** — and the error *grows as training converges*, i.e. it is worst exactly where we want a clean signal.

Measured gradient error vs. an exact float64 reference, feeding **identical bf16 logits** to both paths so the only variable is the compute dtype (V=32k):

| acceptance α | `kl_div` (default) | `tv` | `ce` (control) |
|---|---|---|---|
| 0.88 | 3.0% | 5.7% | 0.14% |
| 0.95 | 9.0% | 7.1% | 0.13% |
| 0.98 | 19.4% | 31.1% | 0.11% |

`ce_loss` is immune and is left unchanged — its gradient is `p - onehot`, which never cancels. Computing the softmax in fp32 drives the error to a flat **~0.15%** (the bf16 gradient-storage floor) at every α.

The error is unbiased noise rather than systematic bias (best-fit gain 0.999), so this is not "training is broken" — but it is 10-20% of gradient magnitude thrown away for free, and every comparable library upcasts here:

- **HF transformers** — `logits = logits.float()` in every CausalLM loss (`loss/loss_utils.py:59`), commented *"Upcast to float ... to avoid potential precision issues"*
- **SpecForge** — `logits.float()` in the eager loss, `.cast(tl.float32)` in the Triton kernel
- **LLaMA-Factory** — `logits.float()` at three sites in `train/trainer_utils.py`
- **unsloth** — fused CE casts to `tl.float32` on load
- **TRL** — `selective_log_softmax` branches on dtype: *"logsumexp approach is unstable with bfloat16"*

Complementary to #765 (fp32 master weights), which fixes bf16 rounding at `optimizer.step()` and explicitly leaves forward/backward in bf16. Same disease, different organ.

## End-to-end drafter quality

Because the fix corrects an *unbiased-noise* error in the gradient (not a bias), it does **not** move the training loss — bf16 and fp32 loss curves overlap. The effect surfaces in the drafter's **acceptance / EAL**, not the loss.

Both refs continue-trained from the published [`RedHatAI/Qwen3-8B-speculator.eagle3`](https://huggingface.co/RedHatAI/Qwen3-8B-speculator.eagle3) on 18k on-policy ultrachat samples (`tv` loss, 15 epochs, lr 5e-5, seed 42; the two arms differ **only** by the softmax dtype):

| metric | base (bf16) | head (fp32, #788) |
| --- | --- | --- |
| `cond_acc` (d0 / d1 / d2) | 0.388 / 0.337 / 0.315 | 0.406 / 0.356 / 0.346 |
| **EAL** | **0.559** | **0.601 (+7.5%)** |
| `loss_epoch` | 1.997 | 1.951 |
| peak HBM / GPU | 16.3 GB | 16.3 GB |

fp32 is higher at every depth (not a single-depth artifact), compounding to **+7.5% EAL**. 

## Tests

`TestBf16GradientPrecision` feeds identical bf16 logits to each loss and to an exact float64
reference, and asserts the gradients agree within 2%. On `main` it fails for every changed loss:

| | kl_div | rkl | tv | nla | lk_hybrid |
|---|---|---|---|---|---|
| bf16 gradient error on `main` | 8.1% | 9.3% | 23.2% | 23.1% | 23.0% |
| with this change | <0.2% | <0.2% | <0.2% | <0.2% | <0.2% |

`tests/unit`: 436 passed, 3 skipped. ruff clean.

## Checklist

### DIFF
--- a/src/speculators/models/metrics.py
+++ b/src/speculators/models/metrics.py
@@ -1,3 +1,13 @@
+"""Metrics and loss functions shared across draft-model architectures.
+
+The divergence losses take their softmax in float32 (``dtype=torch.float32``)
+even under bf16 training. Their gradient w.r.t. the draft logits reduces to
+``p_draft - p_target``; once the draft is well trained the two are nearly equal,
+so in bf16 (8-bit mantissa) that subtraction is a catastrophic cancellation and
+the gradient error grows as training succeeds. ``ce_loss`` needs no upcast --
+its gradient is ``p - onehot``, which never cancels.
+"""
+
 import json
 import math
 from collections.abc import Callable
@@ -90,8 +100,8 @@ def kl_div_loss(
     Returns:
         Per-position KL divergence with shape [1, seq_len].
     """
-    logits = torch.nn.functional.log_softmax(logits, dim=-1)
-    target_p = torch.nn.functional.softmax(targets, dim=-1)
+    logits = torch.nn.functional.log_softmax(logits, dim=-1, dtype=torch.float32)
+    target_p = torch.nn.functional.softmax(targets, dim=-1, dtype=torch.float32)
     elementwise_loss = torch.nn.functional.kl_div(
         logits, target_p, reduction="none", log_target=False
     ).sum(dim=-1)  # shape: [1, seq_len]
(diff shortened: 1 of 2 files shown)


orestis-z (commented):
LGTM. The catastrophic cancellation analysis is solid, the fix uses PyTorch's canonical `dtype` parameter on `softmax`/`log_softmax` (consistent with HF Transformers, SpecForge, TRL, etc.), and the regression test is well-designed as a guard that breaks if the upcast is removed. Recommend approving.

shanjiaz (approved):
Thanks for running the ablation! Will let Fynn know.

PR #790, opened 2026-07-14
refactor(tests): extract pipeline runners into scripts/

WindChimeRan:
## Purpose

The e2e tests hold the runners for our pipeline scripts — data prep, vLLM launch, hidden-state generation, training, stitching. They live under tests, so nothing else can use them.

I'm building a mini-training A/B harness next. It needs to run the same pipeline and read the acceptance numbers back. It can't today: the runners aren't importable from outside tests, and the vLLM runner asserts its results instead of returning them.

## What changed

One behavior change: `run_vllm_engine` now returns its metrics instead of asserting them. The assertions move to a thin wrapper, `run_vllm_engine_and_assert`, so the tests behave the same. This is the reason for the PR.

Everything else is a move, shown as two renames:

- `tests/e2e/utils.py` → `scripts/pipeline_runners.py` — the runners
- `tests/e2e/run_vllm.py` → `scripts/run_vllm.py` — the standalone vLLM CLI

The test-only helpers (`purge_newfiles`, `setup_dummy_sharegpt4v_coco`, and the new `run_vllm_engine_and_assert`) move to `tests/e2e/conftest.py`. Nothing in `src/` changes.

## Why scripts, not the package

The runners shell out to the scripts folder, so they can't run from an installed wheel. Putting them in the package would also publish them as public API. Next to the scripts they drive, they just work, and both the tests and the new harness import them directly.

## Tests

- unit: 458 passed, 3 skipped
- make quality: clean
- e2e, offline eagle3 Qwen3-0.6B smoke test on one GPU: passed in 2:46. This runs the full pipeline — prepare, vLLM launch, hidden-state generation, training, inference — so it exercises every moved runner and the changed vLLM runner end to end.

CI runs unit and integration, not e2e, so the rest of the e2e suite does not run on this PR.

## Follow-up

Some tests still call the scripts by hand. For example:

```python
# a test today:
subprocess.run([sys.executable, "scripts/stitch_mtp.py", ckpt, verifier, "--output-path", out], ...)

# could just be:
run_stitch_mtp(ckpt, verifier, out)
```

These work and are untouched here, so nothing is broken. Switching them to the runners is an easy win. It is a separate PR because one of them needs a new distributed-training option first, and they are GPU tests.

## Checklist

I have filled in:

### DIFF
--- a/scripts/pipeline_runners.py
+++ b/scripts/pipeline_runners.py
@@ -1,3 +1,11 @@
+"""Subprocess runners for the sibling pipeline scripts.
+
+Each runner wraps one stage — data preparation, vLLM server launch, offline
+hidden-state generation, training, MTP stitching, and vLLM inference — building
+the argv and raising on a non-zero exit. Shared by the e2e tests and by sibling
+scripts, so the invocations are kept correct in one place.
+"""
+
 import json
 import os
 import subprocess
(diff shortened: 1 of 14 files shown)


WindChimeRan:
close as not planned.

PR #794, opened 2026-07-15
[Data] Derive off-policy loss masks from vLLM render boundaries

WindChimeRan:
# Purpose

Derive the **off-policy** training loss mask from a **render boundary**, with the vLLM instance as the single tokenization source. For each assistant turn we render the prompt and the prompt-plus-turn via `/v1/chat/completions/render`, and supervise the tokens the second render adds past the first.

**Then we deprecate the fallback hf tokenization & fallback regex paths**

This is the same prompt/completion boundary the engine already reports for on-policy regeneration (#729), reconstructed from the chat template. One tokenizer now produces the loss mask, the target hidden states, and the serving prompt — no separate local tokenization that can drift from serving.

Design write-up (worked examples, rationale): https://claude.ai/code/artifact/6cd3bf57-3d20-4a06-b27d-5b08e5fce6b4

# What changes

- **Deleted:** the two-tier local masking path — the HF `{% generation %}` tag-mask probe and the regex span-detector — plus the local render adapters and `--assistant-pattern`.
- **Added:** a vLLM `/render` client and the boundary derivation.
- **Unchanged:** on-policy pre-tokenized rows (#729) still pass straight through.

**Why `token_ids`, not the server mask:** `/render` can also return `assistant_tokens_mask`, but it's only populated when the template carries `{% generation %}` tags, which the common models don't ship. We request only `token_ids` and compute the mask from the boundary — so it works on every template, and isn't blocked on the unreleased `return_loss_mask` (`/render` itself shipped in v0.15.0).

# Requirement

`--render-endpoint` is now required for off-policy data (pre-tokenized input still works without it). Point it at the vLLM instance already used for hidden-state extraction, or a GPU-less `vllm launch render`. Single source of truth holds only if that instance's `tokenizer_mode` matches serving's.

# How to review

~68% of the diff is pure deletion; the real surface is ~440 lines of source:

| Bucket | What | How to read |
|---|---|---|
| **Deletion** (~1,560) | the old two-tier mask functions and the tests that covered them | skim |
| **New** (~530) | `render_client.py`, the boundary functions, `test_render_boundary.py` | review in isolation |
| **Rewiring** (~200) | `_preprocess_batch`, `build_eagle3_dataset`, `load_and_preprocess_dataset` switching to the boundary, plus the CLI swap | **the logic change** |

The other preprocessing helpers are byte-identical to `main`. The delete-plus-rewire core is atomic; `render_client.py` is the one piece reviewable on its own.

# Tests

```bash
# Serve any model with a chat template
vllm serve Qwen/Qwen3-0.6B --port 8000

# Derive masks from its render boundaries, on real data
python scripts/prepare_data.py \
  --model Qwen/Qwen3-0.6B \
  --data sharegpt \
  --render-endpoint http://localhost:8000 \
  --output /tmp/prepared \
  --max-samples 20
```

It prints the derived mask inline (blue = trainable, grey = masked):

```
<|im_start|>user
I have a location independent business that generates 1 miljon per year...<|im_end|>
<|im_start|>assistant
<think>

</think>

1/5: A clear target market and understanding of their needs is essential...<|im_end|>
```

The boundary is the thing to check: everything from `<think>` on is blue, while the user turn and the `<|im_start|>assistant` header stay grey. 

Or automated — launches its own vLLM and asserts the masks:

```bash
pytest tests/e2e/smoke/test_render_boundary.py -v
```

# Deferred / follow-ups

- **Concurrency** — the boundary does 2–3 renders per turn, today riding `datasets.map(num_proc)`. A batched/async client if throughput needs it.
- **Multimodal** — prefix-stability not yet validated on an image model.
- **Reasoning emission** — fan-out (on-policy, more rows) vs. a future single-row mode (cheaper, off-policy at turn seams); undecided.
- **Render-failure handling** — a dead endpoint currently drains the dataset to empty per-row instead of aborting fast; should distinguish endpoint failure from a single bad conversation.
- **Shared tool-call parsing** with regen (#750), once both paths speak vLLM messages.

## Checklist
- [x] Purpose of the PR
- [x] Test plan / results
- [x] Documentation update (CLI doc updated; user-guide tutorial pending)
- [x] I (a human) have reviewed this code to the best of my ability.


src/speculators/data_generation/preprocessing.py
```diff
@@ -496,56 +331,90 @@ def _parse_conv_tools(conv_tools: object, idx: int) -> list | None:
...
+
     results: dict[str, list] = {"input_ids": [], "loss_mask": [], "seq_len": []}
     conversations: list[dict] = examples.get("conversations", [])
 
-    # MM inputs must use Chat Completions API
-    if isinstance(processor, ProcessorMixin):
+    # MM inputs are extracted via the Chat Completions API, which needs the
+    # original messages -- token ids alone cannot carry the images.
+    if is_multimodal:
         results["messages"] = []
```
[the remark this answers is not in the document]
WindChimeRan: defer to followup PR

src/speculators/data_generation/preprocessing.py
```diff
@@ -194,283 +167,145 @@ def _adapt_conv_for_vllm(normalized_conv: list[dict]):
...
+        prompt_ids = _encode_render(
+            normalized_conv[:j],
+            render_endpoint,
+            add_generation_prompt=True,
+            tools=tools,
+        )
+        if len(prompt_ids) >= max_length:
+            # Context fills the window; this and later turns yield only
+            # unsupervised rows.
+            break
```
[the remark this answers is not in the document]
WindChimeRan: prompt length is monotonic

src/speculators/data_generation/render_client.py
```diff
@@ -0,0 +1,67 @@
...
+        raise RenderError(
+            f"Render endpoint returned {resp.status_code}: {resp.text[:500]}"
+        )
+
+    data = resp.json()
+    if "token_ids" not in data:
+        raise RenderError(
+            f"Render endpoint response missing 'token_ids': {str(data)[:500]}"
+        )
+    return data["token_ids"]
```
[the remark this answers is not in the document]
WindChimeRan: over-validation

speculatorsbot (commented):
LGTM. Clean replacement of the two-tier local masking path (HF template probe + regex span detection) with a single render-boundary mechanism. The boundary derivation is correct, the LCP fallback for scaffold-divergent templates (DeepSeek-R1 / Qwen3.5) is load-bearing and well-validated, and the monotonicity argument for the early-exit `break` holds. Net deletion of ~1k lines. Agree with deferring the multimodal tools round-trip and the async/batched client to follow-ups. Recommend approving.

speculatorsbot (commented):
Re-reviewed after the 7 commits since my previous pass (`ab44652`..`63cfcf5`). The always-fan-out simplification is a good call -- removing the packed/append-only branch deletes `_Turn`, the two-pass chain check, and the dual routing logic in exchange for a measured 2.4x token cost on append-only templates, which is acceptable given the complexity reduction. The e2e test against a live Qwen3-0.6B server (`test_render_boundary.py`) is the most valuable addition: it is the only test that exercises the actual `/render` transport and catches contract drift, which the stubbed unit tests structurally cannot. The transient-4xx retry fix (`TRANSIENT_STATUSES` for 408/429) correctly threads through the existing `with_retries`/`InvalidResponseError` short-circuit. All coderabbit findings from the first round are either addressed in these commits or explicitly deferred (multimodal tools round-trip). DCO sign-offs present on all 8 commits. LGTM. Recommend approving.

src/speculators/data_generation/render_client.py
```diff
@@ -0,0 +1,75 @@
...
+
+    if (
+        HTTPStatus.BAD_REQUEST <= resp.status_code < HTTPStatus.INTERNAL_SERVER_ERROR
+        and resp.status_code not in TRANSIENT_STATUSES
+    ):
+        # Deterministic client error (bad request, wrong URL) -- retrying wastes
+        # requests without changing the outcome. InvalidResponseError short-
+        # circuits @with_retries (see vllm_client._handle_retry_error).
+        raise InvalidResponseError(
+            f"Render endpoint returned {resp.status_code}: {resp.text[:500]}"
```
orestis-z: why do we truncate the error? (same below)

src/speculators/data_generation/render_client.py
```diff
@@ -0,0 +1,75 @@
...
+    tools: list[dict] | None = None,
+    chat_template_kwargs: dict | None = None,
+    timeout: float = DEFAULT_RENDER_TIMEOUT,
+) -> list[int]:
+    """POST to ``/v1/chat/completions/render`` and return the token ids.
+
+    No truncation: boundary detection needs full lengths; over-length rows are
+    clipped downstream.
+    """
+    url = f"{endpoint.rstrip('/')}/v1/chat/completions/render"
```
orestis-z, line 42: nit: the URL will be equivalent with double `//`, so that given I think `rstrip` is not necessary.

src/speculators/data_generation/render_client.py
```diff
@@ -0,0 +1,75 @@
...
+
+    body: dict = {
```
orestis-z: nit: `dict` (`dict[str, Any]`) is already implied, I think only worth explicitly annotating if the type can be different from what can already be inferred from the assigned object.

tests/integration/datagen/test_render_boundary.py
```diff
@@ -0,0 +1,158 @@
...
+guard (needs a template that rewrites history), and the client's error paths.
+"""
+
+import time
+
+import pytest
+from datasets import Dataset as HFDataset
+
+import speculators.data_generation.preprocessing as P  # noqa: N812
+from speculators.data_generation import render_client as RC  # noqa: N812
```
orestis-z: nit: If the user just sees `P` he has to find the definition of `P` to figure out what it is. I think `preprocessing` / `render_client` is fine. Though this is "just" a test so feel (very) free to ignore.

```suggestion
import speculators.data_generation.preprocessing as preprocessing
from speculators.data_generation import render_client
```

orestis-z (commented):
That's a great delta -1.1k lines :)

Some nits on `render_client.py` (feel free to ignore them)

src/speculators/data_generation/preprocessing.py
```diff
@@ -194,283 +165,110 @@ def _adapt_conv_for_vllm(normalized_conv: list[dict]):
...
-            hf_conv,
-            tokenize=False,
-            tools=tools,  # type: ignore[arg-type]
-            add_generation_prompt=False,
+        rows.append(
+            {
+                "input_ids": full_ids,
+                "loss_mask": [0] * boundary + [1] * (len(full_ids) - boundary),
+                "conv": normalized_conv[: j + 1],
+            }
```
orestis-z, line 279: nit: when a dict is having the same structure always (e.g. same keys), I find it to produce cleaner code if using explicit data structures like namedtuple or dataclasses.

src/speculators/data_generation/preprocessing.py
```diff
@@ -560,62 +404,67 @@ def _preprocess_batch(
         )
         tools_col = None
 
+    num_unsupervised = num_clipped = 0
+    num_convs_in = 0
+    num_convs_empty = 0
```
orestis-z: nit: would stick to one style but not mix both (either assigned all to 0 on one line or each on a new line, latter preferred)

src/speculators/data_generation/preprocessing.py
```diff
@@ -560,62 +404,67 @@ def _preprocess_batch(
...
+        # One row the render endpoint or boundary derivation can't handle must
+        # not kill the run. The failure modes can't be enumerated -- templates
+        # are swappable and raise arbitrary types -- so catch broadly and skip.
         except Exception as e:
-            log.error(
-                f"Failed to process conversation {idx} "
-                f"(assistant_pattern={assistant_pattern is not None}): "
-                f"{type(e).__name__}: {e}"
-            )
+            log.error(f"Failed to process conversation {idx}: {type(e).__name__}: {e}")
```
orestis-z: nit: I think better to use `logger.exception` instead since this already appends the exception msg out of the box

> exception(msg, *args, **kwargs)
Logs a message with level [ERROR](https://docs.python.org/3/library/logging.html#logging.ERROR) on this logger. The arguments are interpreted as for [debug()](https://docs.python.org/3/library/logging.html#logging.debug). Exception info is added to the logging message. This method should only be called from an exception handler.

src/speculators/data_generation/preprocessing.py
```diff
@@ -560,62 +404,67 @@ def _preprocess_batch(
         )
         tools_col = None
 
+    num_unsupervised = num_clipped = 0
+    num_convs_in = 0
+    num_convs_empty = 0
+
     for idx, conv in enumerate(conversations):
```
orestis-z, line 486: nit: this method seems to be doing quite a lot, and there's a loop inside a loop. Could extract helper method(s) to hide some of the logic.

orestis-z (commented):
Another round of style nits (feel free to ignore and/or pick some to fix you agree with / add value)

src/speculators/data_generation/preprocessing.py
WindChimeRan: decline 

because `PipelineLogger`has no `exception ` method. 

Actually why did we handroll `PipelineLogger` at the beginning? Seems like a re-implementing std lib logger to me.

src/speculators/data_generation/render_client.py
WindChimeRan: decline: Real vLLM: single slash returns token_ids, double slash 404s.

speculatorsbot (commented):
Re-reviewed after the 7 commits since my last pass (`63cfcf5`..`11a1fc1`). These address @orestis-z's style nits (explicit `BoundaryRow` TypedDict, full error messages, spelled-out test imports, counter formatting) and fix the `break` -> `continue` for over-length prompts -- the author correctly reconsidered the monotonicity argument: reasoning templates (Qwen3, DeepSeek-R1) strip `<think>` from history, so later prompts *can* shrink. The two new regression tests (`test_over_length_turn_does_not_drop_later_turns`, `test_over_length_first_turn_yields_no_rows`) cover both directions. The `list[list[dict]]` annotation for the conversations batch is also a correct fix. DCO sign-offs present on all 15 commits. LGTM. Recommend approving.

src/speculators/data_generation/preprocessing.py
WindChimeRan: Not sure how to do that cleanly.

docs/cli/prepare_data.md
```diff
@@ -42,7 +45,7 @@ python scripts/prepare_data.py \
 
 - **`--token-freq-path`** (str, default: `{output}/token_freq.pt`) Path to save token frequency distribution. Defaults to `token_freq.pt` in the output directory.
 
-- **`--assistant-pattern`** (str, default: `None`) Custom regex pattern for matching assistant responses. If not provided, auto-detected from chat template.
+- **`--render-endpoint`** (str, default: `None`) Base URL of a running vLLM server (e.g. `http://localhost:8000`), typically the same instance used for hidden-state extraction. Conversations are tokenized by its `/v1/chat/completions/render` endpoint and the loss mask is derived from the render boundary. Required unless the input is already pre-tokenized.
```
shanjiaz: Maybe let's make it clear here that this end point can be shared with the hidden states extraction one.

shanjiaz (commented):
Taking a look now. Sorry about the delay.

src/speculators/data_generation/preprocessing.py
shanjiaz: nit: let's rename this since it's not for eagle3 anymore

WindChimeRan: new: build_speculator_training_dataset

docs/cli/prepare_data.md
WindChimeRan: updated.

shanjiaz (commented):
I think the diff makes sense and checking the diff for loss masks is indeed smart. Verified locally. Could you please address Orestis' style concerns? we could merge soon. Would want to make sure this works on multi-modal data tho. Especially the coco dataset we claim to support.

src/speculators/data_generation/preprocessing.py
WindChimeRan: extracted `_render_conversation_rows` and `_append_boundary_rows`

orestis-z (approved):
Thanks for your work!

shanjiaz (approved):

PR #803, opened 2026-07-17
Add timing metrics to response regeneration pipeline

orestis-z:
## Summary
- Instruments the response regeneration script with performance timing using `time.perf_counter()`
- Adds live throughput to tqdm postfix (`rps` = vLLM requests/s, `tps` = generated tokens/s) alongside existing ok/err/trunc counters
- Tracks per-request latency, queue wait time, and token counts; logs a final summary with aggregate stats
- Uses plain logging + tqdm (not `speculators.metrics`), per the approach agreed in [#717](https://github.com/vllm-project/speculators/issues/717)

Companion to the offline data generation timing PR (https://github.com/vllm-project/speculators/pull/804). Both address [INFERENG-8704](https://redhat.atlassian.net/browse/INFERENG-8704).

## Test plan
- [x] Existing unit tests pass (`tests/unit/scripts/test_response_regeneration.py` — 43 tests)
- [x] Updated `_NullProgress` stub and stats dict in tests to match new signature
- [x] `ruff check` and `ruff format` pass
- [ ] Manual run against a vLLM endpoint to verify tqdm postfix and final summary output

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #717, opened 2026-07-03
[RFC]: Add Timing Metrics and Performance Tracing to Training and Data Generation Pipelines


shanjiaz:
### Motivation.

Speculators currently has a well-structured metric logging system (`speculators.metrics` logger with TensorBoard/W&B/Trackio/MLflow backends) that tracks model quality metrics (loss, accuracy, per-position acceptance rates), but has no instrumentation for performance. Other important parts of the pipeline like response regeneration also does not record any meaningful performance metric. There are lots of ongoing work for preprocessing & regeneration pipeline hardening, multi-node training and they can all benefit from a better performance tracing system.

### Proposed Change.

Add a lightweight step timer to the trainer that breaks each training step into phases and emits per-window averages through the existing `speculators.metrics` logger (TensorBoard / W&B / Trackio / MLflow).
Metrics:

- `profile/fetch_ms` -- data loading + host-to-device transfer
- `profile/fwd_ms` -- model forward pass
- `profile/bwd_ms` -- loss.backward + grad clipping
- `profile/opt_ms` -- optimizer step + LR scheduler
- `profile/step_ms` -- total step wall-clock
- `profile/tokens_per_s` -- throughput
- `profile/fetch_frac` -- fraction of step spent waiting on data (fetch-bound indicator)

Similarly, add timing to the response regeneration pipeline (`scripts/response_regeneration/`) and offline data generation (`scripts/data_generation_offline.py`):

- Per-request latency (already partially exists in response regeneration via `time.time()`, but not aggregated)
- Batch/window throughput (requests/s, tokens/s)
- Queue/wait time when the vLLM server is saturated
- Safetensors write time for data generation

Or any other reasonably design.

omerap12:
Hey @shanjiaz , I have opened this #722 to address some of the tasks above :)

orestis-z:
Thanks @omerap12 !

I'll take the remaining two items (response regeneration pipeline timing + offline data generation timing). These are independent of #722.

orestis-z:
@shanjiaz one thought on the response regeneration + offline data generation timing :

The RFC mentions emitting through the `speculators.metrics` logger, which makes sense for the trainer (step-indexed scalars → TensorBoard/W&B). For the batch scripts though, I think plain logging + tqdm is a better fit:

- These are one-shot pipeline runs, not training loops — there's no natural step axis for experiment tracking
[...]

PR #804, opened 2026-07-17
Add timing metrics to offline data generation pipeline

orestis-z:
## Summary
- Instruments the offline data generation script with performance timing using `time.perf_counter()`
- Times each `generate_hidden_states_async()` call (vLLM request latency) and each `shutil.move()` call (file write latency) independently
- Adds live throughput to tqdm postfix (`rps` = samples/s, `vllm` = avg request ms, `write` = avg write ms)
- Logs a final timing summary with aggregate stats after pipeline completion
- Uses plain logging + tqdm (not `speculators.metrics`), per the approach agreed in [#717](https://github.com/vllm-project/speculators/issues/717)

Companion to the response regeneration timing PR (#803). Both address [INFERENG-8704](https://redhat.atlassian.net/browse/INFERENG-8704).

## Test plan
- [x] Existing unit tests pass (`tests/unit/data_generation/test_offline.py` — 18 tests)
- [x] `ruff check` and `ruff format` pass
- [ ] Manual run against a vLLM endpoint to verify tqdm postfix and final summary output

### DIFF
--- a/scripts/data_generation_offline.py
+++ b/scripts/data_generation_offline.py
@@ -275,7 +275,6 @@ async def worker(  # noqa: C901
                 ) from e
         else:
             stats["ok"] += 1
-            stats["requests"] += 1
             stats["total_vllm_s"] += vllm_s
             stats["total_write_s"] += write_s
             logger.debug(
@@ -289,7 +288,7 @@ async def worker(  # noqa: C901
         finally:
             elapsed = time.perf_counter() - stats["start_time"]
             postfix = {"ok": stats["ok"], "err": stats["errors"]}
-            if elapsed > 0 and stats["requests"] > 0:
+            if elapsed > 0 and stats["ok"] > 0:
                 postfix["rps"] = f"{stats['requests'] / elapsed:.1f}"
                 postfix["vllm"] = (
                     f"{stats['total_vllm_s'] / stats['requests'] * 1000:.0f}ms"
@@ -375,7 +374,6 @@ async def generate_and_save_hidden_states(args, dataset):
     stats: dict[str, Any] = {
         "ok": 0,
         "errors": 0,
-        "requests": 0,
         "total_vllm_s": 0.0,
         "total_write_s": 0.0,
         "start_time": time.perf_counter(),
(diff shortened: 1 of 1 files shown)


WindChimeRan (commented):
the deletion of        `"requests": 0,` break the code (this is from the fix to the coderabbit). : KeyError: 'requests' hangs the pipeline on the first successful sample

speculatorsbot (commented):
Clean, well-scoped change — the design fits the RFC #717 discussion (plain logging + tqdm for one-shot pipeline scripts). No concerns beyond the `stats['requests']` KeyError already flagged by @WindChimeRan; once that's fixed, recommend approving.

speculatorsbot (commented):
Clean, well-scoped instrumentation that aligns with the approach agreed in #717. Agree with @WindChimeRan's finding — the three `stats['requests']` references in the `finally` block need to be replaced with `stats['ok']` to fix the KeyError introduced by the second commit. Recommend approving once that's fixed.

speculatorsbot (commented):
The `stats['requests']` KeyError flagged by @WindChimeRan is fixed in `7e608eb` — all three f-string references now correctly use `stats['ok']`. No other issues. Recommend approving.

scripts/data_generation_offline.py
```diff
@@ -226,21 +228,25 @@ async def worker(  # noqa: C901
...
             lock_path = hidden_states_path + ".lock"
             if Path(lock_path).exists():  # noqa: ASYNC240
                 await wait_for_lock_async(lock_path)
 
             async with write_semaphore:  # Limit number of active disk writes
+                t_write = time.perf_counter()
                 await asyncio.to_thread(
                     shutil.move, hidden_states_path, target_hidden_states_path
                 )
+                write_s = time.perf_counter() - t_write
```
WindChimeRan, line 249: write_s excludes validation, which still holds the write semaphore. 

Suggest change: 

```suggestion
                if validate_outputs:

                    def _load_and_check(
                        path=target_hidden_states_path,
                        tokens=item["input_ids"],
                    ):
                        loaded = load_file(path)
                        check_hidden_states(loaded, tokens)

                    await asyncio.to_thread(_load_and_check)
                write_s = time.perf_counter() - t_write
```

WindChimeRan (commented):
One semaphore issues. then LGTM

----

I found this script overcomplicated. e.g., some of the semaphore usages are over-engineering. need to clean it up later.

scripts/data_generation_offline.py
orestis-z: Good catch on the semaphore scope, but I think the current placement is intentional — `write_s` is meant to measure `shutil.move` (the actual disk write), not the optional validation read-back.

The tqdm postfix and summary log label this as "write" / "avg file write", so it should reflect write latency specifically. Validation (`load_file` + `check_hidden_states`) is a read + integrity check — bundling it in would inflate the metric and make it misleading, especially since it's gated behind `--validate-outputs`.

The write semaphore limits concurrent disk I/O (both writes and validation reads), but it's a concurrency limiter, not a semantic boundary for timing. If validation timing is useful it could be its own metric in a follow-up.

WindChimeRan (commented):
LGTM. Recommend accept.

shanjiaz (approved):
Thanks for adding this!

scripts/data_generation_offline.py
rahul-tuli, line 249: Q: Won't this just measure the rename time? Since the server has written the hidden_states_already?

orestis-z: Yes — when source and target are on the same filesystem, `shutil.move` is just `os.rename()` and this will show ~0 ms. That's still useful: it confirms writes aren't the bottleneck.

When the paths span different filesystems (e.g. vLLM writing to a tmpdir on local NVMe while the output dir is on CephFS/NFS), it becomes a full copy+delete and can be significant. The metric catches that case, plus any I/O contention under high concurrency. The cost of collecting it is effectively zero either way.

PR #805, opened 2026-07-17
fix(dspark/metrics): include slot0 in accept_len and confidence metri…

minziyu:
## Summary

DSpark with `sample_from_anchor=True` treats slot 0 as the first real prediction (not an anchor), but the `accept_len` and `confidence_cumprod_bias` metrics were hardcoded to skip slot 0 (`[:, 1:]`). This caused reported metrics to undercount for DSpark models.

## Fix

Move `start_pos = 0 if sample_from_anchor else 1` earlier in `compute_metrics()` so it is shared by:
  - `accept_prefix` / `draft_mask` in the accept-length block
  - `conf_prefix` in the confidence block
  - (per-position accuracy already used `start_pos`; its behavior is unchanged)

For DFlash (`sample_from_anchor=False`): `start_pos=1` — behavior unchanged.
For DSpark (`sample_from_anchor=True`): `start_pos=0` — slot 0 is now counted.
This is a pure metrics-display fix; loss, gradients, and checkpoints are unaffected.

## Testing

Compared eval output before/after the fix on a DSpark checkpoint (block_size=8, sample_from_anchor=True). Metrics that previously reported lower accept_len due to the hardcoded skip now reflect the correct per-position breakdown including slot 0.


src/speculators/models/dspark/metrics.py
```diff
@@ -63,6 +63,8 @@ def compute_metrics(
     device = logits.device
     seq_len = logits.shape[1]
     pos_idx = (torch.arange(seq_len, device=device) % block_size).unsqueeze(0)
+    # Start position: 0 if sample_from_anchor (DSpark) else 1 (DFlash anchor).
```
fynnsu: Remove comment

fynnsu (approved):
Looks good, thanks

src/speculators/models/dspark/metrics.py
minziyu: done

minziyu:
@fynnsu Sorry for the repeated quality-check failures. The issue may be that my local ruff version differs from the one used in CI, so ruff format produces slightly different outputs on my machine, causing the check to fail on your side.

If the check fails again, could you please let me know the exact ruff version the CI is running? I'll match it locally and push a corrected formatting commit.

Apologies for the churn, and thanks for your patience!

fynnsu:
@minziyu No worries, looks like it passed this time.

And for future reference you should be able to run `uv pip install -e .[dev]` (i.e. install the optional dev dependencies) and those should be pinned to match CI.

fynnsu:
@minziyu looks like a couple of tests are failing and will need to be updated to handle the new logic. https://buildkite.com/vllm/speculators/builds/981/list?jid=019f7167-9e13-4e59-949e-af038bb9219e&tab=output

minziyu:
> @minziyu looks like a couple of tests are failing and will need to be updated to handle the new logic. https://buildkite.com/vllm/speculators/builds/981/list?jid=019f7167-9e13-4e59-949e-af038bb9219e&tab=output

@fynnsu The unit-test failures were caused by two existing tests asserting the old hardcoded [:, 1:] behavior (which always excluded slot 0 from accept-length/confidence metrics). Since these tests treat position 0 as the anchor, they only hold under sample_from_anchor=False, so I've made that explicit.

Would you like me to also add a test covering the default sample_from_anchor=True path (where slot 0 is a real prediction and accept_len counts all draft slots)? I have it ready but held off to keep this fix minimal — happy to add it if you think it's worth the coverage!

fynnsu:
> Would you like me to also add a test covering the default sample_from_anchor=True path (where slot 0 is a real prediction and accept_len counts all draft slots)? I have it ready but held off to keep this fix minimal — happy to add it if you think it's worth the coverage!

@minziyu yes that'd be great if you can add that!

minziyu:
> > Would you like me to also add a test covering the default sample_from_anchor=True path (where slot 0 is a real prediction and accept_len counts all draft slots)? I have it ready but held off to keep this fix minimal — happy to add it if you think it's worth the coverage!
> 
> @minziyu yes that'd be great if you can add that!

Done! I've pushed the test case for the default sample_from_anchor=True path. Please let me know if it looks good to you!

fynnsu (approved):

fynnsu:
Thanks @minziyu

Issue #781, opened 2026-07-13
Q3 Roadmap 


dsikka:
### Status: In Progress

Support multi-node training using the Mooncake backend for hidden states. This initiative also involves updating the project structure to create a dedicated folder / package for various KV Cache connectors (such as the Mooncake connector) and other modifications that could easily be substituted in during training. 

Relevant PRs: 
1. https://github.com/vllm-project/speculators/pull/735
2. https://github.com/vllm-project/speculators/pull/836

### Status: In Progress 

Several improvements have been proposed to extend the vanilla DFlash algorithm, most notably the DSpark algorithm (supported as of: https://github.com/vllm-project/speculators/commit/ff71b1e6c89feb62d77a492075164d207989e538) and the DPace loss function: (supported as of: https://github.com/vllm-project/speculators/pull/736). Q3 will include active work to support these extensions and compare their performances.

Relevant PRs / RFCs:
1. https://github.com/vllm-project/speculators/pull/773
2. https://github.com/vllm-project/speculators/issues/757
3. https://github.com/vllm-project/speculators/pull/736
4. https://github.com/vllm-project/speculators/pull/798
5. https://github.com/vllm-project/speculators/pull/677

### Status: In Progress
1. Long Context Training Support through SP Ulysses: https://github.com/vllm-project/speculators/pull/640
3. Sliding Window Attention for Eagle3 and P-Eagle: https://github.com/vllm-project/speculators/pull/684

### Status: In Progress

Relevant PRs / RFCs: 
1. https://github.com/vllm-project/speculators/issues/652
2. https://github.com/vllm-project/speculators/pull/665

#### Tool Call Improvements
1. https://github.com/vllm-project/speculators/pull/771
2. https://github.com/vllm-project/speculators/pull/769
3. https://github.com/vllm-project/speculators/pull/750
4. https://github.com/vllm-project/speculators/pull/777

### Status: In Progress

With the rapid growth of the project, targeted refactoring and cleanup efforts are needed to simplify and improve areas of the codebase that have become more complex over time. Focusing on improving code organization, removing outdated or redundant implementations, and standardizing existing patterns will improve the overall quality, maintainability, and readability of the codebase.

Areas of improvements include: 
- [ ] Training Args handling (currently at over 70 args)
   - https://github.com/vllm-project/speculators/pull/841
- [ ]  Legacy code removal (Eagle), potentially remove older converters?
[...]

PR #812, opened 2026-07-18
Remove deprecated Eagle3 legacy model from converter

guan404ming:
## Purpose

Part of the legacy code removal item in #781.

- Delete deprecated `eagle3_legacy_model.py`; its `forward()` was never implemented and it only served as a weight skeleton for the Eagle3 converter.
- Convert with the training-side `Eagle3DraftModel` instead, following the DFlash converter pattern. Net -387 lines.
- Side fix: the legacy skeleton hardcoded a single decoder layer, silently dropping `layers.1.*` weights of multi-layer checkpoints. `Eagle3DraftModel` builds all layers, so they are now preserved.

## Tests

- New unit test in `test_eagle3_converter.py`: converts a tiny 2-layer checkpoint and asserts the saved state dict keeps every decoder layer, contains only source-checkpoint keys, and round-trips values exactly.
- `tests/unit/convert/`: 52 passed. `make quality` (ruff + mypy) passes.

## Checklist

I have filled in:


speculatorsbot (commented):
LGTM. Clean removal of the deprecated legacy model, correctly replaces it with `Eagle3DraftModel` following the DFlash converter pattern. The side fix for multi-layer checkpoint preservation is correct and well-tested. Recommend approving.

speculatorsbot (commented):
Clean removal of dead code with a multi-layer bug fix as a bonus. The legacy `Eagle3Speculator` had an unimplemented `forward()` and hardcoded a single decoder layer, silently dropping `layers.1.*` weights from multi-layer checkpoints. Switching to `Eagle3DraftModel` follows the established DFlash converter pattern, the `_VERIFIER_FILLED_KEYS` frozenset is consistent with `Eagle3DraftModel._keys_to_ignore_on_load_missing`, and the save-filter approach correctly avoids persisting NaN-initialized verifier weights. Test coverage for the multi-layer fix is solid. LGTM. Recommend approving.

speculatorsbot (commented):
The approach is sound -- replacing the deprecated `Eagle3Speculator` weight skeleton with `Eagle3DraftModel` is a clean simplification (net -387 lines), and the multi-layer fix that preserves `layers.1.*` weights is a welcome correctness improvement. One concern about an internal inconsistency between the save and validate paths when source checkpoints lack drafter embeddings.

src/speculators/convert/eagle/eagle3_converter.py
```diff
@@ -223,20 +216,23 @@ def _save_converted_checkpoint(
...
-                verifier_attachment_mode="detached",
-            )
-            logger.success("Validation succeeded")
+            model = Eagle3DraftModel.from_pretrained(checkpoint_path)
         except (OSError, ValueError, RuntimeError) as e:
             logger.error(f"Validation failed: {e}")
             raise
+        for name in ("fc.weight", "lm_head.weight", "embed_tokens.weight"):
+            if torch.isnan(model.state_dict()[name]).any():
+                raise ValueError(f"Converted checkpoint has NaN in {name}")
```
speculatorsbot, line 237: `_save_converted_checkpoint` treats absent `embed_tokens.weight` as expected (it is in `_VERIFIER_FILLED_KEYS`, so no warning is logged and the key is simply omitted from the saved checkpoint). But this NaN check then rejects the same checkpoint during validation.

This matters for source checkpoints that do not include drafter embeddings -- the case the old `has_drafter_embedding = "embed_tokens.weight" in weights` flag handled. For such a checkpoint, `_save_converted_checkpoint` succeeds (correctly saving only available weights), but `_validate_converted_checkpoint` fails with `Converted checkpoint has NaN in embed_tokens.weight`.

The [DFlash converter avoids this](https://github.com/vllm-project/speculators/blob/e591d85efb0ac1c80089f488ae960f9acf885c69/src/speculators/convert/dflash/converter.py#L166-L170) by calling `model.load_verifier_weights()` before saving, so verifier-sourced weights (`embed_tokens`, `lm_head`, `verifier_lm_head`, `verifier_norm`) are always populated and the NaN check is guaranteed to pass.

Two options to resolve the inconsistency:
1. If all Eagle3 source checkpoints are guaranteed to include `embed_tokens.weight`, remove it from `_VERIFIER_FILLED_KEYS` so the missing-key warning surfaces early in `_save_converted_checkpoint`.
2. If some checkpoints legitimately omit it, either call `load_verifier_weights()` before saving (matching DFlash) or remove `"embed_tokens.weight"` from this NaN check tuple.

PR #813, opened 2026-07-18
Batch val metric reduction to one sync per epoch

guan404ming:
## Purpose

The val loop issued one `all_reduce` and one `.item()` GPU sync per metric per batch. Accumulate metrics on device and reduce/sync once per epoch instead. Metric values are unchanged.

Benchmark on Modal, driving `Trainer.val_epoch` with a stub model (200 batches × 15 metrics, median of 5 rounds):

| setup | before | after | speedup |
| --- | --- | --- | --- |
| 1× L4 | 389.6 ms | 364.2 ms | 1.07× |
| 2× L4 (NCCL) | 642.5 ms | 439.8 ms | 1.46× |

## Tests

- Benchmark asserts old and new implementations produce identical metrics.
- `make quality` passes; trainer unit tests pass.

## Checklist

I have filled in:

### DIFF
--- a/src/speculators/train/trainer.py
+++ b/src/speculators/train/trainer.py
@@ -522,8 +522,6 @@ class Trainer:
         if self.rank == 0:
             val_loader = tqdm(val_loader, desc=f"Epoch {epoch}")  # type: ignore[assignment]
 
-        # Accumulate metrics on device and reduce/sync once per epoch to avoid
-        # per-batch, per-metric all_reduce and .item() GPU syncs.
         accumulated: dict[str, torch.Tensor] = {}
         num_batches = len(val_loader)
         for batch in val_loader:
--- a/src/speculators/train/trainer.py
+++ b/src/speculators/train/trainer.py
@@ -522,7 +522,9 @@ class Trainer:
         if self.rank == 0:
             val_loader = tqdm(val_loader, desc=f"Epoch {epoch}")  # type: ignore[assignment]
 
-        val_metrics: dict[str, float] = {}
+        # Accumulate metrics on device and reduce/sync once per epoch to avoid
+        # per-batch, per-metric all_reduce and .item() GPU syncs.
+        accumulated: dict[str, torch.Tensor] = {}
         num_batches = len(val_loader)
         for batch in val_loader:
             gpu_batch = {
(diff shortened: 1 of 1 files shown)


src/speculators/train/trainer.py
```diff
@@ -539,12 +541,16 @@ def val_epoch(self, epoch: int) -> dict[str, float] | None:
...
-                val_metrics[k] = val_metrics.get(k, 0.0) + v.item()
+                acc = accumulated.get(k)
+                accumulated[k] = v.float() if acc is None else acc + v.float()
+
+        val_metrics: dict[str, float] = {}
+        if accumulated:
+            stacked = torch.stack(list(accumulated.values()))
+            if self.is_distributed:
+                dist.all_reduce(stacked, op=dist.ReduceOp.SUM)
+            val_metrics = dict(zip(accumulated, stacked.tolist(), strict=True))
```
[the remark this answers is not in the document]
guan404ming: Metric keys and their order come from run-level config, so they are identical on every rank; the pre-change per-batch all_reduce relied on the same invariant. All metric values are 0-D scalars, so stack + tolist already yields floats, and a mixed shape would make torch.stack fail loudly rather than corrupt silently. Keeping as is.

speculatorsbot (commented):
LGTM. The reduction-order change is correct (summation commutes across batches and ranks), the scope is tight, and the benchmarks are convincing. Agree with the author's response to the dict-ordering concern — metric keys are deterministic from run config. Recommend approving.

speculatorsbot (commented):
LGTM. The batch-then-reduce approach is mathematically equivalent to the per-batch reduction (addition is associative/commutative), and the benchmarks confirm identical metric values. Agree with the author's response to @coderabbitai — metric key ordering is determined by model config, so it is consistent across ranks. Recommend approving.

src/speculators/train/trainer.py
```diff
@@ -522,7 +522,9 @@ def val_epoch(self, epoch: int) -> dict[str, float] | None:
         if self.rank == 0:
             val_loader = tqdm(val_loader, desc=f"Epoch {epoch}")  # type: ignore[assignment]
 
-        val_metrics: dict[str, float] = {}
+        # Accumulate metrics on device and reduce/sync once per epoch to avoid
+        # per-batch, per-metric all_reduce and .item() GPU syncs.
```
fynnsu: Remove comment.

fynnsu (approved):

src/speculators/train/trainer.py
guan404ming: Sure, let me remove them.

guan404ming:
Thanks!

PR #815, opened 2026-07-18
Preallocate collate buffer to avoid double copy

guan404ming:
## Purpose

Speed up the training dataloader collate path. Each batch was assembled with `torch.cat` followed by `F.pad`, which materializes two full-size intermediate copies per key. Samples are now copied once into a preallocated `[max_len, ...]` buffer. Output is unchanged (verified `torch.equal` against the old implementation); the now-unused `slice_and_pad_to_length` is removed.

Micro-benchmark (8192 seq × 3×4096 hidden, 6 samples, median of 30 runs):

| | per batch | speedup |
| --- | --- | --- |
| before (cat + pad) | 10.33 ms | |
| after (preallocated buffer) | 7.92 ms | 1.30× |

## Tests

- `tests/unit/train/test_data.py`: 10 passed (covers collate basics and length truncation).
- `make quality` (ruff + mypy) passes.

## Checklist

I have filled in:

### DIFF
--- a/src/speculators/train/data.py
+++ b/src/speculators/train/data.py
@@ -10,7 +10,6 @@ from typing import Any, Literal, cast
 
 import openai
 import torch
-import torch.nn.functional as F  # noqa: N812
 from datasets import load_from_disk
 from torch.utils.data import Dataset
 
@@ -37,13 +36,6 @@ def list_files(path):
             datapath.append(file_path)
 
     return datapath
-
-
-def slice_and_pad_to_length(tensor, length):
-    sliced_tensor = tensor[:length]
-    padding = [0, 0] * sliced_tensor.dim()
-    padding[-1] = length - sliced_tensor.shape[0]
-    return F.pad(sliced_tensor, padding)
 
 
 def split_files(datapath: str, ratio: float = 0.9, seed: int = 0):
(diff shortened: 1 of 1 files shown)


WindChimeRan (commented):
From my understanding, this is mainly a hidden state memory optimization in dataloader. 

For the speedup, the collate runs in prefetched dataloader workers overlapped with GPU compute, so it's largely hidden end-to-end.

I tested it locally on memory saving:

Qwen3-8B dims, one packed 8192-tok batch ≈ 5 ShareGPT convs):

  ~105 MB/batch saved (compared with main). most of them are from hidden states. 

---

The gain is small but It's ok to accept.

speculatorsbot (commented):
LGTM. Clean, minimal optimization — removes a helper + import, replaces two intermediate allocations with a single preallocated buffer. Output is equivalent (both take the first `max_len` rows of the concatenated sequence and zero-pad the rest). Agree with @WindChimeRan's assessment. Recommend approving.

speculatorsbot (commented):
LGTM. The preallocated-buffer approach is correct and eliminates the two intermediate copies; `slice_and_pad_to_length` is only used here so removal is safe. Recommend approving.

fynnsu:
Closing this, because we will merge #816 instead.

PR #819, opened 2026-07-19
refactor: replace torch.cuda.synchronize() with device-agnostic torch.accelerator.synchronize()

CarterDuan:
## Purpose
This PR replaces torch.cuda.synchronize() calls with PyTorch hardware-agnostic unified accelerator API torch.accelerator.synchronize(). A version fallback wrapper is added to maintain compatibility with PyTorch versions below 2.5, eliminating hard-coded CUDA dependencies and enabling native support for heterogeneous accelerators such as NPU and XPU.

## Tests
tested on cuda GPU environment and Ascend NPU evironment

## Checklist
I have filled in:
- [ Y] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ Y] The test plan/results, such as providing test command and pasting the results.
- [Y ] I (a human) have written or reviewed the code in this pr to the best of my ability.

### DIFF
--- a/src/speculators/train/trainer.py
+++ b/src/speculators/train/trainer.py
@@ -54,10 +54,7 @@ class _StepTimer:
 
     def mark(self, name: str) -> None:
         if self.enabled:
-            if hasattr(torch, "accelerator"):
-                torch.accelerator.synchronize()
-            else:
-                torch.cuda.synchronize()
+            torch.accelerator.synchronize()
             self._marks[name] = time.perf_counter()
 
     def mark_value(self, name: str, value: float) -> None:
@@ -67,10 +64,7 @@ class _StepTimer:
     def now(self) -> float | None:
         if not self.enabled:
             return None
-        if hasattr(torch, "accelerator"):
-            torch.accelerator.synchronize()
-        else:
-            torch.cuda.synchronize()
+        torch.accelerator.synchronize()
         return time.perf_counter()
 
     def profile(self, num_tokens: int) -> dict[str, float] | None:
--- a/src/speculators/train/trainer.py
+++ b/src/speculators/train/trainer.py
@@ -41,7 +41,7 @@ metric_logger = logging.getLogger("speculators.metrics")
 
 
 class _StepTimer:
-    # Each mark()/now() forces a cuda.synchronize to capture true GPU time.
+    # Each mark()/now() forces an accelerator.synchronize to capture true GPU time.
     # This serialises the CUDA pipeline, so profiled steps are slower; keep
     # log_freq > 1 in perf-sensitive runs.
     def __init__(self, enabled: bool = False):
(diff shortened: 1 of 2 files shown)


WindChimeRan (commented):
References: 
vllm rfc https://github.com/vllm-project/vllm/issues/30679
vllm pr (merged) https://github.com/vllm-project/vllm/pull/36085

**I agree with the direction.** 

But fallback is overcomplicated `if hasattr(torch, "accelerator"):`. Because the RFC marks torch.accelerator.synchronize "torch 2.9 ready.". 

one line swap should be fine `torch.accelerator.synchronize()`

cc: @shanjiaz

speculatorsbot (commented):
LGTM directionally. Agree with @WindChimeRan — `speculators` requires `torch>=2.9.0` (see `pyproject.toml`), so `torch.accelerator` is always available and the `hasattr` fallback is dead code. A straight swap to `torch.accelerator.synchronize()` is cleaner. Recommend approving once simplified.

speculatorsbot (commented):
LGTM on the direction. Agree with @WindChimeRan that the `hasattr` fallback can be dropped — the project already requires `torch>=2.9.0` and uses `torch.accelerator` directly elsewhere (e.g., [`distributed.py:151-152`](https://github.com/vllm-project/speculators/blob/1d6b2c335388fcf6a369234f252790fffb767ce2/src/speculators/train/distributed.py#L151-L152), [`utils/util.py`](https://github.com/vllm-project/speculators/blob/1d6b2c335388fcf6a369234f252790fffb767ce2/src/speculators/utils/util.py#L15-L47)).

One additional note: the class comment at [`trainer.py:43`](https://github.com/vllm-project/speculators/blob/1d6b2c335388fcf6a369234f252790fffb767ce2/src/speculators/train/trainer.py#L43) still says *"forces a cuda.synchronize"* — worth updating to reference the accelerator API for consistency with this change.

Recommend approving once simplified per WindChimeRan's feedback.

fynnsu (changes requested):
Hi, thanks for submitting the change. Could you please remove the if/else logic and just call `torch.accelerator.synchronize()` directly. Speculators only supports `torch>=2.9` so we don't need the fallback logic.

CarterDuan:
> Hi, thanks for submitting the change. Could you please remove the if/else logic and just call `torch.accelerator.synchronize()` directly. Speculators only supports `torch>=2.9` so we don't need the fallback logic.

Ok, makes sense. I removed the fallback and call torch.accelerator.synchronize() directly.

fynnsu (approved):

fynnsu:
@CarterDuan looks like there was a small issue with one of tests which was checking for `torch.cuda.synchronize` calls. I put up a fix and the pr has now landed.

PR #820, opened 2026-07-19
[Clarification] Rename launch_vllm.py -> launch_vllm_hidden_states.py

WindChimeRan:
## Purpose

**TLDR**: `scripts/launch_vllm.py` is slow because it always producing hidden states. So the name should not be generic. 

`scripts/launch_vllm.py` always configures vLLM as a hidden-states producer (writes tensors to `--hidden-states-path` on every request), but its generic name reads like a plain-server launcher. So it gets reached for in workflows that only want `vllm serve` (e.g. response regeneration). 

Rename it to `launch_vllm_hidden_states.py`, update all references, and add a one-line module docstring plus a one-sentence CLI-ref note. Pure rename + reference updates; no behavior change.

## Tests

- `git grep 'launch_vllm\.py'` / `launch_vllm.md` / `launch_vllm/` → all empty.

## Checklist

I have filled in:

### DIFF
--- a/docs/.nav.yml
+++ b/docs/.nav.yml
@@ -35,7 +35,7 @@ nav:
           - cli/index.md
           - prepare_data.py: cli/prepare_data.md
           - data_generation_offline.py: cli/data_generation_offline.md
-          - launch_vllm.py: cli/launch_vllm.md
+          - launch_vllm_hidden_states.py: cli/launch_vllm_hidden_states.md
           - train.py: cli/train.md
           - response_regeneration: cli/response_regeneration.md
     - Community:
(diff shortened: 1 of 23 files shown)


fynnsu:
I'm not entirely sold on this. We already have docstrings and an argparse description making it clear this is for hidden states extraction. I get that we could make this more explicit, but I'm not convinced its necessary.

WindChimeRan:
@fynnsu 


```
docstrings and an argparse description
```
No one actually read it (both human and agent).

issue: people & agent frequently use `launch_vllm.py` for regen, which is wrong. (it'll be super slow due to saving hidden state to disk) (I think I saw it 3 times, one by my own claude agent, one by a user in an issue, one by #847)


Also

`scripts/` is not in the wheel. so the only valid way is installing from main from source.

I recommend ship the file renaming first (file renaming is the smallest fix for it), then re-design the UX, by merging some scripts functionalities to the src/

WindChimeRan:
I'll close my PR and leave it to you @fynnsu

PR #826, opened 2026-07-20
fix(trainer): device-agnostic synchronize

ys2025-AI:
## Purpose

Replace hardcoded torch.cuda.synchronize() in _StepTimer with _synchronize_device(), which falls back to torch.npu.synchronize() on non-CUDA builds.

## Tests

Verified on the Ascend NPU 910B3.

## Checklist

I have filled in:

- [] (Optional) The necessary documentation update.

### DIFF
--- a/src/speculators/train/trainer.py
+++ b/src/speculators/train/trainer.py
@@ -40,6 +40,14 @@ root_logger = logging.getLogger("speculators")
 metric_logger = logging.getLogger("speculators.metrics")
 
 
+def _synchronize_device() -> None:
+    """Device-agnostic synchronize: CUDA when available, otherwise NPU."""
+    if torch.cuda.is_available():
+        torch.cuda.synchronize()
+    elif hasattr(torch, "npu") and torch.npu.is_available():
+        torch.npu.synchronize()
+
+
 class _StepTimer:
     # Each mark()/now() forces a cuda.synchronize to capture true GPU time.
     # This serialises the CUDA pipeline, so profiled steps are slower; keep
@@ -54,7 +62,7 @@ class _StepTimer:
 
     def mark(self, name: str) -> None:
         if self.enabled:
-            torch.cuda.synchronize()
+            _synchronize_device()
             self._marks[name] = time.perf_counter()
 
     def mark_value(self, name: str, value: float) -> None:
@@ -64,7 +72,7 @@ class _StepTimer:
     def now(self) -> float | None:
         if not self.enabled:
             return None
-        torch.cuda.synchronize()
+        _synchronize_device()
         return time.perf_counter()
 
     def profile(self, num_tokens: int) -> dict[str, float] | None:
--- a/tests/unit/train/test_step_timer.py
+++ b/tests/unit/train/test_step_timer.py
@@ -1,8 +1,8 @@
-from unittest.mock import patch
+from unittest.mock import MagicMock, patch
 
 import pytest
 
-from speculators.train.trainer import _StepTimer
+from speculators.train.trainer import _StepTimer, _synchronize_device
 
 
 def test_disabled_timer_returns_none():
@@ -16,7 +16,7 @@ def test_disabled_timer_returns_none():
     assert timer.profile(num_tokens=1024) is None
 
 
-@patch("speculators.train.trainer.torch.cuda.synchronize")
+@patch("speculators.train.trainer._synchronize_device")
 def test_enabled_timer_returns_profile(mock_sync):
     timer = _StepTimer(enabled=True)
 
@@ -79,7 +79,7 @@ def test_disabled_to_enabled_transition():
     timer.mark_value("start", t_before_fetch)
 
     with (
-        patch("speculators.train.trainer.torch.cuda.synchronize"),
+        patch("speculators.train.trainer._synchronize_device"),
         patch(
             "speculators.train.trainer.time.perf_counter",
             side_effect=[2.1, 2.4, 2.5, 2.6, 2.6],
@@ -106,7 +106,7 @@ def test_zero_step_ms_returns_zero_throughput():
     timer = _StepTimer(enabled=True)
     timer.mark_value("start", 1.0)
     with (
-        patch("speculators.train.trainer.torch.cuda.synchronize"),
+        patch("speculators.train.trainer._synchronize_device"),
         patch("speculators.train.trainer.time.perf_counter", return_value=1.0),
     ):
         timer.mark("fetch")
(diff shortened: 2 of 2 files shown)


ys2025-AI:
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> 
> ⚠️ Outside diff range comments (1)
> > tests/unit/train/test_step_timer.py (1)> `19-40`: _🎯 Functional Correctness_ | _🟡 Minor_ | _⚡ Quick win_
> > > **Add direct tests for `_synchronize_device()` dispatch.**
> > > These tests only verify that `_StepTimer` calls the wrapper; they do not catch regressions in CUDA/NPU selection or the no-device fallback. Add focused tests for CUDA available, NPU-only available, and neither backend available.
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against current code. Fix only still-valid issues, skip the
> > > rest with a brief reason, keep changes minimal, and validate.
> > > 
> > > In `@tests/unit/train/test_step_timer.py` around lines 19 - 40, Add focused unit
> > > tests for the `_synchronize_device()` wrapper covering CUDA availability,
> > > NPU-only availability, and neither backend being available. Mock the relevant
> > > device-availability checks and synchronization functions, then assert the
> > > correct backend dispatch or no-device fallback directly rather than testing only
> > > `_StepTimer` through `test_enabled_timer_returns_profile`.
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > 
> > > _Source: Path instructions_
> 
> 🤖 Prompt for all review comments with AI agents
> ```
> Verify each finding against current code. Fix only still-valid issues, skip the
> rest with a brief reason, keep changes minimal, and validate.
> 
> Outside diff comments:
> In `@tests/unit/train/test_step_timer.py`:
> - Around line 19-40: Add focused unit tests for the `_synchronize_device()`
> wrapper covering CUDA availability, NPU-only availability, and neither backend
> being available. Mock the relevant device-availability checks and
> synchronization functions, then assert the correct backend dispatch or no-device
> fallback directly rather than testing only `_StepTimer` through
> `test_enabled_timer_returns_profile`.
> ```
> 
> ℹ️ Review info

Added three focused dispatch tests in test_step_timer.py:
test_synchronize_device_uses_cuda_when_available — mocks cuda.is_available()=True, asserts torch.cuda.synchronize() is called.
test_synchronize_device_falls_back_to_npu — mocks CUDA unavailable + NPU available, asserts torch.npu.synchronize() is called.
test_synchronize_device_no_backend_available — mocks both unavailable, asserts no error and no sync call.

speculatorsbot (commented):
This PR and [#819](https://github.com/vllm-project/speculators/pull/819) solve the same problem. Since `speculators` requires `torch>=2.9.0`, [`torch.accelerator.synchronize()`](https://docs.pytorch.org/docs/stable/generated/torch.accelerator.synchronize.html) is always available and is PyTorch's official device-agnostic API — it dispatches to CUDA, NPU, XPU, MPS, or whatever the current accelerator is, without manual if-else chains. A maintainer [confirmed this direction on #819](https://github.com/vllm-project/speculators/pull/819#pullrequestreview-4731091532), noting a one-line swap to `torch.accelerator.synchronize()` is sufficient (no fallback needed), and linked the [vLLM RFC](https://github.com/vllm-project/vllm/issues/30679) that standardizes on `torch.accelerator` across the project.

I'd recommend either closing this in favor of #819 (after it adds tests), or rebasing this PR to use `torch.accelerator.synchronize()` instead of the CUDA/NPU if-else — that way the tests from this PR (which #819 is missing) would carry over.

src/speculators/train/trainer.py
```diff
@@ -40,6 +40,14 @@
 metric_logger = logging.getLogger("speculators.metrics")
 
 
+def _synchronize_device() -> None:
+    """Device-agnostic synchronize: CUDA when available, otherwise NPU."""
+    if torch.cuda.is_available():
+        torch.cuda.synchronize()
+    elif hasattr(torch, "npu") and torch.npu.is_available():
+        torch.npu.synchronize()
```
speculatorsbot, line 48: Since `speculators` requires `torch>=2.9.0` ([pyproject.toml](https://github.com/vllm-project/speculators/blob/b264de613e8a75b4638076156e35a94b16ea256d/pyproject.toml)), `torch.accelerator.synchronize()` is unconditionally available and handles CUDA, NPU, XPU, MPS, and any future backend in a single call — no if-else chain needed.

A maintainer already endorsed this approach on the competing PR #819 ([comment](https://github.com/vllm-project/speculators/pull/819#pullrequestreview-4731091532)), and it aligns with the [vLLM RFC](https://github.com/vllm-project/vllm/issues/30679) to migrate from `torch.cuda` to `torch.accelerator`.

Suggested replacement for this entire function:

```python
def _synchronize_device() -> None:
    """Device-agnostic synchronize via PyTorch's unified accelerator API."""
    torch.accelerator.synchronize()
```

This also simplifies the tests — no need to mock per-backend availability.

speculatorsbot, line 48: `torch.accelerator.synchronize()` is PyTorch's built-in device-agnostic synchronization API ([docs](https://docs.pytorch.org/docs/stable/generated/torch.accelerator.synchronize.html)). It is available in all PyTorch versions this project supports (`torch>=2.9.0`) and handles every accelerator backend (CUDA, NPU, XPU, MPS, etc.) without manual dispatch.

This entire helper could be replaced by calling `torch.accelerator.synchronize()` directly at the two call sites, which is both simpler and more general than maintaining a custom `if cuda / elif npu` chain.

```python
# In mark():
torch.accelerator.synchronize()

# In now():
torch.accelerator.synchronize()
```

Note: the `torch.accelerator.synchronize()` function is a no-op when no accelerator is initialized, so the "neither backend available" case is handled automatically.

speculatorsbot (commented):
This PR replaces hardcoded `torch.cuda.synchronize()` with a custom `_synchronize_device()` helper that checks CUDA, then falls back to NPU. The intent is right, but there is a simpler and more general solution already available in PyTorch.

Since `speculators` requires `torch>=2.9.0` (see `pyproject.toml`), [`torch.accelerator.synchronize()`](https://docs.pytorch.org/docs/stable/generated/torch.accelerator.synchronize.html) is always available. It handles CUDA, NPU, XPU, MPS, and any future accelerator backend — no custom dispatch needed. Using it would eliminate the `_synchronize_device()` helper entirely, the three new unit tests for its dispatch logic, and the maintenance cost of extending the `if/elif` chain for each new backend.

PR #819 (also open, by a different author) already takes this approach. I'd suggest coordinating with that PR to avoid duplicate work and converge on the `torch.accelerator` API.

PR #827, opened 2026-07-20
chore(evaluate): upgrade guidellm from 0.6.0 to 0.7.1

orestis-z:
## Summary

- Upgrade `guidellm` from pinned `==0.6.0` to `>=0.7.1` (3 releases behind, same vllm-project org)
- Adapt `run_guidellm()` CLI invocation to 0.7.x syntax: `benchmark` → `run`, typed `key=value` flags for `--backend`, `--data`, `--profile`, `--constraint`, `--output`
- Update JSON output parsers (`_load_json`, `parse_gen_len_file`) for new top-level structure (`config.spec.data` instead of `args.data_args`) and request stats (`output_metrics.text_tokens` instead of `output_tokens`)
- Replace `build_backend_args()` with `parse_gen_kwargs()` — backend args are now folded into the `--backend` flag
- Update `--data-column-mapper` default to guidellm 0.7.x typed format
- Add 24 unit tests covering all changed code paths (verified they fail against old code)

## Test plan

- [x] `ruff check` and `ruff format --check` pass
- [x] 24 unit tests pass (`pytest tests/unit/scripts/test_perf_utils.py --noconftest`)
- [x] Tests verified to fail against pre-upgrade code, confirming change-path coverage
- [ ] Run `python evaluate.py throughput --target <server>/v1 --subsets HumanEval` against a live vLLM server to verify end-to-end CLI construction and output parsing

### DIFF
--- a/scripts/evaluate/requirements.txt
+++ b/scripts/evaluate/requirements.txt
@@ -1,6 +1,6 @@
 # Core dependencies for performance benchmarking with guidellm
 vllm>=0.12.0
-guidellm>=0.7.1
+guidellm~=0.7.1
 
 # Visualization dependencies
 matplotlib
--- a/scripts/evaluate/requirements.txt
+++ b/scripts/evaluate/requirements.txt
@@ -1,6 +1,6 @@
 # Core dependencies for performance benchmarking with guidellm
 vllm>=0.12.0
-guidellm==0.6.0
+guidellm>=0.7.1
 
 # Visualization dependencies
 matplotlib
--- a/docs/user_guide/tutorials/evaluating_performance.md
+++ b/docs/user_guide/tutorials/evaluating_performance.md
@@ -39,7 +39,8 @@ Both `throughput` and `sweep` share the same options:
   --gen-len-rate N           Request rate for gen-len estimation (default: 128)
   --sweep-rate N             Number of sweep rate points (default: 10)
   --gen-kwargs JSON          Generation kwargs, e.g. '{"temperature":0.6}'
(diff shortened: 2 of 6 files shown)


speculatorsbot (commented):
LGTM. Clean, well-scoped upgrade with good test coverage. Agree with @coderabbitai's suggestion to cap the upper bound. Recommend approving.

scripts/evaluate/requirements.txt
```diff
@@ -1,6 +1,6 @@
 # Core dependencies for performance benchmarking with guidellm
 vllm>=0.12.0
-guidellm==0.6.0
+guidellm>=0.7.1
```
rahul-tuli: Couldn't >= resolve to another version with different usage? Should we pin to exactly 0.7.1 for now, since we've tested that?

rahul-tuli (commented):
Looks good pending two things:
1) Maybe we should pin `guidellm` to the exact version we've tested since the API could change b/w versions

2) nit: Since we've updated `--data-column-mapper` to text, I think it'll be nice if we can update our evaluate example to explicitly pass that arg, so users can see it's usage and potentially update for their own flows?

scripts/evaluate/requirements.txt
orestis-z: I agree. How about mid-ground `guidellm~=0.7.1`?  Which would allow patch updates.

rahul-tuli: That will be perfect!

rahul-tuli (approved):

PR #834, opened 2026-07-21
Enable FA4 Flex attention backend

fynnsu:
## Purpose
Pytorch added support for an Flash Attention 4 backend to Flex Attention which has much better performance on Hopper and Blackwell gpus: https://pytorch.org/blog/flexattention-flashattention-4-fast-and-flexible/

Requires manually installing fa4 first:

```
uv pip install flash-attn-4 --prerelease allow
```
or
```
uv pip install "flash-attn-4[cu13]" --prerelease allow
```

Note: this is experimental. 

By default, if the flash attention package is available and we're using a Hopper or newer gpu, this backend will get used. You can also explicitly enable/disable it with `--fa4 on` or `--fa4 off`

## Tests

I've added some explicit tests for FA4 on the Hopper runners. I've also updated the buildkite pipeline to install FA4 so that the tests will run. 

## Checklist

I have filled in:

### DIFF
--- a/src/speculators/models/attention.py
+++ b/src/speculators/models/attention.py
@@ -4,7 +4,11 @@ This module contains attention functions and utilities shared across different
 speculator architectures (EAGLE3, DFlash, etc.) to avoid code duplication.
 """
 
+import importlib.util
+import logging
 from collections.abc import Callable
+from functools import lru_cache
+from typing import Literal
 
 import torch
 from torch.nn.attention.flex_attention import (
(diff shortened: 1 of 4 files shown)


src/speculators/models/attention.py
```diff
@@ -16,6 +20,50 @@
 )
 from transformers.modeling_utils import AttentionInterface
 
+logger = logging.getLogger(__name__)
+
+
+@lru_cache(maxsize=1)
+def fa4_is_available() -> bool:
```
WindChimeRan, line 27: crashed on GB10 Blackwell (maybe it's an upstream bug)

`fa4_is_available()` returns True on GB10 (sm 12.1 ≥ 9.0 + package present), so --fa4 auto would enable FA4 and then crash at the first masked flex_attention, i.e., it breaks DFlash/Eagle-3 training on GB10 out of the box.

because: 

case | result on GB10
-- | --
FA4, dense (no mask) | runs fwd+bwd
FA4, causal mask | AssertionError: Block sparsity not supported on SM 12.0
FA4, DFlash anchor mask | AssertionError: Block sparsity not supported on SM 12.0

WindChimeRan (commented):
Could you please run some quick e2e benchmark on, e.g., 100 step speedup on a/b?

speculatorsbot (commented):
Clean, well-scoped addition. The implementation is minimal — extends the existing `flex_attention_forward` with an optional `kernel_options` parameter, adds the CLI plumbing, CI setup, and hardware-gated tests. No design concerns.

Agree with @WindChimeRan's finding that `fa4_is_available()` would return `True` on Blackwell GPUs (e.g. GB10, SM 12.x >= 9.0) where FA4's block-sparsity support is incomplete, causing masked `flex_attention` (causal, anchor) to crash under `--fa4 auto`. Recommend resolving that before merge.

WindChimeRan:
I tried this PR on a H100 machine:

On an H100 it makes the draft model's attention kernels 1–4% faster,  but attention is a small enough slice of a training step that end-to-end gain lands near 1%, and the default setting breaks DFlash training outright.

full report: 
https://claude.ai/code/artifact/e07582ec-5e60-4a26-813d-be0868d4fefb

PR #836, opened 2026-07-21
Add Mooncake hidden states extraction backend

fynnsu:
## Purpose

Adds a mooncake backend to the hs_connectors sub-package. This will enable training on multi-node setups, or single node without using disk as an intermediary.

Requires the `mooncake-transfer-engine` package. 

## Tests

Comprehensive testing is still in progress.

I have successfully completed a basic regression test training using this backend. 

## Checklist

I have filled in:


hs_connectors/src/hs_connectors/mooncake_hidden_states_connector.py
```diff
@@ -0,0 +1,329 @@
...
+        self._is_tp_rank_zero = get_tensor_model_parallel_rank() == 0
+
+        from vllm.model_executor.models.extract_hidden_states import (  # noqa: PLC0415
+            CacheOnlyAttentionLayer,
+        )
+
+        layers = get_layers_from_vllm_config(
+            self._vllm_config, CacheOnlyAttentionLayer, list(kv_caches.keys())
+        )
+        cache_layers = list(layers.keys())
```
speculatorsbot, line 197: The upstream `ExampleHiddenStatesConnector` validates that the derived block size matches the actual KV cache buffer shape in [`register_kv_caches`](https://github.com/vllm-project/vllm/blob/e3f9e32/vllm/distributed/kv_transfer/kv_connector/v1/example_hidden_states_connector.py#L313-L318):

```python
if self._block_size != self._kv_cache.shape[1]:
    raise ValueError(
        f"Hidden-states block-size mismatch: derived {self._block_size} "
        f"but buffer block size is {self._kv_cache.shape[1]}; read slots "
        "would be wrong (likely a hybrid block-size resolution bug)."
    )
```

This connector is missing that check. Since `_write_sample` computes slot offsets using `self._block_size` but `extract_from_kv_cache` indexes with `kv_cache.shape[1]`, a mismatch would silently produce wrong hidden states. Consider adding the same validation after `self._kv_cache = kv_caches[cache_layers[0]]`.

speculatorsbot (commented):
Design looks good -- the Mooncake connector follows the `ExampleHiddenStatesConnector` patterns closely (group/block-size resolution via spec, scheduler-side pending-saves, worker-side async extraction) and the backend registration through `HiddenStatesBackend.register` is clean. One missing safety check below.

WindChimeRan (commented):
I love the direction of `single node without using disk as an intermediary` and I think we should use mooncake as the default. 

However, I couldn't get it runnable (our A100 machine): 

My claude code's diagnosis: 

_0.3.12's prebuilt mooncake_master segfaults on startup, and the last working master (0.3.11) is manylinux_2_35-only — so mooncake can't run on glibc-2.34 hosts (RHEL 9); please ship a fixed master on a lower-glibc wheel or pin a known-good version._

---

Also Please run some benchmarks to show the speedup from the saving of the original disk lock data transfer.

rahul-tuli:
@fynnsu  #841 just landed — the Mooncake backend needs its args mirrored into the schema 👋

Heads up: #841 (config-file-first CLI) is now in `main`. It changes how `scripts/train.py` gets its flags — instead of calling each backend's `add_train_args`, the CLI is now **generated from a pydantic schema** (`src/speculators/train/config/schema.py`). `hs_connectors` stays argparse-based and standalone (so vLLM can keep using it), so each backend's train-args are now **mirrored** as schema fields.

This PR doesn't touch `scripts/train.py`, so it'll rebase onto `main` with **no git conflict** — but it'll be quietly broken until the mirror is added:

- `tests/unit/train/config/test_backend_reconciliation.py` will **fail** — it asserts every dest a backend registers via `add_train_args` has a matching schema field.
- At runtime the `--mooncake-*` flags won't exist (train.py's parser is schema-generated now), and `MooncakeBackend.from_train_args` reading `args.mooncake_*` would hit an `AttributeError`.

### The fix — mirror the 3 args into the schema

Your `MooncakeBackend.add_train_args` / `from_train_args` stay exactly as-is. You just need the same three knobs mirrored as schema fields so they survive resolution (field names must equal the argparse dests: `mooncake_master`, `mooncake_metadata_server`, `mooncake_protocol`). Two ways to do it — pick based on how many backends we expect. Note the *group* only affects `--help` grouping and `run.yaml` nesting; the flag names, flat dests, `from_train_args`, and the reconciliation test are identical either way.

**Option A — mirror into `DataArgs` (minimal, matches today's precedent).** Drop them next to the existing `hidden_states_path` (the `file` backend's mirror already lives there):

```python
# src/speculators/train/config/schema.py — in class DataArgs
    mooncake_master: str = Field(
        default="127.0.0.1:50051",
        description="Mooncake master server address. "
        "Used with --hidden-states-backend mooncake.",
    )
    mooncake_metadata_server: str = Field(
        default="P2PHANDSHAKE",
        description="Mooncake metadata server (or P2PHANDSHAKE). "
        "Used with --hidden-states-backend mooncake.",
    )
    mooncake_protocol: Literal["tcp", "rdma"] = Field(
        default="tcp",
        description="Mooncake transport protocol. "
        "Used with --hidden-states-backend mooncake.",
    )
```

Cheapest change, but `DataArgs` (dataset + dataloader config) starts to accumulate unrelated transport config as more backends land.

**Option B — a dedicated `BackendArgs` group (cleaner if more backends are coming).** Collect all backend-mirrored args in one group and move `hidden_states_path` in too, so the group is the single home for the mirror:

```python
# src/speculators/train/config/schema.py
class BackendArgs(_Group):
    """Per-backend transport args, mirrored from each HiddenStatesBackend's
    add_train_args (hs_connectors stays argparse-based). Only the fields for the
    selected --hidden-states-backend apply."""
    hidden_states_path: str | None = Field(default=None, ...)      # 'file' (moved from DataArgs)
    mooncake_master: str = Field(default="127.0.0.1:50051", ...)   # 'mooncake'
    mooncake_metadata_server: str = Field(default="P2PHANDSHAKE", ...)
    mooncake_protocol: Literal["tcp", "rdma"] = Field(default="tcp", ...)

# register in _GROUPS (order defines --help + run.yaml layout), right after "data"
    "data": DataArgs,
    "backend": BackendArgs,

# and add the field on TrainConfig
    backend: BackendArgs = Field(default_factory=BackendArgs)
```

More churn on the just-landed schema, and it shifts `hidden_states_path`'s `run.yaml` nesting from `data:` → `backend:` (trivial break — `run.yaml` only just shipped), but it's the natural "one group per concern" home and scales for the NIXL/RDMA backends on the roadmap.

**My take:** since the reconciliation test itself notes NIXL/RDMA are coming, I'd lean **B** (it's cheapest to extract now, with only one field to move) — but **A** is totally fine if you want to keep this PR minimal; just file a follow-up to extract `BackendArgs` when the second backend lands. Either way:
- `Literal["tcp", "rdma"]` auto-generates the argparse `choices=["tcp", "rdma"]`, matching your `--mooncake-protocol`. (`Literal` is already imported in `schema.py`.)
- **No change needed for `--hidden-states-backend mooncake`** — its choices come from the live `HiddenStatesBackend.registry`, so registering the backend makes `mooncake` a valid choice automatically.
- Neither option *gates* the flags (warn if `--mooncake-*` is set under `--hidden-states-backend file`). That'd mirror `_ALGORITHM_GROUP_USERS` / `_warn_mismatched_algorithm_blocks`; B is the natural foundation for it. Worth a follow-up, not needed here.

### Sanity check after adding them

```bash
python -m pytest tests/unit/train/config/test_backend_reconciliation.py
python scripts/train.py --help | grep mooncake   # all three flags show up
```

Ping me if you'd like the change as a ready-to-apply patch. Or if you'd like me to push the commit

hs_connectors/src/hs_connectors/mooncake_hidden_states_connector.py
```diff
@@ -0,0 +1,367 @@
...
+        mooncake_key = self._request_keys.pop(req_id, sanitize_key(req_id))
+
+        kv_params = request.kv_transfer_params or {}
+        if kv_params.get("include_output_tokens", False):
+            # Drop the final token: it was output, never an input to a forward
+            # pass, so its hidden state is not in the cache.
+            token_ids = torch.tensor(list(request.all_token_ids)[:-1])
+        elif request.prompt_token_ids is not None:
+            token_ids = torch.tensor(request.prompt_token_ids)
+        else:
```
shanjiaz, line 335: I feel like this case should never happen? Maybe we skip samples like this since we can't test alignment on such samples.

hs_connectors/src/hs_connectors/mooncake_hidden_states_connector.py
```diff
@@ -0,0 +1,367 @@
...
+        cache_layers = list(layers.keys())
+        assert len(cache_layers) == 1, (
+            f"Expected 1 CacheOnlyAttentionLayer, got {len(cache_layers)}"
+        )
+        self._kv_cache = kv_caches[cache_layers[0]]
+        self._copy_stream = torch.cuda.Stream()
+
+    def _get_copy_stream(self) -> torch.cuda.Stream:
+        """Return the dedicated copy stream, lazily creating it if needed."""
+        if self._copy_stream is None:
```
shanjiaz: why would this happen of we always create stream in `register_kv_caches`?

shanjiaz (approved):
Looks clean! I think speculators side support changes make sense. Left some small comments on mooncake_hidden_states_connector, but non-blocking. Only thing left is to make it compatible with the new train args system.

Please add an example and update docs once we've done tests on GB300

hs_connectors/src/hs_connectors/mooncake_hidden_states_connector.py
fynnsu: I think this case might get hit if the user sends a request without `return_token_ids=True`.

Our options are then:
1. Fail in the vllm connector and raise an error
2. Silently skip sending the data
3. Return some kind of marker (e.g. an empty tensor) in place of the token ids

I think 3 (the current behaviour) is best because then we can fail in the training process and notify that the prompt tokens received were empty. If we do 1 we block use cases that don't need this data (maybe not a big deal) and 2 will result in an ambiguous error since the training process will block and then timeout when it tries to receive this tensor.

hs_connectors/src/hs_connectors/mooncake_hidden_states_connector.py
fynnsu: Removed

hs_connectors/src/hs_connectors/mooncake_store.py
```diff
@@ -0,0 +1,123 @@
...
+    local_buffer_size: int = 2 * 1024 * 1024 * 1024
+    protocol: str = "tcp"
+    device_name: str = ""
+    num_writer_threads: int = 16
+
+    @classmethod
+    def from_dict(cls, d: dict | None) -> MooncakeStoreConfig:
+        d = d or {}
+        known = set(cls.__dataclass_fields__)  # type: ignore[attr-defined]
+        return cls(**{k: v for k, v in d.items() if k in known})
```
orestis-z, line 41: minor note: this silently ignores entries that don't exist in the config. Do we want to inform/warn the user?

hs_connectors/src/hs_connectors/mooncake_store.py
```diff
@@ -0,0 +1,123 @@
...
+            return self
+        try:
+            from mooncake.store import (  # type: ignore[import-not-found] # noqa: PLC0415
+                MooncakeDistributedStore,
+            )
+        except ImportError as e:  # pragma: no cover - optional dependency
+            raise ImportError(
+                "Mooncake is required for the Mooncake hidden-states backend. "
+                "Install it with `pip install mooncake-transfer-engine` or "
+                "`pip install mooncake-transfer-engine-cuda-13`."
```
orestis-z, line 71: nit: hardcoding cuda-13 means we have to update this here when there's a cuda upgrade. We could also just remove the `except ImportError` clause and let it raise the default import error.
nit: `raise ImportError(...) from e`

hs_connectors/src/hs_connectors/mooncake_store.py
```diff
@@ -0,0 +1,123 @@
...
+        if not raw:
+            return
+        names = json.loads(raw)
+        keys_to_remove = [f"{key}:{name}" for name in names] + [f"{key}:meta"]
+        self._store.batch_remove(keys_to_remove, force=True)
+
+    def get_sample(
+        self, key: str, timeout: float = 120.0, poll_interval: float = 0.05
+    ) -> dict[str, torch.Tensor]:
+        assert self._store is not None, "call setup() first"
```
orestis-z: minor: better practice to not use assert in production code but raise a value error instead. 

Edit: I just saw you added `S101` to pyproject.toml, but this is not an invariant since it depends on `setup` being called first, is it?

hs_connectors/src/hs_connectors/mooncake_store.py
```diff
@@ -0,0 +1,123 @@
...
+        names = json.loads(self._wait_for(f"{key}:meta", timeout, poll_interval))
+        result = {}
+        for name in names:
+            tensor = self._store.get_tensor(f"{key}:{name}")
+            if tensor is None:
+                raise RuntimeError(f"Mooncake tensor evicted for key={key}:{name}")
+            result[name] = tensor
+        return result
+
+    def _wait_for(self, key: str, timeout: float, poll_interval: float) -> bytes:
```
orestis-z, line 123: nit: check if poll_interval > 0

hs_connectors/src/hs_connectors/transfer.py
```diff
@@ -197,3 +199,132 @@ def build_kv_transfer_config(args: argparse.Namespace) -> dict[str, Any]:
...
+    def add_launch_args(parser: argparse.ArgumentParser) -> None:
+        MooncakeBackend._add_mooncake_args(parser)
+
+    @staticmethod
+    def from_train_args(
+        args: argparse.Namespace,
+        data_path: str,  # noqa: ARG004
+    ) -> MooncakeTransfer:
+        import os  # noqa: PLC0415
+        import socket  # noqa: PLC0415
```
orestis-z: I don't see why those are imported within this method and not module level. They have no import overhead and are built-in pkgs.
Same also below.

hs_connectors/src/hs_connectors/transfer.py
```diff
@@ -197,3 +199,132 @@ def build_kv_transfer_config(args: argparse.Namespace) -> dict[str, Any]:
...
+                    "metadata_server": mooncake_cfg.metadata_server,
+                    "master_server_address": mooncake_cfg.master_server_address,
+                    "global_segment_size": mooncake_cfg.global_segment_size,
+                    "local_buffer_size": mooncake_cfg.local_buffer_size,
+                    "protocol": mooncake_cfg.protocol,
+                    "device_name": mooncake_cfg.device_name,
+                    "num_writer_threads": mooncake_cfg.num_writer_threads,
+                }
+            },
+        }
```
orestis-z, line 308: do you think worth using a more "structured" datastructure here like a dataclass would add value or not applicable?

src/speculators/train/config/schema.py
```diff
@@ -680,15 +681,23 @@ def flatten(self) -> dict[str, Any]:
...
                 values[fname] = getattr(group, fname)
+        values.update(self._backend_args)
         return values
 
     @classmethod
     def from_flat(cls, flat: dict[str, Any]) -> "TrainConfig":
         """Inverse of :meth:`flatten`: recover the typed config from a flat
         working-dict. Non-config keys are dropped; the value validators are
         idempotent on already-resolved values."""
+        from speculators.train.config.resolution import _BACKEND_DESTS  # noqa: PLC0415
```
orestis-z: minor: it's not clear to me why we're not importing on module level here, maybe a small comment would shine some light (like heavy import only on demand or similar)

hs_connectors/src/hs_connectors/mooncake_store.py
```diff
@@ -0,0 +1,123 @@
...
+import torch
+
+
+@dataclass
+class MooncakeStoreConfig:
+    """Connection settings, passed straight to ``MooncakeDistributedStore.setup``."""
+
+    local_hostname: str = "localhost"
+    metadata_server: str = "http://localhost:8080/metadata"
+    master_server_address: str = "localhost:50051"
```
orestis-z: `50051` (and the others) could live in a constant like in `tests/e2e/smoke/test_mooncake_online_training.py` since they're also used elsewhere (e.g. `hs_connectors/src/hs_connectors/transfer.py`), to have single modification point

src/speculators/train/config/schema.py
```diff
@@ -680,15 +681,23 @@ def flatten(self) -> dict[str, Any]:
             group = getattr(self, gname)
             for fname in type(group).model_fields:
                 values[fname] = getattr(group, fname)
+        values.update(self._backend_args)
         return values
 
     @classmethod
     def from_flat(cls, flat: dict[str, Any]) -> "TrainConfig":
         """Inverse of :meth:`flatten`: recover the typed config from a flat
         working-dict. Non-config keys are dropped; the value validators are
```
orestis-z, line 690: Non-config keys are silently dropped, is this wanted or worth informing the user?

orestis-z (commented):
Looks great to me, my comments below are minor ignorable nits.

I checked those two 2 suggestions of `speculatorsbot` and `coderabbitai` and they seem valid, what are your thoughts?
- Missing block-size validation: https://github.com/vllm-project/speculators/pull/836#discussion_r3628613769
- Failed writes reported as done: https://github.com/vllm-project/speculators/pull/836#discussion_r3624590624

hs_connectors/src/hs_connectors/mooncake_store.py
fynnsu: I think we will deal with this when we need to. These are just the two versions available right now.

hs_connectors/src/hs_connectors/mooncake_store.py
fynnsu: Will log a warning.

hs_connectors/src/hs_connectors/transfer.py
fynnsu: Agreed. Fixing now

hs_connectors/src/hs_connectors/transfer.py
fynnsu: Yeah, the outer dict (with kv_connector, kv_role, etc) is defined by vLLM and I don't think is worthwhile trying to reproduce here, but I've replaced this with `"mooncake": dataclasses.asdict(mooncake_cfg)` so that we aren't manually unpacking the mooncake config.

src/speculators/train/config/schema.py
fynnsu: I think this is mostly used for record keeping (where we're just round-tripping the flattened args) and testing. I'm not entirely sold on this design either, but it seems like there are tests enforcing the keys are dropped, so presumably its intentional.

hs_connectors/src/hs_connectors/mooncake_store.py
fynnsu: Yeah, I think it's best to keep it this way. The ones we define here are default values for users. Maybe we could skip repeating them in tests but at least some of those tests are explicitly checking that they get set correctly.

hs_connectors/src/hs_connectors/mooncake_store.py
fynnsu: This is currently not user configurable (just hardcoded in) and this isn't a user facing function, so I think it's okay not to check

orestis-z (approved):
LGTM, conditioned on checking out bot comments on https://github.com/vllm-project/speculators/pull/836#pullrequestreview-4828872209

PR #839, opened 2026-07-21
docs: fix --target-layer-ids default and clarify aux/final layer split

WindChimeRan:
## The issue:

The two scripts disagree about what the flag's default is, and the disagreement is correct — each is right for its own side:

Script | Default layer ids | Count | Because
-- | -- | -- | --
launch_vllm.py | [2, n//2, n−3, n] | 4 | Captures the three auxiliary layers and the final layer n, since both are needed downstream.
train.py | [2, n//2, n−3] | 3 | Names only the auxiliary layers. The final layer arrives as a separate input, not as a member of this list.

## This PR: What

Docs and help text only; no behavior change.

## Detailed explaination:

https://claude.ai/code/artifact/32b15cef-e893-4816-9612-9a072bdb3cb3

## Tests

No behavior change, so no new tests. `ruff check`, `ruff format --check`, and `mdformat --check` pass. `python scripts/train.py --help` renders:

```
  --target-layer-ids TARGET_LAYER_IDS [TARGET_LAYER_IDS ...]
                        (Optional) A (space separated) list of integer layer
                        ids for the auxiliary hidden states. Defaults to [2,
                        num_hidden_layers // 2, num_hidden_layers - 3]. Note:
                        if custom values were used to launch vllm, pass the
                        same ids here, excluding the final layer
                        launch_vllm.py appends.
```

## Checklist

I have filled in:

### DIFF
--- a/docs/cli/launch_vllm.md
+++ b/docs/cli/launch_vllm.md
@@ -20,9 +20,9 @@ python scripts/launch_vllm.py meta-llama/Llama-3.1-8B-Instruct
 
 - **`--target-layer-ids`** (int list, default: auto-select) Space-separated list of integer layer IDs from which to capture hidden states. Note: if `--include-last-layer` is enabled (default), the model's last layer will be appended to this list. Default: `[2, num_layers//2, num_layers-3]`
 
-  **Important:** If set, you must also pass the same layer ids to the training script using `--target-layer-ids`.
+  **Important:** If set, you must also pass the same layer ids to the training script using `--target-layer-ids`, **excluding** the final layer — training takes the auxiliary layers only. For the [full example](#full-example) below, that is `--target-layer-ids 5 20 40`.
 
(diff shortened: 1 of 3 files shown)


shanjiaz (approved):
Thanks, this bit is indeed confusing.

WindChimeRan:
@shanjiaz This PR is to just fix a wrong doc. But the flags are still confusing after this PR. 

In the long run, I think we should not explicitly expose this flag to the users. And since this is a flag across stages, the plumbing refactoring should be happened in `speculators-cli`

shanjiaz:
This flag is available to the users, but I'm not sure if we should.

WindChimeRan:
I think we should allow users to change them (for research purpose, there could be a reason on, e.g., better feature fusion?)

But most of the users don't care about it (low level implementation detail). A consistent default across stages should be fine to them.

PR #841, opened 2026-07-22
feat(train): config-file-first CLI for scripts/train.py

### LINKED ISSUE #833: [RFC]:  Config-file-first CLI for `scripts/train.py`
### Motivation.

Training a drafter is driven by `scripts/train.py`, whose tunable surface has grown past **80 flat `argparse` flags** declared in a single `parse_args`, with `main()` reading `args.*` directly and threading them as `**kwargs` into the five `SpeculatorModel` classes. That shape hurts three ways today:

- **Reproducibility.** There is no first-class, re-runnable artifact of *the exact configuration that produced a checkpoint*. A run is reconstructed after the fact by re-reading a bash script and hoping its variables were not edited since — there is nothing next to the checkpoint that says how it was made.
- **Discoverability.** A wall of ~80 flat flags gives no sense of what belongs together or what a given run actually needs. `--help` is one long undifferentiated list.
- **Maintainability.** Every new knob is another hand-written `add_argument`, and the flag set drifts easily from how `main()` consumes it.

The real workflow makes this concrete. A user copies an `examples/train/*.sh` **recipe**, edits ~15 bash variables in its `# ==== Configuration ====` block, and the script threads them into a curated ~15-flag invocation of `train.py`; the other ~65 flags stay at their defaults. So a per-run configuration file *already exists* — it just lives in bash, covers only a slice of the surface, and produces no durable record of the resolved run. We should make that configuration a first-class, typed, re-runnable artifact without breaking a single existing recipe.

This RFC covers `scripts/train.py` only.

### Proposed Change.

Make `train.py` **config-file-first** while keeping the full flag surface backward-compatible. A run is described by an optional YAML file, overridable on the CLI, and every run writes back a clean, re-runnable copy of the configuration it actually resolved to.

1. **`--config run.yaml` (new).** A run is specified by an optional YAML config file. The file is **stage-shaped**: everything nests under a top-level `train:` key, so a future pipeline config (`prepare_data:`, `launch_vllm:`) extends the same file rather than replacing it. Loading is lenient — a bare top-level mapping is still accepted — and unknown keys warn and are ignored rather than silently accepted.
2. **All ~80 flags remain, byte-for-byte (existing).** Every flag is generated from a single typed schema (one typed field == one flag, with its type / choices / bool-style / help), shown in a flat, **grouped** `--help`. Existing `examples/train/*.sh` recipes and workflows keep working unchanged.
3. **Precedence `flag > YAML > default` (new).** Two explicit layers over the schema default — no third layer, no escape hatch. Predictable: a flag always beats the file, the file always beats the default.
4. **`--dump-config` (new).** Prints the fully-resolved configuration as a valid, `train:`-wrapped config file, so a working invocation can be turned into a `run.yaml` to version and share.
5. **Reproducibility artifacts (new).** Each run writes, next to its checkpoints at rank 0: a resolved **`run.yaml`** (clean, re-runnable via `--config`, no provenance inlined) and a **`train_command.txt`** manifest (exact argv, git SHA, world size, and `speculators` / `transformers` / `torch` versions). Re-running an experiment becomes `--config run.yaml`.
6. **One public type, `TrainConfig` (new).** The configuration subsystem presents essentially one type that parses, layers, validates, and serializes itself. `train.py` collapses to `main(TrainConfig.resolve())`.
   - `TrainConfig.resolve(argv)` — the impure CLI boundary: parse argv, layer sources, validate, handle `--dump-config`, convert a config error into a clean `SystemExit(2)`.
   - `TrainConfig.from_sources(*, cli, config_path, argv)` — the **pure, argv-free, exception-raising core** and primary test seam. Validation runs here, including rejecting conflicting draft-init options (`--from-pretrained` vs `--draft-config` vs decoder-shaping flags) whether the conflict is expressed on the CLI or entirely inside the YAML file.
   - `cfg.flatten()` — the **anti-corruption adapter**: returns the `vars(args)`-shaped flat dict the five `SpeculatorModel` classes already consume via `**kwargs`, with byte-identical ordering to today's parser. The model layer is **not touched**.

#### Old flow

```mermaid
flowchart LR
    argv["argv<br/>(~80 flat flags)"] --> parse["parse_args()<br/>single argparse"]
    parse --> ns["args<br/>(flat Namespace)"]
    ns --> main["main()<br/>reads args.* directly"]
    main --> models["5 SpeculatorModel classes<br/>(**kwargs)"]
    main -. "no artifact" .-> void["(nothing written:<br/>run is not reproducible)"]
```

#### New flow

```mermaid
flowchart LR
    cfg["--config run.yaml"] --> resolve
    flags["CLI flags"] --> resolve
    resolve["TrainConfig.resolve(argv)"] --> fs["from_sources()<br/>layer: defaults -> YAML -> flags<br/>(flag &gt; YAML &gt; default)<br/>validate"]
    fs --> tc["TrainConfig<br/>(typed groups)"]
    tc --> flat["cfg.flatten()<br/>anti-corruption adapter"]
    flat --> models["5 SpeculatorModel classes<br/>(unchanged, **kwargs)"]
    tc --> save["cfg.save(dir)"]
    save --> art["run.yaml + train_command.txt<br/>(reproducible)"]
```

The contrast is the point: today `argv` flows straight through a flat `Namespace` into the model classes and nothing durable is written; the proposal routes the same inputs through one typed `TrainConfig` that layers sources, validates once, hands the model layer the *identical* flat dict through an anti-corruption adapter, and writes back the resolved config so the run is reproducible.

### Any Other Things.

**Backward compatibility is the binding constraint.** Every existing `examples/train/*.sh` invocation must run unchanged; passing no `--config` at all behaves identically to today.

**Test plan.** The suite is anchored in real happy paths rather than a synthetic flag-by-flag snapshot: each `examples/train/*.sh` recipe's invocation is run through the new config layer and its flags asserted to resolve to the expected values — so any flag a recipe relies on is guarded against a silent rename, drop, or retype, and the schema stays the single source of truth for the long tail of flags no recipe exercises. On top of that, at the pure `from_sources(...)` seam: layering/precedence (`flag > YAML > default`) resolves to the right winner; `dump_yaml()` → `--config` round-trips to an identical resolved config; and validation/error fidelity holds (a broken `--config` → clean exit 2; a missing required value names the flag to set; conflicting draft-init options are rejected whether set on the CLI or in YAML; a mismatched algorithm block warns). Process-level behavior (`--dump-config` output, exit codes) is checked at the `resolve(argv)` boundary.

**Considered but out of scope:**

- **`typer` for the generated CLI** — considered for its cleaner grouped `--help`, but `typer`/`click` cannot reproduce the current space-separated `nargs='+'` flags (`--full-attention-indices 2 18 33`, `--target-layer-ids …`) without click-level surgery beneath its vendored parser — a back-compat break under "zero breakage beats polished look." The schema generates `argparse` flags instead.
- **A pydantic-native / `pydantic-settings` CLI** (its built-in argv parsing as the surface) — rejected: it does not reproduce the exact flat flag surface (dotted group names, space-separated `nargs`, the current bool style) byte-for-byte, so it too breaks recipes. (`pydantic-settings` is still used internally for source ordering.)
- **Phase 2** — pushing typed groups into the five `SpeculatorModel` classes and retiring the flat-dict adapter. Separate future PR.
- **Whole-pipeline config** — one file driving `prepare_data` + `launch_vllm` + `train`. The `train:` wrapper only reserves room for it.
- **Env-var / dotenv config sources** — the trainer is configured via CLI + YAML only.

rahul-tuli:
## Purpose

Implements the config-file-first CLI RFC (#833): replaces the ~600-line hand-written `argparse` surface in `scripts/train.py` with a typed, layered configuration system in a new `src/speculators/train/config/` package built on `pydantic-settings`.

This is **phase 1** of the migration. The schema is now the single source of truth for every `train.py` tunable, and `main()` bridges to the existing model layer through a single `argparse.Namespace(**cfg.flatten())` adapter (retired in a later phase). No new dependencies — pydantic, pydantic-settings, and PyYAML are already in the stack.

> **Reviewing?** The four commits are each self-contained and pass their own tests — read commit-by-commit using the **Reviewer reading order** below. `scripts/train.py` is a net **−660 lines** of parser; `main()`'s only changes are the `flatten()` bridge and dropping a now-redundant `float16` runtime guard (the rejection lives in the schema validator), so it can be skimmed.

## What's new for users

- `--config PATH` — load a stage-shaped YAML config; precedence is `flag > yaml > default`.
- `--dump-config` — print the fully-resolved config as a round-trippable `run.yaml` and exit.
- Every run writes `run.yaml` + `train_command.txt` next to its checkpoints, so a run reloads via `--config run.yaml` without hand-reconstructing flags.

### Example

Existing flag-based recipes are unchanged — every flag, env var, and default is preserved:

```bash
scripts/train.py \
  --data-path ./data --save-path ./ckpt \
  --speculator-type eagle3 --epochs 3 --lr 1e-4 --total-seq-len 4096
```

`--dump-config` emits only the values a non-default layer supplied (stage-shaped, clean of provenance):

```yaml
train:
  trainer:
    save_path: ./ckpt
    epochs: 3
  data:
    total_seq_len: 4096
    data_path: ./data
  optimizer:
    lr: 0.0001
  speculator_type: eagle3
  verifier:
    verifier_name_or_path: meta-llama/Llama-3.1-8B-Instruct
```

That file reloads identically via `--config run.yaml`, and re-dumping it is **byte-stable** (emission follows the pinned `flatten()` order, independent of flag-vs-yaml source).

## Generated CLI (`--help`)

The flag surface is generated from the schema — one typed field becomes one flag, in a `--help` grouped by concern (`general`, `verifier`, `draft`, `data`, `generation`, `loss`, `optimizer`, `scheduler`, `trainer`, `logging`, and the per-algorithm groups `dflash` / `dspark` / `peagle` / `mtp`). `--config` / `--dump-config` lead the surface; every other flag, choice, bool style, and default is preserved from the legacy parser.

```bash
$ scripts/train.py --help
```

<details>
<summary>Full generated <code>--help</code> output</summary>

```text
usage: train.py [-h] [--config PATH] [--dump-config]
                [--speculator-type SPECULATOR_TYPE] [--dry-run] [--seed SEED]
                [--deterministic-cuda]
                [--verifier-name-or-path VERIFIER_NAME_OR_PATH]
                [--trust-remote-code] [--from-pretrained FROM_PRETRAINED]
                [--draft-config DRAFT_CONFIG] [--num-layers NUM_LAYERS]
                [--draft-arch {llama,qwen3}]
                [--draft-hidden-act DRAFT_HIDDEN_ACT]
                [--draft-mrope-full-head-hack | --no-draft-mrope-full-head-hack]
                [--target-layer-ids TARGET_LAYER_IDS [TARGET_LAYER_IDS ...]]
                [--token-freq-path TOKEN_FREQ_PATH]
                [--draft-vocab-size DRAFT_VOCAB_SIZE] [--d2t-path D2T_PATH]
                [--t2d-path T2D_PATH] [--mask-token-id MASK_TOKEN_ID]
                [--norm-before-residual | --no-norm-before-residual]
                [--embed-requires-grad | --no-embed-requires-grad]
                [--norm-before-fc | --no-norm-before-fc] [--fc-norm]
                [--norm-output | --no-norm-output]
                [--sliding-window SLIDING_WINDOW]
                [--full-attention-indices FULL_ATTENTION_INDICES [FULL_ATTENTION_INDICES ...]]
                [--sliding-window-non-causal]
                [--draft-attn-impl {simple_flex_attention,sdpa,eager}]
                [--data-path DATA_PATH] [--hidden-states-backend {file}]
                [--hidden-states-path HIDDEN_STATES_PATH]
                [--total-seq-len TOTAL_SEQ_LEN]
                [--train-data-ratio TRAIN_DATA_RATIO] [--noise-std NOISE_STD]
                [--legacy-data] [--hidden-states-dtype HIDDEN_STATES_DTYPE]
                [--num-workers NUM_WORKERS]
                [--prefetch-factor PREFETCH_FACTOR]
                [--max-anchors MAX_ANCHORS] [--vllm-endpoint VLLM_ENDPOINT]
                [--on-missing {generate,skip,warn,raise}]
                [--on-generate {cache,delete}]
                [--request-timeout REQUEST_TIMEOUT]
                [--max-retries MAX_RETRIES] [--loss-fn LOSS_FN]
                [--ttt-steps TTT_STEPS]
                [--ttt-step-loss-decay TTT_STEP_LOSS_DECAY]
                [--optimizer {adamw,muon}] [--lr LR]
                [--weight-decay WEIGHT_DECAY] [--muon-lr MUON_LR]
                [--muon-momentum MUON_MOMENTUM]
                [--muon-weight-decay MUON_WEIGHT_DECAY]
                [--muon-ns-steps MUON_NS_STEPS]
                [--muon-adjust-lr-fn {original,match_rms_adamw}]
                [--scheduler-type {linear,cosine,none}]
                [--scheduler-warmup-steps SCHEDULER_WARMUP_STEPS]
                [--scheduler-warmup-ratio SCHEDULER_WARMUP_RATIO]
                [--scheduler-total-steps SCHEDULER_TOTAL_STEPS]
                [--scheduler-num-cosine-cycles SCHEDULER_NUM_COSINE_CYCLES]
                [--epochs EPOCHS] [--checkpoint-freq CHECKPOINT_FREQ]
                [--save-best] [--no-resume-from-checkpoint]
                [--log-freq LOG_FREQ] [--save-path SAVE_PATH] [--fsdp-shard]
                [--logger LOGGER] [--log-dir LOG_DIR] [--run-name RUN_NAME]
                [--block-size BLOCK_SIZE]
                [--sample-from-anchor | --no-sample-from-anchor]
                [--dflash-decay-gamma DFLASH_DECAY_GAMMA]
                [--per-position-loss-weight {fixed-exp-decay,dpace}]
                [--dpace-alpha DPACE_ALPHA] [--markov-rank MARKOV_RANK]
                [--markov-head-type {vanilla,gated,rnn}]
                [--enable-confidence-head | --no-enable-confidence-head]
                [--confidence-head-with-markov | --no-confidence-head-with-markov]
                [--confidence-head-alpha CONFIDENCE_HEAD_ALPHA]
                [--num-depths NUM_DEPTHS]
                [--down-sample-ratio DOWN_SAMPLE_RATIO]
                [--down-sample-ratio-min DOWN_SAMPLE_RATIO_MIN]
                [--num-speculative-steps NUM_SPECULATIVE_STEPS]
                [--step-weight-beta STEP_WEIGHT_BETA]

options:
  -h, --help            show this help message and exit
  --config PATH         YAML config file (stage-shaped: trainer keys under a
                        top-level 'train:' key; a bare mapping is also
                        accepted). CLI flags override it.
  --dump-config         Print the fully-resolved config as a valid run.yaml to
                        stdout and exit, so a working invocation can be
                        scaffolded into a shareable config.

general:
  --speculator-type SPECULATOR_TYPE
                        Type of speculator model to train (eagle3, dflash,
                        dspark, peagle, mtp).
  --dry-run             Build the speculator, initialize weights, save a
                        checkpoint to --save-path, then exit before training.
                        Useful to validate the config and weights (e.g. in
                        vLLM) before a full run. Can be combined with --draft-
                        config or --from-pretrained.
  --seed SEED           Random seed for reproducibility.
  --deterministic-cuda  Set cuda to deterministic mode. This may impact
                        performance.

verifier:
  --verifier-name-or-path VERIFIER_NAME_OR_PATH
                        Path or HF id of the verifier/target model.
  --trust-remote-code   Allow executing code from HF Hub when loading the
                        verifier's tokenizer.

draft:
  --from-pretrained FROM_PRETRAINED
                        Path or HF id of a pretrained draft. May also point to
                        a local directory containing only a config.json, in
                        which case a fresh draft is initialized from that full
                        speculator config. Takes precedence over and is
                        mutually exclusive with --draft-config and the
                        decoder-shaping flags (--num-layers, --draft-arch,
                        --draft-hidden-act, --sliding-window, --full-
                        attention-indices).
  --draft-config DRAFT_CONFIG
                        HF id, directory, or JSON path of a decoder config
                        (LlamaConfig for eagle3/peagle, Qwen3Config for
                        dflash) to use as the draft transformer_layer_config;
                        the rest of the speculator is built from the other CLI
                        args. Mutually exclusive with --from-pretrained and
                        the decoder-shaping flags (--num-layers, --draft-arch,
                        --draft-hidden-act, --sliding-window, --full-
                        attention-indices).
  --num-layers NUM_LAYERS
                        Number of draft decoder layers to synthesize.
  --draft-arch {llama,qwen3}
                        Architecture for draft decoder layers (default:
                        'llama' for eagle3, 'qwen3' otherwise).
  --draft-hidden-act DRAFT_HIDDEN_ACT
                        Activation function for draft decoder layers. Defaults
                        to 'silu'. Qwen3 layers of dflash expect 'silu' for
                        vLLM deployment. Leave as None to fall back to the
                        verifier's activation function.
  --draft-mrope-full-head-hack, --no-draft-mrope-full-head-hack
                        For MRoPE configs with partial_rotary_factor < 1,
                        rescale mrope_section and set
                        partial_rotary_factor=1.0 so HF training and vLLM
                        inference use equivalent full-head rotary semantics.
  --target-layer-ids TARGET_LAYER_IDS [TARGET_LAYER_IDS ...]
                        (Optional) space-separated list of integer layer ids.
                        Defaults to [2, n//2, n-3, n]. Must be set explicitly
                        if custom values were used to launch vllm.
  --token-freq-path TOKEN_FREQ_PATH
                        Path to token frequency distribution file (.pt). Used
                        with --draft-vocab-size to build vocab mappings at
                        training time. Falls back to '<data-
                        path>/token_freq.pt'. If neither exists nor --draft-
                        vocab-size is set, vocab mapping is skipped and the
                        full verifier vocab is used.
  --draft-vocab-size DRAFT_VOCAB_SIZE
                        Vocabulary size for the draft model. Must be provided
                        together with a token frequency file to generate vocab
                        mappings; otherwise a no-op.
  --d2t-path D2T_PATH   Draft-to-target vocab mapping file (.npy).
  --t2d-path T2D_PATH   Target-to-draft vocab mapping file (.npy).
  --mask-token-id MASK_TOKEN_ID
                        Token id used to mask positions during training.
                        Resolved from the verifier tokenizer when unset.
  --norm-before-residual, --no-norm-before-residual
                        Toggle normalization before residual connections
                        (default: True).
  --embed-requires-grad, --no-embed-requires-grad
                        Whether to train embedding layer weights (default:
                        False).
  --norm-before-fc, --no-norm-before-fc
                        Apply a single RMSNorm to the concatenated auxiliary
                        hidden states before the FC projection (gpt-oss
                        style). See --fc-norm for the per-layer alternative
                        from the Eagle 3.1 paper. (default: True for eagle3,
                        False otherwise).
  --fc-norm             Apply per-layer RMSNorm to each auxiliary hidden state
                        before concatenation and FC projection (Eagle 3.1
                        paper approach).
  --norm-output, --no-norm-output
                        Feed post-norm hidden states back across TTT steps to
                        stabilize magnitude drift across speculation depths
                        (default: True for eagle3, False otherwise).
  --sliding-window SLIDING_WINDOW
                        Sliding window size for sliding window attention
                        layers (default: 2048). All draft layers use sliding
                        window by default (except mtp).
  --full-attention-indices FULL_ATTENTION_INDICES [FULL_ATTENTION_INDICES ...]
                        (Optional) space-separated draft layer indices that
                        should use full attention instead of sliding window.
                        All draft layers use sliding window by default (except
                        mtp). (e.g. '--full-attention-indices 0 2' makes
                        layers 0 and 2 use full attention; the rest use
                        sliding window).
  --sliding-window-non-causal
                        Use non-causal (bidirectional) masking within draft
                        blocks for sliding window attention layers. Full
                        attention layers are always bidirectional. Note: vLLM
                        currently doesn't support these models.
  --draft-attn-impl {simple_flex_attention,sdpa,eager}
                        Attention implementation for draft layers. Use 'sdpa'
                        or 'eager' for hardware that doesn't support flex
                        attention. Not supported for MTP.

data:
  --data-path DATA_PATH
                        Root data directory with the preprocessed dataset,
                        vocab mappings (d2t.npy, t2d.npy), token frequencies
                        (token_freq.pt), and hidden states.
  --hidden-states-backend {file}
                        Hidden states transfer backend. Each backend may add
                        its own CLI arguments. Default: 'file'.
  --hidden-states-path HIDDEN_STATES_PATH
                        Path where cached hidden states are stored (default:
                        <data-path>/hidden_states). Contributed by the 'file'
                        backend.
  --total-seq-len TOTAL_SEQ_LEN
                        Maximum training sequence length, in tokens.
  --train-data-ratio TRAIN_DATA_RATIO
                        Fraction of the dataset used for training; the
                        remainder is held out for validation.
  --noise-std NOISE_STD
                        Standard deviation for noise augmentation.
  --legacy-data         DEPRECATED. Use the old data format which stores
                        hidden states alongside token_ids and assistant_masks
                        in data_i.pt files. Will be removed soon.
  --hidden-states-dtype HIDDEN_STATES_DTYPE
                        Data type for dataloader hidden states and autocast
                        compute. Model master weights are always kept in fp32.
                        Options: float32, bfloat16 (recommended). float16 is
                        not supported (requires gradient scaling).
  --num-workers NUM_WORKERS
                        Number of dataloader workers.
  --prefetch-factor PREFETCH_FACTOR
                        Dataloader prefetch factor.
  --max-anchors MAX_ANCHORS
                        Maximum anchor positions for DFlash, DSpark, and
                        P-EAGLE training (default: 3072).

generation:
  --vllm-endpoint VLLM_ENDPOINT
                        vLLM endpoint used to generate hidden states on
                        demand. Only needed if --on-missing=generate and
                        samples are missing. The vLLM instance must cache
                        hidden states to a location reachable from the
                        training instance.
  --on-missing {generate,skip,warn,raise}
                        Dataloader behaviour when a sample has no cached
                        hidden states. Default 'generate' generates them on
                        demand via the vLLM endpoint; the others skip the
                        sample, skip with a warning, or raise.
  --on-generate {cache,delete}
                        Behaviour after generating a hidden state (only if
                        --on-missing=generate). 'delete' discards it once
                        loaded; 'cache' stores it in the hidden states path,
                        enabling hybrid online/offline training.
  --request-timeout REQUEST_TIMEOUT
                        Timeout in seconds for each individual vLLM request.
                        Only applies if --on-missing=generate.
  --max-retries MAX_RETRIES
                        Maximum retry attempts per vLLM request on failure.
                        Only applies if --on-missing=generate.

loss:
  --loss-fn LOSS_FN     Loss function specification. A name (kl_div, rkl, jsd,
                        ce, tv, nla, lk_hybrid) or a JSON dict for a weighted
                        combination, e.g. '{"ce": 0.1, "tv": 0.9}'.
  --ttt-steps TTT_STEPS
                        Number of test-time-training (TTT) steps the draft is
                        unrolled over during training.
  --ttt-step-loss-decay TTT_STEP_LOSS_DECAY
                        Multiplicative loss weight decay applied per TTT step
                        (1.0 = weight every step equally).

optimizer:
  --optimizer {adamw,muon}
                        Optimizer to use. 'muon' applies Muon to 2D weight
                        matrices and AdamW to the remaining params (norms,
                        biases, embeddings, lm_head).
  --lr LR               Learning rate (AdamW / base group).
  --weight-decay WEIGHT_DECAY
                        Weight decay for the AdamW optimizer (and the AdamW
                        group in muon mode).
  --muon-lr MUON_LR     LR for the Muon (2D weights) group. Only used with
                        --optimizer muon. Defaults to 10*lr (and --lr defaults
                        to 1e-4).
  --muon-momentum MUON_MOMENTUM
                        Momentum for the Muon group.
  --muon-weight-decay MUON_WEIGHT_DECAY
                        Weight decay for the Muon group.
  --muon-ns-steps MUON_NS_STEPS
                        Newton-Schulz iteration steps for Muon.
  --muon-adjust-lr-fn {original,match_rms_adamw}
                        Muon LR adjustment. 'match_rms_adamw' matches AdamW's
                        update RMS.

scheduler:
  --scheduler-type {linear,cosine,none}
                        LR scheduler type.
  --scheduler-warmup-steps SCHEDULER_WARMUP_STEPS
                        Warmup steps (default: scheduler-dependent).
  --scheduler-warmup-ratio SCHEDULER_WARMUP_RATIO
                        Warmup as a fraction of total scheduler steps, in [0,
                        1]. Ignored (with a warning) when --scheduler-warmup-
                        steps is also set.
  --scheduler-total-steps SCHEDULER_TOTAL_STEPS
                        Total scheduler steps (default: inferred).
  --scheduler-num-cosine-cycles SCHEDULER_NUM_COSINE_CYCLES
                        Number of cosine cycles for the cosine scheduler.

trainer:
  --epochs EPOCHS       Number of training epochs.
  --checkpoint-freq CHECKPOINT_FREQ
                        Save a checkpoint every N epochs. Values < 1 enable
                        sub-epoch checkpointing (e.g. 0.5 = every half epoch).
  --save-best           Also point a checkpoint at the lowest validation loss.
  --no-resume-from-checkpoint
                        Do not resume training from an existing checkpoint.
  --log-freq LOG_FREQ   Log training metrics every N steps (default: 1).
  --save-path SAVE_PATH
                        Directory to write checkpoints and the resolved
                        run.yaml.
  --fsdp-shard          Shard model parameters across GPUs with FSDP. By
                        default parameters are fully replicated (DDP-like).
                        Enable when the model does not fit in a single GPU's
                        memory.

logging:
  --logger LOGGER       One of 'trackio', 'wandb', 'tensorboard', 'mlflow' or
                        a comma separated list.
  --log-dir LOG_DIR     Directory for training logs.
  --run-name RUN_NAME   Name for this run (used by metric loggers). Auto-
                        generated when unset.

dflash:
  --block-size BLOCK_SIZE
                        Block size for DFlash model (default: 8).
  --sample-from-anchor, --no-sample-from-anchor
                        Sample from the anchor position (all positions
                        predict). Default: False for dflash, True for dspark.
  --dflash-decay-gamma DFLASH_DECAY_GAMMA
                        Decay gamma for DFlash/DSpark loss weighting.
  --per-position-loss-weight {fixed-exp-decay,dpace}
                        Per-position loss weight option for D-PACE support
                        (default: fixed-exp-decay).
  --dpace-alpha DPACE_ALPHA
                        Smoothing constant for the D-PACE loss (default: 0.5).
                        Must be in (0, 1] when --per-position-loss-
                        weight=dpace.

dspark:
  --markov-rank MARKOV_RANK
                        DSpark: low-rank dim of the Markov logit-bias head (0
                        disables it).
  --markov-head-type {vanilla,gated,rnn}
                        DSpark: sequential head variant.
  --enable-confidence-head, --no-enable-confidence-head
                        DSpark: attach the per-position acceptance confidence
                        head.
  --confidence-head-with-markov, --no-confidence-head-with-markov
                        DSpark: feed the Markov previous-token embedding into
                        the confidence head alongside the backbone hidden
                        state.
  --confidence-head-alpha CONFIDENCE_HEAD_ALPHA
                        DSpark: weight of the confidence-head BCE term.

peagle:
  --num-depths NUM_DEPTHS
                        Number of parallel prediction depths for P-EAGLE
                        (default: 8).
  --down-sample-ratio DOWN_SAMPLE_RATIO
                        Geometric decay ratio for COD sampling in P-EAGLE.
  --down-sample-ratio-min DOWN_SAMPLE_RATIO_MIN
                        Minimum retention ratio for COD sampling in P-EAGLE.

mtp:
  --num-speculative-steps NUM_SPECULATIVE_STEPS
                        Number of MTP prediction steps (default: 3).
  --step-weight-beta STEP_WEIGHT_BETA
                        Exponential decay factor for MTP step weights; higher
                        weights earlier prediction steps more. Only used with
                        MTP.
```

</details>

## Back-compat

- Every existing flag, env, and default is preserved. `flatten()` derives the flat `vars(args)`-shaped dict the model layer consumes directly from the schema (consumers bind by name), so a new tunable is a one-field edit and there is no hand-maintained key list to keep in sync.
- The example recipes under `examples/train/*.sh` are replayed through `resolve()` as the central back-compat guard — a renamed/dropped/retyped flag breaks a real recipe.
- `hs_connectors` stays speculators-agnostic (argparse-based, usable by vLLM standalone). Backend train-args are mirrored as schema fields; `test_backend_reconciliation.py` fails loudly if a backend registers a train-arg with no matching schema field.
- **Reproducibility anchor.** `run.yaml` intentionally emits only non-default values, so it is *not* pinned against future changes to the schema defaults. `train_command.txt` (which records the exact argv **plus the resolving git SHA**, world size, and library versions) is the durable anchor: a reload always names the code version it was produced against.

Closes #833

## Reviewer reading order

1. **`feat(train/config): typed schema as the single source of truth`** — `schema.py`. The typed groups and the `flatten()`/`from_flat()` seam. `flatten()` derives the flat surface straight from the schema (declaration order), so adding a tunable is a one-field edit. Value validators (e.g. the `hidden_states_dtype` `float16` rejection) run at construction, so bad values fail as clean config errors.
2. **`feat(train/config): generate the CLI from the schema; layered resolution`** — `resolution.py`. Argparse generation from annotations (choices/required/bool markers are declarative via `json_schema_extra`), the `flag > yaml > default` walk, provenance, the post-merge required-flag check, draft-init conflict check, and the backend reconciliation guard.
3. **`feat(train/config): reproducibility artifacts`** — `artifacts.py`. YAML dump + `run.yaml`/`train_command.txt` writers; only non-default values are emitted, in the pinned order, so the artifact round-trips byte-stably (and `dump_yaml()` also handles a `from_flat` config, which carries no provenance).
4. **`refactor(train): cut train.py over to TrainConfig; drop legacy argparse`** — the `scripts/train.py` diff (mostly deletions, −660 net) plus removal of the now-orphaned `argparse_utils` helper.

## Validation

- `pytest tests/unit/train/config/` → **65 passed** (schema / resolution / artifacts / backend-reconciliation seams).
- Full `tests/unit/train/` → **221 passed**; the only 2 failures (`test_distributed_resume[ddp]/[fsdp]`) are pre-existing and unrelated — they fail identically on the pre-refactor commit with the same `initial_lr` scheduler-resume error.
- Back-compat guard: all six `examples/train/*.sh` recipes replay through `resolve()` cleanly (`-k recipe` → 6 passed).
- Round-trip verified: `--dump-config` → reload via `--config` → re-dump is **byte-identical** (regression test: `test_dump_yaml_is_byte_stable_across_reload`).
- `ruff check` / `ruff format --check` → clean · `mypy src/speculators/train/config/` → no issues · `scripts/train.py --help` renders the generated grouped help.
- Each of the four commits was checked out individually and imports + passes its own tests.

## Follow-up (phase 2)

- The `argparse.Namespace(**cfg.flatten())` adapter in `main()` is a deliberate phase-1 seam. Phase 2 passes the typed config into the model layer directly and retires the adapter; it is deferred to a separate PR to keep this diff focused on the CLI/resolution boundary.
- **Uniform bool override (deferred, intentional).** Only `embed_requires_grad` among the False-default bools is tagged `BooleanOptionalAction`, so a `run.yaml` that sets another False-default bool `true` (e.g. `save_best`, `fsdp_shard`) has no CLI off-switch. Blanket-tagging every False-default bool would make CLI-over-YAML override uniform, but it is *not* free: `no_resume_from_checkpoint` would generate a confusing double-negative `--no-no-resume-from-checkpoint`. Left as a deliberate follow-up rather than shipping that UX wrinkle.

## Checklist


rahul-tuli:
Tracked in JIRA: https://issues.redhat.com/browse/INFERENG-9334

**JIRA Details:**
- Issue Type: Task
- Priority: Major
- Component: Speculators
- Activity Type: Tech Debt & Quality
- Fix Version: Speculators 0.7.0
- Assignee: Rahul Tuli
- Status: Backlog

src/speculators/train/config/schema.py
```diff
@@ -0,0 +1,769 @@
...
+            raise ValueError(
+                "hidden_states_dtype must be a dtype attribute of torch, e.g. "
+                "`bfloat16`."
+            )
+        if getattr(torch, v) is torch.float16:
+            raise ValueError(
+                "hidden_states_dtype float16 is not supported (requires gradient "
+                "scaling); use bfloat16."
+            )
+        return v
```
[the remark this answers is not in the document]
rahul-tuli: Addressed!

src/speculators/train/config/artifacts.py
```diff
@@ -0,0 +1,76 @@
...
+
+import yaml
+
+from .schema import nest_flat
+
+if TYPE_CHECKING:
+    from .schema import TrainConfig
+
+
+def dump_yaml(cfg: "TrainConfig") -> str:
```
orestis-z, line 18: do we use "string-typing" because there's a circular import or similar?

orestis-z (commented):
tested on a high level and looks good to me, didn't dig into the implementation details yet but looks very clean on first sight.

2 questions: 
- due to many cli args the `--help` output doesn't fit into my terminal, can we instead enter some interactive scrolling (similar to a man page)? (could also be done in a follow up)
- did we decide or are we planning to add cli group support, e.g. --draft.from_pretrained=xyz in a follow-up?

src/speculators/train/config/artifacts.py
rahul-tuli: Yes, you are correct

rahul-tuli:
> tested on a high level and looks good to me, didn't dig into the implementation details yet but looks very clean on first sight.
> 
> 2 questions:
> 
> * due to many cli args the `--help` output doesn't fit into my terminal, can we instead enter some interactive scrolling (similar to a man page)? (could also be done in a follow up)
> * did we decide or are we planning to add cli group support, e.g. --draft.from_pretrained=xyz in a follow-up?

I haven't looked into 1, but we can add both things in a follow up, for now they out of scope for this PR

src/speculators/train/config/resolution.py
```diff
@@ -0,0 +1,455 @@
...
+        origin = get_origin(annotation)
+    if origin is list:
+        (inner,) = get_args(annotation)
+        return inner, True, None
+    if origin is Literal:
+        choices = list(get_args(annotation))
+        elem_types = {type(choice) for choice in choices}
+        base = elem_types.pop() if len(elem_types) == 1 else str
+        return base, False, choices
+    return annotation, False, None
```
rahul-tuli, line 143: nit: Convert to a `NamedTuple`

orestis-z (approved):
LGTM, thanks a lot for walking me through all 🫶.

Only optional minor one is to change from warning to error for params not related to the method used.

Issue #757, opened 2026-07-09
RFC: Ablation Study — Domino vs DSpark Performance Comparison


orestis-z:
## Motivation

Both Domino (#685, [arXiv:2605.29707](https://arxiv.org/abs/2605.29707)) and DSpark add sequential/causal conditioning on top of the DFlash parallel draft backbone, but through fundamentally different mechanisms:

- **Domino**: GRU-based correction head producing additive logit deltas on suffix positions, with a dual-loss lambda curriculum that transitions from 100% backbone loss to 100% refined loss.
- **DSpark**: Low-rank factorized Markov logit-bias head (`W1 @ W2`, rank 256) + optional confidence head (per-position acceptance predictor trained with BCE).

Prior experiments have compared these methods, but never under identical conditions (same dataset, hyperparameters, verifier, and evaluation protocol). Weifan's DPace ablations (#736) and the v6 Domino experiments used different datasets and settings, making direct comparison unreliable. This RFC proposes a controlled ablation to resolve that.

Additionally, the v6 Domino experiments showed that the lambda curriculum improved the **backbone itself** by +9.1% MAL even when the GRU head was dropped at inference — meaning sequential conditioning during training can act as a regularizer that benefits the parallel backbone. Understanding whether DSpark's Markov head produces a similar backbone-strengthening effect is a key question.

**Inference support in vLLM:**
- **DSpark**: Fully supported in vLLM — `DSparkSpeculator` loads the Markov head and performs sequential left-to-right sampling with Markov bias at each step (`vllm/v1/worker/gpu/spec_decode/dspark/`).
- **Domino**: WIP/draft support in vLLM (vllm-project/vllm#48241), not yet merged. Pending testing with a training checkpoint.

**Note on parameter and compute overhead:** The Domino GRU head and DSpark Markov head add different amounts of parameters on top of the DFlash backbone. More importantly, they have different computational profiles: the GRU requires sequential state updates (inherently less parallelizable), while DSpark's factorized Markov bias is a single embedding lookup + matmul. Parameter counts and drafting latency will both be reported alongside quality metrics.

### Shared Configuration

```
--verifier-name-or-path  Qwen/Qwen3-8B
--num-layers             5
--target-layer-ids       1 9 17 25 34
--block-size             8
--max-anchors            3072
--draft-vocab-size       32000
--sliding-window         2048        # SWA default
--optimizer              muon        # Muon default
--lr 1e-4  --muon-lr 1e-3
[...]

PR #846, opened 2026-07-22
fix(dspark): detach confidence head inputs to prevent gradient leakage

orestis-z:
## Summary

Detach `hidden_blocks` and `prev_emb` before feeding them to the confidence head, so its BCE loss only trains the confidence head's own projection — not the backbone or Markov head.

Without `.detach()`, the confidence BCE gradients leak into the backbone transformer and Markov W1 embedding, pushing them toward acceptance-rate prediction instead of generation quality. This explains the A3 vs A4 regression in ablation #757 (1.96 vs 3.20 MAL).

## Test plan

- [x] `ruff check` passes
- [x] DSpark model definition tests pass (8/8)
- [x] Gradient flow test confirms detach blocks backprop without affecting the main loss path

### DIFF
--- a/src/speculators/models/dspark/core.py
+++ b/src/speculators/models/dspark/core.py
@@ -172,14 +172,15 @@ class DSparkDraftModel(DFlashDraftModel):
             )
 
         if self.confidence_head is not None:
-            # confidence_head_with_markov requires markov_rank > 0 (enforced in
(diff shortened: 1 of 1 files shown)


orestis-z:
Duplicate of https://github.com/vllm-project/speculators/pull/848

PR #847, opened 2026-07-22
Add a minimal Verifiers agentic regeneration demo

mgoin:
## What this demonstrates

- run upstream agent environments through the stable Verifiers CLI
- keep the integration in a small UV project pinned to `verifiers==0.2.0` and an exact research-environments commit
- provide Prolog, LiveCodeBench, and R2E-Gym example profiles
- convert native Verifiers traces into exact `input_ids` and `loss_mask`
- replay those exact IDs through vLLM for aligned hidden-state collection

The boundary is intentionally narrow: Verifiers owns environments, tools, Docker, turns, and scoring. Speculators does not add an environment wrapper or rollout loop.

## Live demonstration

A Qwen3-8B Prolog run produced a six-turn trajectory with six assistant tool calls and five tool observations. The unsuccessful `max_turns` trace contained 5,708 exact tokens (4,914 sampled), and replay returned an exact token match with BF16 hidden states shaped `(5708, 4, 4096)`.

The zero reward is retained intentionally because failed trajectories are still valid on-policy data.

## Scope

This omits generated rollout artifacts and does not lock ordinary transitive Python dependencies. The compatibility boundary is expressed through the direct pins in `pyproject.toml`.

## Validation

- 77 passed, 2 skipped
- Ruff passed
- all three Verifiers profiles pass CLI dry-run validation
- `uv sync --project scripts/agentic_regeneration` succeeds


WindChimeRan (commented):
I think this PR conflates several data pipeline scripts. 

* scripts/launch_vllm.py this is for online training hidden state generation, so it’s very slow, and we should not use it for regen.
* regen actually uses scripts/response_regeneration/run_all.sh
* Prepare and replay exact IDs —> this is for off-policy data preparation… we shouldn’t include it for any regen training. 

This is indeed a very confusing UX. 😅 But it's a pre-existing bad design. 

I suggest we fix the UX first. 

cc:  @fynnsu @shanjiaz

PR #852, opened 2026-07-23
Add simple benchmarking script

fynnsu:
## Purpose

A way to compare different implementations and determine if there are real speed improvements.

Accepts real training args and uses them to setup a Trainer object. Then runs a fixed number of steps and records metrics. 

```
# Synthetic benchmark (no dataset / vLLM needed)
python scripts/benchmark.py run --synthetic \\
    -- --verifier-name-or-path Qwen/Qwen3-8B --total-seq-len 4096

# Real data benchmark
python scripts/benchmark.py run \\
    -- --verifier-name-or-path Qwen/Qwen3-8B --data-path ./output \\
    --on-missing skip

# Multi-GPU
torchrun --standalone --nproc_per_node 2 scripts/benchmark.py run \\
    --synthetic -- --verifier-name-or-path Qwen/Qwen3-8B

# Compare two runs
python scripts/benchmark.py compare baseline.json candidate.json
```

## Tests

Manually tested a simple comparison

## Checklist

I have filled in:

### DIFF
--- a/src/speculators/train/trainer.py
+++ b/src/speculators/train/trainer.py
@@ -118,6 +118,7 @@ class TrainerConfig(NamedTuple):
     hidden_states_dtype: torch.dtype = torch.bfloat16
     log_freq: int = 1
     fsdp_shard: bool = False
+    max_steps: int | None = None
 
 
 def _resolve_scheduler_steps(
@@ -501,6 +502,12 @@ class Trainer:
                     extra={"step": self.global_step},
                 )
             self.global_step += 1
+
+            if (
+                self.config.max_steps is not None
+                and self.global_step >= self.config.max_steps
+            ):
+                break
 
             if (
                 step_interval is not None
--- a/scripts/benchmark.py
+++ b/scripts/benchmark.py
@@ -1,0 +1,667 @@
+#!/usr/bin/env python3
+"""Training benchmark harness for speculators.
+
+Measures training loop throughput, timing breakdown, and GPU memory usage
+with full provenance tracking. Results are stored as JSON for comparison
+across code changes.
+
+Uses the real Trainer class so measurements stay in sync with the actual
+training code path.
+
+Subcommands:
+    run       Run a training benchmark
+    compare   Compare two benchmark result files
+
+Examples:
+    # Synthetic benchmark (no dataset / vLLM needed)
+    python scripts/benchmark.py run --synthetic \\
+        -- --verifier-name-or-path Qwen/Qwen3-8B --total-seq-len 4096
+
+    # Real data benchmark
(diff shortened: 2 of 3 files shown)


src/speculators/train/trainer.py
```diff
@@ -502,6 +503,12 @@ def train_epoch(self, epoch: int):
                 )
             self.global_step += 1
 
+            if (
+                self.config.max_steps is not None
+                and self.global_step >= self.config.max_steps
+            ):
+                break
+
```
fynnsu, line 511: This is needed to allow for early stopping during benchmark runs.

speculatorsbot (commented):
**Design-level assessment**: The approach is sound. Reusing the real `Trainer` for benchmarking is the right call -- it keeps measurements in sync with the production code path and avoids a parallel implementation that would drift. The `max_steps` addition to `TrainerConfig` is clean and backward-compatible (defaults to `None`, no existing callers affected). The `_MetricCapture` logging handler is a lightweight way to intercept profiles without modifying the Trainer. The `compare` subcommand is a useful addition.

**One concern: missing tests.** The PR adds 667 lines of new code with no unit tests. The project has an established testing norm for this kind of code -- there is a `tests/unit/scripts/` directory and unit tests for trainer internals (e.g., `test_step_timer.py` for the `_StepTimer` class that this benchmark relies on). The benchmark introduces several pure functions that are straightforward to unit test:

- `compute_statistics` -- verify mean/std/min/max/median/count on known inputs, edge case with a single value (`stdev` fallback to 0.0)
- `compare_benchmarks` -- feed two JSON fixtures and assert the output includes correct deltas/percentages and comparability warnings
- `_MetricCapture.emit` -- verify it captures profile dicts and ignores non-profile log records
- `create_synthetic_batch` -- verify shapes, dtypes, and device placement of all returned tensors
- Warmup/measured split logic -- verify `all_profiles[warmup_steps:]` produces the correct subset and the fallback warning fires when profiles are fewer than expected

CONTRIBUTING.md requires unit tests for new features. Could you add coverage for the above scenarios?

---

shanjiaz (approved):
This looks good! Example run:
```
Metric                 Mean        Std        Min        Max
----------------------------------------------------------
step_ms              175.08       2.05     170.61     180.08
fwd_ms                65.77       1.73      63.64      70.78
bwd_ms                56.11       0.55      55.10      57.73
opt_ms                43.72       1.05      42.02      46.76
fetch_ms               9.48       0.43       9.23      11.61
tokens_per_s       23398.18     272.96   22745.84   24008.10

Peak memory: 25091.4 MB allocated, 26842.0 MB reserved
```
I think we might need to update our docs for this? @fynnsu

PR #855, opened 2026-07-24
Add `CODEOWNERS`

dsikka:
Adds .github/CODEOWNERS to define review ownership. GitHub uses last-match-wins: the most specific matching path decides reviewers, not a union of all matches. The file is ordered broad → specific accordingly.

Hierarchy:

  1. Package-wide defaults — `/src/speculators/`,`/tests/`, `/scripts/` owned by all core maintainers (@fynnsu @shanjiaz @rahul-tuli @orestis-z), acting as a
  catch-all.
  2. Subsystem overrides — convert/, data_generation/, train/ narrow to whoever owns that subsystem.
  3. Algorithm-specific overrides — convert/mtp/, convert/eagle/, models/mtp/, models/peagle/, models/eagle3/, models/dflash/ route to the individual
  method's owner(s), overriding both subsystem and package-wide rules.
  4. Tests mirror source — `/tests/` follows the same pattern, with e2e/regression/ and e2e/smoke/ also adding @dsikka.
  5. Infra/process files — `.buildkite/`, `.github/`, and root config/docs (README.md, pyproject.toml, etc.) owned separately by @dsikka plus relevant
  maintainers.

Example: a PR touching src/speculators/models/mtp/layer.py matches three rules — `/src/speculators/` (line 4),
`/src/speculators/models/` (line 9), and `/src/speculators/models/mtp/` (line 10) — but only the last, most specific match
applies, so reviewers are @rahul-tuli @fynnsu, not the full set of four maintainers.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,41 @@
+# See https://help.github.com/articles/about-codeowners/
+# for more info about CODEOWNERS file
+
+/src/speculators/ @fynnsu @shanjiaz @rahul-tuli @orestis-z
+/src/speculators/convert/ @shanjiaz @fynnsu @rahul-tuli
+/src/speculators/convert/mtp/ @rahul-tuli @fynnsu
+/src/speculators/convert/eagle/ @shanjiaz @fynnsu
+/src/speculators/data_generation/ @shanjiaz @fynnsu @rahul-tuli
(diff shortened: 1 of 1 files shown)


shanjiaz (approved):
Thanks for adding this! I think the rules make sense 🫡

PR #860, opened 2026-07-24
Restore the --max-steps training flag

eldarkurtic:
## Summary

The config-file-first CLI refactor (#841) moved every `scripts/train.py` flag into the typed schema but did not carry over **`--max-steps`**. `TrainerConfig.max_steps` and the training loop's early-stop check (`self.global_step >= self.config.max_steps`) still exist, but the flag was unreachable from the CLI and left unset in `main()`, so `max_steps` was effectively dead code — `--max-steps N` now errors with `unrecognized arguments: --max-steps`.

This restores it the schema-native way:

- Add a single `max_steps: int | None` `Field` (with `ge=1`) to `TrainerArgs` in `src/speculators/train/config/schema.py`, so the flag, type, `--help`, and validation are all derived from it automatically.
- Wire it into the `TrainerConfig` built in `main()` (`max_steps=args.max_steps`).

No behavior change to the training loop itself — it already honored `TrainerConfig.max_steps`; this just makes it reachable again.

## Test plan

- `python -m pytest tests/unit/train/test_cli_args.py tests/unit/train/config` — 88 passing (adds default/explicit/reject-non-positive cases for `--max-steps`).
- `make quality` (ruff + format + mypy) clean.
- Manual: `--max-steps 15` → `cfg.trainer.max_steps == 15`; omitted → `None`; `--max-steps 0` → clean config error; flag appears in `--help`.

### DIFF
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -675,6 +675,7 @@ def main(cfg: TrainConfig):  # noqa: C901
         hidden_states_dtype=hidden_states_dtype,
         log_freq=args.log_freq,
         fsdp_shard=args.fsdp_shard,
+        max_steps=args.max_steps,
     )
     trainer = Trainer(draft_model, trainer_config, train_loader, val_loader)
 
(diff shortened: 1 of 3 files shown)


shanjiaz (approved):
Good catch!

fynnsu (approved):

PR #867, opened 2026-07-26
fix(train): balance sampler sample counts and bound validation rank skew

eldarkurtic:
## Summary

- **Sampler rotation**: the LPT bin-packer in `MultipackDistributedBatchSamplerV2` tie-breaks on slot id when bins are equally full, so slot 0 always grabs the largest sample and the highest slot absorbs leftover small ones. Over a full epoch this creates a **2.4× sample-count skew** across ranks. The fix rotates the slot→rank mapping per batch (`target_slot = (rank - rotation) % num_replicas`), cycling which rank plays each role. Sample-count ratio drops from 2.40× to 1.00×; token balance stays at 1.002×; shards remain disjoint; batch counts are unchanged.

- **Validation sync barriers**: after PR #813 removed per-batch validation all-reduce, cross-rank speed differences in the validation loop accumulate unboundedly. With the sample-count skew above, the fastest rank finishes validation minutes before the slowest, and the final metrics `all_reduce` trips the NCCL watchdog (10 min default) → **validation hang**. A new `_maybe_val_sync` method issues a `dist.barrier()` every 50 validation batches, bounding drift well below the watchdog window.

**Root cause**: LPT slot-id tie-breaking bias → 2.4× val sample-count skew → no per-batch sync (post-#813) → drift exceeds NCCL watchdog → hang at end-of-epoch validation.

## Test plan

- [x] 19 unit tests for sampler balance (`test_batch_sampler_balance.py`): sample balance, token balance, disjoint partitions, equal batch counts, token budget, rotation mechanism — 5 of these fail on the unfixed sampler
- [x] 8 unit tests for validation sync (`test_val_sync.py`): barrier placement at interval boundaries only, never at batch 0, disabled when single-process or interval=0, deterministic count
- [x] Fast-val reproduction: built a dataset with the exact val split (10k rows, 1864 batches/rank) but tiny train split — reproduced the hang on unfixed code, confirmed fix eliminates it
- [x] Full 100k-sample ablation (accum=1 and accum=4, 3-GPU DDP, ~8h each): both complete validation without hanging

### DIFF
--- a/src/speculators/train/distributed_batch_sampler.py
+++ b/src/speculators/train/distributed_batch_sampler.py
@@ -41,11 +41,16 @@ class _Bin(NamedTuple):
     """Helper named tuple for `lpt_packed_batch`"""
 
     fill: int  # sum of items in _Bin
-    rank: int  # global rank _Bin is associated with
+    slot: int  # heap slot id (0..num_replicas-1)
 
 
 def _lpt_packed_batch(
-    lengths: np.ndarray, max_len: int, num_replicas: int, start_index: int, rank: int
+    lengths: np.ndarray,
+    max_len: int,
+    num_replicas: int,
+    start_index: int,
+    rank: int,
+    rotation: int,
 ) -> None | list:
     """
     Check if lengths can be distributed into `num_replicas` machines with at most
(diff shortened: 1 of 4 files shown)


eldarkurtic:
FYI: before this PR, Qwen3-8B DDP training 1 GPU (vllm server) + 3 GPUs (training) would crash at validation step. After this PR, the experiment runs fine without any issues.

eldarkurtic:
@coderabbitai full review

eldarkurtic:
@coderabbitai full review

fynnsu (approved):
Looks good! Thanks for catching and fixing this

PR #868, opened 2026-07-26
fix(train): derive train/val split boundaries from one expression

eldarkurtic:
## Summary

- The `ArrowDataset` split logic computed the train boundary as `int(n * ratio)` and the val boundary as `int(n * (ratio - 1.0))`. These two expressions disagree in floating point at certain ratios — e.g. at `ratio=0.2` with `n=12501`: `int(12501 * 0.2) == 2500` but `int(12501 * -0.8) → int(12501 * 0.2000…3) == 2501` — putting **one row in both splits** (silent train/val leakage).

- The fix replaces the `split_ratio: float` parameter with `train_ratio: float` + `split: Literal["train", "val"]`. Both splits derive their boundary from the single expression `int(len(data) * train_ratio)`, guaranteeing exact complementarity. Input validation rejects out-of-range ratios and val splits at `train_ratio=1.0`.

## Test plan

- [x] 34 unit tests (`test_split_boundaries.py`): exact complementarity across 10 ratios (including 0.2 which was broken), no row in both splits, union covers full dataset, default takes whole dataset, rejects out-of-range, rejects val at ratio 1.0
- [x] Verified the 0.2 ratio case produces off-by-one on unfixed code, exact complement on fixed code


eldarkurtic:
@coderabbitai full review

fynnsu (approved):

PR #871, opened 2026-07-27
Add Ulysses SP for Training

Roderick-Wu:
## Purpose

** updated description to help with review, got claude to write a lot of it but I reviewed this

Add ulysses sequence parallelism(SP) for speculator training. Long-context training is memory-bound by activations that scale with sequence length — Ulysses SP shards the packed sequence across multiple GPUs using all-to-all communication, reducing per-GPU activation memory proportionally.

Enabled with `--sp-size <N>` (default 1, off). Supported for Eagle3, PEagle, DFlash, and DSpark. Works with both DDP and FSDP (`--fsdp-shard`).

This version uses scoped DDP/FSDP + explicit gradient hooks, which cleanly separates the sum-across-SP and average-across-DP reduction semantics.

## Design

SP groups use contiguous ranks, DP groups use strided ranks. For example, with `world_size=4, sp_size=2`:

```
SP groups: {0,1}, {2,3}    (process same batch, different seq chunks)
DP groups: {0,2}, {1,3}    (process different batches, same seq position)
```

A 2D `DeviceMesh("dp", "sp")` is created when `sp_size > 1` so FSDP can scope its sharding to the DP sub-mesh only. 

DDP/FSDP all-reduce uses averaging, but SP partial gradients must be summed (each rank sees a fraction of the sequence, not a fraction of the batch). These are different operations on different axes:

- `register_sp_gradient_hooks` uses `post_accumulate_grad_hook` to SUM gradients across the SP group (handles `DTensor` from FSDP transparently)
- DDP/FSDP is scoped to the DP group, where averaging is correct

The loss function (`loss_function` in `metrics.py`) all-reduces the numerator and denominator across the SP group in the forward pass (detached from autograd) so every SP rank computes the same scalar loss. The gradient hooks handle the backward side.

### Per-model SP strategy

Models differ in how they handle the sequence dimension, captured by the `_sp_splits_batch` ClassVar:

| Model | `_sp_splits_batch` | Strategy |
|-------|-------------------|----------|
| **Eagle3** | `True` (default) | Trainer splits batch; model sees local seq chunk for all tensors except `document_ids` (kept full for global position info) |
| **DFlash** | `False` | Handles SP internally — builds attention mask at full dims on rank 0, broadcasts anchors, then partitions anchors per SP rank. Uses a custom `dflash_flex_attention` attention impl that handles the context+noise layout. |
| **DSpark** | `False` (inherits DFlash) | Inherits DFlash's SP slicing. Derives `num_blocks` from `anchored_block_indices` shape instead of `max_anchors` arg (correct under SP partitioning). |
| **PEagle** | `False` | COD samples are generated on rank 0 and broadcast to SP group. Samples are padded for divisibility and partitioned, with a `pad_mask` to zero out padded positions in the loss. |

### Attention all-to-all

The Ulysses all-to-all is integrated into `flex_attention_forward`:

1. Q/K/V arrive in sequence-parallel layout: `(B, H, S_local, D)`
2. `ulysses_scatter` transposes to head-parallel layout: `(B, H/sp, S_full, D)` — each rank gets a subset of heads but the full sequence
3. `flex_attention` runs with the full sequence visible per head (standard causal/block mask)
4. `ulysses_gather` transposes back: `(B, H, S_local, D)`

KV heads are replicated when `num_kv_heads < sp_size` (via `maybe_replicate_kv_heads`) to ensure every SP rank gets at least one KV head after the scatter.

### Batch splitting

`split_batch_for_sp` in the trainer splits the batch along the sequence dimension before each forward pass (for models with `_sp_splits_batch=True`). Keys in `_SPLIT_KEYS` (hidden_states, verifier_last_hidden_states, input_ids, loss_mask, position_ids) are chunked; keys in `_KEEP_FULL_KEYS` (document_ids) are kept full. A `full_seq_len` sentinel is injected for downstream code that needs the global length.

## Usage

```bash
# 8 GPUs, sp_size=2: sequences split across pairs of GPUs
torchrun --nproc_per_node=8 scripts/train.py \
  --sp-size 2 \
  --fsdp-shard \
  --draft-attn-impl simple_flex_attention \
  --total-seq-len 65536 \
  ...
```

Constraints:
- `--total-seq-len` must be divisible by `--sp-size`
- `world_size` must be divisible by `--sp-size`
- `--draft-attn-impl` must be `simple_flex_attention` (eager/sdpa don't go through the SP-aware attention path)
- `--max-anchors` must be divisible by `--sp-size` (DFlash/DSpark only)

## Tests

- 15 unit tests for `split_batch_for_sp` and `maybe_replicate_kv_heads`
- 9 unit tests for `normalize_counted_metrics`
- 5 multi-GPU tests: SP group initialization (sp_size=1 and sp_size=2), all-to-all round-trip correctness, all-to-all backward gradient correctness, SP gradient hook all-reduce
- E2E verified on all four architectures (Eagle3, DFlash, DSpark, PEagle) with memory reduction confirmed

## Checklist


src/speculators/train/trainer.py
```diff
@@ -455,6 +465,8 @@ def train_epoch(self, epoch: int):
                 else v
                 for k, v in batch.items()
             }
+            if getattr(self.model, "_sp_splits_batch", True):
+                gpu_batch = split_batch_for_sp(gpu_batch, get_sp_rank(), get_sp_size())
```
speculatorsbot: **Bug: `_sp_splits_batch` is invisible through the DDP wrapper, causing incorrect batch splitting for DFlash, DSpark, and PEagle under DDP + SP.**

When `self.config.fsdp_shard` is `False`, `_setup_model_ddp` reassigns `self.model` to a `DistributedDataParallel` wrapper (line 346). DDP's `__getattr__` only delegates lookups for `_parameters`, `_buffers`, and `_modules` — it does not forward arbitrary class-level attributes to the wrapped module. So `getattr(self.model, "_sp_splits_batch", True)` always returns the default `True` for a DDP-wrapped model, regardless of what the underlying model class defines.

DFlash, DSpark, and PEagle all set `_sp_splits_batch = False` because they receive the full batch and do their own internal sequence slicing. With this bug, the trainer calls `split_batch_for_sp` on the batch *before* passing it to these models, so they receive a pre-split batch and then try to slice it *again* — resulting in either an index-out-of-bounds crash or silently processing the wrong data.

The FSDP path is not affected because `fully_shard` modifies the model in-place without wrapping it, so the class attribute remains accessible.

Fix: access the underlying module through `self.model.module` for DDP, or resolve the flag once during setup:

```python
# In setup_model or __init__:
self._sp_splits_batch = getattr(self.model, "_sp_splits_batch", True)

# In train_epoch / val_epoch:
if self._sp_splits_batch:
    gpu_batch = split_batch_for_sp(gpu_batch, get_sp_rank(), get_sp_size())
```

This same issue is present in `val_epoch` on line 567.

speculatorsbot (commented):
Overall this is a well-structured implementation of Ulysses sequence parallelism. The core all-to-all primitives, the custom op with autograd backward, the DFlash-specific asymmetric attention handling, and the PEagle COD sampling broadcast/partition logic all look correct. The approach of scoping FSDP to the DP sub-mesh while using post-accumulate-grad hooks for the SP gradient all-reduce is sound.

I found one bug that affects the DDP path for DFlash, DSpark, and PEagle (all models with `_sp_splits_batch = False`). See inline comment.

src/speculators/train/trainer.py
rahul-tuli: +1 I think DDP wraps the model into another abstraction so `self._sp_splits_batch` will always be `True`?

src/speculators/models/dflash/core.py
```diff
@@ -359,42 +406,40 @@ def _backbone_forward(
...
         with torch.no_grad():
-            if anchored_block_indices.numel() < total_seq_len:
-                target_indices = (
-                    anchored_block_indices
-                    if self.config.sample_from_anchor
-                    else (anchored_block_indices - 1) % total_seq_len
-                )
-                targets = self.verifier_lm_head(
-                    self.verifier_norm(verifier_last_hidden_states[:, target_indices])
-                )
```
rahul-tuli, line 389: Why did we delete this branch? Do we always need all verifier_logits?

.gitignore
```diff
@@ -173,4 +173,7 @@ cython_debug/
 .DS_Store
 
 # Visual Studio Code
-.vscode/
\ No newline at end of file
+.vscode/
+
+output
+output/*
```
rahul-tuli: nit: Remove?

src/speculators/train/trainer.py
```diff
@@ -303,6 +308,7 @@ def _setup_model_fsdp(self, load_checkpoint: bool):
             full_state_dict = self.model.state_dict()
 
         apply_fully_sharded(self.model, param_dtype=self.config.hidden_states_dtype)
+        self._sp_grad_hooks = register_sp_gradient_hooks(self.model)
```
rahul-tuli, line 318: Do we ever cleanup these hooks? Shouldn't we?

src/speculators/train/distributed.py
```diff
@@ -127,6 +136,13 @@ def _init_sp_process_groups(rank: int, world_size: int, sp_size: int) -> None:
     _dp_size = dp_size
     _dp_rank = rank // sp_size
 
+    if sp_size > 1:
+        _device_mesh = init_device_mesh(
+            "cuda",
```
rahul-tuli: nit: Should we use `torch.accelerator.current_accelerator().type` to stay model agnostic

src/speculators/train/sequence_parallel.py
```diff
@@ -0,0 +1,297 @@
...
+                f"num_kv_heads ({num_kv_heads}) must be divisible by "
+                f"sp_size ({sp_size})"
+            )
+        return key, value
+
+    if sp_size % num_kv_heads != 0:
+        raise ValueError(
+            f"sp_size ({sp_size}) must be divisible by "
+            f"num_kv_heads ({num_kv_heads}) when sp_size > num_kv_heads"
+        )
```
rahul-tuli, line 184: nit: Should we also check if `num_attention_heads % sp_size == 0` for the query all-to-all, or is it fine?

rahul-tuli (commented):
I did not run the full code e2e, but looks good on first pass pending minor nits and questions. Great Job on this!

src/speculators/models/dflash/core.py
```diff
@@ -68,7 +71,18 @@ def __init__(
...
+        from speculators.train.distributed import get_sp_size  # noqa: PLC0415
+
+        # Fix for config, so that the model trained with sp. Reset config before saving
+        orig_attn_impl = config.transformer_layer_config._attn_implementation  # noqa: SLF001
+        if get_sp_size() > 1:
+            config.transformer_layer_config._attn_implementation = (  # noqa: SLF001
+                "dflash_flex_attention"
+            )
+
         super().__init__(config=config)
```
speculatorsbot: **Config restore undoes the runtime attention dispatch, not just the serialization.**

`PreTrainedModel.__init__` stores `self.config = config` by reference (no copy). The restore on line 84 therefore also changes `self.config.transformer_layer_config._attn_implementation` back to the original value. Since every `Qwen3DFlashDecoderLayer` shares this same config object, at forward time `Qwen3DFlashAttention.forward` dispatches to `ALL_ATTENTION_FUNCTIONS["simple_flex_attention"]` -- i.e., the generic `flex_attention_forward` -- instead of `dflash_flex_attention_forward`.

The generic path handles asymmetric Q/KV (`kv_len > q_len`) by splitting K into chunks of `q_len`. For DFlash, K is `[context | noise]` where `ctx_len != noise_len` in general, so this chunking splits at the wrong boundary, producing an incorrect token ordering that doesn't match the `BlockMask` built by `create_anchor_block_mask_mod`.

To fix both the serialization leak and the runtime dispatch, store the override on each attention module rather than on the shared config:

```python
# After layer creation, still in __init__:
if get_sp_size() > 1:
    for layer in self.layers:
        layer.self_attn._attn_impl_override = "dflash_flex_attention"
```

Then in `Qwen3DFlashAttention.forward`, prefer the override:

```python
impl = getattr(self, "_attn_impl_override", None) or self.config._attn_implementation
if impl is not None and impl != "eager":
    attn_fn = ALL_ATTENTION_FUNCTIONS[impl]
```

This keeps the persisted config clean while routing DFlash layers to the correct SP attention function at runtime.

scripts/train.py
```diff
@@ -693,6 +735,8 @@ def main(cfg: TrainConfig):  # noqa: C901
 if __name__ == "__main__":
     main(TrainConfig.resolve())
 
+    print(f"Peak GPU memory: {torch.cuda.max_memory_allocated() / 1e9:.2f} GB")
```
speculatorsbot: This bare `print` runs on **every** rank in distributed training and bypasses the logger. It looks like leftover debug code -- consider removing it, or moving it inside `main()` guarded by `rank == 0` and using `logger.info`.

speculatorsbot (commented):
The DDP `_sp_splits_batch` bug from my previous review has been fixed correctly -- the flag is now resolved once in `Trainer.__init__` before DDP wrapping. The hook cleanup via `_cleanup()` and the DDP sub-group scoping look good.

I found one issue in how the config save/restore interacts with DFlash's runtime attention dispatch, and one piece of leftover debug code.

PR #875, opened 2026-07-28
fix(regen): create parent dirs for --outfile before opening

jialefu:
## Purpose

`open(path, "a")` creates the file but not its parent directories, so the regen script crashes with `FileNotFoundError` when `--outfile` points into a missing directory (e.g. no `output/` on a fresh checkout) — and only *after* the vLLM server has finished starting up.                                                 
                                                                                               
Create the parent dirs (`os.makedirs(..., exist_ok=True)`) before opening. No behavior change when the directory already exists.

## Tests

Ran `scripts/response_regeneration/script.py` with `--outfile` pointing into a non-existent directory: reproduced the crash on main; with this fix the file is created and generation starts normally.

## Checklist

I have filled in:

### DIFF
--- a/scripts/response_regeneration/script.py
+++ b/scripts/response_regeneration/script.py
@@ -47,6 +47,16 @@ def _dataset_choice(name: str) -> str:
 # ---------------------------------------------------------------------------
 
 
+def ensure_parent_dirs(*paths: str) -> None:
+    """Create parent directories for output files.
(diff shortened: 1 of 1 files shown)


fynnsu (approved):

fynnsu:
Thank you for the fix @jialefu!

PR #882, opened 2026-07-29
Add AGENTS.md with vLLM launch instructions

orestis-z:
## Summary
- Adds `AGENTS.md` at the repo root with instructions for AI coding agents on how to correctly launch vLLM
- **Training**: always use `scripts/launch_vllm.py` (configures `--speculative_config` and `--kv_transfer_config` for hidden state extraction)
- **Eval**: launch vLLM directly with `--spec-model` and `--spec-tokens`

## Test plan
- [ ] Verify the file renders correctly on GitHub
- [ ] Confirm AI agents pick up the instructions

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,6 @@
+# Speculators Agent Instructions
+
+## Launching vLLM
+
+- **For training** (hidden state extraction): always use `scripts/launch_vllm.py`. Never use `python -m vllm.entrypoints.openai.api_server` directly — it misses `--speculative_config` and `--kv_transfer_config` which are required for hidden state extraction.
+- **For eval**: launch vLLM directly with `--spec-model` and `--spec-tokens` (e.g. `python -m vllm.entrypoints.openai.api_server --model <target> --spec-model <checkpoint> --spec-tokens <N>`).


AGENTS.md
```diff
@@ -0,0 +1,6 @@
+# Speculators Agent Instructions
+
+## Launching vLLM
+
+- **For training** (hidden state extraction): always use `scripts/launch_vllm.py`. Never use `python -m vllm.entrypoints.openai.api_server` directly — it misses `--speculative_config` and `--kv_transfer_config` which are required for hidden state extraction.
+- **For eval**: launch vLLM directly with `--spec-model` and `--spec-tokens` (e.g. `python -m vllm.entrypoints.openai.api_server --model <target> --spec-model <checkpoint> --spec-tokens <N>`).
```
speculatorsbot, line 6: The eval instruction diverges from every other example in this repo and is internally inconsistent with the training bullet.

**1. Pattern mismatch with repo docs/examples.** The [README](https://github.com/vllm-project/speculators/blob/af34d0145dc59d83c9f757995c6a82c0c3217367/README.md#L176-L180) documents eval as `vllm serve <speculator_model>`, the [eval example script](https://github.com/vllm-project/speculators/blob/af34d0145dc59d83c9f757995c6a82c0c3217367/examples/evaluate/example_qwen3_8b_dflash_humaneval.sh#L46) does `vllm serve "$MODEL"`, and the [response regeneration script](https://github.com/vllm-project/speculators/blob/af34d0145dc59d83c9f757995c6a82c0c3217367/scripts/response_regeneration/run_all.sh#L91) does the same. But this file tells agents to use `python -m vllm.entrypoints.openai.api_server --model <target> --spec-model <checkpoint> --spec-tokens <N>` — a different entrypoint and a different model-specification pattern (`--model target + --spec-model draft` vs. `vllm serve <speculator>`). An agent following this file will produce commands that look nothing like what the rest of the codebase does.

**2. Internal contradiction.** Line 5 warns "Never use `python -m vllm.entrypoints.openai.api_server` directly" (for training), but line 6 uses exactly that entrypoint for eval. Agents will reasonably wonder whether `api_server` is forbidden or fine.

Both approaches (`vllm serve <speculator>` vs. `--model <target> --spec-model <draft>`) are valid in vLLM, but if there's a reason agents should use the explicit split-model pattern (e.g. intermediate checkpoints lacking a `speculators_config`), it would help to say so. Otherwise, aligning with the repo's existing examples would avoid confusion:

```markdown
- **For eval**: launch vLLM directly (e.g. `vllm serve <checkpoint>`).
```

speculatorsbot (commented):
Small docs-only PR — the training guidance is accurate and useful. One concern about the eval instruction diverging from the rest of the repo's documentation.

PR #892, opened 2026-08-01
add support for gemma3

tonyay163:
## Purpose

Gemma3 has two things which are not supported:

1. Hybrid attention where some layers have sliding attention and some layers have full attention. If we are using sliding attention in the drafter, we probably want to pull the RoPE theta from the sliding attention layer.
2. Gemma3RMSNorm weights are +1 over Qwen3RMSNorm. I don't love how I did this, but given we explicitly load specify, I think this is the best way. Another way is loading the original model and then just keeping the norm from the model.

## Checklist

I have filled in:

### DIFF
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -169,6 +169,7 @@ def create_transformer_layer_config(  # noqa: C901
         "full_attention" if i in full_attention_indices else "sliding_attention"
         for i in range(num_layers)
     ]
+    has_sliding = "sliding_attention" in layer_types
 
     config = config_class(
         vocab_size=verifier_config.vocab_size,
@@ -183,8 +184,8 @@ def create_transformer_layer_config(  # noqa: C901
         rms_norm_eps=verifier_config.rms_norm_eps,
         head_dim=head_dim,
         tie_word_embeddings=False,
-        sliding_window=sliding_window,
-        use_sliding_window="sliding_attention" in layer_types,
+        sliding_window=sliding_window if has_sliding else None,
+        use_sliding_window=has_sliding,
         layer_types=layer_types,
     )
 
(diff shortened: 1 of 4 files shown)


scripts/train.py
```diff
@@ -194,12 +195,16 @@ def create_transformer_layer_config(  # noqa: C901
...
             if isinstance(rope_params, dict) and "rope_theta" not in rope_params:
                 sub = (
-                    rope_params.get("sliding_attention")
-                    or rope_params.get("full_attention")
+                    rope_params.get(
+                        "sliding_attention" if has_sliding else "full_attention"
+                    )
+                    or rope_params.get(
+                        "full_attention" if has_sliding else "sliding_attention"
+                    )
```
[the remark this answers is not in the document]
tonyay163: Existing tests already cover both cases though doesn't it?

src/speculators/models/dflash/core.py
```diff
@@ -106,7 +107,20 @@ def __init__(
...
+            else []
+        )
+        verifier_norm_cls = (
+            Gemma3RMSNorm
+            if any(
+                architecture.startswith("Gemma3")
+                for architecture in verifier_architectures
+            )
+            else Qwen3RMSNorm
+        )
```
speculatorsbot, line 122: The `startswith("Gemma3")` check only covers Gemma3, but Gemma (`GemmaForCausalLM`) and Gemma2 (`Gemma2ForCausalLM`) use the same `(1.0 + weight)` RMSNorm pattern:

```python
# Gemma3RMSNorm, Gemma2RMSNorm, and GemmaRMSNorm all do:
output = self._norm(x.float())
output = output * (1.0 + self.weight.float())  # weight init'd to zeros
```

If someone uses a Gemma2 verifier, the loaded norm weights (near-zero, intended for `1+w`) would be fed into `Qwen3RMSNorm` which uses them directly as `w * x`, producing near-zero outputs.

Using `startswith("Gemma")` instead of `startswith("Gemma3")` would cover the entire family without matching any unrelated architectures:

```python
verifier_norm_cls = (
    Gemma3RMSNorm
    if any(
        architecture.startswith("Gemma")
        for architecture in verifier_architectures
    )
    else Qwen3RMSNorm
)
```

(Using `Gemma3RMSNorm` as the concrete class is fine -- all three Gemma variants have identical RMSNorm implementations.)

speculatorsbot (commented):
The approach here is sound. Both changes fix real issues: (1) the verifier norm mismatch would cause numerically incorrect outputs when using a Gemma3 verifier since Gemma's `(1+w)` norm semantics differ from Qwen3's `w` norm, and (2) the RoPE selection now correctly matches the draft's attention mode rather than always preferring sliding-attention parameters.

One correctness concern with the norm dispatch scope, and a process note on DCO.

**DCO sign-off**: Commit `35cd856` is missing the required `Signed-off-by` line. CI will also flag this, but noting it here to save a round-trip. Fix with `git commit --amend --signoff`.

src/speculators/models/dflash/core.py
tonyay163: ok but gemma4 doesn't need this and i don't know what is missing for gemma1/gemma2 so i don't think we should support it unless it's needed

PR #894, opened 2026-08-03
fix(train): derive head_dim for verifiers that omit it from config

yuekaizhang:
Qwen2-style verifier configs (e.g. yuekai/cosyvoice2_llm) have no head_dim field, so create_transformer_layer_config passed head_dim=None into the draft transformer config. dflash/qwen3 draft configs validate head_dim as int and crash with "Field 'head_dim' expected int, got NoneType". Fall back to hidden_size // num_attention_heads.

### DIFF
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -121,6 +121,11 @@ def create_transformer_layer_config(  # noqa: C901
         if num_attention_heads % num_key_value_heads != 0:
             num_key_value_heads = num_attention_heads
 
+    if head_dim is None:
+        # Verifiers like Qwen2 omit head_dim from their config; derive the
+        # implicit value so draft configs that require an int don't reject None.
+        head_dim = verifier_config.hidden_size // num_attention_heads
+
     if full_attention_indices and (
         min(full_attention_indices) < 0 or max(full_attention_indices) >= num_layers
     ):


scripts/train.py
```diff
@@ -121,6 +121,11 @@ def create_transformer_layer_config(  # noqa: C901
         if num_attention_heads % num_key_value_heads != 0:
             num_key_value_heads = num_attention_heads
 
+    if head_dim is None:
+        # Verifiers like Qwen2 omit head_dim from their config; derive the
+        # implicit value so draft configs that require an int don't reject None.
+        head_dim = verifier_config.hidden_size // num_attention_heads
+
```
speculatorsbot, line 128: **Missing regression test.** This new code path has no test coverage. The existing infrastructure in `tests/unit/train/test_draft_config_init.py` (`_make_verifier_namespace`, `_create_layer_config_for`) makes it easy to add one:

```python
def test_create_layer_config_derives_head_dim_when_missing():
    """Qwen2-style verifiers omit head_dim; the builder must derive it."""
    verifier = _make_verifier_namespace()
    del verifier.head_dim
    config = _create_layer_config_for(verifier)
    assert config.head_dim == 32 // 4  # hidden_size // num_attention_heads
```

This fails without the fix (`Qwen3Config` rejects `head_dim=None`) and passes with it.

speculatorsbot (commented):
Clean, well-motivated fix. I confirmed that `Qwen3Config(head_dim=None)` raises `StrictDataclassFieldValidationError: Field 'head_dim' expected int, got NoneType`, so the crash described in the PR body is real. The fallback `hidden_size // num_attention_heads` is the standard derivation (matching what transformers uses internally for models without an explicit `head_dim`).

One ask: please add a regression test — details in the inline comment.

PR #897, opened 2026-08-03
Fix eager-path draft-token mask extension inverting additive mask semantics

khazic:
## Problem

`extend_dense_mask_for_draft_tokens` appends a 1/0 diagonal block regardless of the input mask's semantics. The eager/SDPA path feeds it the additive float mask from `create_float_mask` (0 = attend, -inf = blocked), so on every TTT step >= 1 the appended draft-KV columns are 0.0 off-diagonal (fully attended) and +1.0 on the diagonal (a spurious score bias). Every query then attends all previous-step draft tokens across document boundaries and future positions, so training or validation with `draft_attn_impl="eager"` silently diverges from the flex-attention backend and learns from leaked context.

## Fix

The appended block now follows the input mask's semantics: 0 on the diagonal and -inf elsewhere for floating-point masks, and the previous True/False (or 1/0) diagonal for boolean and integral masks.

## Tests

`test_extend_dense_float_mask_keeps_additive_semantics` extends a two-document additive float mask over three steps and asserts the new columns are 0 on the diagonal and -inf off it (fails on main), plus `test_extend_dense_integral_mask_keeps_one_zero_diagonal` pins the integral-mask behavior. Existing boolean-mask tests are unchanged and pass.

### DIFF
--- a/src/speculators/models/eagle3/attention.py
+++ b/src/speculators/models/eagle3/attention.py
@@ -117,10 +117,16 @@ def extend_dense_mask_for_draft_tokens(mask: torch.Tensor, total_seq_len: int):
 
     Dense-mask equivalent of extend_mask_for_draft_tokens (which operates on
     BlockMask). Appends total_seq_len new KV columns where only position
-    (q, new_offset + q) is True — the same diagonal pattern.
+    (q, new_offset + q) is attended (the same diagonal pattern).
+
+    The appended block follows the input mask's semantics: 0 / -inf for an
+    additive float mask (the eager / SDPA path built by create_float_mask,
+    where 0 means attend and -inf blocked), and True/False (or 1/0) for a
+    boolean or integral mask.
 
     Args:
-        mask: Dense boolean mask of shape [1, 1, total_seq_len, kv_len].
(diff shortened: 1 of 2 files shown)


WindChimeRan (commented):
LGTM. 

Confirmed the bug and the fix locally. --draft-attn-impl eager routes mask creation through create_float_mask. 

ran a softmax over uniform logits, 8 tokens packed as two documents, one TTT step:


main | PR #897
-- | --
draft-KV mass off the diagonal | 4.62 | 0.0000
q=0 attending draft token 7 (next doc, future) | 0.093 | 0.0000

khazic:
@WindChimeRan thanks for confirming the fix locally. Mergify still lists this PR as waiting on a review, and a COMMENTED review does not satisfy the protection. Could you convert your LGTM into an approving review, or could another maintainer take a look?

WindChimeRan:
cc @shanjiaz 

---

Please join our vllm slack channel #speculator ! @khazic

fynnsu (approved):

fynnsu:
Thanks for the fix!

PR #902, opened 2026-08-03
Count only evaluable positions in the conditional accuracy denominator

khazic:
## Problem

For eagle3 TTT steps >= 1, the caller slices `loss_mask[:, ttt_step:]` against `prev_correct[:, :-ttt_step]`, so a position can be prev-correct yet not evaluable at this step. `compute_accuracy_single_step` counted the conditional denominator from `prev_correct` alone, before the loss mask applies, while the numerator is loss-masked, so `cond_acc_1/2` are systematically biased low at every assistant-span boundary.

## Fix

The conditional denominator now counts positions that are both prev-correct and inside the loss mask, snapshotted before the in-place `logical_and` update. With `prev_correct=None` and a mask, the denominator is now the masked count for the same reason (previously the unmasked numel). Note for dashboards: logged `cond_acc_*` values step-change across this fix by design.

## Tests

New unit tests pin the masked denominator with and without `prev_correct` (both fail on main); the existing unmasked chain test is unchanged and passes.

### DIFF
--- a/src/speculators/models/metrics.py
+++ b/src/speculators/models/metrics.py
@@ -32,15 +32,27 @@ def compute_accuracy_single_step(
         counts suitable for distributed reduction before computing ratios.
     """
     correct = pred_ids == target_ids
-    cond_total = torch.tensor(correct.numel(), dtype=torch.float, device=correct.device)
+    eval_mask = loss_mask.to(torch.bool) if loss_mask is not None else None
+    cond_total = None
     if prev_correct is not None:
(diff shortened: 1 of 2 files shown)


khazic:
Gentle follow-up on this one. It corrects the conditional accuracy denominator to count only positions that can actually be evaluated; checks are green and the PR is small, so a review whenever convenient would be appreciated. Happy to rebase or adjust if anything is off.

PR #903, opened 2026-08-03
Warn when a cached token frequency distribution is silently reused

khazic:
## Problem

`save_token_frequency_distribution` silently returns when the output path already exists, with no signal that the cached counts may belong to a different dataset or target tokenizer. Retraining a draft for a new target model or dataset in the same working directory silently reuses the old `token_freq.pt`, so the d2t/t2d vocab mappings are built from the wrong corpus and the draft trains with a silently mismatched vocabulary.

## Fix

Reusing an existing file now logs a prominent warning naming the path and telling the user to delete it (or pass a different output path) when the dataset or target model changed. The cache behavior itself is unchanged to stay backward compatible; validating the cache against a dataset fingerprint would require a file-format change and is left as a follow-up if you want it.

## Tests

Unit tests cover the fresh-count path and assert the warning fires and the cache is preserved on reuse. The warning test fails on main.


shanjiaz (commented):
No need to add test for this warning.

khazic:
Dropped the test in de22c7d, the PR is now just the warning in `vocab_mapping.py`.

khazic:
Gentle follow-up: the test was dropped as suggested and the PR is now only the warning in vocab_mapping.py, with CI green. Mergify still needs an approving review from the maintainer list before it can merge, so an approval whenever convenient would be appreciated. Thanks for the earlier look.

fynnsu (approved):

PR #904, opened 2026-08-03
Align CLI convert algorithm choices with the conversion backend

khazic:
## Problem

The CLI `convert` command offers `--algorithm eagle`, which `convert_model` rejects (`ValueError: Unsupported algorithm: eagle`), and omits `dflash`, which the backend supports. The help text also documents eagle-v1-only kwargs.

## Fix

The supported list now has a single source of truth: `SUPPORTED_ALGORITHMS` in `convert/entrypoints.py`, derived from the same `Literal` that types `convert_model`, and consumed by the CLI's `click.Choice`, so the two cannot drift again. Help text and docstring examples are updated (eagle v1 section removed, dflash added).

Note on scope: this removes the user-visible (but non-functional) `eagle` choice; since selecting it only ever produced a backend error, no working invocation changes.

## Tests

CLI-runner tests assert every supported algorithm reaches `convert_model` and that `eagle` is rejected by the parser instead of exploding in the backend. Both fail on main.


khazic:
Gentle follow-up on this one. It is a small change that keeps the CLI's accepted algorithm names in sync with what the conversion backend supports; DCO and the other checks are green, and Mergify only needs an approving review from the maintainer list. Happy to rebase or adjust if anything is off.

PR #905, opened 2026-08-03
Match sample data files by the .pt extension, not a bare pt suffix

khazic:
## Problem

`list_files` matches sample files with `endswith("pt")`, so any filename merely ending in the letters "pt" (`model.ckpt`, `notes.gpt`) is treated as a sample and crashes the dataloader worker mid-training. The documented data layout also stores `token_freq.pt` (the vocab-mapping cache) in the same directory as the samples, and it matches even a correct `.pt` check.

## Fix

Match the `.pt` extension exactly and skip `token_freq.pt` explicitly.

## Tests

A unit test in `test_data.py` lays out sample files next to `model.ckpt`, `notes.gpt`, and `token_freq.pt` and asserts only the samples are returned. Fails on main.


shanjiaz (commented):
@khazic Could you resolve merge conflict? otherwise looks good to me.

src/speculators/train/data.py
WindChimeRan, line 29: `list_files` is only reachable in the legacy path https://github.com/vllm-project/speculators/blob/c667ee19ed38c9f2cae9d95c4046b3f94af05769/src/speculators/train/dataloader.py#L126

"deprecated and will be removed soon" warning in #353

Recommend close this PR and remove legacy data path.

WindChimeRan:
legacy path removal: https://github.com/vllm-project/speculators/pull/914

khazic:
Agreed, closing. `list_files` is reachable only through `split_files` / `SampleFileDataset`, and both sit behind the `legacy_data` branch that #353 already marked for removal, so hardening the suffix match there is not worth carrying.

For whoever picks up the removal: the same bare-suffix match is what would pull `model.ckpt`, `notes.gpt` and the `token_freq.pt` cache into the sample list, so it is a small argument for deleting the path rather than keeping it around unmaintained.

PR #910, opened 2026-08-03
feat(#880): add reproducibility artifacts to eval runs and checkpoints

### LINKED ISSUE #880: [RFC] Add reproducibility artifacts to eval runs and checkpoints
## Motivation

Evaluation results are currently not reproducible from their output alone. The eval script (`scripts/evaluate/evaluate.py`) produces only benchmark data (`acceptance.csv`, `perf_results.csv`, raw guidellm JSONs) — no metadata about the environment, command, or code state that produced them.

This caused unnecessary re-work in #757 where we couldn't determine which vLLM patches had been applied to original eval runs. Training already records provenance via `save_train_command()` — evaluation has no equivalent.

## Artifact Summary

| Artifact | Script | Default Location | Format |
|---|---|---|---|
| `train_command.txt` | `train/utils.py` | `<checkpoint_dir>/` | `shlex.join(sys.argv)` |
| `eval_command.txt` | `evaluate.py` | `<output_dir>/` | Header (timestamp, git SHA, package versions) + `shlex.join(sys.argv)` |
| `vllm_command.txt` | `launch_vllm.py` | `vllm_<model>_<ts>/` | Header (timestamp, python, git SHA, vllm version) + `shlex.join(cmd)` |
| `vllm.patch` | `launch_vllm.py` | `vllm_<model>_<ts>/` | `# repo: <path> (<sha>)` + `git diff HEAD` |
| `checkpoint_sha256.txt` | `launch_vllm.py` | `vllm_<model>_<ts>/` | `<sha256>  <filename>` per `.safetensors` file |
| `speculators.patch` | `train/utils.py` | `<checkpoint_dir>/` | `git diff HEAD` |

`launch_vllm.py` provenance location is overridable with `--provenance-dir` to co-locate with eval results.

### 1. `eval_command.txt` — eval metadata file

Mirror the existing `save_train_command()` pattern. Write to the eval output directory:
- Timestamp (UTC ISO 8601)
- Speculators git SHA
- Package versions (speculators, vllm, transformers, torch, compressed-tensors)
- Full `sys.argv` eval command

**Integration point:** `run_benchmark()` in `evaluate.py`, right after `artifacts_dir.mkdir()`.

### 2. `vllm_command.txt` + `vllm.patch` + `checkpoint_sha256.txt` — vLLM provenance

Written by `launch_vllm.py` (not the eval script) to keep concerns separated — the eval script doesn't know how vLLM was launched.

- `vllm_command.txt`: full constructed command + metadata header (timestamp, python path, git SHA, vllm version)
- `vllm.patch`: `git diff HEAD` from the vLLM source tree (auto-discovered by walking up from the installed package, falling back to `~/vllm` and `/workspace/vllm`). Header-only when the checkout is clean. Records version string when installed from a wheel.
- `checkpoint_sha256.txt`: SHA256 of all `.safetensors` files for local checkpoints, or a comment noting the model is a remote reference.

`launch_vllm.py` always saves provenance — defaults to `vllm_<model>_<YYYYMMDD_HHMMSS>/`, overridable with `--provenance-dir`.

### 3. `speculators.patch` — speculators diff artifact

Add `git diff HEAD` from the speculators repo root to `<checkpoint_dir>/` alongside the existing `train_command.txt`. This captures uncommitted changes at training time so the diff travels with the checkpoint (useful for HF uploads).

## Implementation Notes

- All artifacts are best-effort (wrapped in try/except) — failure to capture provenance never blocks training or evaluation.
- Atomic writes via tempfile + rename, matching the existing `save_train_command()` pattern.
- For wheel installs (no git checkout), the version string is recorded and no diff is available.

shotsan:
Implements the reproducibility artifacts RFC (#880) to save exact eval commands, vLLM launch metadata, safetensors SHA256 hashes, and uncommitted git diffs.

Closes #880.

### DIFF
--- a/scripts/launch_vllm.py
+++ b/scripts/launch_vllm.py
@@ -86,6 +86,17 @@ def parse_args():
         ),
     )
     parser.add_argument(
+        "--provenance-dir",
+        type=str,
+        default=None,
+        help="Directory to save provenance artifacts (vllm_command.txt, vllm.patch, checkpoint_sha256.txt).",
+    )
+    parser.add_argument(
+        "--no-hash-checkpoints",
+        action="store_true",
+        help="Skip hashing the .safetensors checkpoints in the provenance directory.",
+    )
+    parser.add_argument(
         "--dry-run",
         action="store_true",
         help="Print the command that would be executed without running it",
@@ -151,6 +162,53 @@ def main():
 
     print("Running command:")
     print(" ".join(cmd))
+
+    import datetime
+    import shlex
+    from pathlib import Path
+    from speculators.utils.provenance import (
+        get_git_sha,
+        get_git_diff,
+        get_package_root,
+        get_package_versions,
+        hash_safetensors,
+        write_atomic,
+    )
+
+    prov_dir = args.provenance_dir
+    if not prov_dir:
+        safe_model = args.model.replace("/", "_")
+        ts = datetime.datetime.now(datetime.timezone.utc).strftime("%Y%m%d_%H%M%S")
+        prov_dir = f"vllm_{safe_model}_{ts}"
+    prov_path = Path(prov_dir)
+
+    spec_root = get_package_root("speculators") or Path.cwd()
+    vllm_root = get_package_root("vllm")
+    
+    spec_sha = get_git_sha(spec_root)
+    versions = get_package_versions(["vllm"])
+    vllm_ver = versions[0] if versions else "unknown"
+
+    vllm_cmd_header = "\n".join([
+        f"# Timestamp: {datetime.datetime.now(datetime.timezone.utc).isoformat()}",
+        f"# Python: {sys.executable}",
+        f"# Speculators Git SHA: {spec_sha}",
+        f"{vllm_ver}",
+    ])
+    vllm_cmd_content = f"{vllm_cmd_header}\n{shlex.join(cmd)}\n"
+    write_atomic(prov_path, "vllm_command.txt", vllm_cmd_content)
+
+    if vllm_root and (vllm_root.parent / ".git").exists():
+        vllm_repo = vllm_root.parent
+        diff = get_git_diff(vllm_repo)
+        patch_header = f"# repo: {vllm_repo} ({get_git_sha(vllm_repo)})"
+        write_atomic(prov_path, "vllm.patch", f"{patch_header}\n{diff}\n")
+    else:
+        write_atomic(prov_path, "vllm.patch", f"# Installed from wheel or non-git source\n{vllm_ver}\n")
+
+    if not args.no_hash_checkpoints:
+        hashes = hash_safetensors(args.model)
+        write_atomic(prov_path, "checkpoint_sha256.txt", hashes + "\n")
 
     if not args.dry_run:
         os.execvp(cmd[0], cmd)  # noqa: S606
--- a/scripts/evaluate/evaluate.py
+++ b/scripts/evaluate/evaluate.py
@@ -258,6 +258,29 @@ def run_benchmark(args: argparse.Namespace) -> None:
     artifacts_dir = output_dir / "artifacts"
     artifacts_dir.mkdir(parents=True, exist_ok=True)
 
+    from datetime import timezone
+    import shlex
+    from speculators.utils.provenance import (
+        get_git_sha,
+        get_package_root,
+        get_package_versions,
+    )
+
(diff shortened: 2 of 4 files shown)


scripts/launch_vllm.py
```diff
@@ -152,6 +163,53 @@ def main():
...
+        f"# Timestamp: {datetime.datetime.now(datetime.timezone.utc).isoformat()}",
+        f"# Python: {sys.executable}",
+        f"# Speculators Git SHA: {spec_sha}",
+        f"{vllm_ver}",
+    ])
+    vllm_cmd_content = f"{vllm_cmd_header}\n{shlex.join(cmd)}\n"
+    write_atomic(prov_path, "vllm_command.txt", vllm_cmd_content)
+
+    if vllm_root and (vllm_root.parent / ".git").exists():
+        vllm_repo = vllm_root.parent
```
speculatorsbot, line 202: `get_package_root("vllm")` returns the package directory (e.g. `.../vllm/vllm/`), and this line assumes `.parent` is the git root. That holds for vllm's current flat layout but breaks for any package using a `src/` layout — and notably, speculators itself uses `src/` layout, so applying this same pattern to speculators would fail.

The RFC described “walking up from the installed package” — consider using `git rev-parse --show-toplevel` (which you already shell out to for SHA/diff) to find the actual repo root reliably:

```python
def _get_git_root(start: Path) -> Optional[Path]:
    try:
        root = subprocess.run(
            ["git", "rev-parse", "--show-toplevel"],
            capture_output=True, text=True, check=True,
            cwd=str(start),
        ).stdout.strip()
        return Path(root)
    except (OSError, subprocess.CalledProcessError):
        return None
```

This would replace both `get_package_root` + the `.parent / ".git"` check with a single reliable lookup, and it generalizes to any repository layout.

speculatorsbot (commented):
This PR consolidates provenance logic into a shared `provenance.py` module and integrates it into eval, training, and vLLM launch scripts to implement RFC #880. The refactoring into a shared module is a sound approach, and the `write_atomic` pattern is correct.

### Duplicate PR

PR #883 (opened 5 days earlier by @orestis-z, a listed maintainer) implements the same RFC #880, modifies the same set of files (`evaluate.py`, `launch_vllm.py`, `train/utils.py`), and also closes #880. That PR additionally includes three unit test files (`test_eval_provenance.py`, `test_vllm_provenance.py`, `test_provenance_integration.py`). These two PRs will conflict with each other. Per [CONTRIBUTING.md](https://github.com/vllm-project/speculators/blob/main/CONTRIBUTING.md), coordination with maintainers is recommended for changes affecting 3+ files to avoid duplicate effort. @shotsan — are you aware of #883? It would be worth discussing with @orestis-z how to proceed (adopt one, merge the approaches, etc.).

### Missing tests

`provenance.py` introduces 6 utility functions that are now shared across three integration points. Functions like `write_atomic`, `hash_safetensors`, and `get_package_root` have meaningful edge cases (permissions, missing directories, namespace packages, non-git installs). Since a regression in the shared module would silently break all three workflows, unit tests are warranted.

### Dry-run creates provenance artifacts

The provenance block in `launch_vllm.py` (lines 166–211) runs unconditionally, including when `--dry-run` is passed. This creates directories and files on disk even though no server is launched — the `--dry-run` help text says "Print the command that would be executed without running it," which implies no side effects. Is this intentional? If so, it would be worth documenting in the help text; if not, consider gating the provenance block on `not args.dry_run`.

### Note: stale base

The diff context shows `save_train_command(save_path, argv=...)` but `main` has since removed the `argv` parameter. A rebase is likely needed before merge.

Minor: `hash_safetensors` reads files in 8 KB chunks. For multi-GB safetensors files, a larger buffer (e.g. `1 << 20`) would meaningfully reduce syscall overhead, since the RFC discussion explicitly flagged checkpoint hashing I/O cost as a concern.

orestis-z:
Hey @shotsan thanks a lot for your work. There was this ongoing work already in https://github.com/vllm-project/speculators/pull/883. Would you like to review that instead?

santosh-agebold:
Ah, I completely missed PR #883! Closing this duplicate and I will head over to review yours. Thanks for pointing it out!

santosh-agebold:
Ah, I completely missed PR #883! I'll leave this up for now but consider it abandoned, and I will head over to review yours. Thanks for pointing it out!

orestis-z:
no worries @shotsan and thanks for your efforts. I closed this PR for now.

PR #915, opened 2026-08-04
Add DFlash Qwen3-8B online training example for Ascend NPU devices

Simba2017:
## Purpose
The official speculators examples are GPU-centric: they assume CUDA availability, use CUDA_VISIBLE_DEVICES, , and rely on the vllm-ascend backend for large models. There is no reference configuration or script for running DFlash on Ascend NPU with small dense models.

## Tests
training_curves

## Checklist

I have filled in:

### DIFF
--- a/examples/train/dflash_qwen3_8b_sharegpt_online_5k_npu.sh
+++ b/examples/train/dflash_qwen3_8b_sharegpt_online_5k_npu.sh
@@ -5,7 +5,7 @@
 # and training (with hidden states generated on-the-fly from the live server).
 #
 # Usage: Copy this script, modify the configuration variables below, then run:
-#   bash examples/train/dflash_qwen3_8b_sharegpt_online_10k_npu.sh
+#   bash examples/train/dflash_qwen3_8b_sharegpt_online_5k_npu.sh
 #
 # For a detailed walkthrough, see 
 # https://docs.vllm.ai/projects/speculators/en/latest/user_guide/tutorials/train_dflash_online/
--- a/examples/train/dflash_qwen3_8b_sharegpt_online_5k_npu.sh
+++ b/examples/train/dflash_qwen3_8b_sharegpt_online_5k_npu.sh
@@ -1,0 +1,100 @@
+#!/bin/bash
+# Online DFlash Training Script
+#
+# Runs the full online DFlash training pipeline: data preparation, vLLM server launch,
+# and training (with hidden states generated on-the-fly from the live server).
+#
+# Usage: Copy this script, modify the configuration variables below, then run:
+#   bash examples/train/dflash_qwen3_8b_sharegpt_online_10k_npu.sh
+#
+# For a detailed walkthrough, see 
+# https://docs.vllm.ai/projects/speculators/en/latest/user_guide/tutorials/train_dflash_online/
+
+### Example E2E run for DFlash Qwen3-8B on Ascend NPU device from ShareGPT ###
+
+# Note: With just 5k samples, the model performance will not be very good, however there
+# are enough samples to verify that the pipeline is working correctly and that the model
+# is learning something. This is a good sanity check when creating a drafter for a new
+# target model.
+
+# Experiments on 6x Ascend 910B 64GB NPUs, DP=2
+
+set -euo pipefail
+
+# ============ Configuration ============
+MODEL="Qwen/Qwen3-8B"
+DATASET="sharegpt"                # sharegpt, ultrachat, or path to custom data
+OUTPUT_DIR="./output/dflash_qwen3_8b_sharegpt"
+VLLM_PORT=8010
+MAX_SAMPLES=5000
+SEQ_LENGTH=8192
+EPOCHS=10
+LR=3e-4
+
+# DFlash-specific parameters
+SPECULATOR_TYPE="dflash"
+BLOCK_SIZE=8
+MAX_ANCHORS=1024
+NUM_LAYERS=5
+DRAFT_VOCAB_SIZE=32000
+TARGET_LAYER_IDS="2 18 33"  # Must match vLLM's eagle_aux_hidden_state_layer_ids
+
+# NPU assignments (online training needs separate NPUs for vLLM and training)
+VLLM_NPUS="0,1"
+TRAIN_NPUS="2,3"
+NUM_TRAIN_NPUS=2
+# =======================================
+
+# Step 1: Prepare data
+echo "=== Step 1: Preparing data ==="
+python scripts/prepare_data.py \
+    --model "$MODEL" \
+    --data "$DATASET" \
+    --output "$OUTPUT_DIR" \
+    --max-samples "$MAX_SAMPLES" \
+    --seq-length "$SEQ_LENGTH"
+
+# Step 2: Launch vLLM server in the background
+echo "=== Step 2: Launching vLLM server ==="
+ASCEND_RT_VISIBLE_DEVICES="$VLLM_NPUS" python scripts/launch_vllm.py "$MODEL" \
+    --target-layer-ids $TARGET_LAYER_IDS \
+    -- --data-parallel-size 2 --port "$VLLM_PORT" &
+VLLM_PID=$!
+
+# Ensure vLLM is cleaned up on exit
+cleanup() {
+    echo "Stopping vLLM server..."
+    kill "$VLLM_PID" 2>/dev/null || true
+    wait "$VLLM_PID" 2>/dev/null || true
+}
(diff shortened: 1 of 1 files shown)


examples/train/dflash_qwen3_8b_sharegpt_online_5k_npu.sh
```diff
@@ -0,0 +1,100 @@
...
+    echo "Stopping vLLM server..."
+    kill "$VLLM_PID" 2>/dev/null || true
+    wait "$VLLM_PID" 2>/dev/null || true
+}
+trap cleanup EXIT
+
+echo "Waiting for vLLM server to be ready..."
+until curl -sf "http://localhost:${VLLM_PORT}/health" > /dev/null 2>&1; do
+    sleep 2
+done
```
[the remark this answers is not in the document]
Simba2017: this code followed the file examples/train/dflash_qwen3_8b_sharegpt_online_5k.sh

examples/train/dflash_qwen3_8b_sharegpt_online_5k_npu.sh
```diff
@@ -0,0 +1,100 @@
...
+# https://docs.vllm.ai/projects/speculators/en/latest/user_guide/tutorials/train_dflash_online/
+
+### Example E2E run for DFlash Qwen3-8B on Ascend NPU device from ShareGPT ###
+
+# Note: With just 5k samples, the model performance will not be very good, however there
+# are enough samples to verify that the pipeline is working correctly and that the model
+# is learning something. This is a good sanity check when creating a drafter for a new
+# target model.
+
+# Experiments on 6x Ascend 910B 64GB NPUs, DP=2
```
speculatorsbot: The comment says "6x Ascend 910B 64GB NPUs" but the script only assigns 4 (NPUs 0,1 for vLLM + 2,3 for training). In the GPU script, the hardware count matches the GPUs used ("4x NVIDIA H100 80GB GPUs" with 4 GPUs assigned). Here the mismatch could confuse users about the minimum hardware needed.

Should this say "4x Ascend 910B 64GB NPUs" to match the actual assignment? Or if the intent is to document the full node spec, consider noting that only 4 are used, e.g.:
```suggestion
# Experiments on 6x Ascend 910B 64GB NPUs (4 used), DP=2
```

examples/train/dflash_qwen3_8b_sharegpt_online_5k_npu.sh
```diff
@@ -0,0 +1,100 @@
...
+    --draft-vocab-size "$DRAFT_VOCAB_SIZE" \
+    --epochs "$EPOCHS" \
+    --lr "$LR" \
+    --total-seq-len "$SEQ_LENGTH" \
+    --speculator-type "$SPECULATOR_TYPE" \
+    --block-size "$BLOCK_SIZE" \
+    --max-anchors "$MAX_ANCHORS" \
+    --num-layers "$NUM_LAYERS" \
+    --target-layer-ids $TARGET_LAYER_IDS \
+    --draft-attn-impl sdpa \
```
speculatorsbot: This is the key functional difference from the GPU script (which omits the flag and defaults to `simple_flex_attention`). A brief inline comment explaining *why* `sdpa` is needed on Ascend NPU would help future readers who compare this script against the GPU variant. Something like:
```suggestion
    --draft-attn-impl sdpa \  # FlexAttention not available on Ascend NPU
```

speculatorsbot (commented):
This PR adds the first Ascend NPU example script for DFlash training, closely mirroring the existing GPU script (`dflash_qwen3_8b_sharegpt_online_5k.sh`). The structure is clean and the NPU-specific adaptations (`ASCEND_RT_VISIBLE_DEVICES`, `--draft-attn-impl sdpa`) are the right changes for the platform. Two minor items below.

examples/train/dflash_qwen3_8b_sharegpt_online_5k_npu.sh
```diff
@@ -0,0 +1,100 @@
...
+    --lr "$LR" \
+    --total-seq-len "$SEQ_LENGTH" \
+    --speculator-type "$SPECULATOR_TYPE" \
+    --block-size "$BLOCK_SIZE" \
+    --max-anchors "$MAX_ANCHORS" \
+    --num-layers "$NUM_LAYERS" \
+    --target-layer-ids $TARGET_LAYER_IDS \
+    --draft-attn-impl sdpa \   # FlexAttention not available on Ascend NPU
+    --on-missing generate \
+    --on-generate delete
```
speculatorsbot: **Bug: broken line continuation — script will fail at runtime.**

In bash, `\` must be immediately followed by a newline to continue a command across lines. Here, `\` is followed by spaces and a `#` comment, so the continuation is broken: `\` escapes the next space character, `#` starts a comment, and the command terminates at this line. The `--on-missing` and `--on-generate` flags are never passed to `torchrun` — instead bash tries to execute `--on-missing` as a standalone command, which fails immediately under `set -euo pipefail`.

You can verify this yourself:
```bash
$ bash -c '
set -euo pipefail
echo \   # comment
    --on-missing generate
' 2>&1
# Output: bash: line 4: --on-missing: command not found
```

(Apologies — my earlier suggestion had the same problem. You can't put a comment after `\` on a continuation line in bash.)

Fix: move the comment to its own line above the flag:
```suggestion
    # FlexAttention not available on Ascend NPU
    --draft-attn-impl sdpa \
    --on-missing generate \
    --on-generate delete
```

speculatorsbot (commented):
Thanks for addressing the previous findings. The NPU count and filename are fixed, and the `sdpa` comment is a good addition. However, the way the comment was added on line 96 breaks the bash line continuation — the script will fail at runtime. One inline comment below.

sunny-infra:
For the sake of repo cleanliness, I recommend including the NPU-related details as comments within `dflash_qwen3_8b_sharegpt_online_5k.sh`, rather than introducing a new NPU-specific file.

zihanlin-ai:
Validation on Atlas A3. The script does not run on `main` for three separate reasons, only one of which is NPU-specific:

| # | What breaks | Scope |
|---|---|---|
| 1 | `prepare_data.py` requires `--render-endpoint` since #975, and every example still calls it *before* the server is up | 7 of the `examples/train/*.sh`, CUDA ones included — reproduced on 8× H20 with no NPU involved |
| 2 | `ExampleHiddenStatesConnector` + chunked prefill is rejected by the vLLM pinned in `vllm-ascend:nightly-releases-v0.23.0-a3` | that image only; upstream dropped the restriction after v0.24.0 |
| 3 | fused CE loss overflows the NPU Unified Buffer at `--draft-vocab-size 32000` | Ascend only — fixed by #993, which we validated separately |

On #1, the failure is `ValueError: render_endpoint is required to convert natural-language conversations to speculator training rows` at `preprocessing.py:563`. It predates this PR: the `raise` came in with `f4cf910` (#975, Aug 14), this PR is from Aug 5. A two-line conversations `.jsonl` plus any tokenizer directory reproduces it with no dataset, GPU or server.

Two changes to the script make it work; nothing else touched:

```diff
-# Step 1 ... (moved below, after the server is up)
-    -- --data-parallel-size 2 --port "$VLLM_PORT" &
+    -- --data-parallel-size 2 --no-enable-chunked-prefill --port "$VLLM_PORT" &
 echo "vLLM server ready."
+python scripts/prepare_data.py --model "$MODEL" --data "$DATASET" --output "$OUTPUT_DIR" \
+    --render-endpoint "http://localhost:${VLLM_PORT}" --max-samples "$MAX_SAMPLES" --seq-length "$SEQ_LENGTH"
```

With that, on A3 (4 chips: vLLM DP=2 plus DDP training on 2), speculators `main` @ `c05b95b` + #993:

- server ready in ~2.5 min; 5000 samples prepared in ~11 min through `POST /v1/chat/completions/render`
- 10 epochs, 319 steps/epoch, ~1.09 s/step, ~6.98e3 tok/s; validation loss `1.581 → 1.485 → 1.423 → 1.412 → 1.385 → 1.368 → 1.356 → 1.343 → 1.331 → 1.322`; a self-contained checkpoint per epoch, `checkpoint_best` updated, exit code 0
- image `vllm-ascend:nightly-releases-v0.23.0-a3`, python 3.12.13, torch 2.10.0 / torch_npu 2.10.0.post2, CANN 9.0.1, driver 25.2.1, transformers 5.5.4

Suggestions: reorder step 1 and add the flag with a short version note; mention #993 as a prerequisite rather than an optional companion. The same reorder is needed for the six other examples. This supersedes the reproduction in #531.

PR #918, opened 2026-08-04
fix(dflash): use conditional_torch_compile to avoid BiSheng IR error on Ascend NPU devices

Simba2017:
## Purpose
src/speculators/models/dflash/core.py
line 30: _compiled_create_block_mask = torch.compile(create_block_mask)
NVIDIA GPU → Inductor/Triton Backend; Ascend NPU → BiSheng Compiler Backend
The BiSheng Compiler cannot process the complex computation graph generated by create_block_mask (from torch.nn.attention.flex_attention), throwing an error: fail to run BiShengIR Pipeline.

use conditional_torch_compile(reuse existing interfaces) to avoid BiSheng IR error on Ascend NPU.

## Tests
work

## Checklist

I have filled in:


src/speculators/models/dflash/core.py
```diff
@@ -31,7 +31,7 @@
 
 # Compile so the mask builds block-sparse instead of materializing DFlash's huge
 # dense [Q, KV] grid every step. (No benefit for EAGLE3's small autoregressive mask.)
-_compiled_create_block_mask = torch.compile(create_block_mask)
+_compiled_create_block_mask = conditional_torch_compile(create_block_mask)
```
rahul-tuli, line 34: Question: Does this fix enable DFlash training on Ascend NPU, or does it only prevent the import-time crash? if latter, could the PR description clarify the scope, e.g. `prevents import-time crash so non-DFlash models can be used on NPU`?

rahul-tuli (commented):
Looks good, pending a clarification and CI

sunny-infra:
The root cause of the current issue is that Ascend PTA does not yet support `flex_attentio`n. Therefore, the recommended parameter for running speculators on NPU at present is `--draft-attn-impl sdpa`.
Changes introduced in this PR do not fundamentally resolve the NPU adaptation problem, and the flex_attention‑related error persists. We suggest using the SDPA implementation as a temporary workaround instead of modifying the upstream repository code. We will roll out an update as soon as Ascend adds support for `flex_attention` (expected in mid‑September). cc @shanjiaz

fynnsu:
@sunny-infra would you recommend we get rid of the `conditional_torch_compile` logic then? We can also add a warning/raise an error if the user select flex_attention with npu accelerators. 

As it currently stands, I believe the `conditional_torch_compile` will be disabling `torch.compile` on all NPU training runs, including those that correctly use the `sdpa` backend, which I'd expect to slow down NPU training.

sunny-infra:
@fynnsu This is a good idea. I will first test whether removing conditional_torch_compile works in other scenarios. If it does, I will add an NPU‑related warning for cases where flex_attention is triggered.
Meanwhile, I would like to ask whether I can add usage guidance for NPU to the documentation at [https://docs.vllm.ai/projects/speculators/en/latest/user_guide/tutorials/train/](https://link.wtturl.cn/?target=https%3A%2F%2Fdocs.vllm.ai%2Fprojects%2Fspeculators%2Fen%2Flatest%2Fuser_guide%2Ftutorials%2Ftrain%2F&scene=im&aid=497858&lang=zh).

PR #921, opened 2026-08-04
fix(dflash): flatten Laguna-style nested rope_parameters

orestis-z:
## Summary
- Laguna models use nested per-layer-type `rope_parameters` (with `full_attention` and `sliding_attention` sub-dicts), which `Qwen3RotaryEmbedding` doesn't understand — it expects a flat dict with `rope_theta` at the top level.
- Since all DFlash draft layers use sliding-window attention, this PR flattens to the `sliding_attention` variant before constructing the rotary embedding.
- Without this fix, training a DFlash/DSpark draft model against a Laguna verifier crashes with `KeyError: 'rope_theta'`.

## Test plan
- [x] Verified fix by training a DSpark model for `poolside/Laguna-S-2.1` from scratch — training runs without errors, loss decreasing normally
- [ ] Existing tests pass (`pytest tests/`)


shotsan:
I verified this fix locally. It correctly flattens the rope parameters and passes all quality checks.

src/speculators/models/dflash/core.py
```diff
@@ -95,7 +96,13 @@ def __init__(
...
             eps=config.transformer_layer_config.rms_norm_eps,  # type: ignore[arg-type]
         )
-        self.rotary_emb = Qwen3RotaryEmbedding(config.transformer_layer_config)  # type: ignore[arg-type]
+        rotary_config = config.transformer_layer_config
+        rope_params = getattr(rotary_config, "rope_parameters", None)
+        if rope_params and "sliding_attention" in rope_params:
+            # Flatten nested rope_parameters to the sliding_attention variant.
+            rotary_config = deepcopy(rotary_config)
+            rotary_config.rope_parameters = rope_params["sliding_attention"]
+        self.rotary_emb = Qwen3RotaryEmbedding(rotary_config)  # type: ignore[arg-type]
```
speculatorsbot, line 105: Eagle3 ([`core.py:107`](https://github.com/vllm-project/speculators/blob/079b38658b6dfd2b72ff768a860fcdb48c6fe60f/src/speculators/models/eagle3/core.py#L107)) and MTP ([`core.py:86`](https://github.com/vllm-project/speculators/blob/079b38658b6dfd2b72ff768a860fcdb48c6fe60f/src/speculators/models/mtp/core.py#L86)) construct their rotary embeddings the same way and would hit the same `KeyError` with Laguna configs. Consider extracting this flattening into a shared helper (e.g., in `models/utils.py`) so all model types benefit — rather than having to duplicate this block when fixing Eagle3/MTP.

speculatorsbot (commented):
Clean, well-scoped fix that correctly prevents a `KeyError: 'rope_theta'` crash when training DFlash/DSpark against Laguna-style verifiers with nested `rope_parameters`. The deepcopy + flatten approach is sound.

**Missing regression test**: Since this is a bug fix, a regression test that fails without the fix and passes with it would be valuable. A unit test with a mock config containing nested `rope_parameters` (matching [Laguna's structure](https://github.com/vllm-project/speculators/pull/649)) could verify the flattening in `DFlashDraftModel.__init__`. The existing `tests/unit/train/test_rope_config.py` covers the config builder but not the model init path.

shanjiaz (approved):
Thanks for putting up the changes!

PR #922, opened 2026-08-04
feat(dflash): remap fused QKV weights for Laguna warm-start

orestis-z:
## Summary
- Adds `_remap_weights()` to `DFlashConverter` that splits Laguna-style fused `qkv_proj` into separate `q_proj`/`k_proj`/`v_proj`, drops incompatible keys (`g_proj`, `aux_hidden_norms`), and slices `fc.weight` when source has more target layers than needed.
- Enables warm-starting DSpark training from published Laguna DFlash checkpoints (e.g. `poolside/Laguna-S-2.1-DFlash`) instead of training from scratch (~13.5 day ETA → significantly faster convergence).
- No-op when checkpoint already uses separate Q/K/V projections (Qwen3-style).

## Test plan
- [x] 5 new unit tests covering QKV splitting, key dropping, fc slicing, passthrough, and key preservation
- [x] All 57 converter tests pass
- [x] End-to-end conversion of `poolside/Laguna-S-2.1-DFlash` succeeds (6 fused qkv_proj split, 12 keys dropped, fc sliced 18432→15360)


shanjiaz (commented):
Looks good pending quality!

src/speculators/models/dflash/core.py
```diff
@@ -95,7 +96,13 @@ def __init__(
             config.transformer_layer_config.hidden_size,
             eps=config.transformer_layer_config.rms_norm_eps,  # type: ignore[arg-type]
         )
-        self.rotary_emb = Qwen3RotaryEmbedding(config.transformer_layer_config)  # type: ignore[arg-type]
+        rotary_config = config.transformer_layer_config
+        rope_params = getattr(rotary_config, "rope_parameters", None)
+        if rope_params and "sliding_attention" in rope_params:
+            # Flatten nested rope_parameters to the sliding_attention variant.
```
speculatorsbot, line 102: This unconditionally selects the `sliding_attention` variant of `rope_parameters` for the shared `rotary_emb`, but `DFlashDraftModel` supports mixed layer types — `self.uses_full_attn` on line 92 is explicitly computed and used to create separate attention masks for full-attention layers.

The commit message says "all DFlash draft layers use sliding-window attention," which is true for Laguna but is not a general DFlash invariant. If a future model (e.g. Qwen3.5-based) has nested `rope_parameters` with differing configs for `full_attention` vs. `sliding_attention`, the full-attention layers would silently use incorrect rope scaling.

Two suggestions:
1. Add a defensive check — e.g. log a warning or assert when `self.uses_full_attn` is true and we're flattening to `sliding_attention` rope params.
2. Add a unit test for this code path (it's currently the only change in the PR without test coverage). A minimal test that constructs a `DFlashDraftModel` with a config containing nested `rope_parameters` and verifies the rotary embedding initializes correctly would prevent regressions.

speculatorsbot (commented):
The weight remapping logic in `_remap_weights` is clean and well-tested — the QKV split, key dropping, and fc slicing all make sense for the Laguna warm-start use case. The `_build_config` change to exclude the last verifier layer from `aux_hidden_state_layer_ids` is correct and pairs well with the fc slicing.

One concern about the `rope_parameters` fix in `core.py` — see inline comment.

shanjiaz (approved):

PR #926, opened 2026-08-05
fix a ddp ckpt bug

husky23333:
When using ddp, ckpt is not resumed correctly. Only rank0 loads ckpt, which caused DDP is not initialized properly.
image

The relevant code is as follows, only rank0 loaded ckpt
image

### DIFF
--- a/src/speculators/train/checkpointer.py
+++ b/src/speculators/train/checkpointer.py
@@ -390,8 +390,13 @@ class DistributedCheckpointer(BaseCheckpointer):
             weights_only=True,
             map_location="cpu",
         )
+
+        if isinstance(model, DistributedDataParallel):
+            model_dtype = model.module.dtype
+        else:
+            model_dtype = model.dtype
         full_state_dict = convert_float_dtype(
-            full_state_dict, float_dtype or model.dtype
+            full_state_dict, float_dtype or model_dtype
         )
 
         set_optimizer_state_dict(
--- a/src/speculators/train/trainer.py
+++ b/src/speculators/train/trainer.py
@@ -324,8 +324,7 @@ class Trainer:
         self.model.to(self.local_rank)  # type: ignore[arg-type]
 
         if load_checkpoint:
-            if dist.get_rank() == 0:
-                self.checkpointer.load_model_state_dict(self.model)
+            self.checkpointer.load_model_state_dict(self.model)
         else:
             # Fresh init: broadcast rank 0's random initialization to all ranks
             for param in self.model.parameters():


src/speculators/train/checkpointer.py
```diff
@@ -390,8 +390,13 @@ def load_optimizer_state_dict(
             weights_only=True,
             map_location="cpu",
         )
+
+        if isinstance(model, DistributedDataParallel):
+            model_dtype = model.module.dtype
+        else:
+            model_dtype = model.dtype
```
speculatorsbot, line 397: `DistributedCheckpointer` is only instantiated when `fsdp_shard=True` ([`trainer.py` L179-L183](https://github.com/vllm-project/speculators/blob/322f8c00b6a8d3fb7625e628f2a348ce5fcd6c87/src/speculators/train/trainer.py#L179-L183)). In that FSDP path, the model is never wrapped in `DistributedDataParallel`, so `isinstance(model, DistributedDataParallel)` will always be `False` here.

DDP mode uses `SingleGPUCheckpointer`, whose [`load_optimizer_state_dict`](https://github.com/vllm-project/speculators/blob/322f8c00b6a8d3fb7625e628f2a348ce5fcd6c87/src/speculators/train/checkpointer.py#L327-L330) already handles DDP unwrapping correctly:
```python
raw_model = (
    model.module if isinstance(model, DistributedDataParallel) else model
)
dtype = float_dtype or raw_model.dtype
```

Is this change intended as a defensive measure for future code paths, or was it targeting a specific observed bug? If defensive, consider adding a brief comment explaining when it would apply.

speculatorsbot (commented):
The `trainer.py` change correctly ensures all ranks load the checkpoint model state in DDP mode. Relying solely on rank 0 to load and then depending on the DDP constructor to broadcast creates a window where non-zero ranks wait idle while rank 0 performs disk I/O, which can surface as a timeout or synchronization issue. Removing the `get_rank() == 0` guard is the right fix.

One process item: the commit (`322f8c0`) is missing the required `Signed-off-by` line (ref: [CONTRIBUTING.md — DCO Signoff](https://github.com/vllm-project/speculators/blob/main/CONTRIBUTING.md#dco-signoff)). The DCO CI check is already failing. Fix with:
```bash
git commit --amend --signoff
git push --force-with-lease
```

fynnsu:
Hi @husky23333, could you share a bit more about your setup? I'm trying to reproduce this.

My understanding is that DDP uses the SingleGPUCheckpointer and then uses the `DistributedDataParallel` initialization to broadcast the weight states from rank0 to the other ranks. So we can avoid actually loading the weights on all ranks.

The error you're showing is suggesting there's a model mismatch between the different ranks, but at this stage they should all be initialized and setup the same way. If loading the checkpoint somehow changes the saved weights, then that would be the bug we need to resolve.

husky23333:
Hi @fynnsu , after double-checking, I realized I was looking at the wrong code. DDP correctly uses SingleGPUCheckpointer, so this isn't actually a bug. Really appreciate your review—closing this PR now!

PR #942, opened 2026-08-06
fix(loss): normalize training loss by global (cross-rank) token count

Sawyer117:
## Purpose

`loss_function` (`src/speculators/models/metrics.py`) normalizes the masked training loss by the **per-rank** supervised-token count:

```python
denominator = loss_mask.sum(dim=1) + _EPS
batch_loss  = torch.sum(elementwise_loss, dim=1) / denominator
return batch_loss.mean()
```

Under DDP/FSDP the per-rank gradients are then **mean-averaged**, so the effective objective is the rank-local mean of ratios

```
(1/R) · Σ_r ( Σ_t loss_r / Σ_t mask_r )
```

rather than the token-weighted objective

```
( Σ_r Σ_t loss ) / ( Σ_r Σ_t mask )
```

These differ whenever the per-rank supervised-token counts are uneven: a token's weight in the update ends up depending on which rank it landed on.

This is not prevented by packing. `_lpt_packed_batch` balances **total** tokens against a budget (and its docstring notes that "slot 0 always takes the largest sample and the highest slot absorbs the small ones", so sample counts stay skewed), while the loss denominator counts **supervised** tokens — the packer never looks at `loss_mask`. Two ranks can hold the same token budget and still differ substantially in how many positions are unmasked: one rank holding a single long sample has one masked prompt, while a rank absorbing many small samples pays that prompt cost many times over.

For reference, DeepSpec computes its backward loss this way (`_build_loss`): each term is `local_numerator / all_reduced_denominator`, and the sum is scaled by `world_size`; the rank-local denominators are used only for reported metrics.

## Fix

When distributed, all-reduce the denominator to the global token count and scale by `world_size`, so the subsequent gradient mean-average reconstructs the token-weighted objective.

- **Gradient-identical when ranks are token-balanced** (single rank, or balanced supervised-token counts) — a no-op for that case, so it does not shift the effective LR of tuned recipes.
- Only changes behavior — correctly — when per-rank supervised-token counts are imbalanced.
- The all-reduce is on the **detached denominator only**; autograd flows through this rank's numerator exactly as before.
- Uses the default process group, matching `fully_shard`'s gradient reduction; correct under sequence parallelism as well, where summing across shards recovers the full sequence's token count.
- Batch is `[1, seq_len]` after collation, so within-rank semantics are unchanged; only the cross-rank normalization changes.
- Cost is one scalar all-reduce per loss term per step; the call is rank-symmetric, like the collectives the trainer already performs.
- Device-agnostic; no new dependencies.

## Tests

New `tests/unit/models/test_loss_global_reduce.py`:

- `test_single_process_is_unchanged` — the non-distributed path still equals the existing masked mean.
- `test_ddp_averaged_grad_equals_global_objective_under_imbalance` — 2-rank gloo with **uneven** per-rank token counts; the mean-averaged gradient equals the analytically-computed gradient of the global objective.

```bash
pytest tests/unit/models/test_loss_global_reduce.py -v
# 2 passed
```

`ruff check` / `ruff format` clean — no new findings (the two pre-existing `PLR0917` in `metrics.py` are unchanged from `main`).

## Checklist

I have filled in:


Sawyer117:
Noting the overlap flagged above: #871 (Ulysses SP) touches the same lines, but the two address different problems — #871 all-reduces numerator and denominator over the **SP group** to reassemble a sequence sharded across SP ranks, while this PR corrects the **cross-DP** normalization (uneven supervised-token counts across data-parallel ranks give a mean-of-ratios objective instead of a token-weighted one). They're complementary rather than duplicate.

They will conflict textually, and the scaling factor here (`world_size`) assumes gradients are mean-averaged over the default group. Happy to rebase this on top of #871 once it lands and adjust the group/scaling to match its SP gradient handling.

speculatorsbot (commented):
The math is correct: all-reducing the detached denominator and scaling by `world_size` reconstructs the global token-weighted objective `(Σ_r Σ_t loss) / (Σ_r Σ_t mask)` after DDP/FSDP gradient mean-averaging. The implementation is clean — one scalar all-reduce per loss term, detached denominator keeps autograd on the numerator only, and the existing per-rank path remains for non-distributed use.

The 2-rank gloo test directly asserts the core invariant (averaged gradient under uneven per-rank token counts equals the analytic global objective gradient), which is the right thing to test. Single-process invariance test covers the regression angle.

One consistency concern flagged inline.

src/speculators/models/metrics.py
```diff
@@ -536,6 +537,16 @@ def loss_function(
...
 
+    if dist.is_available() and dist.is_initialized() and dist.get_world_size() > 1:
+        # Normalize by the GLOBAL token count: the per-rank mean below, once
+        # DDP/FSDP mean-averages the grads, yields the rank-local mean of ratios
+        # instead of Σ_r Σ_t loss / Σ_r Σ_t mask. The ×world_size counteracts that
+        # averaging, so this is grad-identical for token-balanced ranks and only
+        # corrects the imbalanced case. Denominator is detached (a normalizer).
+        global_den = loss_mask.sum().detach()
+        dist.all_reduce(global_den, op=dist.ReduceOp.SUM)
+        return elementwise_loss.sum() / (global_den + _EPS) * dist.get_world_size()
```
speculatorsbot: This correctly normalizes `loss_function`, but DSpark's combined loss includes a confidence BCE term computed via `_masked_decayed_mean` ([`dspark/metrics.py:31-45`](https://github.com/vllm-project/speculators/blob/b387464524ed3f5448f54b9c4bc1c4bdc15b1aaa/src/speculators/models/dspark/metrics.py#L31-L45)), which has its own per-rank masked-mean normalization:

```python
denominator = loss_mask.sum(dim=1) + _EPS
return (weighted.sum(dim=1) / denominator).mean()
```

After this PR, DSpark's combined loss `loss + confidence_head_alpha * conf_loss` has two terms with *different* normalization: the speculative loss (via `compound_loss` → `loss_function`) is globally normalized, but the confidence loss remains per-rank normalized. Under uneven per-rank token counts, the confidence loss gradient would still carry the mean-of-ratios bias.

Would you want to address this here (e.g., by extracting the global-reduce pattern into a shared helper that `_masked_decayed_mean` can also call), or track it as a follow-up?

eldarkurtic:
@Sawyer117 could you please check training runtime implications of this update? And some short training runs with accuracy evals would be good as well.

src/speculators/models/metrics.py
Sawyer117: Folded in — good catch, and the "different normalization for two terms that get added together" framing is the right one: fixing only `loss_function` would have left DSpark optimizing a token-weighted term plus a mean-of-ratios term, which is arguably worse than fixing neither.

`ad43c34` extracts `global_token_weighted_mean(numerator, mask_sum)` into `models/metrics.py` and calls it from both `loss_function` and `dspark/metrics.py::_masked_decayed_mean`. It returns `None` when there is no process group or `world_size == 1`, so each caller keeps its existing per-rank path verbatim for single-process runs. Same two-rank gloo test added for the confidence term.

speculatorsbot (commented):
The follow-up commit cleanly addresses the consistency concern from the previous review. Extracting the global-reduce pattern into `global_token_weighted_mean` and applying it to both `loss_function` and `_masked_decayed_mean` is the right call -- DSpark's combined loss `loss + confidence_head_alpha * conf_loss` now has both terms under the same normalization, so the sum is no longer mixing a token-weighted term with a mean-of-ratios term under imbalance.

Implementation notes:

- The helper's `None`-return convention keeps the existing per-rank path completely untouched for single-process runs -- good for not breaking tuned recipes.
- `.detach().clone()` on the denominator before `all_reduce` is the defensive choice: the helper cannot assume its `mask_sum` argument shares no storage with a graph-tracked tensor, even though callers currently pass `loss_mask.sum()` (a fresh tensor). Correct.
- Placement in `_masked_decayed_mean` is after the decay weighting, so the globally normalized sum includes position-decay -- matches the per-rank path's semantics.
- The new confidence-term tests mirror the `loss_function` tests structurally (single-process invariance + 2-rank gloo gradient check), and the test docstring correctly explains why covering both call sites matters.

No remaining concerns. LGTM.

🤖 Generated with [Claude Code](https://claude.ai/code)

Sawyer117:
@eldarkurtic Thanks for the suggestion — ran both arms: a DeepSeek-V4-Flash DSpark draft on 8 Ascend NPUs (FSDP2 + expert-parallel MoE), same recipe / data / steps / build, only this normalization toggled.

**Runtime: no measurable cost.** One scalar all-reduce per loss term per step; step time came out +20 / +50 / +40 ms across three samples on a ~2080 ms step, within this cluster's ~2.6% run-to-run drift.

**Accuracy: unchanged.** Acceptance length at 0.5 epoch, full 5-dataset eval, greedy, `num_speculative_tokens=5`:

| dataset | off | on | Δ |
|---|---:|---:|---:|
| gsm8k | 4.309 | 4.298 | −0.011 |
| math500 | 4.068 | 4.048 | −0.020 |
| humaneval | 4.298 | 4.326 | +0.028 |
| mbpp | 3.908 | 3.911 | +0.003 |
| mt-bench | 2.627 | 2.626 | −0.001 |
| **mean** | **3.8420** | **3.8418** | **−0.0002** |

Although the results look identical, the ranks really were imbalanced in this run: cumulative supervised tokens ran from 5,731,147 (r0) down to 4,801,747 (r7) — 1.194× heaviest to lightest, consistently the same ranks — so a token that landed on r7 carried ~19% more weight in the loss than an identical token on r0, for the whole run. 24 of 2095 logged steps had a rank with zero supervised tokens.

Run is at 64% (3.2 of 5 epochs) — I will post later checkpoints as they are evaluated.

Sawyer117:
One thing I should flag: my tree predates #867 — my merge base with main is from 17 July. The skew I reported above (monotone in rank index, the same ranks heavy every batch) looks like the LPT slot-id tie-break you fixed there, so 1.194× is probably larger than what current main would show.

It isn't quite the same quantity — #867 balances sample count (2.40× → 1.00×), and total tokens were already at 1.002×, whereas what I measured is supervised tokens (`loss_mask.sum()`). My guess at the mechanism is per-sample prompt overhead: a rank holding many short samples spends more of its budget on unsupervised prefix. I haven't verified that, and I don't know how much skew survives the rotation.

The case for this PR doesn't rest on it — any residual imbalance reintroduces the same bias, and a rank with zero supervised tokens stays possible either way. But better you hear it from me than assume the number carries over. I can re-measure on a tree that includes #867 if that would be useful.

zihanlin-ai:
Two things on this change, from carrying an equivalent fix downstream:

**1. A deterministic unit test for the gradient, not just the loss value.** Two-rank Gloo test with deliberately uneven supervised-token counts: same tiny model on both ranks, rank 0 gets a batch with 1 valid token and rank 1 a batch with 3, run the loss, let DDP average the gradients, compare the resulting `param.grad` against the gradient of the single-process global weighted mean over all 4 tokens. With per-rank normalization the numbers differ by construction (each rank's tokens get weight 1/(2·n_rank) instead of 1/4); with the global denominator they match to fp32 precision. ~40 lines.

The downstream fix is a `distributed_weighted_mean` that all-reduces the denominator and scales by world size so the DDP/FSDP-averaged gradient equals the gradient of the global weighted mean; that test is what caught it: on a two-rank CUDA probe with uneven valid-token counts the DDP-averaged gradient came out 2.999986 where the single-process global reference was 4.0. *(Edited: an earlier version said 8 ranks; the preserved evidence is the two-GPU probe.)*

**2. Same shape in one more place.** `dspark/metrics.py::_masked_decayed_mean` (confidence-head BCE) has the same per-rank `loss_mask.sum()` denominator; routing it through the same helper keeps `loss + conf_alpha * conf_loss` consistently normalized. The confidence term is the one whose per-rank token counts vary most once anchors are sampled.

PR #941, opened 2026-08-05
fix(dflash): resolve quality failures and PR feedback for Laguna converter

shotsan:
This PR fixes the quality checks and implements the CodeRabbit review feedback for PR #922. It ensures mixed fused/separate projections raise a ValueError.

PR #945, opened 2026-08-06
fix(dflash): resolve quality failures and PR feedback for Laguna converter (clean)

santosh-agebold:
This PR fixes the quality checks and implements the CodeRabbit review feedback for PR #922. It ensures mixed fused/separate projections raise a ValueError. (Replacing PR #941 which had merge conflicts).

### DIFF
--- a/src/speculators/convert/dflash/converter.py
+++ b/src/speculators/convert/dflash/converter.py
@@ -173,10 +173,19 @@ class DFlashConverter:
 
         for key, tensor in weights.items():
             if "qkv_proj" in key:
+                q_key = key.replace("qkv_proj", "q_proj")
+                k_key = key.replace("qkv_proj", "k_proj")
+                v_key = key.replace("qkv_proj", "v_proj")
+                for dest_key in (q_key, k_key, v_key):
+                    if dest_key in remapped or dest_key in weights:
+                        raise ValueError(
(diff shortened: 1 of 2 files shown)


speculatorsbot (commented):
LGTM. Clean, focused change that adds a necessary guard.

**What it does**: Validates that when splitting fused `qkv_proj` into separate `q/k/v_proj`, the destination keys don't already exist -- either in the original `weights` dict or in the already-remapped output. This catches malformed checkpoints that mix fused and separate projection formats, which would otherwise silently overwrite data.

**Correctness**: The dual check (`dest_key in remapped or dest_key in weights`) correctly handles both dict iteration orderings:
- If `qkv_proj` is iterated before `q_proj`: the `in weights` check catches the conflict.
- If `q_proj` is iterated first (added to `remapped` via the `else` branch): the `in remapped` check catches it when `qkv_proj` is processed.

The pre-computation of `q_key`/`k_key`/`v_key` also eliminates three redundant `str.replace()` calls per fused key, which is a nice cleanup.

**Test coverage**: `test_rejects_mixed_fused_and_separate_projections` exercises the `in weights` path. The `in remapped` path is implicitly covered by the ordering guarantee but could optionally get its own test in a follow-up.

No issues found.

PR #951, opened 2026-08-06
[Memory] Fuse every loss through one Triton online-softmax kernel

WindChimeRan:
## Summary

Extend the Triton online-softmax loss fusion from #752 to every configured dense loss.

Eager losses materialize several full-vocabulary fp32 tensors. With bf16 inputs, `V=151936`, and `T=8192`, the default `kl_div` loss peaks at 23.2 GB beyond its inputs. The fused path uses 2.3 GB, saving approximately 21 GB per GPU, while making the losses 1.2–4× faster.

## New change

- Keep the eager loss implementations for compatibility and numerical validation, noting that they may OOM with DFlash or DSpark.
- Remove automatic fallback between fused and eager losses.
- Add `--loss-implementation {fused,eager}` for explicit selection, with fused as the default.
- Refactor the loss implementations and shared utilities into `src/speculators/losses/`.

## Implementation

- A shared forward/backward kernel handles KL, reverse KL, JSD, CE, and TV; a compile-time `OP` selects the reduction.
- Backward recomputes probabilities from five saved scalars per row, so no `[T, V]` intermediate is saved.
- NLA and LK-hybrid compose the fused primitives using small `[1, T]` operations.
- Users can explicitly select the unchanged eager implementations with `--loss-implementation eager`; fused remains the default.
- JSD now computes in fp32, reducing bf16 gradient error from 15.1% to 0.14%.
- DSpark computes its confidence target as `1 - TV`, reusing the fused kernel under `no_grad`.

## Results

Measured on GB10 with bf16, `V=151936`, `T=8192`, and forward + backward under autocast:

| loss | peak memory | speedup |
|---|---|---|
| kl_div | 23.2 → 2.3 GB | 2.7× |
| rkl | 23.2 → 2.3 GB | 2.7× |
| jsd | 37.1 → 2.3 GB | 1.2× |
| ce | 11.6 → 2.3 GB | 1.4× |
| tv | 20.9 → 2.3 GB | 3.9× |
| nla | 20.9 → 2.3 GB | 4.0× |
| lk_hybrid | 32.5 → 7.0 GB | 3.1× |

## Not planned

- **In-place backward.** SpecForge overwrites the logits buffer with its gradient, eliminating one remaining bf16 buffer (~2.3 GiB at `T=8192`, `V=151936`). This requires the logits storage to be dead and to have a single differentiable consumer. Speculators cannot guarantee that: `compound_loss` can apply multiple losses to the same logits, and LK-hybrid itself composes KL and TV. One backward could therefore corrupt logits needed by another.

- **Mask-aware forward.** Losses intentionally return per-token values, with `loss_mask` and decay applied afterward. Masked rows already reach the fused backward with `grad_output == 0`, which skips probability recomputation while zero-filling the gradient row. Passing the mask into the fused forward would optimize only the forward pass and require special-casing the fused and eager implementations, so it is not included.

## Tests

```text
pytest tests/unit -q  # 721 passed, 5 skipped
pytest tests/unit/models/test_fused_losses.py -q
```

Per-loss CUDA tests compare values and gradients with eager execution across fp32 edge cases, bf16 training inputs, and a 151936-wide vocabulary. They also cover explicit eager selection, zero upstream gradients, and `torch.compile(fullgraph=True)`.

## Checklist

- [x] Purpose and motivation
- [x] Test plan and results
- [x] Documentation update, if needed
- [x] Human-authored or reviewed


speculatorsbot (commented):
## Summary

This PR generalizes the fused Triton TV/NLA kernel from #752 into a single kernel pair that serves all seven losses in `_LOSS_FN_MAP`. The approach is clean: one forward/backward kernel with a `tl.constexpr OP` selector that Triton compiles into separate specialized instances (untaken branches are pruned). The memory savings are dramatic -- from 11-37 GB down to 2.3 GB per GPU at 8k tokens -- because no `[T, V]` intermediate is ever materialized or saved; backward recomputes probabilities from 5 saved scalars per row.

## Correctness

I verified all five per-row gradient formulas by hand against the analytical derivatives of the respective losses through softmax:

- **KL**: `dp - tp` -- correct for `d/d(logit) KL(tp || dp)`
- **RKL**: `dp * (log(dp/tp) - L)` where L is the row loss -- correct for `d/d(logit) KL(dp || tp)`, using the standard softmax Jacobian identity
- **JSD**: `0.5 * dp * (log(dp/mix) - KL(dp||mix))` -- correct; the `KL(dp||mix)` term (stored in `stats[4]`) arises from the softmax sum-to-one constraint
- **CE**: `dp - onehot(argmax tp)` -- standard softmax cross-entropy gradient
- **TV**: `-dp * (1[dp <= tp] - s_s)` -- correct; the indicator derivative through `min(dp, tp)` combined with the softmax Jacobian

The composed losses (`nla` and `lk_hybrid`) correctly chain through autograd rather than adding kernel branches, which is the right design choice since they combine row-level outputs with torch ops.

## Design

The `_fused_or_eager` factory in metrics.py is a good improvement over the hand-written per-loss dispatchers. The `targets.requires_grad` guard is important since the fused backward returns no target gradient -- this is well-documented and tested (`test_dispatcher_falls_back_to_eager_for_differentiable_targets`).

The commit organization is clean: rename, then kernel + dispatch, then the dspark derivation.

## Tests

Test coverage is comprehensive:
- Three distinct failure-mode regimes (fp32 with saturated point masses, bf16, 151936-wide multi-block vocab)
- Upstream gradient with exact zeros to exercise the `go == 0` early-out path
- Dispatcher guards (fused == bitwise-equal on CUDA, differentiable targets fall back to eager)
- `torch.compile(fullgraph=True)` for every loss (guards the compiled training path)
- The dspark accept-rate pinned against the explicit softmax-overlap formula

## On the CodeRabbit suggestions

**Re: broadening `_fused_kernel` to catch `Exception`** -- I think the current `except ImportError` is the right choice. The lazy import loads a Python module; Triton JIT compilation happens later when a kernel is first called with specific arguments, not at module import time. If `import fused_losses` raises something other than `ImportError`, that signals a genuine bug in the module (e.g., `NameError`, `SyntaxError`) that should propagate rather than be silently swallowed. Catching `Exception` would mask real problems and make debugging harder.

**Re: duplicated `_EPS` and `eta` constants** -- agree this is a minor maintenance risk, but it is a nitpick. Creating a shared constants module for two values that each have clear comments pointing to their counterpart seems like overengineering. A cross-reference comment (already present: `# matches models/metrics.py`) is sufficient.

## Minor observation (non-blocking)

The CE path stores `argmax(targets)` as float32 in the stats buffer, with the comment "exact: vocab < 2**24". This is safe for all current LLM vocabs (max ~152k), but if a future model exceeds 16M vocab entries, the round-trip through float32 would silently truncate. Consider adding an assertion like `assert V < 2**24` in the forward path (or in tests) to document and enforce this invariant.

LGTM -- well-designed, well-tested, and the memory/performance wins are substantial.

src/speculators/models/metrics.py
```diff
@@ -381,49 +381,62 @@ def dpace_loss_decay(
...
+
+    def loss_fn(logits: torch.Tensor, targets: torch.Tensor) -> torch.Tensor:
+        if logits.is_cuda and not targets.requires_grad:
+            fused = _fused_kernel(fused_name)
+            if fused is not None:
+                return fused(logits, targets)
+        return eager_fn(logits, targets)
+
+    loss_fn.__name__ = f"{eager_fn.__name__}_fused_or_eager"
+    return loss_fn
```
mgoin: Setting `__name__` does not change `__qualname__`, so `pickle.dumps(resolve_loss_config("kl_div"))` fails with `Can't pickle local object '_fused_or_eager.<locals>.loss_fn'`. 

Previously all mapped functions were module-level and picklable. This breaks loss configs passed through spawn-based multiprocessing or similar serialization. Use module-level callables, a picklable callable class/partial, or correctly publish the generated functions.

src/speculators/losses/fused.py
```diff
@@ -0,0 +1,328 @@
...
+
+
+def fused_lk_hybrid_loss(logits, targets, eta: float = 3.0):
+    """Per-position hybrid LK ``[1, T]``; fused KL + TV composed as in
+    ``lk_hybrid_loss`` (blend weight detached).
+    """
+    kl = fused_kl_div_loss(logits, targets)
+    tv = fused_tv_loss(logits, targets)
+    weight = torch.exp(-eta * (1.0 - tv).detach())
+    return weight * kl + (1.0 - weight) * tv
```
mgoin, line 330: It seems the `eta` argument here has no chance to get passed in. The new dispatcher only accepts `(logits, targets)`, so `resolve_loss_config("lk_hybrid")["lk_hybrid"][0](x, y, eta=1.0)` would now raise.

mgoin:
Overall I feel the callable factory is over-abstracted. I’d favor explicit top-level wrappers with one shared dispatch helper like this

```python
def _dispatch_fused(
    fused_name: str,
    eager_fn: Callable,
    logits: torch.Tensor,
    targets: torch.Tensor,
    *args,
    **kwargs,
):
    if logits.is_cuda and not targets.requires_grad:
        fused_fn = _fused_kernel(fused_name)
        if fused_fn is not None:
            return fused_fn(logits, targets, *args, **kwargs)
    return eager_fn(logits, targets, *args, **kwargs)


def kl_div_loss_fused_or_eager(logits, targets):
    return _dispatch_fused(
        "fused_kl_div_loss", kl_div_loss, logits, targets
    )


def lk_hybrid_loss_fused_or_eager(logits, targets, eta: float = 3.0):
    return _dispatch_fused(
        "fused_lk_hybrid_loss",
        lk_hybrid_loss,
        logits,
        targets,
        eta=eta,
    )
```

The cached lazy import is fine, although doing it while Dynamo traces produces an annoying warning. 
A module-level optional import would be simpler if importing Triton is reliably harmless (or just include a util that does `from importlib.util import find_spec; HAS_TRITON = find_spec("triton") is not None`)
```
try:
    from speculators.models import fused_losses as _fused_losses
except ImportError:
    _fused_losses = None
```

WindChimeRan:
After talking with @fynnsu, we agreed to keep both the eager and fused paths without automatic fallback. Fused remains the default, while users can explicitly pass `--loss-implementation eager` for debugging and numerical validation, e.g.,, when testing support on a particular NPU stack.

The follow-up changes are intentionally small. Much of the large LOC diff is code movement: the eager losses were moved out of `models/metrics.py`, and the loss implementations and shared utilities were organized under `src/speculators/losses/`.

New commits:

- Keep the eager implementations for compatibility and numerical validation, while noting that they may OOM with DFlash or DSpark.
- Remove automatic fallback between fused and eager losses.
- Add `--loss-implementation {fused,eager}` for explicit selection, defaulting to fused.
- Move the implementations and shared utilities into `src/speculators/losses/`.

fynnsu (approved):
Great work! Thank you

PR #958, opened 2026-08-07
feat(launch_vllm): add vLLM provenance artifacts (RFC #880)

orestis-z:
## Summary
- Add `vllm_command.txt`, `vllm.patch`, and `checkpoint_sha256.txt` to `launch_vllm.py` provenance directory
- Each artifact writer is independent and best-effort — failures warn but never block the vLLM launch
- Add `--provenance-dir`, `--no-hash-checkpoints`, and `--dry-run` flags
- Add vllm_command.txt hint in `evaluate.py` when no co-located provenance exists
- Uses `.exists()` for `.git` detection to support git worktrees

Part 2 of the RFC #880 provenance stack. Depends on #957.

## Test plan
- [x] `tests/unit/scripts/test_provenance_integration.py` — tests for provenance file creation, metadata content, checkpoint handling
- [x] `make quality` passes
- [x] All existing tests pass

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -2,9 +2,10 @@
 
 ## Provenance
 
-Training and eval scripts write reproducibility artifacts so runs can be recreated later.
+Training, eval, and vLLM-launch scripts write reproducibility artifacts so runs can be recreated later.
 
 - **Training** (`src/speculators/train/utils.py`): `save_train_command()` writes `train_command.txt` (timestamp, git SHA, world size, package versions, full argv) into the checkpoint directory.
 - **Eval** (`scripts/evaluate/evaluate.py`): `save_eval_provenance()` writes `eval_command.txt` (timestamp, git SHA, package versions, full argv) into the output directory.
+- **vLLM launch** (`scripts/launch_vllm.py`): `_save_vllm_provenance()` writes `vllm_command.txt`, `vllm.patch`, and `checkpoint_sha256.txt` into `--provenance-dir`.
 
-When publishing a model or eval results (HuggingFace, GitHub comments, etc.), always include the provenance artifacts so the run can be reproduced.
+When publishing a model or eval results (HuggingFace, GitHub comments, etc.), always include the provenance artifacts (`train_command.txt`, `eval_command.txt`, `vllm_command.txt`, patches) so the run can be reproduced.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -8,4 +8,4 @@ Training, eval, and vLLM-launch scripts write reproducibility artifacts so runs 
 - **Eval** (`scripts/evaluate/evaluate.py`): `save_eval_provenance()` writes `eval_command.txt` (timestamp, git SHA, package versions, full argv) into the output directory.
 - **vLLM launch** (`scripts/launch_vllm.py`): `_save_vllm_provenance()` writes `vllm_command.txt`, `vllm.patch`, and `checkpoint_sha256.txt` into `--provenance-dir`.
 
-When publishing a model or eval results (HuggingFace, GitHub comments, etc.), always include the provenance artifacts (`train_command.txt`, `eval_command.txt`, `vllm_command.txt`, patches) so the run can be reproduced.
+When publishing a model or eval results (HuggingFace, GitHub comments, etc.), always include the provenance artifacts (`train_command.txt`, `eval_command.txt`, `vllm_command.txt`, `checkpoint_sha256.txt`, patches) so the run can be reproduced.
--- a/scripts/launch_vllm.py
+++ b/scripts/launch_vllm.py
@@ -251,21 +251,15 @@ def _save_vllm_provenance(
     writers = [
         (
             "vllm_command.txt",
-            lambda: _save_vllm_command(
-                prov_dir, cmd, sha, diff, vllm_ver
-            ),
+            lambda: _save_vllm_command(prov_dir, cmd, sha, diff, vllm_ver),
         ),
         (
             "vllm.patch",
-            lambda: _save_vllm_patch(
-                prov_dir, vllm_repo, sha, diff, vllm_ver
-            ),
+            lambda: _save_vllm_patch(prov_dir, vllm_repo, sha, diff, vllm_ver),
         ),
         (
             "checkpoint_sha256.txt",
-            lambda: _save_checkpoint_sha256(
-                prov_dir, model, skip_hash=skip_hash
-            ),
+            lambda: _save_checkpoint_sha256(prov_dir, model, skip_hash=skip_hash),
         ),
     ]
     for artifact, write in writers:
--- a/scripts/launch_vllm.py
+++ b/scripts/launch_vllm.py
@@ -183,7 +183,7 @@ def _save_vllm_patch(
     vllm_ver: str,
 ) -> None:
     if vllm_repo:
-        content = f"# repo: {vllm_repo} ({sha})\n{diff}"
+        content = f"# repo: {vllm_repo} ({sha})\n{diff}\n"
     else:
         content = f"# vllm {vllm_ver} (wheel install, no git repo found)\n"
     atomic_write(prov_dir / "vllm.patch", content)
@@ -326,7 +326,7 @@ def main():
 
     if not args.provenance_dir:
         sanitized = args.model.replace("/", "_").replace(" ", "_")
-        ts = datetime.datetime.now().strftime("%Y%m%d_%H%M%S")
+        ts = datetime.datetime.now(tz=datetime.timezone.utc).strftime("%Y%m%d_%H%M%S")
         args.provenance_dir = f"vllm_{sanitized}_{ts}"
     _save_vllm_provenance(
         cmd,
--- a/scripts/launch_vllm.py
+++ b/scripts/launch_vllm.py
@@ -1,15 +1,22 @@
 import argparse
 import datetime
 import hashlib
-import importlib.metadata
-import importlib.util
 import json
 import os
 import shlex
-import subprocess
 import sys
-import tempfile
 import warnings
+from pathlib import Path
+
+from speculators.provenance import (
+    atomic_write,
+    find_package_repo,
+    git_diff,
+    pkg_version,
+)
+from speculators.provenance import (
+    git_sha as _git_sha,
+)
 
 try:
     from hs_connectors import HiddenStatesBackend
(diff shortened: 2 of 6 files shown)


scripts/launch_vllm.py
```diff
@@ -93,6 +120,222 @@ def parse_args():
...
+    Best-effort — failures warn but never block the vLLM launch.
+    """
+    try:
+        os.makedirs(provenance_dir, exist_ok=True)
+    except OSError as exc:
+        _warn(f"could not create provenance dir: {exc}")
+        return
+
+    vllm_repo = _find_vllm_repo()
+    git_sha = _run_git(["git", "rev-parse", "HEAD"], vllm_repo or ".") or "unknown"
```
speculatorsbot, line 290: **Misleading git SHA when vLLM is installed from a wheel.** When `_find_vllm_repo()` returns `None` (no source checkout), this falls back to `cwd="."`, so `git rev-parse HEAD` returns the SHA of whatever repo the CWD happens to be in (e.g., the speculators repo). That SHA then gets recorded in `vllm_command.txt` as `# Git SHA: <speculators-sha>`, which is misleading — a reader would assume it refers to the vLLM commit.

The `diff` computation on [lines 291–298](https://github.com/vllm-project/speculators/blob/a70583c3204465eabad8a246eecf8158968601ad/scripts/launch_vllm.py#L291-L298) already handles this correctly with `if vllm_repo else ""`, but the `git_sha` assignment doesn't have the same guard.

Suggested fix:
```python
git_sha = (_run_git(["git", "rev-parse", "HEAD"], vllm_repo) if vllm_repo else "") or "unknown"
```

speculatorsbot (commented):
Clean implementation that follows RFC #880 and the existing `save_train_command()` patterns well. The best-effort error handling, atomic writes, and test coverage are all solid. One correctness issue with git SHA resolution when vLLM is installed from a wheel.

tests/unit/scripts/test_vllm_provenance.py
```diff
@@ -0,0 +1,193 @@
...
+import sys
+from pathlib import Path
+from unittest.mock import patch
+
+sys.path.insert(0, str(Path(__file__).resolve().parents[3] / "scripts"))
+
+
+from launch_vllm import (  # type: ignore[import-not-found]
+    _find_vllm_repo,
+    _is_vllm_repo,
```
shubhra: It looks like `_is_vllm_repo` doesn't exist in `launch_vllm` but there is a check inside `_find_vllm_repo` that does the same thing. It might be possible to just drop the `_is_vllm_repo` import and delete the `TestIsVllmRepo` class because `_find_vllm_repo` already covers that behavior.

orestis-z: thank you!

AGENTS.md
```diff
@@ -2,9 +2,10 @@
...
 
-Training and eval scripts write reproducibility artifacts so runs can be recreated later.
+Training, eval, and vLLM-launch scripts write reproducibility artifacts so runs can be recreated later.
 
 - **Training** (`src/speculators/train/utils.py`): `save_train_command()` writes `train_command.txt` (timestamp, git SHA, world size, package versions, full argv) into the checkpoint directory.
 - **Eval** (`scripts/evaluate/evaluate.py`): `save_eval_provenance()` writes `eval_command.txt` (timestamp, git SHA, package versions, full argv) into the output directory.
+- **vLLM launch** (`scripts/launch_vllm.py`): `_save_vllm_provenance()` writes `vllm_command.txt`, `vllm.patch`, and `checkpoint_sha256.txt` into `--provenance-dir`.
 
-When publishing a model or eval results (HuggingFace, GitHub comments, etc.), always include the provenance artifacts so the run can be reproduced.
+When publishing a model or eval results (HuggingFace, GitHub comments, etc.), always include the provenance artifacts (`train_command.txt`, `eval_command.txt`, `vllm_command.txt`, patches) so the run can be reproduced.
```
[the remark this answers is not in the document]
orestis-z: fixed

tests/unit/scripts/test_provenance_integration.py
```diff
@@ -0,0 +1,81 @@
...
+
+import subprocess
+import sys
+from pathlib import Path
+
+import pytest
+
+SCRIPTS_DIR = Path(__file__).resolve().parents[3] / "scripts"
+LAUNCH_VLLM = str(SCRIPTS_DIR / "launch_vllm.py")
+MODEL = "gpt2"
```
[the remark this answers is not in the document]
orestis-z: gpt2 is a tiny config (~700 bytes). Fully airgapped CI worker would fail. Skipping this unless we see it fail in the CI/

src/speculators/train/utils.py
```diff
@@ -107,11 +107,22 @@ def save_train_command(save_path: str, argv: list[str] | None = None) -> None:
...
+    repo_root = None
+    try:
+        d = Path(__file__).resolve().parent
+        while d != d.parent:
+            if (d / ".git").exists():
+                repo_root = d
+                break
+            d = d.parent
+    except OSError:
+        pass
```
rahul-tuli: Should this use `find_repo_root(Path(__file__))` from `speculators.provenance` instead of reimplementing the walk-up loop?

scripts/launch_vllm.py
```diff
@@ -148,6 +324,17 @@ def main():
     print("Running command:")
     print(" ".join(cmd))
 
+    if not args.provenance_dir:
+        sanitized = args.model.replace("/", "_").replace(" ", "_")
+        ts = datetime.datetime.now().strftime("%Y%m%d_%H%M%S")
```
rahul-tuli: Could we make the default provenance-dir timestamp also use UTC (e.g. `datetime.datetime.now(datetime.timezone.utc)`) to be consistent with the UTC timestamp written inside `vllm_command.txt` (or the opposite)? I believe right now this gives us the local time:

```python
In [1]: import datetime

In [2]: datetime.datetime.now().strftime("%Y%m%d_%H%M%S")
Out[2]: '20260818_131943'

In [3]: datetime.datetime.now(tz=datetime.timezone.utc).isoformat() # Used by _save_vllm_command
Out[3]: '2026-08-18T13:20:23.222252+00:00'
```

scripts/launch_vllm.py
```diff
@@ -1,8 +1,22 @@
...
+
+from speculators.provenance import (
+    atomic_write,
+    find_package_repo,
+    git_diff,
+    pkg_version,
+)
+from speculators.provenance import (
+    git_sha as _git_sha,
+)
```
rahul-tuli, line 19: nit: Would it be cleaner to merge the two `from speculators.provenance import ...` blocks into one? The `git_sha as _git_sha` alias can be included in the first import group.

scripts/launch_vllm.py
```diff
@@ -93,6 +127,148 @@ def parse_args():
     return parser.parse_known_args()
 
 
+def _warn(msg: str) -> None:
+    print(f"Warning: {msg}", file=sys.stderr)
```
rahul-tuli, line 131: Should `_warn()` use `warnings.warn(..., stacklevel=2)` since its already imported? or is there a deliberate reason to bypass the warnings machinery here?

scripts/launch_vllm.py
```diff
@@ -93,6 +127,148 @@ def parse_args():
...
+
+def _save_vllm_patch(
+    prov_dir: Path,
+    vllm_repo: str | None,
+    sha: str,
+    diff: str,
+    vllm_ver: str,
+) -> None:
+    if vllm_repo:
+        content = f"# repo: {vllm_repo} ({sha})\n{diff}"
```
rahul-tuli: When `vllm_repo` is set and `diff` is non-empty, the content is `f"# repo: {vllm_repo} ({sha})\n{diff}"`. Since `git_diff()` strips its output, the resulting file will not end with a newline character. Patch files conventionally end with a trailing newline, and tools like `git apply` may warn about the missing newline.
Could the content ensure a trailing newline (e.g. `f'# repo: {vllm_repo} ({sha})\n{diff}\n'`) so the generated `.patch` file is well-formed for `git apply`?"

scripts/launch_vllm.py
```diff
@@ -93,6 +127,148 @@ def parse_args():
...
+def _save_checkpoint_sha256(
+    prov_dir: Path, model: str, *, skip_hash: bool = False
+) -> None:
+    dest = prov_dir / "checkpoint_sha256.txt"
+    model_path = os.path.expanduser(model)
+    if not os.path.isdir(model_path):
+        atomic_write(dest, f"# model: {model} (not a local path)\n")
+        return
+    safetensors = sorted(
+        f for f in os.listdir(model_path) if f.endswith(".safetensors")
```
rahul-tuli, line 201: nit: Could we add a comment noting that only top-level `.safetensors` files are hashed? since `os.listdir` is non-recursive

scripts/launch_vllm.py
```diff
@@ -148,6 +324,17 @@ def main():
     print("Running command:")
     print(" ".join(cmd))
 
+    if not args.provenance_dir:
+        sanitized = args.model.replace("/", "_").replace(" ", "_")
+        ts = datetime.datetime.now().strftime("%Y%m%d_%H%M%S")
+        args.provenance_dir = f"vllm_{sanitized}_{ts}"
+    _save_vllm_provenance(
```
rahul-tuli, line 331: When `--dry-run` is passed, the vLLM server is not launched, but provenance artifacts are still written to disk, is that the intended behaviour? Maybe we should move this call below after the `if not args.dry_run` check

rahul-tuli (commented):
Great Job on this! LGTM pendings nits/suggestions

scripts/launch_vllm.py
orestis-z: Ruff's isort wants the aliased import separate. :(

scripts/launch_vllm.py
orestis-z: fixed

scripts/launch_vllm.py
orestis-z: fixed

scripts/launch_vllm.py
orestis-z: yes this is intentional to test the artifact creation behavior. It is also what the tests rely on

scripts/launch_vllm.py
orestis-z: good point. I wanted to have a user-warning that starts with `Warning: ...` and not the dev-facing `UserWarning: could not create provenance dir: ....` that can be filtered in code. Can change it to `warnings.warn` if you have a strong opinion on this

src/speculators/train/utils.py
orestis-z: valid point thank you!

orestis-z: fixed

scripts/launch_vllm.py
orestis-z: i assumed they are always flat on the top-level. do we have use-cases with subdirs?

scripts/launch_vllm.py
rahul-tuli: Aah, I forgot about that

scripts/launch_vllm.py
rahul-tuli: They are for the most part, I'm fine leaving this as is; we don't need to be so pedantic

scripts/launch_vllm.py
rahul-tuli: Sounds good, thank you!

scripts/launch_vllm.py
rahul-tuli: No strong opinion, just wanted to know the reason, resolving this!

rahul-tuli (approved):
Approved, pending rebase! Thanks for the quick turn around

PR #954, opened 2026-08-07
[RFC] feat(launch_vllm): add eval subcommand for speculative decoding serving

orestis-z:
## Summary

- Add `eval` subcommand to `launch_vllm.py` so eval vLLM launches also capture provenance artifacts (`vllm_command.txt`, `vllm.patch`, `checkpoint_sha256.txt`)
- `launch_vllm.py eval MODEL --spec-model DRAFTER [--spec-tokens N] [--spec-method M]` builds a vLLM serve command with speculative decoding flags
- `launch_vllm.py train MODEL` (default) is unchanged — omitting the subcommand defaults to `train` for backward compatibility
- Update `AGENTS.md` to document both subcommands

Depends on #883.

## Test plan

- [x] All 47 provenance tests pass (27 unit + 7 train integration + 6 eval integration + 7 eval provenance)
- [x] `test_implicit_train_subcommand` verifies backward compat (no subcommand → train)
- [x] `test_eval_mode_requires_spec_model` verifies argparse rejects missing `--spec-model`
- [x] Lint clean (`make quality` shows no errors in changed files)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #959, opened 2026-08-07
feat(launch_vllm): add eval subcommand for spec-decode serving

orestis-z:
## Summary
- Add `launch_vllm.py eval MODEL --spec-model DRAFTER` subcommand for launching vLLM in speculative decoding serve mode
- Refactor CLI to use argparse subcommands (`train`/`eval`) with shared args via `_add_shared_args()`
- Backward-compatible: bare `launch_vllm.py MODEL` defaults to `train`
- Default provenance dir: `vllm_<mode>_<model>_<timestamp>/`

Replaces #954. Part 3 of the RFC #880 provenance stack. Depends on #958.

## Test plan
- [x] `TestLaunchVllmEvalMode` — 6 tests covering provenance creation, spec flags, no train flags, requires spec-model, spec-method passthrough, extra vllm args
- [x] `test_implicit_train_subcommand` — backward compat test
- [x] `test_default_provenance_dir_created` — default dir naming
- [x] `make quality` passes

### DIFF
--- a/scripts/launch_vllm.py
+++ b/scripts/launch_vllm.py
@@ -85,9 +85,6 @@ def _add_shared_args(parser: argparse.ArgumentParser) -> None:
     )
 
 
-_SUBCOMMANDS = {"train", "eval"}
-
-
 def parse_args():
     parser = argparse.ArgumentParser(
         description="Launch vLLM for training or evaluation",
@@ -170,8 +167,9 @@ def parse_args():
     )
     _add_shared_args(eval_parser)
 
+    subcommands = set(sub.choices)
     argv = sys.argv[1:]
-    if not argv or (argv[0] not in _SUBCOMMANDS and not argv[0].startswith("-")):
+    if not argv or (argv[0] not in subcommands and not argv[0].startswith("-")):
         argv = ["train", *argv]
     args, vllm_args = parser.parse_known_args(argv)
     if args.subcommand is None:
--- a/scripts/launch_vllm.py
+++ b/scripts/launch_vllm.py
@@ -319,13 +319,17 @@ def _save_vllm_provenance(
     ]
     if spec_model is not None:
         drafter_dest = "drafter_checkpoint_sha256.txt"
-        writers.append((
-            drafter_dest,
-            lambda: _save_checkpoint_sha256(
-                prov_dir, spec_model,
-                skip_hash=skip_hash, dest_name=drafter_dest,
-            ),
-        ))
+        writers.append(
+            (
+                drafter_dest,
+                lambda: _save_checkpoint_sha256(
+                    prov_dir,
+                    spec_model,
+                    skip_hash=skip_hash,
+                    dest_name=drafter_dest,
+                ),
+            )
+        )
     for artifact, write in writers:
         try:
             write()
--- a/scripts/launch_vllm.py
+++ b/scripts/launch_vllm.py
@@ -173,7 +173,10 @@ def parse_args():
     argv = sys.argv[1:]
     if not argv or (argv[0] not in _SUBCOMMANDS and not argv[0].startswith("-")):
         argv = ["train", *argv]
-    return parser.parse_known_args(argv)
+    args, vllm_args = parser.parse_known_args(argv)
+    if args.subcommand is None:
+        args, vllm_args = parser.parse_known_args(["train"] + sys.argv[1:])
+    return args, vllm_args
 
 
 def _warn(msg: str) -> None:
--- a/scripts/launch_vllm.py
+++ b/scripts/launch_vllm.py
@@ -171,7 +171,7 @@ def parse_args():
     _add_shared_args(eval_parser)
 
     argv = sys.argv[1:]
-    if argv and argv[0] not in _SUBCOMMANDS and not argv[0].startswith("-"):
+    if not argv or (argv[0] not in _SUBCOMMANDS and not argv[0].startswith("-")):
         argv = ["train", *argv]
     return parser.parse_known_args(argv)
 
--- a/scripts/launch_vllm.py
+++ b/scripts/launch_vllm.py
@@ -56,56 +56,15 @@ if "file" not in _backend_registry:
     _backend_registry["file"] = _InlineFileBackend  # type: ignore[assignment]
 
 
-def parse_args():
-    parser = argparse.ArgumentParser(
-        description="Launch vLLM for hidden states extraction",
-        usage=(
-            "launch_vllm.py [-h] MODEL [--hidden-states-backend BACKEND] "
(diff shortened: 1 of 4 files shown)


speculatorsbot (commented):
Clean refactoring of `launch_vllm.py` into train/eval subcommands. The shared-args pattern is tidy, backward compat for bare `MODEL` invocations works, and the test coverage is thorough (backward compat, required args, flag passthrough, negative cases). Two items below.

scripts/launch_vllm.py
```diff
@@ -49,75 +49,124 @@ def build_kv_transfer_config(args: argparse.Namespace) -> dict:
...
+        "--spec-method",
+        type=str,
+        default=None,
+        help="Speculative decoding method (optional)",
+    )
+    _add_shared_args(eval_parser)
+
+    argv = sys.argv[1:]
+    if argv and argv[0] not in _SUBCOMMANDS and not argv[0].startswith("-"):
+        argv = ["train", *argv]
```
speculatorsbot, line 168: The guard `argv and ...` skips the transform when `argv` is empty. On Python 3.10–3.11 (where `add_subparsers` creates optional subparsers by default), running `launch_vllm.py` with no arguments sets `subcommand=None`. `main()` then falls through the `else` branch into `_build_eval_cmd`, which crashes with `AttributeError` instead of showing a clean usage message.

The project supports Python 3.10+ per CONTRIBUTING.md. Before this PR the same invocation produced a clear argparse error (`the following arguments are required: model`).

Easiest fix — default empty argv to train so argparse shows the train subparser’s own required-arg error:

```suggestion
    if not argv or (argv[0] not in _SUBCOMMANDS and not argv[0].startswith("-")):
        argv = ["train", *argv]
```

Alternatively, `sub.required = True` right after `add_subparsers` works on all Python versions.

scripts/launch_vllm.py
```diff
@@ -392,13 +437,45 @@ def main():
...
     print("Running command:")
     print(" ".join(cmd))
 
     if not args.provenance_dir:
         sanitized = args.model.replace("/", "_").replace(" ", "_")
         ts = datetime.datetime.now().strftime("%Y%m%d_%H%M%S")
-        args.provenance_dir = f"vllm_{sanitized}_{ts}"
+        mode = args.subcommand
+        args.provenance_dir = f"vllm_{mode}_{sanitized}_{ts}"
     _save_vllm_provenance(
```
speculatorsbot, line 479: In eval mode the drafter model (`args.spec_model`) is the primary artifact being evaluated, but only the target model (`args.model`) gets its checkpoint hashed here. The drafter’s path does appear in `vllm_command.txt`, but its binary identity isn’t captured in `checkpoint_sha256.txt`.

Is this intentional? If a local drafter checkpoint changes between runs, the current provenance wouldn’t detect it. A follow-up or quick addition could pass `spec_model` to a second `_save_checkpoint_sha256` call (e.g. writing to `drafter_checkpoint_sha256.txt`).

scripts/launch_vllm.py
```diff
@@ -56,75 +56,124 @@ def build_kv_transfer_config(args: argparse.Namespace) -> dict:
...
+        type=str,
+        default=None,
+        help="Speculative decoding method (optional)",
+    )
+    _add_shared_args(eval_parser)
+
+    argv = sys.argv[1:]
+    if not argv or (argv[0] not in _SUBCOMMANDS and not argv[0].startswith("-")):
+        argv = ["train", *argv]
+    return parser.parse_known_args(argv)
```
speculatorsbot: The guard handles `launch_vllm.py MODEL` correctly, but misses the case where flags precede the model without an explicit subcommand — e.g. `launch_vllm.py --dry-run gpt2` or `launch_vllm.py --provenance-dir /foo gpt2`.

Because `argv[0]` starts with `-`, the guard skips the transform. `parse_known_args` then runs on the top-level parser with no subcommand matched, so `args.subcommand` is `None` and `args` has no `model`/`dry_run`/etc. attributes. In `main()`, the `else` branch calls `_build_eval_cmd(args, ...)`, which crashes with `AttributeError: 'Namespace' object has no attribute 'model'`.

In the base-branch code (pre-subcommands), `launch_vllm.py --dry-run gpt2` worked because all args lived on a single parser.

A fallback after parsing fixes this without changing the `-h` behavior (argparse exits on `-h` during the first parse, so the fallback is never reached for help):

```suggestion
    args, vllm_args = parser.parse_known_args(argv)
    if args.subcommand is None:
        args, vllm_args = parser.parse_known_args(["train"] + sys.argv[1:])
    return args, vllm_args
```

speculatorsbot (commented):
The empty-argv fix from the previous review looks good (commit 65bc692). One remaining backward-compat edge case below.

scripts/launch_vllm.py
```diff
@@ -321,13 +369,43 @@ def main():
...
+
+
+def main():
+    args, vllm_args = parse_args()
+    if "--" in vllm_args:
+        vllm_args.remove("--")
+
+    if args.subcommand == "train":
+        cmd = _build_train_cmd(args, vllm_args)
+    else:
```
rahul-tuli: nit: Should we make it an `elif` branch instead of a blanket else?

scripts/launch_vllm.py
rahul-tuli: Should `_save_checkpoint_sha256` also record hashes for `args.spec_model` in eval mode, perhaps as a separate `drafter_sha256.txt` artifact or appended to the existing file?

scripts/launch_vllm.py
```diff
@@ -56,75 +56,127 @@ def build_kv_transfer_config(args: argparse.Namespace) -> dict:
...
+    eval_parser.add_argument(
+        "--spec-method",
+        type=str,
+        default=None,
+        help="Speculative decoding method (optional)",
+    )
+    _add_shared_args(eval_parser)
+
+    argv = sys.argv[1:]
+    if not argv or (argv[0] not in _SUBCOMMANDS and not argv[0].startswith("-")):
```
rahul-tuli: Is there a way to derive `_SUBCOMMANDS` from the parser (e.g. `sub.choices.keys()` after registering subparsers) rather than maintaining it as a separate constant?

rahul-tuli (commented):
LGTM pending nits/suggestions!

scripts/launch_vllm.py
orestis-z: good point

scripts/launch_vllm.py
orestis-z: good catch

scripts/launch_vllm.py
orestis-z: good point, missed that. added

rahul-tuli, line 545: Could we gate these as well? I believe we don't need them for eval serving

rahul-tuli (approved):
LGTM, pending a comment

PR #956, opened 2026-08-07
feat(train): write speculators.patch alongside train_command.txt

orestis-z:
## Summary

- `save_train_command()` now also writes `speculators.patch` (uncommitted diff via `git diff HEAD`) into the checkpoint directory
- Refactored repo-root discovery and git SHA into reusable helpers (`_find_repo_root`, `_git_sha`, `_atomic_write`)
- Best-effort — patch failures log a warning and never block training
- Updated `AGENTS.md` to document `speculators.patch` in training provenance

Depends on #954.

## Test plan

- [x] All 21 `test_save_train_command.py` tests pass (12 existing + 5 new patch tests + 4 checkpointer)
- [x] `test_creates_patch_file` — verifies `speculators.patch` is written
- [x] `test_patch_contains_repo_header` — verifies `# repo:` header
- [x] `test_patch_contains_sha` — verifies SHA in header
- [x] `test_no_patch_when_no_repo` — verifies no file when repo not found
- [x] `test_patch_failure_does_not_block` — verifies `train_command.txt` still written when git diff fails
- [x] Lint clean (`make quality`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #960, opened 2026-08-07
feat(train): write speculators.patch alongside train_command.txt

orestis-z:
## Summary
- Add `speculators.patch` capture to training provenance — runs `git diff HEAD` in the speculators repo at launch
- Refactor `save_train_command()` extracting `_find_repo_root()`, `_git_sha()`, `_atomic_write()`
- Patch header includes repo path and SHA for traceability
- Failures propagate (no silent swallowing) — consistent with existing training provenance behavior

Replaces #956. Part 4 of the RFC #880 provenance stack. Depends on #959.

## Test plan
- [x] `TestSpeculatorsPatch` — 5 tests covering file creation, header content, SHA presence, no-repo skip, and error propagation
- [x] All existing `test_save_train_command.py` tests pass
- [x] `make quality` passes

### DIFF
--- a/src/speculators/train/utils.py
+++ b/src/speculators/train/utils.py
@@ -129,16 +129,19 @@ def _atomic_write(dest: Path, content: str) -> None:
 def _save_speculators_patch(save_dir: Path, repo_root: Path | None, sha: str) -> None:
     if repo_root is None:
         return
-    diff = subprocess.run(
-        ["git", "diff", "HEAD"],  # noqa: S607
-        capture_output=True,
-        text=True,
-        cwd=repo_root,
-        timeout=30,
-        check=False,
-    ).stdout.strip()
-    content = f"# repo: {repo_root} ({sha})\n{diff}"
-    _atomic_write(save_dir / "speculators.patch", content)
+    try:
+        diff = subprocess.run(
+            ["git", "diff", "HEAD"],  # noqa: S607
+            capture_output=True,
+            text=True,
+            cwd=repo_root,
+            timeout=30,
+            check=False,
+        ).stdout.strip()
+        content = f"# repo: {repo_root} ({sha})\n{diff}"
+        _atomic_write(save_dir / "speculators.patch", content)
+    except (OSError, subprocess.TimeoutExpired):
+        logger.warning("Failed to save speculators.patch", exc_info=True)
 
 
 def save_train_command(save_path: str, argv: list[str] | None = None) -> None:
--- a/src/speculators/train/utils.py
+++ b/src/speculators/train/utils.py
@@ -101,16 +101,9 @@ def normalize_counted_metrics(
     return metrics
 
 
-def save_train_command(save_path: str, argv: list[str] | None = None) -> None:
-    """Write the launch command and provenance header to save_path/train_command.txt.
-
-    ``argv`` is the exact command the run was resolved from (``TrainConfig`` records
-    it during resolution); it falls back to the live ``sys.argv`` when a caller has
-    no recorded argv, so a direct call is unchanged.
-    """
-    repo_root = find_repo_root(Path(__file__))
+def _git_sha(repo_root: Path | None) -> str:
     try:
-        sha = subprocess.run(
+        return subprocess.run(
             ["git", "rev-parse", "HEAD"],  # noqa: S607
             capture_output=True,
             text=True,
@@ -118,7 +111,45 @@ def save_train_command(save_path: str, argv: list[str] | None = None) -> None:
             check=True,
         ).stdout.strip()
     except (OSError, subprocess.CalledProcessError):
-        sha = "unknown"
+        return "unknown"
+
+
+def _atomic_write(dest: Path, content: str) -> None:
+    fd, tmp = tempfile.mkstemp(dir=dest.parent, prefix=f".{dest.name}_", suffix=".tmp")
+    tmp_path = Path(tmp)
+    try:
+        with os.fdopen(fd, "w") as f:
+            f.write(content)
+        tmp_path.replace(dest)
+    finally:
+        if tmp_path.exists():
+            tmp_path.unlink()
+
+
+def _save_speculators_patch(save_dir: Path, repo_root: Path | None, sha: str) -> None:
+    if repo_root is None:
+        return
+    diff = subprocess.run(
+        ["git", "diff", "HEAD"],  # noqa: S607
+        capture_output=True,
+        text=True,
+        cwd=repo_root,
+        timeout=30,
+        check=False,
+    ).stdout.strip()
+    content = f"# repo: {repo_root} ({sha})\n{diff}"
+    _atomic_write(save_dir / "speculators.patch", content)
+
+
+def save_train_command(save_path: str, argv: list[str] | None = None) -> None:
+    """Write train_command.txt and speculators.patch to *save_path*.
+
+    ``argv`` is the exact command the run was resolved from (``TrainConfig`` records
+    it during resolution); it falls back to the live ``sys.argv`` when a caller has
+    no recorded argv, so a direct call is unchanged.
+    """
+    repo_root = find_repo_root(Path(__file__))
+    sha = _git_sha(repo_root)
 
     pkg_versions: list[str] = []
     for pkg in ("speculators", "vllm", "transformers", "torch", "compressed-tensors"):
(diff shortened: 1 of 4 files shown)


speculatorsbot (commented):
Clean refactoring — extracting `_find_repo_root`, `_git_sha`, and `_atomic_write` into reusable helpers is a good call, and anchoring git commands to the repo root via `cwd=repo_root` addresses the RFC #880 feedback from @rahul-tuli. One concern about error handling semantics vs the RFC spec.

src/speculators/train/utils.py
```diff
@@ -100,33 +100,67 @@ def normalize_counted_metrics(
...
+    diff = subprocess.run(
+        ["git", "diff", "HEAD"],  # noqa: S607
+        capture_output=True,
+        text=True,
+        cwd=repo_root,
+        timeout=30,
+        check=False,
+    ).stdout.strip()
+    content = f"# repo: {repo_root} ({sha})\n{diff}"
+    _atomic_write(save_dir / "speculators.patch", content)
```
speculatorsbot, line 152: `_save_speculators_patch` does not handle `OSError` (git not found / not in PATH) or `subprocess.TimeoutExpired` (git hangs for > 30 seconds). These exceptions will propagate through `save_train_command` → `artifacts.save` → `cfg.save` and crash training.

This contradicts the [RFC #880 implementation note](https://github.com/vllm-project/speculators/issues/880):

> All artifacts are best-effort (wrapped in try/except) — failure to capture provenance never blocks training or evaluation.

Within the same RFC stack, `_save_vllm_provenance` in [`launch_vllm.py`](https://github.com/vllm-project/speculators/blob/ec356c2e46ac0e23b8ad6903230ae42184510d63/scripts/launch_vllm.py#L381-L385) wraps each writer in `try/except Exception` explicitly for this reason. And within this same file, `_git_sha` follows best-effort by catching `(OSError, CalledProcessError)` and returning `"unknown"`.

The replaced PR #956 also described best-effort behavior ("patch failures log a warning and never block training"). The PR description says this change is intentional ("Failures propagate — consistent with existing training provenance behavior"), but I think the consistency argument is weaker than it appears: existing `_atomic_write` failures for `train_command.txt` typically signal disk/permission issues that would also break checkpoint saving, whereas a missing `git` binary or a hanging `git diff` on CephFS is an unrelated issue that shouldn't take down a multi-GPU training job.

Would it make sense to match the `launch_vllm.py` pattern here — wrap the `subprocess.run` + `_atomic_write` in a `try/except` and warn on failure? Or if propagation is the intended contract, should the RFC be updated to reflect that?

tests/unit/train/test_save_train_command.py
```diff
@@ -87,6 +90,54 @@ def test_overwrites_existing(self, tmp_path: Path):
...
+            return original_run(*args, **kwargs)
+
+        with (
+            patch(
+                "speculators.train.utils.subprocess.run",
+                side_effect=_fail_on_diff,
+            ),
+            pytest.raises(OSError, match="git broke"),
+        ):
+            save_train_command(str(tmp_path))
```
speculatorsbot: `test_patch_failure_raises` expects `OSError("git broke")` to propagate out of `save_train_command` (line 136: `pytest.raises(OSError, match="git broke")`), but `_save_speculators_patch` now catches `(OSError, subprocess.TimeoutExpired)` at [`utils.py:154-155`](https://github.com/vllm-project/speculators/blob/95f5ced903d487749d61440cf91574d61c391ed0/src/speculators/train/utils.py#L154-L155) and logs a warning. The test will fail because `save_train_command` returns normally — `pytest.raises` won't see any exception.

Suggestion: rename to something like `test_patch_failure_does_not_block` and assert the new best-effort contract instead:

```python
def test_patch_failure_does_not_block(self, tmp_path: Path):
    original_run = subprocess.run

    def _fail_on_diff(*args, **kwargs):
        if "diff" in args[0]:
            raise OSError("git broke")
        return original_run(*args, **kwargs)

    with patch(
        "speculators.train.utils.subprocess.run",
        side_effect=_fail_on_diff,
    ):
        save_train_command(str(tmp_path))  # should not raise
    # train_command.txt is still written
    assert (tmp_path / "train_command.txt").exists()
    # patch file is not created on failure
    assert not (tmp_path / "speculators.patch").exists()
```

speculatorsbot (commented):
The error-handling fix in `95f5ced` addresses the concern from the previous review — `_save_speculators_patch` now catches `(OSError, subprocess.TimeoutExpired)` and logs a warning, aligning with the RFC #880 best-effort requirement. However, the test for this behavior still asserts the old contract (propagation), so it will fail.

Minor note: the PR description still says "Failures propagate (no silent swallowing)" which no longer matches the code after the fix commit. Worth updating for clarity.

src/speculators/train/utils.py
```diff
@@ -100,33 +100,70 @@ def normalize_counted_metrics(
...
+        diff = subprocess.run(
+            ["git", "diff", "HEAD"],  # noqa: S607
+            capture_output=True,
+            text=True,
+            cwd=repo_root,
+            timeout=30,
+            check=False,
+        ).stdout.strip()
+        content = f"# repo: {repo_root} ({sha})\n{diff}"
+        _atomic_write(save_dir / "speculators.patch", content)
```
rahul-tuli: Should `_save_speculators_patch` skip writing the file when the diff is empty? That way the presence of `speculators.patch` would be a clear signal that uncommitted changes existed at training time?

rahul-tuli (approved):
LGTM!

src/speculators/train/utils.py
orestis-z: We have vllm.patch that writing also when there's no diff. I'd lean toward keeping both consistent. Always writing is simpler and lets you confirm the artifact was generated (vs. something failing silently)

src/speculators/train/utils.py
```diff
@@ -101,24 +101,58 @@ def normalize_counted_metrics(
...
 
-def save_train_command(save_path: str, argv: list[str] | None = None) -> None:
-    """Write the launch command and provenance header to save_path/train_command.txt.
-
-    ``argv`` is the exact command the run was resolved from (``TrainConfig`` records
-    it during resolution); it falls back to the live ``sys.argv`` when a caller has
-    no recorded argv, so a direct call is unchanged.
-    """
-    repo_root = find_repo_root(Path(__file__))
+def _git_sha(repo_root: Path | None) -> str:
```
rahul-tuli: nit: Flagging this as an edge case, but if `repo_root` is `None`, subprocess will inherit training process's CWD, if the training command is launched from any other git repo, `git rev-parse HEAD` succeeds and records the
 **wrong repo's SHA** as the speculators SHA in `train_command.txt`

Should we just do `if repo_root is None: return "unknown"`

rahul-tuli: Also, don't we have the same function in `provenance.py`?

src/speculators/train/utils.py
```diff
@@ -101,24 +101,58 @@ def normalize_counted_metrics(
...
             text=True,
             cwd=repo_root,
             check=True,
         ).stdout.strip()
     except (OSError, subprocess.CalledProcessError):
-        sha = "unknown"
+        return "unknown"
+
+
+def _atomic_write(dest: Path, content: str) -> None:
```
rahul-tuli: Why not use `provenance.atomic_write`? seems like a 1-to-1 copy

src/speculators/train/utils.py
```diff
@@ -101,24 +101,58 @@ def normalize_counted_metrics(
...
+    finally:
+        if tmp_path.exists():
+            tmp_path.unlink()
+
+
+def _save_speculators_patch(save_dir: Path, repo_root: Path | None, sha: str) -> None:
+    if repo_root is None:
+        return
+    try:
+        diff = subprocess.run(
```
rahul-tuli: nit: Should we use `provenance.git_diff`?

rahul-tuli (changes requested):
Overall looks good, except that some functions are duplicated?

src/speculators/train/utils.py
```diff
@@ -101,24 +101,58 @@ def normalize_counted_metrics(
...
+            ["git", "diff", "HEAD"],  # noqa: S607
+            capture_output=True,
+            text=True,
+            cwd=repo_root,
+            timeout=30,
+            check=False,
+        ).stdout.strip()
+        content = f"# repo: {repo_root} ({sha})\n{diff}"
+        _atomic_write(save_dir / "speculators.patch", content)
```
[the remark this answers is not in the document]
orestis-z: The patch is written to the local checkpoint/output directory only — it is never uploaded automatically. This mirrors the existing `vllm.patch` behavior, and always writing (rather than opt-in) keeps the artifact set predictable and lets you confirm provenance was captured. Redacting a diff generically is unreliable, so we treat it as the user's responsibility not to publish a diff that contains secrets when sharing a run. Resolving as working-as-intended.

rahul-tuli (approved):
LGTM!

PR #962, opened 2026-08-08
Omit verifier-owned weights from saved DFlash/DSpark checkpoints

qianlihuang:
## Purpose

Reduce DFlash/DSpark checkpoint size by omitting verifier-owned weights that can be reconstructed on load.

This PR:

- omits `embed_tokens.weight`;
- omits `lm_head.weight` for full-vocabulary drafts;
- keeps reduced-vocabulary `lm_head.weight` and `t2d`/`d2t`;
- reconstructs verifier-owned weights from the verifier on load; and
- preserves compatibility with existing self-contained checkpoints.

The ownership policy is applied through `save_pretrained`.

## Tests

```
pytest -q tests/unit/models/test_checkpoint_key_ownership.py
```
Full unit suite: 664 passed.

- Full-vocabulary DSpark slim checkpoint E2E validated with vLLM 0.26.1rc1.dev403.
- Real Qwen3.6-27B-NVFP4 verifier target with TP2 serving.
- Slim checkpoint reduced from 8.2 GB to 3.5 GB.
- Removed verifier-owned tensors:
  embed_tokens.weight,
  lm_head.weight,
  verifier_lm_head.weight,
  verifier_norm.weight.
- Greedy AR baseline and DSpark speculative decoding produced identical token IDs for 64 generated tokens.

## Checklist

I have filled in:


src/speculators/model.py
```diff
@@ -154,9 +154,13 @@ def load_verifier_weights(self):  # noqa: C901
...
-        # Load embed_tokens if not already loaded (NaN means uninitialized)
-        if self.embed_tokens.weight.isnan().any():
-            self.embed_tokens.load_state_dict({"weight": embed_tokens_weight})
+        # embed_tokens and lm_head are frozen verifier-derived parameters, so
+        # they are always overwritten from the verifier. The previous NaN guard
+        # is unreliable: transformers re-materializes parameters absent from a
+        # checkpoint with uninitialized storage (see
+        # _move_missing_keys_from_meta_to_device), so NaN no longer means
+        # "uninitialized" and a slim checkpoint would load garbage.
+        self.embed_tokens.load_state_dict({"weight": embed_tokens_weight})
```
speculatorsbot: nit (not on this line): The simplification in this PR (removing two `if` branches for NaN guards) drops the cyclomatic complexity of `load_verifier_weights` below the C901 threshold, so the `# noqa: C901` annotation on the function definition (line 123) is now unused. This triggers a `RUF100` lint warning:

```
RUF100 [*] Unused `noqa` directive (unused: `C901`)
   --> src/speculators/model.py:123:39
```

Suggested fix - change line 123 from:
```python
def load_verifier_weights(self):  # noqa: C901
```
to:
```python
def load_verifier_weights(self):
```

speculatorsbot (commented):
## Review Summary

This PR reduces DFlash/DSpark checkpoint sizes by omitting verifier-owned weights (`embed_tokens.weight`, full-vocab `lm_head.weight`, `verifier_lm_head.weight`, `verifier_norm.weight`) that can be deterministically reconstructed from the verifier at load time. The design is sound and the implementation is clean.

**What the PR does well:**
- The rationale for removing the NaN guard is correct: transformers' `_move_missing_keys_from_meta_to_device` re-materializes checkpoint-absent parameters with uninitialized (not necessarily NaN) storage, so the old guard was unreliable for slim checkpoints.
- Per-instance copying of `_keys_to_ignore_on_save` / `_keys_to_ignore_on_load_missing` before appending is the right approach to prevent a full-vocab model from leaking `lm_head` omission rules into a reduced-vocab sibling created later in the same process.
- The ownership policy correctly distinguishes between full-vocab `lm_head` (frozen verifier projection, omittable) and reduced-vocab `lm_head` (runtime-required, must be serialized with `t2d`/`d2t`).
- DSpark inherits correctly: it has no `_keys_to_ignore_*` overrides and its `__init__` calls `super().__init__`, so the per-instance setup in `DFlashDraftModel.__init__` applies to DSpark too.
- Tests cover the three important scenarios (save-key ownership, slim roundtrip, legacy self-contained checkpoint compatibility), and the reduced-vocab roundtrip correctly loads `t2d`/`d2t` from the checkpoint rather than passing them in.
- Both prior reviewer concerns (CodeRabbit re: passing t2d/d2t explicitly, Copilot re: asserting verifier_lm_head/verifier_norm reconstruction) have been addressed in this revision.

I left two minor comments below. Overall the change looks correct and well-tested.

tests/unit/models/test_checkpoint_key_ownership.py
```diff
@@ -0,0 +1,213 @@
...
+    assert "lm_head.weight" in keys
+
+    fake = _fake_verifier()
+    monkeypatch.setattr(
+        "speculators.utils.loading.load_model_layers",
+        _make_fake_loader(fake),
+    )
+    loaded = DFlashDraftModel.from_pretrained(tmp_path, local_files_only=True)
+    assert torch.equal(loaded.embed_tokens.weight, fake["embed_tokens.weight"])
+    assert torch.equal(loaded.lm_head.weight, fake["lm_head.weight"])
```
speculatorsbot: nit: The full-vocab slim roundtrip test (`test_full_vocab_slim_roundtrip_reconstructs_from_verifier`) asserts reconstruction of all four verifier-owned tensors including `verifier_lm_head.weight` and `verifier_norm.weight`. This legacy-compat test only checks `embed_tokens` and `lm_head`. Consider adding the same two assertions here for completeness:

```python
assert torch.equal(loaded.verifier_lm_head.weight, fake["lm_head.weight"])
assert torch.equal(loaded.verifier_norm.weight, fake["model.norm.weight"])
```

This would ensure that when a legacy checkpoint supplies all keys, the verifier source-of-truth policy still overrides them all.

src/speculators/model.py
```diff
@@ -154,9 +154,13 @@ def load_verifier_weights(self):  # noqa: C901
         embed_tokens_weight = verifier_weights["embed_tokens.weight"]
         lm_head_weight = verifier_weights.get("lm_head.weight", embed_tokens_weight)
 
-        # Load embed_tokens if not already loaded (NaN means uninitialized)
-        if self.embed_tokens.weight.isnan().any():
-            self.embed_tokens.load_state_dict({"weight": embed_tokens_weight})
+        # embed_tokens and lm_head are frozen verifier-derived parameters, so
```
fynnsu: Currently P-Eagle and Eagle-3 allow for training the embed tokens (and P-Eagle enables this by default). This change will result in those pretrained embed tokens getting overriden with the verifier weights.

fynnsu (changes requested):
Thanks for the pr! I think this makes sense to reduce our checkpoint size, but there's a small correctness issue for P-Eagle/Eagle-3.

I think it's important that if there is a weight in the checkpoint, that it gets used, and otherwise we re-initialize from the verifier.

shanjiaz (approved):
Thanks for addressing the comment!

shanjiaz (dismissed):
@qianlihuang Could you resolve quality, looks like some typechecks are failing now. Thanks!

tests/unit/models/test_checkpoint_key_ownership.py
```diff
@@ -0,0 +1,166 @@
...
+    assert torch.equal(loaded.verifier_lm_head.weight, fake["lm_head.weight"])
+    assert torch.equal(loaded.verifier_norm.weight, fake["model.norm.weight"])
```
speculatorsbot, line 139: The previous revision had a `test_reduced_vocab_slim_roundtrip_reconstructs_head_through_t2d` test that was removed in this commit. This leaves a coverage gap for the reduced-vocab DFlash roundtrip, which exercises a distinct code path from the full-vocab test above:

1. `overwrite_lm_head=False` with a checkpoint-loaded (non-NaN) reduced lm_head -- the NaN guard must correctly *preserve* the checkpoint weight rather than overwrite it.
2. `t2d`/`d2t` loaded from the checkpoint itself (not passed explicitly).
3. The reduced `lm_head` shape `(draft_vocab_size, hidden_size)` round-trips correctly.
4. `verifier_lm_head` gets the t2d-masked verifier head while `lm_head` stays as-is from the checkpoint.

The full-vocab roundtrip test here only exercises `overwrite_lm_head=True` (both missing from checkpoint), and `test_checkpoint_weights_take_precedence` only exercises the bare `DraftVocabMixin` base class (not DFlash's override with `overwrite_embed_tokens=True, overwrite_lm_head=False`). Neither covers this combination.

Consider restoring a reduced-vocab roundtrip test -- it guards against regressions in the more complex vocab-masking path.

speculatorsbot (commented):
## Review Summary

The new commit cleanly addresses fynnsu's concern about P-Eagle/Eagle-3 trained `embed_tokens` being overwritten. The keyword-only `overwrite_embed_tokens` / `overwrite_lm_head` flags on the base class are a good design: the default preserves checkpoint-loaded weights (NaN guard), while DFlash/DSpark explicitly opt in to verifier overwrite for their frozen parameters. Eagle3/PEagle inherit the base class behavior and correctly preserve their trained weights.

`test_checkpoint_weights_take_precedence` is a solid addition that validates the base-class NaN-guard contract.

One test coverage gap flagged below.

qianlihuang:
@shanjiaz @fynnsu I pushed a fix for the quality issues. It looks like the new workflow run is awaiting approval - could you approve it when you get a chance? Thanks!

fynnsu (approved):
LGTM! Thanks for making the fix

PR #963, opened 2026-08-10
trust remote code

fynnsu:
## Purpose

Currently `--trust-remote-code` only applies to the tokenizer setup. This pr extends it to also be used when initializing configs with `AutoConfig` which is required by some models (e.g. Kimi K3)

## Tests

Confirmed this works for Kimi K3

## Checklist

I have filled in:

---

### DIFF
--- a/src/speculators/models/utils.py
+++ b/src/speculators/models/utils.py
@@ -17,8 +17,14 @@ def conditional_torch_compile(func=None, *args, **kwargs):
     return func
 
 
-def get_verifier_config(verifier_name_or_path: str) -> PretrainedConfig:
-    verifier_config = AutoConfig.from_pretrained(verifier_name_or_path)
+def get_verifier_config(
+    verifier_name_or_path: str,
+    trust_remote_code: bool = False,
+) -> PretrainedConfig:
+    verifier_config = AutoConfig.from_pretrained(
+        verifier_name_or_path,
+        trust_remote_code=trust_remote_code,
+    )
     if hasattr(verifier_config, "text_config"):
         verifier_config = verifier_config.text_config
     return verifier_config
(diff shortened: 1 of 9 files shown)


src/speculators/models/utils.py
```diff
@@ -30,11 +36,15 @@ def get_verifier_config(verifier_name_or_path: str) -> PretrainedConfig:
 def resolve_target_layer_ids(
     target_layer_ids: list[int] | None,
     verifier_name_or_path: str,
+    trust_remote_code: bool = False,
```
speculatorsbot, line 39: Three `from_training_args` callers of this function don't forward `trust_remote_code` from kwargs, so training without explicit `--target-layer-ids` will still fail for remote-code models:

- [`eagle3/core.py:386-387`](https://github.com/vllm-project/speculators/blob/2c631cb034b88900bd1ada109abf09d0bece8817/src/speculators/models/eagle3/core.py#L385-L388)
- [`dflash/core.py:186-187`](https://github.com/vllm-project/speculators/blob/2c631cb034b88900bd1ada109abf09d0bece8817/src/speculators/models/dflash/core.py#L185-L188)
- [`peagle/core.py:233-234`](https://github.com/vllm-project/speculators/blob/2c631cb034b88900bd1ada109abf09d0bece8817/src/speculators/models/peagle/core.py#L232-L235)

Each receives `trust_remote_code` via `**vars(args)` in `kwargs` but doesn't pass it through. Suggested fix (same for all three):

```python
target_layer_ids = resolve_target_layer_ids(
    kwargs.get("target_layer_ids"),
    kwargs["verifier_name_or_path"],
    trust_remote_code=kwargs.get("trust_remote_code", False),
)
```

speculatorsbot (commented):
Clean, well-scoped bug fix that threads `trust_remote_code` through the config-loading helpers (`get_verifier_config`, `resolve_target_layer_ids`, `get_target_vocab_size`) and their call sites in `scripts/train.py` and `scripts/launch_vllm.py`. The approach is correct.

One gap: `resolve_target_layer_ids` now accepts `trust_remote_code`, but three callers inside model `from_training_args` methods don't forward it -- see inline comment.

A related item for follow-up: [`Eagle3DraftModel.load_verifier_weights()`](https://github.com/vllm-project/speculators/blob/2c631cb034b88900bd1ada109abf09d0bece8817/src/speculators/models/eagle3/core.py#L169-L185) also calls `AutoConfig.from_pretrained` without `trust_remote_code` (line 175). This is harder to fix -- the persisted `VerifierConfig` doesn't carry the flag -- but it means training eagle3/peagle with a remote-code model like Kimi K3 would still fail at model construction time even with this PR. Consider adding `trust_remote_code` as a parameter to `load_verifier_weights()` and threading it from the caller.

src/speculators/models/utils.py
fynnsu: Fixed

shanjiaz (approved):

orestis-z (approved):

PR #964, opened 2026-08-10
feat(data): accept pre-tokenized datasets without a chat template

fynnsu:
## Purpose

Currently we always require a chat template (or for one to be on the tokenizer). However, if response regen was used to generate the dataset, we already have fully tokenized sequences and we can skip initializing the tokenizer. 

## Tests

Validated this works on a response regenerated dataset. 

## Checklist

I have filled in:

---


speculatorsbot (commented):
LGTM. Clean, well-scoped change.

**Summary**: Previously, `load_and_preprocess_dataset` unconditionally required the processor to have a chat template, which prevented using pre-tokenized datasets (e.g., from response regeneration). This PR defers the chat-template check until after loading the dataset: if the data already contains `input_ids` and `loss_mask` columns, no chat template is needed.

**What I checked**:

1. **Downstream safety**: `build_eagle3_dataset` (line 652) and `_preprocess_batch` (line 545) already detect pre-tokenized data via the same `{"input_ids", "loss_mask"} <= set(columns)` check and route to `_passthrough_pretokenized`, which never touches the chat template. So the processor is never asked for a chat template on the pre-tokenized path -- this PR aligns the upstream validation with the existing downstream behavior.

2. **`_visualize_sample`**: This calls `processor.decode()`, not `apply_chat_template`, so it works fine with a template-less processor.

3. **Mixed dataset paths**: If `train_data_paths` contains both pre-tokenized and conversational datasets, each is validated independently inside the loop. A conversational dataset will still correctly fail if the processor lacks a chat template.

4. **Robustness improvement**: The `getattr(processor, "chat_template", None) is not None` pattern is safer than the original `processor.chat_template is None`, which could raise `AttributeError` if the attribute is absent but `apply_chat_template` exists.

5. **Tests**: Two new tests cover both directions -- pre-tokenized data accepted without a chat template, and conversational data still rejected without one. The error message update helpfully mentions the pre-tokenized alternative.

src/speculators/data_generation/preprocessing.py
```diff
@@ -776,6 +776,11 @@ def load_and_preprocess_dataset(
     log.subsection("Loading processor")
     processor = load_processor(target_model_path, trust_remote_code=trust_remote_code)
 
+    processor_has_chat_template = (
+        hasattr(processor, "apply_chat_template")
+        and getattr(processor, "chat_template", None) is not None
```
orestis-z, line 781: if a processor has the `apply_chat_template` attribute, doesn't it have to have the `chat_template` too? Or are we checking both to be extra careful?

orestis-z (approved):

src/speculators/data_generation/preprocessing.py
fynnsu: So the first check for `apply_chat_template` is mostly verifying that we have a valid tokenizer instance since this method should be defined on the base class. The `chat_template is not None` actually checks if it's available.

PR #965, opened 2026-08-10
feat(train): log the resolved training config to the metric backend

fynnsu:
## Purpose

After parsing the training config, this pr adds support for publishing it to the metric backend (if the backend supports config values). 

## Tests

Validated on the W&B backend. 

## Checklist

I have filled in:

---

### DIFF
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -41,6 +41,7 @@ from speculators.train.vocab_mapping import (
 from speculators.utils.loading import is_config_only_dir
 
 logger = logging.getLogger(__name__)
+metric_logger = logging.getLogger("speculators.metrics")
 
 DRAFT_ARCH_CONFIGS: dict[str, type] = {
     "llama": LlamaConfig,
@@ -546,6 +547,12 @@ def main(cfg: TrainConfig):  # noqa: C901
 
     # Setup distributed training
     maybe_setup_distributed()
+
+    # Publish train config to metric backends that support hyperparameter logging
+    resolved_config = cfg.model_dump(mode="json")
+    if cfg.backend_args:
+        resolved_config["hidden_states_backend_args"] = dict(cfg.backend_args)
+    metric_logger.info(resolved_config, extra={"hparams": True})
 
     if args.fsdp_shard and not is_distributed():
         raise ValueError(
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -31,7 +31,11 @@ from speculators.train.distributed import (
     maybe_destroy_distributed,
     maybe_setup_distributed,
 )
-from speculators.train.logger import setup_metric_logger, setup_root_logger
+from speculators.train.logger import (
+    log_run_config,
+    setup_metric_logger,
+    setup_root_logger,
+)
 from speculators.train.trainer import Trainer, TrainerConfig
 from speculators.train.utils import resolve_mask_token_id
 from speculators.train.vocab_mapping import (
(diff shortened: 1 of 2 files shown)


scripts/train.py
```diff
@@ -547,6 +548,12 @@ def main(cfg: TrainConfig):  # noqa: C901
     # Setup distributed training
     maybe_setup_distributed()
 
+    # Publish train config to metric backends that support hyperparameter logging
+    resolved_config = cfg.model_dump(mode="json")
+    if cfg.backend_args:
+        resolved_config["hidden_states_backend_args"] = dict(cfg.backend_args)
+    metric_logger.info(resolved_config, extra={"hparams": True})
```
speculatorsbot, line 555: `model_dump(mode="json")` serializes list-typed fields as Python lists -- specifically `full_attention_indices` (always a `list[int]`, defaults to `[]`) and `target_layer_ids` (when not `None`). After `_flatten_dict`, these reach TensorBoard's `hparams()` which only accepts `int | float | str | bool | torch.Tensor` and raises `ValueError` for anything else ([source: `torch/utils/tensorboard/summary.py:322-323`](https://github.com/pytorch/pytorch/blob/v2.11.0/torch/utils/tensorboard/summary.py#L322-L323)).

Python's logging framework catches the exception via `Handler.handleError`, so training continues and W&B / Trackio work fine -- but hparams silently fail to appear in TensorBoard.

One fix: stringify non-scalar values before emitting:

```python
import json

resolved_config = cfg.model_dump(mode="json")
if cfg.backend_args:
    resolved_config["hidden_states_backend_args"] = dict(cfg.backend_args)

# TensorBoard add_hparams only accepts scalar types;
# coerce lists so all backends receive the config.
def _scalarise(d: dict) -> dict:
    return {
        k: (json.dumps(v) if isinstance(v, (list, dict)) else v)
        for k, v in d.items()
    }

from speculators.train.logger import _flatten_dict
metric_logger.info(
    _scalarise(_flatten_dict(resolved_config)),
    extra={"hparams": True},
)
```

Alternatively this could be handled inside `TensorBoardHandler.emit` so future callers don't have to worry about it.

speculatorsbot (commented):
Clean, well-scoped change that fills a real gap -- metric backends had no record of the resolved config, so runs couldn't be compared or reproduced from the dashboard.

One compatibility concern with the TensorBoard backend below.

scripts/train.py
fynnsu: Fixed in TensorBoardHandler

src/speculators/train/logger.py
```diff
@@ -313,6 +313,25 @@ def _setup(self):
...
+        receive the full config.
+
+        Args:
+            d: Dictionary potentially containing non-scalar values
+
+        Returns:
+            Dictionary with all non-scalar values converted to JSON strings
+        """
+        return {
+            k: (json.dumps(v) if isinstance(v, (list, dict)) else v)
```
orestis-z: i think inverting this would ensure a catch-all:

`k: (json.dumps(v) if type(v) not in TB_SUPPORTED_TYPES)`

scripts/train.py
orestis-z: minor:  instead of commenting separate this into a helper function with a descriptive name

orestis-z (approved):

rahul-tuli (approved):
LGTM!

PR #966, opened 2026-08-10
feat(data-generation): harden hidden-state payload validation

fynnsu:
## Purpose

Add NaN/finite checks for hidden states loaded by the training process. 

## Tests

Added a unit test, and validated it works in a training scenario. 

## Checklist

I have filled in:

---

### DIFF
--- a/src/speculators/data_generation/offline.py
+++ b/src/speculators/data_generation/offline.py
@@ -10,7 +10,7 @@ def check_hidden_states(data: dict, tokens: list[int]):
     required = {"token_ids", "hidden_states"}
     missing = required - data.keys()
     if missing:
-        raise ValueError(f"Hidden-state payload is missing keys: {sorted(missing)}")
+        raise ValueError(f"Hidden-state payload is missing keys: {missing}")
 
     t_ids = data["token_ids"].tolist()
     if t_ids != tokens:
--- a/src/speculators/data_generation/offline.py
+++ b/src/speculators/data_generation/offline.py
@@ -1,22 +1,56 @@
 import logging
 from pathlib import Path
 
+import torch
+
 logger = logging.getLogger(__name__)
 
 
 def check_hidden_states(data: dict, tokens: list[int]):
+    required = {"token_ids", "hidden_states"}
+    missing = required - data.keys()
+    if missing:
+        raise ValueError(f"Hidden-state payload is missing keys: {sorted(missing)}")
+
     t_ids = data["token_ids"].tolist()
     if t_ids != tokens:
         raise ValueError(f"Token ids don't match expected token ids {tokens}")
 
     hs = data["hidden_states"]
-    if hs.isnan().any():
-        raise ValueError("Hidden states contain NaN values")
+    if not isinstance(hs, torch.Tensor):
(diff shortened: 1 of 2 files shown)


src/speculators/data_generation/offline.py
```diff
@@ -1,23 +1,57 @@
...
+        # Process hidden states in chunks to avoid OOMs
+        chunk = hs[start : start + rows_per_chunk]
+        finite = torch.isfinite(chunk)
+        if finite.all():
+            continue
+        nan_count += int(torch.isnan(chunk).sum().item())
+        inf_count += int(torch.isinf(chunk).sum().item())
+        if hs.ndim >= 3:  # noqa: PLR2004
+            bad_layers = (~finite).flatten(start_dim=2).any(dim=(0, 2))
+            affected_layers.update(
```
speculatorsbot, line 42: The caller in `src/speculators/train/data.py:299` decides whether to re-raise NaN failures with:

```python
if isinstance(e, ValueError) and "NaN" in str(e):
    raise
```

The old message was `"Hidden states contain NaN values"` (contains `"NaN"`). The new message `"Hidden states contain non-finite values (...nan_count=..., inf_count=...)"` does **not** contain the case-sensitive substring `"NaN"` -- only lowercase `"nan_count"`. As a result, NaN validation failures will be caught by the generic handler and silently downgraded to a warning instead of propagating.

Two options:
1. Update the caller in `data.py` to match on `"non-finite"` (or drop the string check altogether and always re-raise `ValueError` from `check_hidden_states`).
2. Keep `"NaN"` in the new message (e.g. `"Hidden states contain NaN / non-finite values"`).

Option 1 is cleaner -- the string-matching was fragile to begin with (added in [`04dd376`](https://github.com/vllm-project/speculators/commit/04dd376)), and this PR is a good opportunity to fix the coupling.

speculatorsbot (commented):
The validation improvements (required-key check, type guard, chunked non-finite scan, layer-slot reporting) are solid and well-motivated. One issue: the new error message breaks an existing caller that pattern-matches on the old wording.

src/speculators/data_generation/offline.py
fynnsu: This is a valid concern but resolved later in stack #969

src/speculators/data_generation/offline.py
```diff
@@ -1,23 +1,57 @@
...
+import torch
+
 logger = logging.getLogger(__name__)
 
 
 def check_hidden_states(data: dict, tokens: list[int]):
+    required = {"token_ids", "hidden_states"}
+    missing = required - data.keys()
+    if missing:
+        raise ValueError(f"Hidden-state payload is missing keys: {sorted(missing)}")
```
orestis-z: nit: `sorted` probably not adding much value here since missing can contain at most 2 elements.

src/speculators/data_generation/offline.py
orestis-z:
```python
if isinstance(e, ValueError) and "NaN" in str(e):
    raise
```

was removed in https://github.com/vllm-project/speculators/pull/994

orestis-z (approved):
LGTM

shanjiaz (approved):

PR #967, opened 2026-08-10
feat(hs-connectors): validate Mooncake samples with a checksummed manifest

fynnsu:
## Purpose

Add additional error handling around hidden states/finite checks (before and after sending) and add a checksum to ensure that the received tensors match the sent ones. 

## Tests

Validated during Kimi K3 training run that this successfully catches and recovers issues. 

## Checklist

I have filled in:

---

### DIFF
--- a/hs_connectors/src/hs_connectors/mooncake_store.py
+++ b/hs_connectors/src/hs_connectors/mooncake_store.py
@@ -204,18 +204,13 @@ class MooncakeHiddenStatesStore:
             return
         try:
             manifest = json.loads(raw)
-            names = (
-                list(manifest)
-                if isinstance(manifest, list)  # legacy manifest
-                else list(manifest.get("tensors", {}))
-            )
+            names = list(manifest.get("tensors", {}))
         except (UnicodeDecodeError, json.JSONDecodeError, AttributeError, TypeError):
-            # The current payload contract has these two tensor names. Removing them
-            # explicitly also lets retry cleanup recover from a corrupt manifest.
             names = ["hidden_states", "token_ids"]
         keys_to_remove = [f"{key}:{name}" for name in names] + [f"{key}:meta"]
-        result = self._store.batch_remove(keys_to_remove, force=True)
-        _check_store_result("batch_remove", key, result)
+        results = self._store.batch_remove(keys_to_remove, force=True)
+        for i, status in enumerate(results):
+            _check_store_result("batch_remove", keys_to_remove[i], status)
 
     def get_sample(
         self, key: str, timeout: float = 120.0, poll_interval: float = 0.05
@@ -237,9 +232,7 @@ class MooncakeHiddenStatesStore:
         return result
 
     @staticmethod
-    def _parse_manifest(
-        key: str, raw_manifest: bytes
-    ) -> dict[str, dict[str, Any] | None]:
+    def _parse_manifest(key: str, raw_manifest: bytes) -> dict[str, dict[str, Any]]:
         try:
             manifest = json.loads(raw_manifest)
         except (UnicodeDecodeError, json.JSONDecodeError) as e:
@@ -247,13 +240,6 @@ class MooncakeHiddenStatesStore:
                 f"Corrupt Mooncake manifest for key={key}: {e}"
             ) from e
 
-        if isinstance(manifest, list):
-            # Backward compatibility with handles produced before checksums were added.
-            if not manifest or not all(isinstance(name, str) for name in manifest):
-                raise MooncakeIntegrityError(
-                    f"Invalid legacy Mooncake manifest for key={key}"
-                )
-            return dict.fromkeys(manifest)
         if not isinstance(manifest, dict):
             raise MooncakeIntegrityError(
                 f"Invalid Mooncake manifest type for key={key}: "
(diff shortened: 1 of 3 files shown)


speculatorsbot (commented):
The design here is solid: replacing the bare name-list manifest with a versioned, checksummed one closes a real reliability gap for disaggregated training, and the error-marker pattern (`put_error`) for fast consumer failure is a good call. Backward compat for legacy list manifests and the cleanup-on-partial-write path are handled carefully.

Two items below.

hs_connectors/src/hs_connectors/mooncake_store.py
```diff
@@ -102,24 +202,116 @@ def delete_sample(self, key: str) -> None:
...
+                else list(manifest.get("tensors", {}))
+            )
+        except (UnicodeDecodeError, json.JSONDecodeError, AttributeError, TypeError):
+            # The current payload contract has these two tensor names. Removing them
+            # explicitly also lets retry cleanup recover from a corrupt manifest.
+            names = ["hidden_states", "token_ids"]
         keys_to_remove = [f"{key}:{name}" for name in names] + [f"{key}:meta"]
-        self._store.batch_remove(keys_to_remove, force=True)
+        result = self._store.batch_remove(keys_to_remove, force=True)
+        _check_store_result("batch_remove", key, result)
```
speculatorsbot, line 218: `_check_store_result` checks `isinstance(result, int)`, but the test fake (and, presumably, the real Mooncake API) returns `list[int]` from `batch_remove` — one status per key. A list fails the `isinstance(result, int)` guard, so the check silently passes regardless of the per-key statuses.

This means `delete_sample` can never surface a Mooncake error, which defeats the purpose of capturing the return code here.

Either teach `_check_store_result` to iterate a list result, or inline a per-key check:

```python
results = self._store.batch_remove(keys_to_remove, force=True)
if isinstance(results, list):
    for i, status in enumerate(results):
        if isinstance(status, int) and status != 0:
            raise RuntimeError(
                f"Mooncake batch_remove failed for key={keys_to_remove[i]} "
                f"with status={status}"
            )
else:
    _check_store_result("batch_remove", key, results)
```

(If the real `batch_remove` actually returns a scalar int, then the test fake's return type is unfaithful and should be fixed instead.)

tests/unit/data_generation/test_mooncake_store.py
```diff
@@ -98,13 +101,66 @@ def test_delete_sample_noop_when_missing(store):
...
+    with pytest.raises(RuntimeError, match="status=-800"):
+        store.put_sample(
+            "req-put-fail",
+            {
+                "hidden_states": torch.zeros(4, 2, 8),
+                "token_ids": torch.arange(4),
+            },
+        )
+
+    assert store._store.get("req-put-fail:meta") == b""
```
speculatorsbot, line 166: The commit message explicitly claims backward compatibility with legacy list manifests (“Legacy list manifests are still accepted so in-flight handles keep working”), but there is no test that exercises the consumer-side legacy path in `_parse_manifest`. A regression here would silently break reads from any in-flight producer that hasn't been upgraded yet.

Worth adding a test that writes the old-format manifest directly and round-trips through `get_sample`:

```python
def test_legacy_list_manifest_roundtrip(store):
    hs = torch.randn(4, 2, 8, dtype=torch.bfloat16)
    tids = torch.arange(4, dtype=torch.int64)
    store._store.put_tensor("req-legacy:hidden_states", hs.cpu().contiguous())
    store._store.put_tensor("req-legacy:token_ids", tids.cpu().contiguous())
    # Old-style manifest: bare list of tensor names, no version/checksum.
    store._store.put(
        "req-legacy:meta",
        json.dumps(["hidden_states", "token_ids"]).encode(),
    )
    out = store.get_sample("req-legacy", timeout=1.0)
    assert torch.equal(out["hidden_states"], hs)
    assert torch.equal(out["token_ids"], tids)
```

Same goes for `delete_sample` with both manifest formats.

hs_connectors/src/hs_connectors/mooncake_store.py
fynnsu: Fixed.

tests/unit/hs_connectors/test_mooncake_store.py
```diff
@@ -98,13 +101,66 @@ def test_delete_sample_noop_when_missing(store):
...
+    with pytest.raises(RuntimeError, match="status=-800"):
+        store.put_sample(
+            "req-put-fail",
+            {
+                "hidden_states": torch.zeros(4, 2, 8),
+                "token_ids": torch.arange(4),
+            },
+        )
+
+    assert store._store.get("req-put-fail:meta") == b""
```
fynnsu: I've removed the backwards support logic. This would only be needed if we update the hs_connectors the training process uses, while using an older one for vllm, but that's not a use case we intend to support

hs_connectors/src/hs_connectors/mooncake_store.py
```diff
@@ -81,19 +137,63 @@ def setup(self) -> MooncakeHiddenStatesStore:
...
+            raise
+
+    def put_error(self, key: str, error: str) -> None:
+        """Publish a small terminal marker so consumers fail fast and can retry."""
+        if self._store is None:
+            raise RuntimeError("call setup() first")
+        manifest = {
+            "version": _MANIFEST_VERSION,
+            "status": "error",
+            "error": error[:4096],
```
orestis-z, line 191: nit: why truncate?

hs_connectors/src/hs_connectors/mooncake_store.py
```diff
@@ -102,24 +202,98 @@ def delete_sample(self, key: str) -> None:
...
+        tensor_specs = self._parse_manifest(key, raw_manifest)
+
         result = {}
-        for name in names:
+        for name, spec in tensor_specs.items():
             tensor = self._store.get_tensor(f"{key}:{name}")
             if tensor is None:
-                raise RuntimeError(f"Mooncake tensor evicted for key={key}:{name}")
+                raise MooncakeIntegrityError(
+                    f"Mooncake tensor unavailable for key={key}:{name}"
```
orestis-z: nit: missing space after `:`?

hs_connectors/src/hs_connectors/mooncake_store.py
```diff
@@ -102,24 +202,98 @@ def delete_sample(self, key: str) -> None:
         raw = self._store.get(f"{key}:meta")
         if not raw:
             return
-        names = json.loads(raw)
+        try:
+            manifest = json.loads(raw)
+            names = list(manifest.get("tensors", {}))
+        except (UnicodeDecodeError, json.JSONDecodeError, AttributeError, TypeError):
```
orestis-z, line 208: silent errors: maybe worth logging them?

hs_connectors/src/hs_connectors/mooncake_store.py
```diff
@@ -102,24 +202,98 @@ def delete_sample(self, key: str) -> None:
...
+            names = ["hidden_states", "token_ids"]
         keys_to_remove = [f"{key}:{name}" for name in names] + [f"{key}:meta"]
-        self._store.batch_remove(keys_to_remove, force=True)
+        results = self._store.batch_remove(keys_to_remove, force=True)
+        for i, status in enumerate(results):
+            _check_store_result("batch_remove", keys_to_remove[i], status)
```
orestis-z: minor: use `zip` instead of `enumerate`

hs_connectors/src/hs_connectors/mooncake_store.py
```diff
@@ -81,19 +137,63 @@ def setup(self) -> MooncakeHiddenStatesStore:
...
+            }
+            result = self._store.put(meta_key, json.dumps(manifest).encode("utf-8"))
+            _check_store_result("put", meta_key, result)
+        except Exception:
+            # Do not leave partially published tensor objects behind. In particular,
+            # never publish the completion manifest after a negative put status.
+            cleanup_keys = [*written_keys, meta_key]
+            if cleanup_keys:
+                try:
+                    self._store.batch_remove(cleanup_keys, force=True)
```
orestis-z, line 179: worth checking the results returned by this with `_check_store_result`? Failed results would then land in the except clause below

hs_connectors/src/hs_connectors/mooncake_store.py
```diff
@@ -102,24 +202,98 @@ def delete_sample(self, key: str) -> None:
...
-        for name in names:
+        for name, spec in tensor_specs.items():
             tensor = self._store.get_tensor(f"{key}:{name}")
             if tensor is None:
-                raise RuntimeError(f"Mooncake tensor evicted for key={key}:{name}")
+                raise MooncakeIntegrityError(
+                    f"Mooncake tensor unavailable for key={key}:{name}"
+                )
+            self._validate_tensor(key, name, tensor, spec)
             result[name] = tensor
```
orestis-z, line 234: for consistency I would move the integrity check / tensor validation into the `_parse_manifest` method below

orestis-z (approved):
LGTM, minor non-blocking comments below

shanjiaz (approved):
Thanks for adding NaN hidden states handling

hs_connectors/src/hs_connectors/mooncake_store.py
fynnsu: Just to limit a really large error from blowing up the store (e.g. some kind of recursive loop).

hs_connectors/src/hs_connectors/mooncake_store.py
fynnsu: I think this is okay, since we're just trying to cleanup the previous keys if there already is an exception. So we might already be in an unusual state, where exceptions are more expected.

hs_connectors/src/hs_connectors/mooncake_store.py
fynnsu: I think it's better to keep these separate since parse_manifest is really just focused on handling the json data.

PR #968, opened 2026-08-10
feat(hs-connectors): expose Mooncake segment, buffer, and writer-thread sizing

fynnsu:
## Purpose

Make mooncake tuning parameters configurable using the cli. 

## Tests

Validated usage of new cli args. 

## Checklist

I have filled in:

---

### DIFF
--- a/hs_connectors/src/hs_connectors/mooncake_store.py
+++ b/hs_connectors/src/hs_connectors/mooncake_store.py
@@ -84,7 +84,7 @@ class MooncakeStoreConfig:
     local_buffer_size: int = 2 * 1024 * 1024 * 1024
     protocol: str = "tcp"
     device_name: str = ""
-    num_writer_threads: int = 16
+    num_writer_threads: int = 4
 
     @classmethod
     def from_dict(cls, d: dict | None) -> MooncakeStoreConfig:
--- a/hs_connectors/src/hs_connectors/transfer.py
+++ b/hs_connectors/src/hs_connectors/transfer.py
@@ -270,12 +270,6 @@ class MooncakeBackend(HiddenStatesBackend):
             default=2.0,
             help="Mooncake client's local staging buffer, in GiB.",
         )
-        parser.add_argument(
-            "--mooncake-writer-threads",
-            type=int,
-            default=4,
-            help="Number of asynchronous Mooncake writer threads in the vLLM client.",
-        )
 
     @staticmethod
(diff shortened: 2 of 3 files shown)


speculatorsbot (commented):
## Review Summary

Clean, well-scoped PR that exposes three previously hard-coded Mooncake tuning parameters (`global_segment_size`, `local_buffer_size`, `num_writer_threads`) as CLI arguments. The GiB-to-bytes conversion is a nice UX touch for command-line usage.

### Observations

**Default consistency** -- The defaults in the new CLI args (4.0 GiB, 2.0 GiB, 4 threads) match the updated `MooncakeStoreConfig` defaults exactly. Good.

**`num_writer_threads` default change (16 -> 4)** -- This is a behavioral change that affects every code path constructing `MooncakeStoreConfig()` without an explicit value, including the vLLM-side connector when `from_dict` receives a dict without the key. The commit message justifies this well (oversubscription), but downstream consumers should be aware.

### Minor suggestions

1. **`--mooncake-writer-threads` is a no-op on the training side.** The `num_writer_threads` field is only consumed by `MooncakeHiddenStatesConnector` (vLLM side, via `build_kv_transfer_config` -> `dataclasses.asdict`). On the training side (`from_train_args`), `MooncakeHiddenStatesStore.setup()` never reads it. Adding it to the shared `_add_mooncake_args` means it appears in both `--help` outputs, which could mislead a user into thinking it tunes something on the training side. Consider either (a) adding it only to `add_launch_args`, or (b) noting in the help text that it applies to the vLLM side only.

2. **Test coverage.** The existing `test_mooncake_defaults_backfilled_for_selected_backend` verifies `mooncake_master`, `mooncake_metadata_server`, and `mooncake_protocol` survive the config pipeline, but doesn't cover the three new args. A follow-up assertion would catch regressions if these args are ever renamed or dropped.

3. **`int()` vs `round()` for GiB conversion.** `int(args.mooncake_global_segment_gib * 1024**3)` truncates rather than rounds. For common power-of-two values this is fine, but `round(...)` would be marginally safer for fractional inputs like 0.1 GiB. Very minor.

None of these are blocking. The change is clean and correctly wired through both construction paths.

orestis-z (approved):
LGTM, agree with checking out minor suggestions 1 + 2 in https://github.com/vllm-project/speculators/pull/968#pullrequestreview-4906435914

shanjiaz (approved):

PR #969, opened 2026-08-10
feat(train): recover from hidden-state generation failures

fynnsu:
## Purpose

Add more error handling and recovery logic for training. In particular, this enables automatic retries on invalid/corrupted training data, with a "circuit breaker" system to catch consistent repeat failures and end training. 

## Tests

Validated that this successfully captures errors and retries them without interrupting model training on a larger training run. 

Also added a number of tests for this feature. 

## Checklist

I have filled in:

---

### DIFF
--- a/src/speculators/train/data.py
+++ b/src/speculators/train/data.py
@@ -1,5 +1,6 @@
 import warnings
 from collections.abc import Callable, Sequence
+from dataclasses import dataclass
 from os import PathLike
 from pathlib import Path
 from typing import Any, Literal, cast
@@ -18,8 +19,23 @@ from speculators.data_generation.vllm_client import (
     generate_hidden_states,
 )
 from speculators.train.noise_transforms import TransformTensors
+from speculators.train.utils import (
+    GENERATION_CIRCUIT_BREAKER_KEY,
+    GENERATION_ERROR_KEY,
+    GENERATION_FAILURE_COUNT_KEY,
+    LOCALLY_EMPTY_BATCH_KEY,
+)
 
 BatchType = dict[str, Any]
+
+
+@dataclass(frozen=True)
+class GenerationFailure:
+    """Picklable sentinel passed from a DataLoader worker to the collator."""
+
+    error: str
+    consecutive_failures: int
+    fatal: bool = False
 
 
 def create_empty_sample(
@@ -106,10 +122,10 @@ class BaseDataset(Dataset):
     def _get_raw_data(self, index):
         raise NotImplementedError
 
-    def __getitem__(self, index) -> BatchType | None:
+    def __getitem__(self, index) -> BatchType | GenerationFailure | None:
         data = self._get_raw_data(index)
 
-        if data is None:
+        if data is None or isinstance(data, GenerationFailure):
             return data
 
         # data structure: {
@@ -159,6 +175,9 @@ class ArrowDataset(BaseDataset):
         model: str | None = None,
         request_timeout: float | None = DEFAULT_REQUEST_TIMEOUT,
         max_retries: int = DEFAULT_MAX_RETRIES,
+        generation_validation_retries: int = 2,
+        max_consecutive_generation_failures: int = 20,
+        raise_on_generate_error: bool = False,
     ):
         self.data = load_from_disk(datapath)
         if not 0.0 < train_ratio <= 1.0:
@@ -188,6 +207,17 @@ class ArrowDataset(BaseDataset):
         self.model = model
         self.request_timeout = request_timeout
         self.max_retries = max_retries
+        if generation_validation_retries < 0:
+            raise ValueError("generation_validation_retries must be >= 0")
+        if max_consecutive_generation_failures <= 0:
+            raise ValueError("max_consecutive_generation_failures must be > 0")
+        self.generation_validation_retries = generation_validation_retries
+        self.max_consecutive_generation_failures = max_consecutive_generation_failures
+        # Each spawned DataLoader worker owns its own dataset copy. This tracks
+        # consecutive failed full round trips within that worker and resets after
+        # any valid sample.
+        self._consecutive_generation_failures = 0
+        self.raise_on_generate_error = raise_on_generate_error
 
         # Delay super init so that `_compute_approx_lengths` has required data
         super().__init__(max_len, transform, hidden_states_dtype)
@@ -196,10 +226,10 @@ class ArrowDataset(BaseDataset):
         return index + self.start_file_idx
 
     def _setup_client(self):
-        self.client = openai.OpenAI(
+        client = openai.OpenAI(
             base_url=self.vllm_endpoint, api_key="EMPTY", max_retries=0
         )
-        list_models = self.client.models.list()
+        list_models = client.models.list()
         model_id = list_models.data[0].id
         if self.model and self.model != model_id:
             raise ValueError(
@@ -209,6 +239,9 @@ class ArrowDataset(BaseDataset):
             )
         self.model = model_id
         self.transfer.setup()
+        # Do not retain a half-initialized client if model discovery or Mooncake
+        # setup failed; the outer full-round-trip retry should redo initialization.
+        self.client = client
 
     def __len__(self):
         return len(self.data)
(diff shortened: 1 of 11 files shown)


src/speculators/train/data.py
```diff
@@ -273,47 +306,125 @@ def _compute_approx_lengths(self) -> list[int]:
...
+                loaded_hs = self.transfer.get_generated(handle)
+                if loaded_hs is None:
+                    raise ValueError(
+                        f"Failed to load hidden states for handle {handle}"
+                    )
 
-        return loaded_hs
+                # Covers token mismatch, shape mismatch, NaN, and infinity. The
+                # Mooncake transfer performs manifest/checksum validation first.
+                check_hidden_states(loaded_hs, dataset_item["input_ids"].tolist())
```
speculatorsbot: The comment says this covers "NaN, and infinity" but [`check_hidden_states`](https://github.com/vllm-project/speculators/blob/e27b78fb8086022ed34b6660afc6ef377c6761cf/src/speculators/data_generation/offline.py#L7-L19) only checks `isnan()`, not `isinf()` or `isfinite()`. Infinity values in hidden states would pass validation silently and could produce NaN gradients downstream.

The config field description in `schema.py` also describes the retries as catching "non-finite" payloads, which likewise overpromises.

Either update the comment to drop the infinity claim, or (preferably) fix `check_hidden_states` to use `torch.isfinite()` — that catches both NaN and infinity in a single call:
```python
if not hs.isfinite().all():
    raise ValueError("Hidden states contain non-finite values")
```

speculatorsbot (commented):
**Design-level assessment**: This is a well-motivated and architecturally sound feature. The core problem — a single rank's data fault stalling a DDP job — is addressed cleanly: retry the full round trip, surface failures as a picklable sentinel rather than a `None` that silently vanishes, and pad an all-failed rank with a zero-loss batch so it participates in collectives without contributing noise. The per-step `all_reduce` for coordinated circuit-breaker shutdown is the right tradeoff (16 bytes, negligible latency) compared to the alternative of one rank vanishing mid-collective.

The `_setup_client` atomicity fix (assigning `self.client` only after `transfer.setup()` succeeds) is a good correctness improvement that prevents a half-initialized state from persisting across retry attempts.

Test coverage is thorough — the actual multi-process DDP test (`test_ddp_local_zero_loss_still_joins_backward_collectives`) is especially valuable for validating the zero-gradient invariant that this design relies on.

One factual inaccuracy in a new code comment noted below.

src/speculators/train/data.py
```diff
@@ -217,45 +250,125 @@ def _compute_approx_lengths(self) -> list[int]:
...
+        index: int,
+        attempt: int,
+        total_attempts: int,
+        error: Exception,
+    ) -> GenerationFailure:
+        self._consecutive_generation_failures += 1
+        fatal = (
+            self._consecutive_generation_failures
+            >= self.max_consecutive_generation_failures
+        )
```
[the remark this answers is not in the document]
shanjiaz: @fynnsu Maybe we update the definition or something?

src/speculators/train/data.py
```diff
@@ -217,45 +250,125 @@ def _compute_approx_lengths(self) -> list[int]:
...
-            check_hidden_states(loaded_hs, dataset_item["input_ids"].tolist())
+    def _record_generation_failure(
+        self,
+        *,
+        index: int,
+        attempt: int,
+        total_attempts: int,
+        error: Exception,
+    ) -> GenerationFailure:
+        self._consecutive_generation_failures += 1
```
rahul-tuli: nit: Not a blocker, but won't this increment for every failure i.e if we retry 3 times per sample and and all of them fail it would count as 3 failures? but the `--max-consecutive-generation-failures` is documented as "consecutive failed complete round trips." we should update the documentation accordingly or increment once per sample?

rahul-tuli (commented):
LGTM pending a response on a nit.

src/speculators/train/data.py
```diff
@@ -18,10 +19,25 @@
...
 BatchType = dict[str, Any]
 
 
+@dataclass(frozen=True)
+class GenerationFailure:
+    """Picklable sentinel passed from a DataLoader worker to the collator."""
+
+    error: str
+    consecutive_failures: int
+    fatal: bool = False
```
orestis-z: Nice and explicit!

src/speculators/train/utils.py
```diff
@@ -8,11 +8,23 @@
...
+from speculators.train.distributed import get_local_rank, get_rank, is_distributed
 
 logger = logging.getLogger("speculators")
 
+# Keys the collator attaches to a batch to report hidden-state generation failures.
+# ``consume_recovery_metadata`` removes them again before the batch reaches the model.
+LOCALLY_EMPTY_BATCH_KEY = "__locally_empty_batch__"
+GENERATION_FAILURE_COUNT_KEY = "__generation_failure_count__"
+GENERATION_CIRCUIT_BREAKER_KEY = "__generation_circuit_breaker__"
+GENERATION_ERROR_KEY = "__generation_error__"
```
orestis-z: instead of four stringly-typed magic keys stuffed into the batch dict, can we carry a single recovery struct under one reserved key, instead of four scalar keys?

```python
@dataclass(frozen=True)
class RecoveryMetadata:
      failure_count: int = 0
      locally_empty: bool = False
      fatal: bool = False
      error: str = ""

RECOVERY_METADATA_KEY = "__recovery__"
```

src/speculators/train/data.py
orestis-z, line 1: `_get_raw_data` / `__getitem__` now return a three-valued union `dict | GenerationFailure | None` which looks a bit like a code smell. Every consumer has to test two of the three states with two different idioms:

```python3
# __getitem__ and _get_raw_data both do:
if data is None or isinstance(data, GenerationFailure):
    return data

# collate does the mirror:
[... for b in batch if b is not None and not isinstance(b, GenerationFailure)]
```


Proposal: Reduce to a two-valued union with a single "sample unavailable" guard that carries "why":

```python3
@dataclass(frozen=True)
class SampleUnavailable:
    reason: str = ""
    counts_as_failure: bool = False   # benign skip vs. real failure
    consecutive_failures: int = 0
    fatal: bool = False
```

-> `_getitem__` / `_get_raw_data` return `BatchType | SampleUnavailable`, and every check becomes one idiom: `if isinstance(data, SampleUnavailable): return data`.

src/speculators/train/trainer.py
```diff
@@ -449,6 +452,7 @@ def train_epoch(self, epoch: int):
             timer.reset(self.global_step % self.config.log_freq == 0)
 
             timer.mark_value("start", t_before_fetch)
+            consume_recovery_metadata(batch, phase="training")
```
orestis-z: this contains an all-reduce to detect a condition that it's almost always false. Maybe we can check every N-steps instead?

src/speculators/train/data.py
```diff
@@ -217,45 +250,125 @@ def _compute_approx_lengths(self) -> list[int]:
...
         try:
-            handle = generate_hidden_states(
-                self.client,  # type:ignore[arg-type]
-                self.model,  # type:ignore[arg-type]
-                client_item,
-                timeout=self.request_timeout,
-                max_retries=self.max_retries,
+            self.transfer.delete(handle)
+        except Exception as cleanup_error:  # noqa: BLE001 - recovery must be best effort
+            warnings.warn(
```
orestis-z: minor: do we want `warnings` here? = developer-filterable?

`utils.py` and `trainer.py` use `logger.warning` which I think would also be appropriate here.

src/speculators/train/data.py
```diff
@@ -159,6 +175,9 @@ def __init__(
         model: str | None = None,
         request_timeout: float | None = DEFAULT_REQUEST_TIMEOUT,
         max_retries: int = DEFAULT_MAX_RETRIES,
+        generation_validation_retries: int = 2,
+        max_consecutive_generation_failures: int = 20,
+        raise_on_generate_error: bool = False,
```
orestis-z: minor: i saw only the tests use this and I don't think they should dictate the prod signature

orestis-z (commented):
Thanks a lot for your work!

I would suggest to add a dedicated `recovery.py` module owning the guard, the (eventual) single batch key, and the consume/raise logic. That would make ownership more obvious and not bloat `utils.py`.

minor note: worth to add also a backoff to validation-retry loop?

shanjiaz (approved):
Much cleaner! Thanks.

PR #970, opened 2026-08-10
feat(benchmark): bound the measured window and report effective throughput

fynnsu:
## Purpose

Improve benchmarking scripts to compute metrics (like throughput) more accurately, and with greater control over the benchmarking setup (e.g. number of samples). 

## Tests

Used to tune training performance for a large distributed model run. 

## Checklist

I have filled in:

---

### DIFF
--- a/scripts/benchmark.py
+++ b/scripts/benchmark.py
@@ -227,6 +227,48 @@ def compute_statistics(values: list[float]) -> dict[str, float]:
     }
 
 
+def select_measured_profiles(
+    all_profiles: list[dict], warmup_steps: int, measured_steps: int
+) -> list[dict]:
+    """Validate the sample count and discard warmup profiles."""
+    total_steps = warmup_steps + measured_steps
+    if len(all_profiles) < total_steps:
+        raise RuntimeError(
+            "Benchmark dataset exhausted before the requested number of steps: "
+            f"got {len(all_profiles)}, requested {total_steps} "
+            f"({warmup_steps} warmup + {measured_steps} measured). Use more "
+            "samples or request fewer benchmark steps."
+        )
+    return all_profiles[warmup_steps:total_steps]
+
+
+def compute_aggregate_throughput(profiles: list[dict]) -> dict[str, float]:
+    """Compute time-weighted throughput for the measured window.
+
+    ``mean(tokens_per_s)`` overweights short, fast steps. Reconstructing each
+    profile's token count and dividing by total elapsed time gives the effective
+    throughput observed by rank 0 over the complete measurement window.
+    """
+    elapsed_s = sum(profile["step_ms"] for profile in profiles) / 1000
+    rank0_tokens = sum(
+        profile["tokens_per_s"] * profile["step_ms"] / 1000 for profile in profiles
+    )
+    return {
+        "measured_time_s": elapsed_s,
+        "rank0_tokens": rank0_tokens,
+        "effective_rank0_tokens_per_s": rank0_tokens / elapsed_s,
+    }
+
+
+def shutdown_dataloader_workers(loader) -> None:
+    """Stop persistent workers before Mooncake/distributed teardown."""
+    iterator = getattr(loader, "_iterator", None)
+    shutdown = getattr(iterator, "_shutdown_workers", None)
+    if callable(shutdown):
+        shutdown()
+        loader._iterator = None
+
+
 # ---------------------------------------------------------------------------
 # Benchmark runner
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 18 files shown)


scripts/benchmark.py
```diff
@@ -227,6 +227,48 @@ def compute_statistics(values: list[float]) -> dict[str, float]:
...
+
+    ``mean(tokens_per_s)`` overweights short, fast steps. Reconstructing each
+    profile's token count and dividing by total elapsed time gives the effective
+    throughput observed by rank 0 over the complete measurement window.
+    """
+    elapsed_s = sum(profile["step_ms"] for profile in profiles) / 1000
+    rank0_tokens = sum(
+        profile["tokens_per_s"] * profile["step_ms"] / 1000 for profile in profiles
+    )
+    return {
```
speculatorsbot, line 256: nit: If `profiles` is empty, `elapsed_s` will be 0 and this line will raise `ZeroDivisionError`. Consider either guarding against empty input or documenting the precondition (e.g. `profiles must be non-empty`).

```python
if not profiles:
    return {"measured_time_s": 0, "rank0_tokens": 0, "effective_rank0_tokens_per_s": 0}
```

scripts/benchmark.py
```diff
@@ -227,6 +227,48 @@ def compute_statistics(values: list[float]) -> dict[str, float]:
...
+        "measured_time_s": elapsed_s,
+        "rank0_tokens": rank0_tokens,
+        "effective_rank0_tokens_per_s": rank0_tokens / elapsed_s,
+    }
+
+
+def shutdown_dataloader_workers(loader) -> None:
```
speculatorsbot, line 263: Minor: This accesses `loader._iterator`, a private CPython implementation detail of `torch.utils.data.DataLoader`. It works correctly today and the `getattr` fallback makes it safe if the attribute disappears, but a brief comment noting the coupling would help future readers.

speculatorsbot (commented):
## Review Summary

This PR improves the benchmark harness with four well-motivated changes:

1. **Strict sample-count validation** (`select_measured_profiles`): Replaces a silent fallback that could produce misleading results when the dataset was too small. Now raises a clear `RuntimeError` with an actionable message.

2. **Time-weighted throughput** (`compute_aggregate_throughput`): The old `mean(tokens_per_s)` equally weighted every step, overweighting short/fast steps. The new metric reconstructs total tokens and divides by total wall time -- a more accurate measure of effective throughput.

3. **Bounded sampler + worker shutdown** (`max_batches`, `shutdown_dataloader_workers`): Caps the DataLoader so persistent workers don't prefetch past the measurement window, and explicitly shuts workers down before distributed teardown to avoid racing Mooncake shutdown.

4. **Richer provenance**: Records `data_path`, `hidden_states_backend`, `num_workers`, and `prefetch_factor` -- parameters that affect results but were previously absent.

Tests are well-structured and cover the new functions thoroughly.

**CI note:** The quality-checks failures are **pre-existing on the base branch** (`generation-recovery`). The mypy errors are in unrelated files (`test_data_recovery.py`, `test_data.py`, `data.py`) -- none touched by this PR.

---

### Minor observations (non-blocking)

**1. Potential ZeroDivisionError in `compute_aggregate_throughput`** -- If every profile has `step_ms == 0` (or `profiles` is empty), `elapsed_s` will be zero and the final division raises. In practice this can't happen with real training steps, but a defensive guard or docstring precondition would make the contract clear.

**2. `shutdown_dataloader_workers` relies on `DataLoader._iterator`** -- `_iterator` is a private attribute of PyTorch's `DataLoader`. The `getattr` fallback makes it safe if the attribute disappears, but a brief comment noting the coupling would help future maintainers.

**3. `max_batches` truncation is post-generation** -- `_generate_batches` still computes all batches before truncating. For benchmarks this is fine, but worth noting it doesn't short-circuit the packing algorithm.

Overall the changes are well-scoped, well-tested, and clearly motivated by real pain points in the benchmark harness. Looks good.

src/speculators/train/distributed_batch_sampler.py
```diff
@@ -244,6 +249,8 @@ def _generate_batches(self, epoch: int) -> list[NDArray]:
         # Translate them so that they are instead relative to the overall unshuffled
         # self.lengths array.
         batches = [indices[batch] for batch in batches]
```
speculatorsbot, line 251: Note: The truncation happens after `_assign_to_packed_batches` has already generated all batches. This means the full packing algorithm runs regardless of `max_batches`. For benchmark use cases this is fine (the overhead is negligible), but worth being aware of if this parameter is ever used with very large datasets.

scripts/benchmark.py
```diff
@@ -227,6 +227,48 @@ def compute_statistics(values: list[float]) -> dict[str, float]:
...
+    throughput observed by rank 0 over the complete measurement window.
+    """
+    elapsed_s = sum(profile["step_ms"] for profile in profiles) / 1000
+    rank0_tokens = sum(
+        profile["tokens_per_s"] * profile["step_ms"] / 1000 for profile in profiles
+    )
+    return {
+        "measured_time_s": elapsed_s,
+        "rank0_tokens": rank0_tokens,
+        "effective_rank0_tokens_per_s": rank0_tokens / elapsed_s,
```
rahul-tuli, line 259: nit: I don't think this will happen in a real flow, but should we guard against division by zero?

rahul-tuli (approved):

scripts/benchmark.py
fynnsu: I think this is okay to skip, especially because its just on the benchmark script

PR #971, opened 2026-08-10
fix(dspark): initialize and optimize the Markov lookup embedding as an embedding

fynnsu:
## Purpose

Change the way we init the markov head and switch it to using the Adam optimizer. These should be better defaults for the embedding layer. 

## Tests

Not extensively validated, and change is primarily theoretical. More experimentation to follow. 

## Checklist

I have filled in:

---

### DIFF
--- a/src/speculators/train/optimizers.py
+++ b/src/speculators/train/optimizers.py
@@ -21,9 +21,9 @@ from torch.nn import Module
 logger = logging.getLogger("speculators")
 
 # Names of parameters that are 2D but should still be optimized with AdamW rather than
-# Muon, following the convention from Keller Jordan's Muon (embeddings and the output
-# head are excluded from the orthogonalized update).
-_ADAMW_NAME_HINTS = ("embed_tokens", "lm_head")
+# Muon, following the convention from Keller Jordan's Muon (embeddings, the Markov
+# lookup embedding, and the output head are excluded from the orthogonalized update).
+_ADAMW_NAME_HINTS = ("embed_tokens", "lm_head", "markov_w1")
 
 # Muon only orthogonalizes 2D weight matrices.
 _MATRIX_NDIM = 2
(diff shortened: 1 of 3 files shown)


speculatorsbot (commented):
The change is well-motivated: `markov_w1` is an embedding lookup table and should be treated as one for both initialization and optimizer routing. Initializing with `std=0.01` keeps the Markov sequential bias small at the start of training (the default `N(0, 1)` would let `W1 @ W2` dominate the DFlash logits from step 1), and routing to AdamW follows the standard Muon convention for embedding layers.

The commit message states "Add tests for both: the Muon/AdamW split had no coverage," but the diff contains no test files. Are those tests in a separate PR in the stack, or were they inadvertently left out?

WindChimeRan (commented):
Theoretically this should be applied to both markov_w1 and markov_w2. 

Please see my previous ablation study on why this is difficult. 

https://github.com/vllm-project/speculators/pull/866. 

We need more ablation study for this bug fix.

src/speculators/train/optimizers.py
```diff
@@ -21,9 +21,9 @@
 logger = logging.getLogger("speculators")
 
 # Names of parameters that are 2D but should still be optimized with AdamW rather than
-# Muon, following the convention from Keller Jordan's Muon (embeddings and the output
-# head are excluded from the orthogonalized update).
-_ADAMW_NAME_HINTS = ("embed_tokens", "lm_head")
+# Muon, following the convention from Keller Jordan's Muon (embeddings, the Markov
+# lookup embedding, and the output head are excluded from the orthogonalized update).
+_ADAMW_NAME_HINTS = ("embed_tokens", "lm_head", "markov_w1")
```
rahul-tuli: Q: Based on the comment should `markov_w2` also follow this exclusion?

rahul-tuli (approved):

shanjiaz (commented):
@fynnsu I think Ranran ablation was showing poor results for this. Have we run any ablations?

src/speculators/train/optimizers.py
fynnsu: Yep, added `markov_w2`

PR #972, opened 2026-08-10
docs(examples): add multi-node Kimi-K3 DSpark training example

fynnsu:
## Purpose

Add a multi-node Kimi K3 training example which uses slurm to setup and run the training jobs. Note: the training example is not meant to be a directly usable user script (i.e. it has a lot of hardcoded parameters and is designed for a specific server setup), but rather intended as an example that can then be adapted to new setups. 

## Tests

Based off of scripts that were used for large scale training run. 

## Checklist

I have filled in:

---

### DIFF
--- a/examples/train/kimi_k3_dspark/README.md
+++ b/examples/train/kimi_k3_dspark/README.md
@@ -1,8 +1,6 @@
 # Kimi-K3 DSpark training (multi-node, online extraction)
 
-Kimi-K3 does not fit on a single node, so this example keeps hidden-state
-extraction and draft training on disjoint node groups and streams hidden states
-between them through a Mooncake store:
+Kimi-K3 does not fit on a single node, so this example keeps hidden-state extraction and draft training on disjoint node groups and streams hidden states between them through a Mooncake store:
 
 ```text
 extraction (2 nodes)                 training (1 node)
@@ -11,21 +9,13 @@ Kimi-K3 vLLM TP8 + expert parallel -> torchrun DSpark DDP, 4 ranks
                   Mooncake streaming store
 ```
 
-This is a worked example for one 4-GPU-per-node NVL72 rack, not a portable
-harness. Values are hardcoded so the commands stay readable; adapt them rather
-than configuring them.
+This is a worked example for one 4-GPU-per-node NVL72 rack, not a portable harness. Values are hardcoded so the commands stay readable; adapt them rather than configuring them.
 
-`k3_draft_layer_config.json` is the draft **decoder** config and nothing else — a plain
-`Qwen3Config` that `--draft-config` loads as the speculator's
-`transformer_layer_config`. Everything above the decoder (block size, vocab
-size, mask token, target layers, the Markov and confidence heads) is set by the
-flags in `launch_train_node.sh`, not by this file.
+`k3_draft_layer_config.json` is the draft **decoder** config and nothing else — a plain `Qwen3Config` that `--draft-config` loads as the speculator's `transformer_layer_config`. Everything above the decoder (block size, vocab size, mask token, target layers, the Markov and confidence heads) is set by the flags in `launch_train_node.sh`, not by this file.
 
 ## Requirements
 
-- Two environments: one with vLLM plus `hs_connectors` and
-  `mooncake-transfer-engine` (extraction), one with `speculators` plus
-  `mooncake-transfer-engine` (training).
+- Two environments: one with vLLM plus `hs_connectors` and `mooncake-transfer-engine` (extraction), one with `speculators` plus `mooncake-transfer-engine` (training).
 - Verifier weights at `/mnt/shared/weights/kimi-k3`, or set `MODEL_PATH`.
 
 ## The components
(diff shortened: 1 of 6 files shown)


examples/train/kimi_k3_dspark/launch_extractor_node.sh
```diff
@@ -0,0 +1,73 @@
...
+repo_root="$(cd "$script_dir/../../.." && pwd)"
+model_path="${MODEL_PATH:-moonshotai/Kimi-K3}"
+vllm_python="${VLLM_PYTHON:-python}"
+
+node_rank="${NODE_RANK:-${SLURM_PROCID:?NODE_RANK is required (0 on the head node)}}"
+extract_addr="${EXTRACT_ADDR:?EXTRACT_ADDR is required (head node fabric IP)}"
+mooncake_master="${MOONCAKE_MASTER:?MOONCAKE_MASTER is required}"
+
+# Every rank must bind to the fast fabric, not the management NIC. Different
+# ranks picking different interfaces hang at rendezvous with no error.
```
speculatorsbot, line 20: The `10.13.84.` subnet is hardcoded here and in `launch_train_node.sh:19`. On any cluster with a different fabric subnet, `fabric_iface` will be empty and the next `ip addr show dev ""` command will fail with a confusing *"Device '' does not exist"* error (under `set -euo pipefail`).

The commit message mentions `fabric.sh` resolving the interface from `FABRIC_IFACE` or `FABRIC_SUBNET`, but neither appears in the diff -- was that extracted out? A one-line env-var fallback would make the example adaptable without hurting readability:

```suggestion
fabric_subnet="${FABRIC_SUBNET:-10.13.84.}"
fabric_iface="$(ip -o -4 addr show | awk -v s="^$fabric_subnet" '$4 ~ s {print $2; exit}')"
```

(Same change needed in `launch_train_node.sh:19`.)

examples/train/kimi_k3_dspark/README.md
```diff
@@ -0,0 +1,95 @@
...
+IP on both. Rank 1 runs `--headless`. Wait for `/health` on port 8000 before
+starting training.
+
+### 4. Trainer
+
+```bash
+MOONCAKE_MASTER=<host>:50051 VLLM_ENDPOINT=http://<head-fabric-ip>:8000/v1 \
+  bash examples/train/kimi_k3_dspark/launch_train_node.sh
+```
+
```
speculatorsbot, line 69: `TRAIN_LOGGER=''` is listed alongside the other env-var overrides (`MODEL_PATH`, `DATA_DIR`, `RUN_NAME`, `MAX_STEPS`), but unlike those, `launch_train_node.sh` never reads a `TRAIN_LOGGER` variable -- there is no `${TRAIN_LOGGER:-...}` pattern for it. Additionally, `TrainConfig` explicitly excludes pydantic-settings' env source in `settings_customise_sources`, so an env var would not reach the `--logger` field either.

Since `--logger` defaults to `""` (no backend), W&B is not enabled by default anyway. If the intent is to let users opt *in* to metric logging, consider adding a `LOGGER` variable to `launch_train_node.sh` (passed as `--logger "$LOGGER"`) and updating this line accordingly, or simply dropping the `TRAIN_LOGGER` mention.

speculatorsbot (commented):
Well-structured multi-node example that fills a real gap -- all existing training examples are single-node. The Slurm orchestration in `run.sbatch` (health-gating, cleanup trap, per-component log routing) is solid, and the README does a good job of separating the four components.

Two concrete issues below: a hardcoded subnet that will break on any cluster other than the original one, and a README override (`TRAIN_LOGGER`) that does not actually work.

examples/train/kimi_k3_dspark/launch_extractor_node.sh
```diff
@@ -0,0 +1,73 @@
...
+fabric_iface="$(ip -o -4 addr show | awk '$4 ~ /^10\.13\.84\./ {print $2; exit}')"
```
shanjiaz, line 21: Not sure we should include specific IP like these ones

examples/train/kimi_k3_dspark/run.sbatch
```diff
@@ -0,0 +1,95 @@
...
+  trap - EXIT INT TERM
+  for pid in "${background_pids[@]}"; do
+    kill "$pid" 2>/dev/null || true
+    wait "$pid" 2>/dev/null || true
+  done
+}
+trap cleanup EXIT INT TERM
+
+srun --overlap --nodes=1 --ntasks=1 --nodelist="${nodes[0]}" \
+  --output="$run_dir/logs/mooncake-master.log" \
```
speculatorsbot, line 65: `srun --output` captures only stdout. The Mooncake master's `--logtostderr=true` (end of this block) explicitly sends its logs to stderr, so `mooncake-master.log` will be empty while the actual logs land in the main sbatch output file (`%x-%j.out`). The same pattern affects the extractor (line 69) and trainer (line 87) srun calls -- Python's `logging` module writes to stderr by default, so `extractor-*.log` and `trainer.log` will also miss most output.

Adding `--error` to each `srun` call would route stderr to the per-component log files:

```suggestion
srun --overlap --nodes=1 --ntasks=1 --nodelist="${nodes[0]}" \
  --output="$run_dir/logs/mooncake-master.log" \
  --error="$run_dir/logs/mooncake-master.log" \
  mooncake_master --rpc_port 50051 --metrics_port 9003 \
  --rpc_thread_num 8 --enable_disk_eviction=false --logtostderr=true &
```

(Same `--error` addition needed for the extractor srun on line 69 and the trainer srun on line 87.)

speculatorsbot (commented):
Both issues from the previous review are resolved: `FABRIC_SUBNET` is now a required env var (no hardcoded subnet), and the non-functional `TRAIN_LOGGER` mention is gone from the README. One new finding below about log routing in the Slurm orchestrator.

rahul-tuli (approved):
LGTM!

shanjiaz (commented):
This looks good but I do want to test this on the rack before we merge.

PR #977, opened 2026-08-11
fix(datagen): store MM messages column as JSON to keep shard schemas …

avshalomman:
## Purpose

Multimodal conversations with tool calls don't survive the round-trip from preprocessing to the vLLM hidden-states request. Three failure modes, all hit on real-world multimodal agent-trajectory data (UI-action SFT corpora with tool calls):

1. **Preprocessing crash with `num_proc > 1`**: `build_eagle3_dataset`'s `dataset.map` declares no `features`, so each worker shard infers its own Arrow schema for the structured `messages` column. Shards holding different conversation shapes (optional `tool_calls`, plain-string vs typed-part-list `content`) disagree, and the final concatenation fails:
   `ValueError: The features can't be aligned because the key messages of features {...} has unexpected type`

2. **Tool-call requests rejected by vLLM**: HF chat templates require assistant `tool_calls[].function.arguments` as dicts, but the OpenAI Chat Completions schema vLLM validates against requires a JSON *string* — so requests carrying dict arguments get `400: Input should be a valid string` and those samples are silently dropped. (Reading the prepared dataset in torch format additionally converts numeric argument leaves, e.g. click coordinates, into tensors, failing request serialization outright with `Object of type Tensor is not JSON serializable`.)

3. **Tool definitions dropped from the request**: chat templates render `tools` into the prompt, so preprocessing tokenizes with them — but only `messages` were sent to vLLM. The re-rendered prompt omits the tools block, and `--validate-outputs` rejects every tool-using multimodal sample with `Prompt token IDs mismatch`.

## Description

- `_preprocess_batch` serializes each row's messages with `json.dumps`, so the column is always `Value("string")` — deterministic regardless of sharding, and immune to torch-format tensor conversion.
- The parsed `tools` each conversation was tokenized with are stored alongside, as a JSON-string column for the same reason.
- `build_client_item` (the single consumer, shared by online training and offline datagen) decodes the JSON string, re-encodes dict tool-call arguments as JSON strings for the API (vLLM decodes them back to dicts server-side before rendering its chat template), and forwards `tools` in the request body.
- Datasets prepared before this change hold structured rows and no `tools` column; both are still accepted.

## Tests

- New regression test `test_build_eagle3_dataset_multiproc_heterogeneous_conversations`: mixed string/typed-part conversations (with a heterogeneous `tools` column) through `build_eagle3_dataset` with `num_proc=2`. Verified it reproduces the alignment crash on `main` and passes with this change.
- New unit tests for `build_client_item`: JSON-string multimodal messages are decoded and forwarded, text-only messages are omitted, legacy structured rows keep working, dict tool-call arguments are stringified (string arguments pass through), and `tools` are forwarded only alongside multimodal messages.
- New unit tests for `_chat_extra_body`: re-render flags, Mistral `continue_final_message` special case, tools included when present / omitted when absent.
- `tests/integration/datagen/`, `tests/unit/train/`, `tests/unit/data_generation/`, `tests/unit/scripts/`: 492 passed, 11 skipped. `ruff check`, `ruff format`, `mypy` clean.

## Checklist

- [x] The purpose of the PR
- [x] The test plan/results
- [ ] (Optional) The necessary documentation update
- [x] I (a human) have written or reviewed the code in this pr to the best of my ability


speculatorsbot (commented):
**Design Assessment**

Well-motivated fix for three real failure modes in multimodal tool-call conversations. The approach is sound:

1. JSON-serializing the  column makes the Arrow schema deterministic across preprocessing shards, eliminating the  crash.
2. Stringifying tool-call arguments at both store time () and read time (, idempotent for new datasets) ensures compliance with the OpenAI Chat Completions schema and prevents torch-format tensor conversion from corrupting arguments.
3. Storing and forwarding tool definitions via  reproduces the tokenized prompt during hidden-states extraction.
4. Canonicalizing tool key order to match vLLM's Pydantic serialization prevents -based templates from producing divergent prompts.

Backward compatibility with pre-existing datasets is correctly preserved:  accepts both structured and JSON-string messages, and  absence is handled gracefully.

The  helper is a clean DRY improvement that deduplicates the extra-body construction across both sync and async paths. Pinning  to  correctly avoids vLLM's  requirement while still rendering tools into the prompt.

Test coverage is comprehensive: regression tests for the schema alignment crash (both normal and empty-shard cases), unit tests for argument stringification, tool forwarding, chat extra body, and  with all legacy/new format combinations.

No blocking issues found in the diff.

**Rebase note**

The PR has merge conflicts with current . Since the PR was opened,  was refactored: the direct message append was replaced by a call through  (line 390 on current main), which appends structured dicts via  without JSON serialization and does not handle . Additionally,  was renamed to . During the rebase, these changes will need to be reconciled:  will need the same  treatment for messages and  propagation to avoid reintroducing the schema alignment issue on the render-boundary code path.

Generated with [Claude Code](https://claude.ai/code) using the `/pr-review` skill

speculatorsbot (commented):
*(Supersedes the previous review where inline code formatting was lost.)*

**Design Assessment**

Well-motivated fix for three real failure modes in multimodal tool-call conversations. The approach is sound:

1. JSON-serializing the `messages` column makes the Arrow schema deterministic across preprocessing shards, eliminating the `num_proc > 1` crash.
2. Stringifying tool-call arguments at both store time (`_adapt_conv_for_vllm`) and read time (`build_client_item`, idempotent for new datasets) ensures compliance with the OpenAI Chat Completions schema and prevents torch-format tensor conversion from corrupting arguments.
3. Storing and forwarding tool definitions via `tools_json` reproduces the tokenized prompt during hidden-states extraction.
4. Canonicalizing tool key order to match vLLM's Pydantic serialization prevents `tojson`-based templates from producing divergent prompts.

Backward compatibility with pre-existing datasets is correctly preserved: `build_client_item` accepts both structured and JSON-string messages, and `tools_json` absence is handled gracefully.

The `_chat_extra_body` helper is a clean DRY improvement that deduplicates the extra-body construction across both sync and async paths. Pinning `tool_choice` to `"none"` correctly avoids vLLM's `--enable-auto-tool-choice` requirement while still rendering tools into the prompt.

Test coverage is comprehensive: regression tests for the schema alignment crash (both normal and empty-shard cases), unit tests for argument stringification, tool forwarding, chat extra body, and `build_client_item` with all legacy/new format combinations.

No blocking issues found in the diff.

**Rebase note**

The PR has merge conflicts with current `main`. Since the PR was opened, `_preprocess_batch` was refactored: the direct message append was replaced by a call through `_append_boundary_rows` (line 390 on current main), which appends structured dicts via `_adapt_conv_for_vllm(row["conv"])` without JSON serialization and does not handle `tools_json`. Additionally, `build_eagle3_dataset` was renamed to `build_speculator_training_dataset`. During the rebase, these changes will need to be reconciled: `_append_boundary_rows` will need the same `json.dumps` treatment for messages and `tools_json` propagation to avoid reintroducing the schema alignment issue on the render-boundary code path.

PR #991, opened 2026-08-13
feat(regen): support local prompt files

WindChimeRan:
## Purpose

Response regeneration currently accepts only registered dataset presets. This PR lets `--dataset` also accept a local JSON or JSONL prompt file without modifying `DATASET_CONFIGS`.

```bash
python scripts/response_regeneration/script.py \
  --dataset ./my_prompts.jsonl
```

Rows may use `messages`, `conversations`, or `prompt` as either a string or a message list. When `--outfile` is omitted, the local file stem is used in the generated output filename.

## Changes

- Accept registered presets or local `.json`/`.jsonl` paths through `--dataset`.
- Stream local JSON/JSONL files through the existing regeneration pipeline.
- Reject `--split` and `--subset` for local files.
- Document the local input mode and add focused CLI and loader coverage.

Split from #924.

## Tests

- `.venv/bin/pytest -q tests/unit/scripts/test_response_regeneration.py` — 48 passed
- `make quality` — passed

## Checklist

- [x] Purpose and motivation
- [x] Test plan and results
- [x] Documentation update
- [x] Human-authored or reviewed

### DIFF
--- a/scripts/response_regeneration/script.py
+++ b/scripts/response_regeneration/script.py
@@ -34,26 +34,24 @@ MULTIMODAL_DATASETS = {"sharegpt4v_coco"}
 REGEN_DATASETS = [name for name in DATASET_CONFIGS if name not in MULTIMODAL_DATASETS]
 
 
-def _dataset_choice(name: str) -> str:
-    """Reject multimodal presets with a reason, not a bare invalid choice."""
+def _dataset_source(name: str) -> str:
+    """Validate a registered preset or local JSON/JSONL file."""
     if name in MULTIMODAL_DATASETS:
         raise argparse.ArgumentTypeError(
             f"{name!r} is multimodal; response regeneration does not support "
             "images yet. Generate target-model responses with a multimodal-capable "
             "workflow, then convert them with `prepare_data.py`."
         )
-    return name
+    if name in REGEN_DATASETS:
+        return name
 
-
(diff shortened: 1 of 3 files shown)


scripts/response_regeneration/script.py
```diff
@@ -72,13 +85,19 @@ def parse_args():
         default=None,
         help="Model name exposed by vLLM (auto-detected if not specified)",
     )
-    parser.add_argument(
+    dataset_group = parser.add_mutually_exclusive_group()
+    dataset_group.add_argument(
```
fynnsu: I kinda think the ux would be better if the user can just pass their `.jsonl` file to `--dataset`. 

E.g. 
`python scripts/response_regeneration/script.py --dataset my_dataset.jsonl`

Then we could detect based on the dataset ending with `jsonl/json` and being a local file.

It also means we can avoid some of the additional logic checking for both fields.

fynnsu (commented):
Looks good, but I think we can improve ux / avoid the extra arg.

scripts/response_regeneration/script.py
WindChimeRan: Agree! Addressed in c98b039: --dataset now accepts either a registered preset or an existing local .json/.jsonl path.

shanjiaz (approved):
Looks great! Thanks for addressing Fynn's comments.

PR #992, opened 2026-08-14
feat(train): add --gradient-checkpointing for DFlash/DSpark

WindChimeRan:
## Summary

Adds opt-in `--gradient-checkpointing` for DFlash and DSpark. It recomputes draft-layer activations during backward, reducing peak memory at the cost of step time.

The flag remains off by default. EAGLE-3, P-EAGLE, and MTP are unchanged.

## H100 results

DFlash with a Qwen3-8B verifier, `--total-seq-len 8192`, and current defaults (`--num-layers 5 --block-size 16`, CE, Muon) on one H100 80GB. Values are medians of three interleaved runs after warmup.

| max-anchors | Step time, off → on | Peak allocated, off → on |
| ---: | ---: | ---: |
| 512 *(default)* | 365 → 395 ms *(+8.2%)* | 33.0 → **25.5 GB** *(-22.6%)* |
| 1024 | 507 → 562 ms *(+10.9%)* | 44.1 → **31.2 GB** *(-29.1%)* |

The added time is in backward recomputation; forward and optimizer times were effectively unchanged.

<details>
<summary>Benchmark command and environment</summary>

```text
tree      007c7c1 (content-identical to current head b9f5e1e)
stack     Python 3.12.9, torch 2.13.0+cu130, CUDA 13.0, transformers 5.14.1
protocol  synthetic batches; 10 warmup + 4 measured steps; 3 interleaved repeats
command   python scripts/benchmark.py run --synthetic --warmup-steps 10 \
            --measured-steps 4 -- --speculator-type dflash \
            --verifier-name-or-path Qwen/Qwen3-8B [--gradient-checkpointing]
```

</details>

## Implementation

- Adds the training flag and CLI documentation.
- Enables non-reentrant checkpointing for `DFlashDraftModel` (inherited by DSpark).
- Records the setting in benchmark metadata.
- Adds a regression test for recomputation and gradient equivalence.

Non-reentrant checkpointing is required because draft layers pass inputs as keyword arguments; the reentrant wrapper forwards only positional arguments and would break draft-layer gradients.

## Validation

- Targeted tests: **490 passed, 5 skipped**
- Ruff and mypy: no findings introduced relative to `main`
- End-to-end training with checkpoint creation

### DIFF
--- a/src/speculators/models/dflash/core.py
+++ b/src/speculators/models/dflash/core.py
@@ -38,6 +38,9 @@ _compiled_create_block_mask = torch.compile(create_block_mask)
 class DFlashDraftModel(DraftVocabMixin, SpeculatorModel):
     config_class: ClassVar[type[DFlashSpeculatorConfig]] = DFlashSpeculatorConfig  # type: ignore[misc]
     _no_split_modules = ["Qwen3DFlashDecoderLayer"]
+    # Qwen3DFlashDecoderLayer already extends GradientCheckpointingLayer; this opts
+    # the model into PreTrainedModel.gradient_checkpointing_enable().
+    supports_gradient_checkpointing = True
     _keys_to_ignore_on_load_missing: ClassVar[list[str]] = [  # type: ignore[misc]
         "embed_tokens.weight",
         "verifier_norm.weight",
--- a/src/speculators/models/dflash/core.py
+++ b/src/speculators/models/dflash/core.py
@@ -38,8 +38,7 @@ _compiled_create_block_mask = torch.compile(create_block_mask)
 class DFlashDraftModel(DraftVocabMixin, SpeculatorModel):
     config_class: ClassVar[type[DFlashSpeculatorConfig]] = DFlashSpeculatorConfig  # type: ignore[misc]
     _no_split_modules = ["Qwen3DFlashDecoderLayer"]
-    # Qwen3DFlashDecoderLayer already extends GradientCheckpointingLayer; this opts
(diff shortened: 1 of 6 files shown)


mgoin:
I feel like `--max-anchors 3072` is an extreme case to exaggerate the memory usage. I'd be curious to see the impact with a more standard 512 or 1024

Still I think this is useful to have as an option for users if not difficult to maintain

src/speculators/models/dflash/core.py
```diff
@@ -35,6 +35,8 @@
 class DFlashDraftModel(DraftVocabMixin, SpeculatorModel):
     config_class: ClassVar[type[DFlashSpeculatorConfig]] = DFlashSpeculatorConfig  # type: ignore[misc]
     _no_split_modules = ["Qwen3DFlashDecoderLayer"]
+    # Decoder layers already implement Transformers' checkpointing protocol.
+    supports_gradient_checkpointing = True
```
reed-meyerson, line 42: Is this attribute actually used by anything by anything? Was it intended to be referenced in the 'configure_gradient_checkpointing' function?

WindChimeRan: * Yes, the attribute is used by transformers’ `gradient_checkpointing_enable()` indirectly. 
* No, it does not need to be referenced directly in `configure_gradient_checkpointing()`, since that function already calls `gradient_checkpointing_enable()`.

WindChimeRan:
superseded by #1100

PR #993, opened 2026-08-14
feat(losses): cap fused-loss BLOCK_SIZE at 4096 on Ascend NPU

PHOEBEMOON0802:
## Purpose
The fused Triton loss kernels perform a double-row load (logits + targets per block) that overflows triton-ascend's ~192 KB Unified Buffer at BLOCK_SIZE >= 8192. Thread torch.device into _calculate_settings and clamp NPU launches to MAX_FUSED_SIZE_NPU=4096; CUDA keeps the original MAX_FUSED_SIZE ceiling.

Parametrize test_fused_matches_eager over CUDA and NPU so value/gradient parity is exercised on both backends, and add
test_npu_block_size_respects_ub_cap to assert the 4096 cap holds across vocab sizes from 512 up to Qwen3's 151936 (multi-block streaming path).

## Tests
Training loss is verified to match pre- and post-commit.


PHOEBEMOON0802:
@shanjiaz Training loss was verified to match pre- and post-commit on both backends. Plz review. Thanks.

WindChimeRan (commented):
* All fused losses fail under torch.compile(fullgraph=True)
* Nit: On a host containing both NPU and CUDA, CUDA tensors receive the NPU settings.

PHOEBEMOON0802:
> * All fused losses fail under torch.compile(fullgraph=True)
> * Nit: On a host containing both NPU and CUDA, CUDA tensors receive the NPU settings.

@WindChimeRan solved this, plz review again

WindChimeRan (commented):
LGTM recommend approve @shanjiaz

tests/unit/models/test_fused_losses.py
rahul-tuli, line 1: The PR description says `test_npu_block_size_respects_ub_cap` is included, and CodeRabbit reviewed an earlier version of it, but this test does not appear in the current diff. Was it accidentally dropped?

src/speculators/losses/fused.py
rahul-tuli, line 71: Does triton-ascend use or ignore the `num_warps` parameter? There is an AMD/HIP special case on that adjusts warp count for the wider wavefront, but no equivalent for NPU. If triton-ascend ignores `num_warps` entirely, could you add a brief comment documenting that? If it does use the value, should there be an NPU-specific path here?

tests/unit/models/test_fused_losses.py
rahul-tuli: Could you also add a unit test that calls `_calculate_settings` directly with mock devices (`SimpleNamespace(type='npu')` and `SimpleNamespace(type='cuda')`) to verify the block-size selection for various vocab sizes? This would ensure the NPU cap is exercised on standard CUDA CI runners where the integration tests only hit the CUDA path

rahul-tuli (changes requested):

tests/unit/models/test_fused_losses.py
PHOEBEMOON0802: addressed in 9b5352b933ef2b1a2c716625d2730e61c934ab2a

src/speculators/losses/fused.py
PHOEBEMOON0802: triton-ascend ignores "num_warps", and I added a  comment in commit: 8a4090e394211a1e693d60676f37fc1ec9e02cb4.

PHOEBEMOON0802:
@rahul-tuli @shanjiaz plz review again, thanks.

rahul-tuli (approved):
LGTM! pending CI, Thanks for the quick turnaround

rahul-tuli:
It seems like quality is failing, could you fix that?

shanjiaz:
@sunny_infra Could you take a look? Thanks!

PHOEBEMOON0802:
> It seems like quality is failing, could you fix that?

@rahul-tuli I fixed this problem in the new commit, plz start the CI again, thanks!

PHOEBEMOON0802:
@rahul-tuli CI passed, plz help to merge, thanks!

rahul-tuli (approved):

shanjiaz:
Can we run e2e on this before merging.

PHOEBEMOON0802:
> Can we run e2e on this before merging.

Of course, how to run?

WindChimeRan:
@PHOEBEMOON0802 we don't have a NPU machine. 

e.g., the kernel fusion is to solve memory issue OOM on DFlash/DSpark. Maybe a e2e test of red main ->green PR test on large batch size or long sequence len or large block or anchor setting?

PHOEBEMOON0802:
> @PHOEBEMOON0802 we don't have a NPU machine.
> 
> e.g., the kernel fusion is to solve memory issue OOM on DFlash/DSpark. Maybe a e2e test of red main ->green PR test on large batch size or long sequence len or large block or anchor setting?

I have trained dspark with batch_size=15, max_anchors=1024, full vocab_size=248320 and seq_len=8192 on a 500k dataset.

shanjiaz:
@PHOEBEMOON0802 We're adding e2e smoke tests to ci, will kick off once that's merged. Mainly just to make sure this doesn't break our existing setup. We'll definitely merge this soon! Thank you~

zihanlin-ai:
Independent validation on Atlas A3.

**Failure on `main`, from the repo's own example**

Running `examples/train/dflash_qwen3_8b_sharegpt_online_5k_npu.sh` (from #915) on `main` @ `c05b95b`, training dies at step 0 of epoch 0 on both ranks:

```
File "src/speculators/losses/utils.py", line 140, in ce_loss
    return _fused_kernel("fused_ce_loss")(logits, targets)
File "src/speculators/losses/fused.py", line 253, in forward
triton.compiler.errors.MLIRCompilationError:
[ConvertLinalgRToBinary] encounters error:
loc("src/speculators/losses/fused.py":92:0): error: Failed to run BiShengHIR pipeline
loc("src/speculators/losses/fused.py":92:0): error: ub overflow, requires 4196608 bits
      while 1572864 bits available! (possible reason: tiling basic block is too large)
```

The arithmetic matches the diagnosis in this PR: the example uses `--draft-vocab-size 32000`, so `triton.next_power_of_2(32000) = 32768`, which is under `MAX_FUSED_SIZE = 131072` and therefore passes through unchanged. 4,196,608 bits is 512 KiB of Unified Buffer against the 192 KiB (1,572,864 bits) the A3 chip has.

**With this PR applied** (clean `git apply` onto `c05b95b`), the same command trains:

- all 10 epochs completed, exit code 0, `319` optimizer steps per epoch, no `ub overflow` and no Triton compilation error at any point;
- forward *and* backward both exercised — `profile/fwd_ms=237`, `profile/bwd_ms=431`, `profile/opt_ms=413`, `profile/step_ms=1.09e+03`, `profile/tokens_per_s=6.98e+03`;
- validation loss decreasing monotonically over all ten epochs: `1.581 → 1.485 → 1.423 → 1.412 → 1.385 → 1.368 → 1.356 → 1.343 → 1.331 → 1.322`;
- a self-contained checkpoint written per epoch (`model.safetensors`, optimizer + scheduler state, `training_state.json`) for all of `0`..`9`, with `checkpoint_best` tracking the lowest validation loss.

(The run spans two A3 hosts: the first host was preempted right after epoch 0, so epochs 1-9 are a `resume_from_checkpoint` continuation on a second host; `Found checkpoint at .../checkpoints/0` / `Resuming training on epoch 1.`, loss series continuous across the seam.)

**`tests/unit/models/test_fused_losses.py` as updated by this PR, one A3 chip**

```
test_fused_matches_eager[kl_div]  PASSED     test_fused_matches_eager[tv]         PASSED
test_fused_matches_eager[rkl]     PASSED     test_fused_matches_eager[nla]        PASSED
test_fused_matches_eager[jsd]     PASSED     test_fused_matches_eager[lk_hybrid]  PASSED
test_fused_matches_eager[ce]      PASSED     test_calculate_settings_respects_device_cap PASSED
test_eager_implementation_supports_differentiable_targets PASSED
```

All seven losses match eager in **value and gradient**, across all three regimes in that test — fp32 with saturated ±30 point-mass rows, bf16 (the training dtype), and the 151936-wide vocab leg, which on NPU is the one that actually exercises the tighter `MAX_FUSED_SIZE_NPU = 4096` multi-block loop with a non-power-of-2 tail.

**7 remaining failures, not from this PR**

The same run reports `7 failed`, all `test_compiles_fullgraph[*]`:

```
torch._dynamo.exc.Unsupported: Attempted to inline function marked as skipped
  `NPUDriver.get_current_device` should not be traced
  filename: triton/backends/ascend/driver.py
```

| environment | `test_compiles_fullgraph` |
|---|---|
| NPU, this PR applied | 7 failed |
| NPU, pristine `main` `fused.py` (patch reverted, same container, same chip) | fails identically |
| CUDA, this PR applied | **7 passed** |

`torch._dynamo` refuses to trace into the triton-ascend driver, independent of `BLOCK_SIZE` and of this change. It is a gap for `torch.compile(fullgraph=True)` on NPU; separate issue against triton-ascend/torch_npu.

**CUDA, no regression**

8× H20 (`vllm/vllm-openai:latest`, torch 2.11.0+cu130), same test file:

- pristine `main`: `16 passed`
- with this PR (and its updated test file): `17 passed` — the extra one is the new `test_calculate_settings_respects_device_cap`

and the CUDA block sizes are unchanged by the patch:

```
vocab  32000  BLOCK_SIZE 32768
vocab 151936  BLOCK_SIZE 131072
caps: cuda 131072   npu 4096
```

i.e. the 4096 cap is reached only through `device.type == "npu"`; the CUDA path still saturates `MAX_FUSED_SIZE` on a 151936 vocab exactly as before.

**Environment**

- speculators `main` @ `c05b95b` + this PR, `pip install -e .`
- Atlas A3 (Ascend 910, 64 GB per chip), driver 25.2.1, CANN toolkit 9.0.1 in-container
- image `quay.io/ascend/vllm-ascend:nightly-releases-v0.23.0-a3`; python 3.12.13, torch 2.10.0, torch_npu 2.10.0.post2, transformers 5.5.4, datasets 5.0.1
- verifier `Qwen/Qwen3-8B`, dflash drafter, `--num-layers 5 --block-size 8 --draft-vocab-size 32000 --target-layer-ids 2 18 33 --total-seq-len 8192 --draft-attn-impl sdpa`, `loss_implementation='fused'`, `loss_fn='ce'`, DDP over 2 chips with vLLM DP=2 on 2 more
- CUDA control: 8× H20, `vllm/vllm-openai:latest`, torch 2.11.0+cu130

**Reproduction note**

Reaching the training step on `main` needed two unrelated fixes first: `prepare_data.py` requires `--render-endpoint` since #975 while every `examples/train/*.sh` still calls it before the server exists, and the vLLM pinned in the Ascend image rejects `ExampleHiddenStatesConnector` together with chunked prefill so the server needs `--no-enable-chunked-prefill`. Both reported on #915.

shanjiaz:
@zihanlin-ai Thanks for running the example! Yes we'll update the examples to use our new render endpoint changes! Will merge soon.

fynnsu (approved):

PR #1000, opened 2026-08-17
perf(losses): stop pinning target logits through the fused ce backward

WindChimeRan:
## Summary

Fused CE retained the full target-logit tensor through backward even though backward only needs the target argmax cached during forward. This PR saves `None` for CE and guards target-pointer use behind the non-CE specialization.

At the measured 32k-vocabulary configuration (`max_anchors=512`, `block_size=16`, bf16), this releases 0.49 GiB. Without a vocabulary mapping, training uses the full verifier vocabulary; for Qwen3's 151,936-token vocabulary, the same configuration releases 2.32 GiB.

The saving scales with `max_anchors × block_size × vocabulary size × element size`, not packed `total_seq_len` at a fixed anchor count. Losses and gradients are unchanged. Distribution losses are unchanged, and compound losses containing one will still retain targets.

Follow-up to #951.

## Results

GB10, 512 anchors × 16 positions × 32k mapped vocabulary:

| | Before | After |
| --- | ---: | ---: |
| Held into backward | 1.31 GiB | 0.83 GiB |
| Backward-phase peak | 1.80 GiB | 1.31 GiB |

Gradient checksum was bit-identical; KL was unchanged.

## Tests

- `pytest tests/unit/models/test_fused_losses.py` — 16 passed
- `pytest tests/unit/models tests/unit/train` — 475 passed, 5 skipped

## Checklist

- [x] The purpose of the PR
- [x] The test plan/results
- [ ] (Optional) The necessary documentation update
- [x] I (a human) have written or reviewed the code in this PR to the best of my ability

### DIFF
--- a/src/speculators/losses/fused.py
+++ b/src/speculators/losses/fused.py
@@ -48,8 +48,6 @@ _OP_TV = tl.constexpr(4)
 # stats [_N_STATS, n_rows]: [0]/[1] draft row max / sum-exp, [2]/[3] same for
 # targets (unused by ce), [4] per-OP -- kl_div/rkl: row loss L | jsd:
 # KL(dp||mix) | tv: s_s | ce: argmax as float32 (exact: vocab < 2**24).
-# ce's [4] is the whole of what its backward needs from the targets, so it
-# alone takes targets_ptr=None there and drops them at the end of the forward.
 _N_STATS = 5
 
 
@@ -206,7 +204,7 @@ def loss_backward_kernel(
     lse_d = m_d + tl.log(z_d)
     extra = tl.load(stats_ptr + 4 * stats_row + pid)
     if OP != _OP_CE:
-        # ce is handed targets_ptr=None, so even the offsetting stays guarded.
+        # CE passes None for targets_ptr.
         targets_ptr += pid * n_cols
         m_t = tl.load(stats_ptr + 2 * stats_row + pid)
         z_t = tl.load(stats_ptr + 3 * stats_row + pid)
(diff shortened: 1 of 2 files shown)


tests/unit/models/test_fused_losses.py
rahul-tuli, line 1: The new test checks that backward runs, but could it also spot-check that `logits.grad` matches the eager CE gradient? Without that, the test proves the kernel does not crash with `None` but not that it produces correct results when targets are released early. Even a loose `assert_close` against the eager path would strengthen the coverage.

src/speculators/losses/fused.py
rahul-tuli, line 249: Minor: the PR description mentions savings of 0.49-2.32 GiB; is that the backward-phase reduction only, or the full forward+backward peak? Since `targets_flat` is still materialized during forward, the peak during forward is unchanged?

src/speculators/losses/fused.py
```diff
@@ -189,7 +189,6 @@ def loss_backward_kernel(
     """Recompute probabilities from the saved stats and apply the OP gradient."""
     pid = tl.program_id(0).to(tl.int64)
     logits_ptr += pid * n_cols
-    targets_ptr += pid * n_cols
```
rahul-tuli, line 192: 💯 nice, Just to confirm my reading: in the original code, this line was already dead for CE (no downstream load), so the change is purely structural, right?

rahul-tuli (commented):
LGTM, pending comments!

src/speculators/losses/fused.py
WindChimeRan: yes!

src/speculators/losses/fused.py
WindChimeRan: Yes. it's backward-phase reduction only.

tests/unit/models/test_fused_losses.py
WindChimeRan: Correctness is already covered by test_fused_matches_eager[ce] from the previous PR. This PR does not add new branches, so it's still covered.

WindChimeRan:
Addressed the comments @rahul-tuli !

fynnsu (approved):
Thanks!

PR #1001, opened 2026-08-17
[hs_connectors] Add mooncake-transfer-engine dependency

dsikka:
Summary:
- Add as optional dependency  - mooncake's default wheel is built against CUDA 12, we need to specify CUDA 13 in a separate install. Adding both but runners / modelopt will need to install based on our cuda version
- Separate out hs-tests into separate modules for easier testing when doing hs_connectors releases 
- Small update to one of the smoke test

Test Run: 
https://github.com/neuralmagic/llm-compressor-testing/actions/runs/32166224680/job/95808415181

Test run is on a branch where we're installing mooncake during e2e test flows so that they are no longer skipped


shanjiaz (approved):

pyproject.toml
```diff
@@ -65,6 +65,12 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
+mooncake-cuda = [
+    "mooncake-transfer",
```
fynnsu:
```suggestion
    "mooncake-transfer-engine",
```

fynnsu (commented):
Fix typo

dhuangnm (approved):
Cool, looks good! Thank you @dsikka !

PR #909, opened 2026-08-03
fix(tests): fix weekly regression CI failures

rahul-tuli:
## Purpose

Fixes two independent failures in the [2026-08-01 weekly regression run](https://github.com/neuralmagic/llm-compressor-testing/actions/runs/30678080008):

### 1. MTP CUDA launch failure (`test_mtp_online_regression`)

**Root cause:** Multi-process CUDA graph contention. Both the vLLM datagen server (explicit CUDA graphs) and the training process (`torch.compile` inductor cudagraph trees) maintain private CUDA memory pools on the same GPU. Asynchronous graph replay overlap causes `cudaErrorLaunchFailure`. This affected all three failing jobs (Py 3.10, 3.11, 3.12/nightly).

**Evidence it's not OOM:** Controlled memory comparison showed the crash at 64.5 GiB peak vs a passing run at 65.9 GiB peak — the crash uses *less* memory, ruling out OOM. `CUDA_LAUNCH_BLOCKING=1` (which serializes kernel launches) eliminates the failure, confirming it's a race condition.

**Fix:** Add `enforce_eager=True` to the regression test's vLLM server config. This disables CUDA graph capture in the server, eliminating the contention. The smoke test (`test_mtp_finetuning.py`) already uses `enforce_eager=True`.

### 2. Eagle3 acceptance threshold miss (`test_eagle3_qwen3_8b_sharegpt`)

**Root cause:** Zero-margin threshold on non-deterministic training. The token-0 threshold `0.40` was set as the bare observed value in PR #389 with no margin. Training is inherently non-deterministic (`deterministic_cuda=False`, CUDA kernels use non-deterministic algorithms), and the test uses only 3 prompts generating ~1000-1500 draft steps — +/- 2-3% fluctuation is expected. The 1.7% miss (0.393 vs 0.400) is within this noise floor. Nightly-only because torch 2.13 shifts logits slightly via changes like `clamp→clamp_` and different CUDA kernels.

**Precedent:** Thresholds have been lowered before for the same reason (PRs #246, #463 — the latter's commit message literally says "These tests still seem flaky").

**Fix:** Lower Eagle3 token-0 threshold from `0.40` → `0.38` in the online regression test. This provides ~5% margin while still catching genuine regressions (which typically drop below 0.30).

### MTP regression test — PASSED

Previously-failing test now completes all 146/146 training steps:

```
CADENCE=weekly VLLM_PYTHON=/workspace/vllm/.venv/bin/python \
[...]

PR #1003, opened 2026-08-18
test(mtp): xfail flaky weekly MTP online regression test

rahul-tuli:
## Purpose

`test_mtp_online_regression` is flaky. It crashes intermittently mid-training (specially with vLLM nightly) with `CUDA error: unspecified launch failure` at varying training steps (step 31, 48, 76 — not deterministic), and has done so roughly one week in three since at least 2026-07-11. This is not a new regression — the weekly cadence (`@requires_cadence("weekly")`) masked the intermittency by providing only ~1 observation per cell per week.

Two prior fixes were attempted and both failed:
- PR #800 (`gpu_memory_utilization` reduction) — crash recurred the following week
- PR #909 (`enforce_eager=True`) — validated by a single green run, then crashed again on 08-15

Local investigation on dedicated 2×H100 GPUs with exact CI-matching environments showed a clear pattern:

| Environment | Attempts | Crashes | Rate |
|---|---|---|---|
| vLLM 0.27.1 stable (py3.13) | 4 | 0 | 0% |
| vLLM nightly 0.27.2rc1.dev109 (py3.12) | 4 | 2 | 50% |

The crash is significantly more frequent on the vLLM nightly cell, though CI history shows both cells fail at ~1-in-3 over time.

This PR marks the test with `pytest.mark.xfail(strict=False)` rather than skipping it. `strict=False` means:
- A crash (expected failure) does not break the suite
- A pass (`xpass`) is allowed and reported — providing a free signal for when the issue resolves upstream
- The test continues to run on every weekly CI invocation

When `xpass` starts appearing consistently in CI, the marker should be removed.

The smoke test (`test_mtp_finetuning_smoke` — 0.8B model, 50 samples) still runs on every commit and covers the full MTP pipeline end-to-end.

Failing CI run: https://github.com/neuralmagic/llm-compressor-testing/actions/runs/31856577844

## Tests

Verified on both CI-matching local environments (dedicated 2×H100 80GB):

| Cell | Python | vLLM | Collected as xfail | Suite integrity |
|---|---|---|---|---|
| A | 3.13.15 | 0.27.1 (stable) | Yes | 982 tests, 0 collection errors |
| B | 3.12.14 | 0.27.2rc1.dev109 (nightly) | Yes | 982 tests, 0 collection errors |

```
$ CADENCE=weekly pytest tests/e2e/regression/test_mtp_online_acceptance.py --collect-only -m xfail -q
========================== 1 test collected in 0.19s ===========================
```

`make quality` clean (no new warnings).

## Checklist

I have filled in:

### DIFF
--- a/tests/e2e/regression/test_mtp_online_acceptance.py
+++ b/tests/e2e/regression/test_mtp_online_acceptance.py
@@ -29,6 +29,7 @@ TRAINING_DATA_REPO = "inference-optimization/Qwen3.5-4B-responses"
 @requires_cuda
 @requires_transformers_version("5.2.0")
 @requires_vllm_version("0.22.0")
+@pytest.mark.xfail(strict=False, reason="Flaky: CUDA launch failure mid-training")
 def test_mtp_online_regression(
     tmp_path: Path,
     prompts: list[list[dict[str, str]]],


rahul-tuli:
Weekly test run on CI: https://github.com/neuralmagic/llm-compressor-testing/actions/runs/32124391151

rahul-tuli:
Tracked in JIRA: https://issues.redhat.com/browse/INFERENG-9932

**JIRA Details:**
- Issue Type: Bug
- Priority: Critical
- Component: Speculators
- Status: Backlog

orestis-z:
Sorry for the duplication, should have commented here in the first place 😅

Would it be possible to relax the condition on nightly only? nighly will always be unstable occasionally and create "failure noise", but we want to always stay strict against the stable vllm release (latest).

rahul-tuli:
> Sorry for the duplication, should have commented here in the first place 😅
> 
> Would it be possible to relax the condition on nightly only? nighly will always be unstable occasionally and create "failure noise", but we want to always stay strict against the stable vllm release (latest).

The vllm stable pathway is also flaky, (it more stable than vllm nightly but still fails intermittently), I believe the right fix would be completely rewrite the forward loop of MTP from the ground up, the data generation cleanups we've been doing will also help alleviate some of the intermittent failures.

orestis-z (approved):

PR #1012, opened 2026-08-19
Keep base dependencies in pyproject.toml instead of setup.py

dsikka:
dependencies still has to be dynamic (hs-connectors' constraint is resolved at build time based on SPECULATORS_BUILD_TYPE), but the static base dependency list no longer needs to live as a Python literal in setup.py. It's now a real TOML array under
[tool.speculators.dependencies].base in pyproject.toml, and setup.py just reads it back and appends the computed hs-connectors requirement.

## Checklist

I have filled in:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -175,6 +175,23 @@ cython_debug/
 # Visual Studio Code
 .vscode/
 
-# hs_connectors build version files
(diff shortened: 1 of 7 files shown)


dsikka:
Closing as integrated by Dan

PR #1016, opened 2026-08-20
[Testing] Enable per commit smoke testing

dsikka:
## Purpose
- Enable e2e smoke tests as part of the per-commit testing. This test will run all e2e smoke test using Python 3.13
- The tests are blocked by default and require the user to manually unblock, based on the content of the PR 

## Tests
- Test flow: https://buildkite.com/vllm/speculators/builds/1305

## Checklist
- Need someone from midstream / Deepak to review before landing

I have filled in:

- [x ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [x ] I (a human) have written or reviewed the code in this pr to the best of my ability.

### DIFF
--- a/.buildkite/gpu-tests/gpu-tests-L4.yml
+++ b/.buildkite/gpu-tests/gpu-tests-L4.yml
@@ -163,6 +163,11 @@ steps:
                     env:
                       - name: PYTHON_VERSION
                         value: "{{matrix.python}}"
+                      - name: HF_TOKEN
+                        valueFrom:
+                          secretKeyRef:
+                            name: hf-token-secret
+                            key: token
 
                 volumes:
                   - name: model-cache
--- a/.buildkite/gpu-tests/gpu-tests-L4.yml
+++ b/.buildkite/gpu-tests/gpu-tests-L4.yml
@@ -57,6 +57,61 @@ steps:
             python: ["3.10", "3.13"]
             transformers: ["latest"]
 
+  - group: ":test_tube: Smoke Tests (L4)"
+    steps:
+      - label: ":python: Smoke Tests py{{matrix.python}}"
+        agents:
+          queue: RedHat-L4-GCP
+
+        retry:
+          automatic:
+            - exit_status: -1
+              limit: 3
+            - exit_status: 255
+              limit: 3
+            - signal_reason: agent_stop
+              limit: 3
+
+        plugins:
+          - kubernetes:
+              podSpec:
+                nodeSelector:
+                  pool: L4solo
+
+                containers:
+                  - name: tests
+                    image: buildkite/agent:3-ubuntu
+                    command:
+                      - chmod +x .buildkite/gpu-tests/scripts/run-tests.sh
+                      - .buildkite/gpu-tests/scripts/run-tests.sh smoke
+
+                    resources:
+                      limits:
+                        nvidia.com/gpu: 1
+
+                    volumeMounts:
+                      - name: model-cache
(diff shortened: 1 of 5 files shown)


dhuangnm:
@rashmigottipati can you please review this PR? Thanks.

.buildkite/gpu-tests/scripts/run-tests.sh
```diff
@@ -31,5 +38,17 @@ if [ -n "${TRANSFORMERS_VERSION:-}" ] && [ "${TRANSFORMERS_VERSION}" != "latest"
   uv pip install "transformers${TRANSFORMERS_VERSION}"
 fi
 
+if [ "${TEST_TYPE}" = "smoke" ]; then
+  echo "--- Setting up vLLM environment"
+  uv venv vllm_venv --python "${PYTHON_VERSION}"
+  VLLM_VENV_PYTHON="$PWD/vllm_venv/bin/python"
+  UV_TORCH_BACKEND=cu130 uv pip install --python "${VLLM_VENV_PYTHON}" vllm
```
deepak-kumar-neu, line 48: lgtm! I just have one note, the smoke test installs vllm without a version pin. Should we add a constraint here to avoid any unexpected break from new vLLM releases?

.buildkite/gpu-tests/scripts/run-tests.sh
fynnsu, line 32: In the interest of keeping our log files short (so that they display fully on the buildkite page), could you add:
```
export TQDM_DISABLE=1
export HF_HUB_DISABLE_PROGRESS_BARS=1
```

to this script. 

And then maybe also add
```
> /dev/null 2>&1
``` to the `apt-get update ...` line above.

Comment (fynnsu) [reply .buildkite/gpu-tests/scripts/run-tests.sh:32]: This pr should also help a bit: https://github.com/vllm-project/speculators/pull/1040

File: .buildkite/gpu-tests/scripts/run-tests.sh (hunk 1 above)
Comment (rashmigottipati) [reply .buildkite/gpu-tests/scripts/run-tests.sh:48]: +1 to @deepak-kumar-neu's suggestion. Could we pin or constrain the vllm version here to keep the smoke-test environment reproducible and avoid unexpected CI failures from new vLLM releases?

File: .buildkite/gpu-tests/gpu-tests-L4.yml [hunk 2]
```diff
@@ -120,3 +120,64 @@ steps:
...
+                        nvidia.com/gpu: 1
+
+                    volumeMounts:
+                      - name: model-cache
+                        mountPath: /model-cache
+                        readOnly: false
+                      - name: devshm
+                        mountPath: /dev/shm
+
+                    env:
```
rashmigottipati, line 163: H100 injects `HF_TOKEN`, but the L4 smoke-test job doesn't. If the smoke tests use a gated Hugging Face model, will the L4 job fail to download it?

rashmigottipati (commented):
lgtm overall, just left a couple of comments. will re-review once they're addressed.

.buildkite/gpu-tests/scripts/run-tests.sh
dsikka: I think we want to stick with latest for testing but we can update if we find that its too noisy

shanjiaz (approved):

fynnsu:
@mergify refresh

rashmigottipati (approved):
/lgtm

PR #1026, opened 2026-08-21
feat(train): checkpoint and restore per-rank RNG state for exact resume

zihanlin-ai:
## Purpose

Mid-epoch resume restores the data position, optimizer and scheduler, but not torch RNG state. Accelerator-side anchor sampling, noise augmentation and dropout therefore diverge after resume.

Save the torch CPU and current accelerator RNG state in one per-rank file. Mid-epoch state is captured at checkpoint time; end-of-epoch state is captured after validation and any `--save-best` update. On resume, it is restored after creating the existing-epoch iterator for a mid-epoch checkpoint and before creating the next-epoch iterator at an epoch boundary. Checkpoints without an RNG file still resume with a warning.

This covers torch RNG only. Exact replay requires `--num-workers 0`, the same rank/device topology, and the remaining checkpoint guarantees discussed in #1027.

Implements #1027.

## Tests

Unit (`tests/unit/train/test_mid_epoch_resume.py`, 14 tests): snapshot format / `weights_only` load; end-of-epoch snapshot written after validation; `--save-best` checkpoints get it; a live iterator that continues vs. a fresh process that resumes mid-epoch, with a dataset that draws on every fetch; the same at an epoch boundary; per-rank files under the distributed rank gate; missing-file fallback.

Hardware, on this branch + #899 (fp32 optimizer state) + #900 (scheduler seeding) with model weights serialized as fp32 — a tiny model with dropout, AdamW, 10 steps/epoch, crash after the mid-epoch checkpoint, then resume in a fresh process and compare against the uninterrupted run:

- H20, torch 2.9.0+cu129 — single process: mid-epoch, end-of-epoch and save_best resumes bitwise identical (loss per step, final params, optimizer state); 2-rank DDP (NCCL): both ranks bitwise identical after resume, one RNG file per rank.
- Ascend A3, torch 2.9.0 + torch_npu 2.9.0.post6 — same results (HCCL for the 2-rank case).
- With the upstream default bf16 weight serialization the resumed run diverges at its first step on both devices, independent of this PR (see #1027).

```
$ python -m pytest tests/unit/train/test_mid_epoch_resume.py -q
14 passed
$ ruff check && ruff format --check && mdformat --check docs/cli/train.md && mypy --check-untyped-defs src/speculators/train/trainer.py tests/unit/train/test_mid_epoch_resume.py
All checks passed!
```

## Checklist

- [x] The purpose of the PR is described.
- [x] The test plan and results are provided.
- [x] I (a human) have written or reviewed the code in this PR to the best of my ability.


src/speculators/train/trainer.py
```diff
@@ -597,6 +670,10 @@ def maybe_save_checkpoint(self, epoch: int | str, local_step: int = 0):
             self.checkpointer.save_scheduler_state_dict(self.schedulers, epoch)
         if isinstance(epoch, int):
             self._save_training_state(epoch, local_step)
+            if local_step > 0:
+                self._save_rng_state(epoch)  # mid-epoch: resume continues from here
+            else:
+                self._epoch_checkpoint_written = True  # snapshot after validation
```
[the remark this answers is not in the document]
zihanlin-ai: Good catch — real gap. Fixed in ec4b75e: both deferral sites (`maybe_save_checkpoint`'s end-of-epoch branch and the `save_best` path) now unlink this rank's RNG file before setting the deferred-snapshot flag, so a resume interrupted during validation falls back to the documented no-RNG warning instead of silently replaying mid-epoch draws. Regression test added (`test_deferred_epoch_snapshot_invalidates_stale_mid_epoch_rng`).

src/speculators/train/trainer.py
```diff
@@ -218,6 +246,48 @@ def _load_training_state(self) -> dict:
...
+            self._save_rng_state(epoch)
+
+    def _epoch_iterator(self, skip_steps: int):
+        """Restore before a new epoch's iterator, or after a resumed one."""
+        if skip_steps == 0:
+            self._maybe_restore_rng_state()
+        loader_iter = iter(self.train_loader)
+        if skip_steps > 0:
+            self._maybe_restore_rng_state()
+        return loader_iter
```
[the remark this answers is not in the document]
zihanlin-ai: Confirmed, and it ran deeper than the RNG ordering: the fallback path had no drain at all — the loop retrained the already-trained batches under shifted `local_step` numbering, with the snapshot applied before the replay. Fixed in 8f45360: `_epoch_iterator` now fetches and discards the skipped batches first (their `__getitem__` calls consume random draws exactly as the original epoch did), then applies the snapshot; training never sees the drained batches and `local_step` stays aligned. Fast-skip samplers keep the existing order. Regression test added with `_NoisyDataset` over a plain `DataLoader` (no fast-skip API), and the test mocks now route through the real `_prepare_resume_skip`.

src/speculators/train/trainer.py
```diff
@@ -218,6 +249,69 @@ def _load_training_state(self) -> dict:
...
+                # num_workers=0 their __getitem__ calls consume random draws
+                # exactly as the original epoch did, so the snapshot must be
+                # applied only after the drain.
+                self._replay_skip_steps = False
+                for _ in range(skip_steps):
+                    next(loader_iter, None)
+            self._maybe_restore_rng_state()
+        return loader_iter
+
+    def _load_rng_state(self) -> dict | None:
```
sunny-infra: `_load_rng_state` calls `torch.load(p, map_location="cpu", weights_only=True)` without any `try/except`. If the RNG state file gets corrupted (disk error, interrupted write, partial flush, etc.), the exception bubbles up to `setup_trainer` and kills training startup.

But in the same file, `_load_training_state` wraps `json.loads` in a `try/except` and falls back with a warning. The two loaders are inconsistent in robustness.

**Suggestion**: Wrap `torch.load` in a `try/except`. On failure, log a warning and return `None` (matching the file‑missing fallback), so it behaves like `_load_training_state`.

src/speculators/train/trainer.py
```diff
@@ -218,6 +249,69 @@ def _load_training_state(self) -> dict:
...
+        """
+        self._rng_state_path(epoch).unlink(missing_ok=True)
+
+    def _save_end_of_epoch_rng_state(self, epoch: int) -> None:
+        """Snapshot after validation for either end-of-epoch save path."""
+        if self._epoch_checkpoint_written:
+            self._epoch_checkpoint_written = False
+            self._save_rng_state(epoch)
+
+    def _epoch_iterator(self, skip_steps: int):
```
sunny-infra: The method signature `_epoch_iterator(self, skip_steps: int)` has no return type. The project runs mypy with `--check-untyped-defs` (mentioned in the PR description), and neighbouring methods do have annotations.

**Suggestion**: Add a return annotation, e.g. `-> Iterator` (or the concrete iterator type), for consistency and to satisfy mypy's strict checks.

src/speculators/train/trainer.py
```diff
@@ -432,9 +528,11 @@ def train_epoch(self, epoch: int):
         # Determine how many batches to skip for mid-epoch resume.
         skip_steps = self._prepare_resume_skip(epoch)
 
-        train_loader = self.train_loader
+        train_loader = self._epoch_iterator(skip_steps)
         if self.rank == 0:
-            train_loader = tqdm(train_loader, desc=f"Epoch {epoch}")  # type: ignore[assignment]
+            train_loader = tqdm(  # type: ignore[assignment]
```
sunny-infra: The progress bar sets `total` to the full epoch length. On a mid‑epoch resume, the iterator only yields `len(loader) - skip_steps` batches, so the bar never reaches 100% – users may think training is stuck. For non‑resume runs (`skip_steps == 0`) this is fine.

**Suggestion**: On resume, set `total = len(self.train_loader) - skip_steps`, or pass `initial=skip_steps` so the bar starts from the already‑trained position. Keep `total = len(...)` only when `skip_steps == 0`.

src/speculators/train/trainer.py
```diff
@@ -163,7 +163,38 @@ def _resolve_scheduler_steps(
...
+    torch.set_rng_state(state["torch_cpu"])
+    if state["device"] is not None:
+        torch.get_device_module(state["device_type"]).set_rng_state(
+            state["device"], state["device_index"]
+        )
+
+
 class Trainer:
+    # Loaded on resume and applied once by _maybe_restore_rng_state().
+    _pending_rng_state: dict | None = None
```
sunny-infra: `_pending_rng_state`, `_epoch_checkpoint_written`, and `_replay_skip_steps` are declared as class variables. Although Python's attribute lookup means `self.x = ...` shadows them per instance (so no cross‑instance leakage actually happens), the code reads like shared state and is stylistically inconsistent with sibling attributes like `self._resume_local_step` which are assigned as instance attributes inside `setup_trainer`.

**Suggestion**: Initialize these three as instance attributes in `__init__` (e.g., `self._pending_rng_state = None`), and keep only the type annotations at class level if you still want them.

PR #1028, opened 2026-08-21
feat(hs-connectors): add FP8-quantized hidden-states backend

shubhra:
## Summary

Revives the intent of #491 (FP8HiddenStatesConnector), which predates two major refactors of the hidden-states transfer layer (`save_kv_layer` -> async `_write_tensors` hook, and the file/mooncake `HiddenStatesBackend` plugin abstraction) and can no longer be rebased as-is.

Adds a new `fp8` backend following the same plugin pattern as the existing file/mooncake backends:

- `FP8HiddenStatesConnector` overrides only vLLM's `ExampleHiddenStatesConnector._write_tensors` staticmethod to quantize hidden states to `float8_e4m3fn` with **per-token scaling** (same granularity as the original #491 design) before writing to safetensors. All scheduler-side bookkeeping, async DtoH copy, and file-locking is inherited unchanged.
- `FP8Transfer` (a `FileTransfer` subclass) transparently dequantizes on read, so the training pipeline consumes bf16/fp32 tensors identically regardless of backend.
- New unit tests (`test_fp8_utils.py`, `test_fp8_transfer.py`), backend-args roundtrip coverage, and an e2e roundtrip test (`test_fp8_roundtrip.py`).

## Test plan / results

29/29 unit tests pass locally (`test_fp8_utils.py`, `test_fp8_transfer.py`, `test_backend_args_roundtrip.py`).

Beyond unit tests, ran a full bf16-vs-FP8 ablation: trained `Qwen/Qwen3-8B` speculators (eagle3, dflash, dspark) on 5K magpie + 5K ultrachat samples (from `inference-optimization/Dataset-Qwen3-235B-Instruct`), once with the existing bf16 file backend and once with this FP8 backend, then compared val_loss and ran `guidellm` throughput evals (weighted across all 9 `RedHatAI/speculator_benchmarks` subsets) for both.

`acceptance_length` = tokens actually produced per verification round (drives wall-clock speedup); `acceptance_rate` = fraction of individually proposed tokens accepted (isolates draft-model quality from block size). Format: `len / rate`.

| Model  | Precision | HumanEval | math_reasoning | summarization | Unweighted mean (9 subsets) |
|--------|-----------|-----------|-----------------|----------------|------------------------------|
| eagle3 | bf16      | 2.2226 / 0.4075 | 2.3401 / 0.4467 | 1.7160 / 0.2387 | 2.0044 / 0.3348 |
| eagle3 | fp8       | 2.2070 / 0.4023 | 2.3280 / 0.4427 | 1.7079 / 0.2360 | 1.9940 / 0.3313 |
| dflash | bf16      | 2.2238 / 0.0816 | 2.4191 / 0.0946 | 1.6739 / 0.0449 | 1.9848 / 0.0657 |
| dflash | fp8       | 2.2453 / 0.0830 | 2.4494 / 0.0966 | 1.6664 / 0.0444 | 1.9867 / 0.0658 |
| dspark | bf16      | 2.4703 / 0.1838 | 2.6817 / 0.2102 | 1.7660 / 0.0958 | 2.1365 / 0.1421 |
| dspark | fp8       | 2.4663 / 0.1833 | 2.6688 / 0.2086 | 1.7427 / 0.0928 | 2.1301 / 0.1413 |

bf16-vs-fp8 gap is within run-to-run noise on both `len` and `rate`, for every subset and every architecture — no measurable quality regression from FP8 hidden-states transfer.

Per-subset breakdown for all 6 runs (one row per `RedHatAI/speculator_benchmarks` subset) is committed in [`docs/fp8_ablation/`](https://github.com/vllm-project/speculators/tree/feat/fp8-hidden-states-connector-v2/docs/fp8_ablation) in this branch — `{model}_{precision}_acceptance.csv` plus `RESULTS_SUMMARY.md`.

### Storage/write-speed numbers for the hidden-states transfer itself

The table above is quality-parity (does FP8 hurt the trained speculator?). Separately measured the thing FP8 actually speeds up — the write path:

**Real data-gen run** (eagle3 config, 300 samples via `scripts/data_generation_offline.py`, otherwise-identical vLLM server, `--concurrency 32`):

| Backend | Total on-disk size (300 files) | Throughput       | avg vLLM request | avg file write |
| ------- | ------------------------------- | ----------------- | ------------------ | ---------------- |
| bf16    | 14.24 GB                        | 25.6 samples/s     | 1114 ms             | 2 ms              |
| fp8     | 7.12 GB (**exactly 50.0%**)     | 25.5 samples/s     | 1115 ms             | 3 ms              |

End-to-end throughput is identical within noise — the write (2-3 ms) is ~3 orders of magnitude smaller than per-sample GPU generation time (~1.1 s) and runs async off the critical path, so the 50% smaller payload doesn't show up as a local generation-throughput win. The real benefit is disk footprint and any bandwidth-constrained transfer (e.g. the `mooncake` backend, or a slower/network filesystem).

Isolated CPU-only microbenchmark (synthetic tensors, same shapes, tmpfs) confirms the same exact 50% size reduction at every scale (128–8192 token chunks), with the quantization compute roughly a wash to ~2x slower than a plain write at small/medium chunk sizes on fast local storage, netting out as a write-time win only at very large single writes. Full numbers in `docs/fp8_ablation/RESULTS_SUMMARY.md`.

## Notes

- Quantization granularity kept identical to the original #491 design (per-token scale).

Made with [Cursor](https://cursor.com)


fynnsu:
@shubhra I pushed a commit so that we can use the same `--hidden-states-path` instead of needing a separate `--fp8-hidden-states-path` for the connector.

docs/fp8_ablation/eagle3_fp8_acceptance.csv
fynnsu, line 1: I'm not sure we want to be publishing these csv files to the repo.

docs/fp8_ablation/RESULTS_SUMMARY.md
fynnsu, line 1: I'm not sure if we want this either, but if we do it should be added properly into the `docs/.nav.yml` file so that it ends up in the nav bar / sidebar.

If we do decide to include stuff like this we may want a new section/page on the docs for Research/Ablation results.

fynnsu (commented):
Generally looks good but left a couple comments regarding the docs.

Also it seems like we get good acceptance rates but do you have any results on training speed/data usage. Presumably for offline/hybrid training we can expect data that is roughly half the size, but is there any speed up (either for offline or online)?

docs/fp8_ablation/eagle3_fp8_acceptance.csv
shubhra: Yes, I'll remove the cvs files. I should've mentioned that these are more for us to look at briefly but will remove before merging. Let me see about the speedup numbers as well.

fynnsu (approved):
LGTM! Thanks

PR #1029, opened 2026-08-23
fix: reject duplicate or negative target_layer_ids, fail clearly on tiny verifiers

zihanlin-ai:
## Purpose

`resolve_target_layer_ids()` returns explicit ids unvalidated and, when unset, uses `[2, n // 2, n - 3]`, which repeats an id for verifiers with 4–6 layers (`4 → [2, 2, 1]`, `5 → [2, 2, 2]`, `6 → [2, 3, 3]`); `launch_vllm.py` builds its extraction default from the same formula. The extraction side deduplicates the ids while training sizes its hidden-state input by `len(target_layer_ids)`, so the mismatch only surfaces as a buffer shape error far from its cause. Real verifiers (≥ 7 layers) are unaffected.

We hit it running the end-to-end pipeline on a single device with a synthetic 4-layer verifier instead of Qwen3-0.6B: the run died in the extraction engine with nothing pointing at layer ids. This raises a clear error for duplicate explicit ids and colliding defaults in `resolve_target_layer_ids()`, and adds one shared non-negative/distinct check in `launch_vllm.py` after either branch so extraction fails before writing data; upper-bound checks against the verifier are left to #569.

## Tests

```
$ python -m pytest tests/unit/models/test_target_layer_ids.py tests/unit/scripts/test_launch_vllm_layer_ids.py tests/unit/models/test_dflash_targets.py tests/unit/test_config.py -q
48 passed
$ ruff check && ruff format --check && mypy --check-untyped-defs src/speculators/models/utils.py tests/unit/models/test_target_layer_ids.py tests/unit/scripts/test_launch_vllm_layer_ids.py
All checks passed!
```

## Checklist

- [x] The purpose of the PR is described.
- [x] The test plan and results are provided.
- [x] I (a human) have written or reviewed the code in this PR to the best of my ability.

### DIFF
--- a/scripts/launch_vllm.py
+++ b/scripts/launch_vllm.py
@@ -132,10 +132,16 @@ def main():
             num_hidden_layers - 3,
             num_hidden_layers,
         ]
-    if min(target_layer_ids) < 0 or len(set(target_layer_ids)) != len(target_layer_ids):
+    # Layer id ``num_hidden_layers`` (the final hidden state) is valid: the
+    # default above and --include-last-layer both emit it.
+    if (
+        min(target_layer_ids) < 0
+        or max(target_layer_ids) > num_hidden_layers
+        or len(set(target_layer_ids)) != len(target_layer_ids)
+    ):
         raise ValueError(
-            f"Invalid target layer ids {target_layer_ids}; "
-            "ids must be non-negative and distinct."
+            f"Invalid target layer ids {target_layer_ids}; ids must be "
+            f"distinct and within [0, {num_hidden_layers}]."
         )
 
     speculative_config = {
--- a/scripts/launch_vllm.py
+++ b/scripts/launch_vllm.py
@@ -132,6 +132,11 @@ def main():
             num_hidden_layers - 3,
             num_hidden_layers,
         ]
+    if min(target_layer_ids) < 0 or len(set(target_layer_ids)) != len(target_layer_ids):
+        raise ValueError(
+            f"Invalid target layer ids {target_layer_ids}; "
+            "ids must be non-negative and distinct."
+        )
 
     speculative_config = {
         "method": "extract_hidden_states",
--- a/src/speculators/models/utils.py
+++ b/src/speculators/models/utils.py
@@ -42,27 +42,42 @@ def resolve_target_layer_ids(
     verifier_name_or_path: str,
     trust_remote_code: bool = False,
 ) -> list[int]:
-    if target_layer_ids is not None:
-        if len(set(target_layer_ids)) != len(target_layer_ids):
-            raise ValueError(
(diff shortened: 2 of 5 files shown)


scripts/launch_vllm.py
```diff
@@ -132,6 +132,11 @@ def main():
             num_hidden_layers - 3,
             num_hidden_layers,
         ]
+    if min(target_layer_ids) < 0 or len(set(target_layer_ids)) != len(target_layer_ids):
```
fynnsu: Hmm maybe we should check if `max(target_layer_ids) > num_hidden_layers` too?

src/speculators/models/utils.py
```diff
@@ -43,13 +43,22 @@ def resolve_target_layer_ids(
     trust_remote_code: bool = False,
 ) -> list[int]:
     if target_layer_ids is not None:
+        if len(set(target_layer_ids)) != len(target_layer_ids):
```
fynnsu: Same here (bounds check). Also maybe we can combine this check with the one below, by moving the if/else block.

fynnsu (commented):
Thanks for the pr! Will be good to make this a little bit more robust.

scripts/launch_vllm.py
zihanlin-ai: Added in 7471f58 — used `> num_hidden_layers` since `num_hidden_layers` itself is a valid id (the default and `--include-last-layer` both emit it).

src/speculators/models/utils.py
zihanlin-ai: Done in 7471f58 — merged into one check by fetching `num_layers` first. Note this means explicit ids now also trigger a verifier config load (needed for the bounds check); the `from_training_args` callers load the verifier right after anyway. Also gave `test_from_training_args_loads_vocab_mappings` explicit ids: its 6-layer verifier hits the default `[2, 3, 3]` collapse, which is what broke buildkite.

zihanlin-ai:
Both review points addressed in 7471f58, plus the buildkite fix (the new validation caught `test_from_training_args_loads_vocab_mappings` resolving to duplicate ids `[2, 3, 3]` on its 6-layer verifier — arguably the check doing its job).

Verified: full `tests/unit` green on transformers 5.8.1 (x86 + H100-class GPU) and 4.57.6 (aarch64 + Ascend NPU).

One unrelated find while testing: `test_distributed_resume[ddp/fsdp]` fails on any 2+ GPU machine with `KeyError: initial_lr` on scheduler resume (torch 2.11) — reproduces on unmodified `main`, invisible in CI since single-GPU runners skip it. Happy to file separately.

fynnsu (approved):

PR #1030, opened 2026-08-23
fix(train): make a zero-std noise transform RNG-neutral

zihanlin-ai:
## Purpose

`AddGaussianNoise` / `AddUniformNoise` with `std=0` still sample a full noise tensor and multiply it by zero: the values are unchanged, but the torch generator advances on every batch. `create_train_val_loaders` always installs the transform, so a run with `--noise-std 0` consumes RNG differently from one with noise on, and the two cannot be compared as exact replays (e.g. when checking that disabling noise is the only difference between two runs).

How we noticed: while profiling the collate path of a downstream online hidden-state pipeline with noise disabled, every row was still paying a full `randn_like` over the `[seq, num_layers × hidden]` tensor, and the generator kept advancing — both wasted work in the loader and a silent difference between noise-on and noise-off runs. Not a failure, a profiling find.

Return the data untouched when `std == 0` — a 4-line change in the base class, behaviour for `std > 0` unchanged.

## Tests

`tests/unit/train/test_noise_transforms.py`: for both transforms, `std=0` returns identical data and leaves `torch.get_rng_state()` unchanged; `std>0` still adds noise.

```
$ python -m pytest tests/unit/train/test_noise_transforms.py -q
4 passed
$ ruff check && ruff format --check && mypy --check-untyped-defs src/speculators/train/noise_transforms.py tests/unit/train/test_noise_transforms.py
All checks passed!
```

## Checklist

I have filled in:

### DIFF
--- a/src/speculators/train/noise_transforms.py
+++ b/src/speculators/train/noise_transforms.py
@@ -7,6 +7,10 @@ class TransformTensors:
         self.std = std
 
     def __call__(self, data):
+        if self.std == 0:
+            # Disabled noise must not consume RNG: sampling and multiplying by
+            # zero leaves the values unchanged but advances the generator.
+            return data
         for tensor in self.tensors:
             data[tensor] = self.transform(data[tensor])
         return data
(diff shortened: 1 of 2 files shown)


zihanlin-ai:
Closing: on reflection this is a minor perf skip (the dead randn_like when noise is disabled) rather than a correctness fix, and it would change the RNG stream of noise-off runs relative to noise-on ones. Not worth review time as is.

PR #1032, opened 2026-08-24
fix(dflash): unify per-block accepted-length metrics

WindChimeRan:
## Summary

This PR gives the DFlash family one definition for each accepted-length metric:

- `eal`: realized greedy accepted length per speculative block, including the verifier's bonus token;
- `accept_len`: analytical TV-overlap accepted length, retained where the unary TV tensor is already computed.

DFlash, DSpark, and DFlash2 now use the same block shape and bonus-token convention. Each metric is exported as an additive `*_sum` / `*_total` pair, so it pools correctly across micro-batches and distributed ranks.

## Why EAL must be counted per block

The old DFlash calculation multiplied per-position marginal accuracies inside each micro-batch and exported the result with a total of `1.0`. That has two problems:

- averaging those products gives batch composition equal weight rather than blocks;
- multiplying marginal position accuracies loses the within-block prefix correlation that determines speculative acceptance.

For accepted runs of 3/2/1/0 grouped as (3, 1) and (2, 0), the true mean drafted length is 1.5. The old per-batch products average to 1.25, while a product formed from globally pooled marginals gives 1.219. Counting the leading correct run in each block gives the exact 1.5 and remains additive.

`normalize_counted_metrics` is unchanged. It remains responsible only for normalizing already-reduced counts; no algorithm-specific pre/post-reduce hook is needed.

### Greedy `eal`

For every valid block, `eal` counts the leading run of correct greedy proposals and adds the verifier's always-emitted bonus token.

- DFlash uses the unary argmax path.
- DSpark uses its Markov-corrected argmax path.
- DFlash2 uses the realized selector path, not the unary argmax.

DFlash2's selector is trained with teacher-forced predecessor tokens. Those predecessor tokens are exact while an accepted greedy prefix remains alive. The metric also gates correctness on `contains_target`, so a target injected only to make the training candidate set well-defined cannot be counted as available during serving.

### TV `accept_len`

`accept_len` keeps DSpark's analytical cumulative TV-overlap estimator. It answers the rejection-sampling/temperature-above-zero question and reuses the per-position overlap tensor needed by DSpark's confidence metrics. It uses the same block shape and includes the same bonus token as `eal`.

## DFlash2 metric composition and follow-up refactor

DFlash2 currently imports DSpark's top-level `compute_metrics` function to obtain unary loss, TV, accuracy, and accepted-length metrics, disables the confidence head, and then appends selector-specific metrics. This reuse is convenient, but the ownership is unclear: a sibling algorithm implicitly controls DFlash2's metric dictionary, and changes to DSpark can silently alter DFlash2 behavior.

This PR stays focused by preserving that existing composition while explicitly replacing the inherited unary `eal` with selector-path `eal`.

A follow-up refactor should extract shared DFlash-family primitives for loss counts, greedy prefix length, and TV prefix length. DFlash, DSpark, and DFlash2 can then compose those primitives directly, removing the DFlash2-to-DSpark top-level dependency without changing the metric names or definitions established here.

## Behavior changes

- `eal` is now an exact mean of per-block greedy accepted lengths rather than a product of batch-level marginal accuracies.
- `eal` includes the verifier's bonus token, matching the convention already used by `accept_len`.
- DFlash2 `eal` measures its selector path rather than its unary argmax.
- Existing `eal_epoch` values are therefore not comparable with runs produced before this PR.

## Validation

- 205 model unit tests pass.
- Focused DFlash, DSpark, and DFlash2 coverage passes.
- Ruff lint, formatting, and Git whitespace checks pass.

### DIFF
--- a/src/speculators/models/dflash/metrics.py
+++ b/src/speculators/models/dflash/metrics.py
@@ -12,7 +12,10 @@ from speculators.losses import (
     dpace_loss_decay,
     kl_div_loss,
 )
-from speculators.models.metrics import compute_accuracy_multi_step
+from speculators.models.metrics import (
+    compute_accepted_length_counts,
+    compute_accuracy_multi_step,
+)
 
 _DEFAULT_LOSS_CONFIG: LossConfig = {"kl_div": (kl_div_loss, 1.0)}
 
@@ -45,7 +48,8 @@ def compute_metrics(
             - loss: Scalar loss value
             - full_acc: Overall accuracy
             - position {i} acc: Accuracy at position i within blocks
-            - eal: Expected Accepted Length (headline speculative-decoding metric)
+            - eal: Expected Accepted Length, the mean per-block accepted run
+              plus the verifier's bonus token (headline metric)
     """
     if loss_config is None:
         loss_config = _DEFAULT_LOSS_CONFIG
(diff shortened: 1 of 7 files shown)


fynnsu (changes requested):
While I agree this is more accurate, putting this calculation in `utils.py:normalize_counted_metrics` is the wrong place for it. This function should be just about correcting the aggregated metrics from different ranks.

What I'd suggest instead is having a `compute_metrics_pre_reduce` `compute_metrics_post_reduce` or similar for dflash. `compute_metrics_pre_reduce` would mostly align with the current `compute_metrics` function, whereas `compute_metrics_post_reduce` would only run on rank0 and handle this kind of logic. 

I'm opening to better names or other approaches, but we should make sure this logic ends up in a reasonable location.

WindChimeRan:
@fynnsu 

Agreed. New version: 

Instead of a pre/post-reduce pair,  makes the metric additive so there's nothing left to post-process: EAL is counted per block and exported as a plain `eal_sum` / `eal_total` pair, so it pools through the existing contract.

Note `eal` now includes the bonus token, so `eal_epoch` won't match older runs. (align with vllm convention). 

I also found dflash2 has some weird dependency to dspark. Will make followup PR to clean it up.

fynnsu (approved):
Thanks @WindChimeRan, this is a lot better!

Two comments:
1. I find the metric naming `eal` and `accept_len` pretty unclear. Ideally we'd rename these, maybe to "greedy_accept_len", and "expected_accept_len" or "rejection_sampling_accept_len" to make their usage more clear.
2. If you wanted to extend this further, we could also extract the `accept_len` calculations out of dspark and apply them to all of the `dflash` family algs.

PR #1033, opened 2026-08-24
fix(train): default DSpark drafts to five layers

WindChimeRan:
Default `--num-layers` to 5 for DFlash, DSpark, and DFlash2; leave all other defaults unchanged.


shanjiaz (approved):
Thanks Ranran!

PR #1035, opened 2026-08-24
[feat]: Log number of broken records

### LINKED ISSUE #1034: [Feature request] Log number of datasets records on which fetch failed
Sometimes the fetch process fails (for example, like in this issue, and there are no way to observe it except for reading through logs.  Sometimes it is not even a Speculators bug, like in this [issue](https://github.com/vllm-project/vllm/issues/44485). 

I would like to see number of records among other graphs on Tensorboard or similar experiment tracking system.

pavelgein:
## Purpose
Log number of broken records. Closes  #1034

## Tests
I have run and see scalars at the experiment tracking system

## Checklist

I have filled in:

### DIFF
--- a/src/speculators/train/data.py
+++ b/src/speculators/train/data.py
@@ -320,13 +320,27 @@ class CollateFn:
         self.dtype = dtype
         self.preprocess = preprocess
 
+    def _clean_batch(self, batch: Sequence[BatchType | None]) -> tuple[list[BatchType], int]:
+        """ Apply per-sample preprocessing and filter failed samples """
+        preprocess = self.preprocess
+
+        none_records = 0
+        new_batch = []
+        for item in batch:
+            if item is None:
+                none_records += 1
+                continue
+            if preprocess:
+                item = preprocess(item)
+            new_batch.append(item)
+
+        return new_batch, none_records
+
     def __call__(self, batch: Sequence[BatchType | None]) -> BatchType:
         max_len = self.max_len
         dtype = self.dtype
-        preprocess = self.preprocess
 
-        # Apply per-sample preprocessing and filter failed samples
-        batch = [preprocess(b) if preprocess else b for b in batch if b is not None]
+        batch, num_dropped = self._clean_batch(batch)
 
         if not batch:
             # Create empty sample which then gets padded to full
@@ -337,8 +351,8 @@ class CollateFn:
             empty = create_empty_sample(
                 self.hidden_size, self.num_target_layers, dtype=dtype
             )
-            if preprocess:
-                empty = preprocess(empty)
+            if self.preprocess:
+                empty = self.preprocess(empty)
             batch = [empty]
 
         collated_data = {}
@@ -389,5 +403,7 @@ class CollateFn:
         ).unsqueeze(0)
         # shape: [1, max_len]
         collated_data["document_ids"] = document_ids
+
+        collated_data["error_records"] = num_dropped
 
         return collated_data
--- a/src/speculators/train/data.py
+++ b/src/speculators/train/data.py
@@ -320,8 +320,10 @@ class CollateFn:
         self.dtype = dtype
         self.preprocess = preprocess
 
-    def _clean_batch(self, batch: Sequence[BatchType | None]) -> tuple[list[BatchType], int]:
-        """ Apply per-sample preprocessing and filter failed samples """
+    def _clean_batch(
+        self, batch: Sequence[BatchType | None]
+    ) -> tuple[list[BatchType], int]:
+        """Apply per-sample preprocessing and filter failed samples"""
         preprocess = self.preprocess
 
         none_records = 0
@@ -330,9 +332,8 @@ class CollateFn:
             if item is None:
                 none_records += 1
                 continue
-            if preprocess:
-                item = preprocess(item)
-            new_batch.append(item)
+
+            new_batch.append(preprocess(item) if preprocess else item)
 
         return new_batch, none_records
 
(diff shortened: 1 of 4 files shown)


src/speculators/train/trainer.py
```diff
@@ -469,6 +469,9 @@ def train_epoch(self, epoch: int):
             loss.backward()
             torch.nn.utils.clip_grad_norm_(self.model.parameters(), 1.0)
 
+            metrics["error_records_sum"] = torch.tensor(batch["error_records"], dtype=torch.int32, device=loss.device)
+            metrics["error_records_total"] = torch.tensor(1.0 if (self.is_distributed and self.rank == 0) else 0, device=loss.device)
```
fynnsu: I think we want
`torch.tensor(1.0 if self.rank == 0 else 0, device=loss.device)`

Otherwise the non-distributed case might have a division by zero.

fynnsu (approved):
Approved but please correct issue mentioned below.

src/speculators/train/trainer.py
pavelgein: Thank you for the review, fixed it and style

src/speculators/train/data.py
```diff
@@ -390,4 +405,6 @@ def __call__(self, batch: Sequence[BatchType | None]) -> BatchType:
         # shape: [1, max_len]
         collated_data["document_ids"] = document_ids
 
+        collated_data["error_records"] = num_dropped
```
fynnsu, line 408: Looks like we actually need this to be a tensor at this stage. (Everything coming out of the collation function should be a tensor).

fynnsu (commented):
The tests are failing, please see below.

src/speculators/train/data.py
pavelgein: Well, it's interesting, because it successfully running. And the [code](https://github.com/vllm-project/speculators/blob/main/src/speculators/train/trainer.py#L546-L551) is written in the way to handle non-tensor items in the batch.

It seems that the problem is with the test `tests/unit/train/test_data.py::test_collate_fn_basic` itself, because it expects that all items in a batch are tensors. 
So I suggest that I fix the test.

fynnsu: Yes, however it's not just the one unit test that's failing. There's also a bunch of integration tests https://buildkite.com/vllm/speculators/builds/1315/list?jid=01a03e69-428c-4c12-ab93-b4c7342a0e9e&tab=output.

These mostly all use the same batch setup logic, so you can certainly fix it there. Either way works.

pavelgein: Yes, this integration tests lacks filtering logic from Trainer class

pavelgein: Fixed in [ce604b7](https://github.com/vllm-project/speculators/pull/1035/commits/ce604b7dda2f3bfdac3f9cd39c707217388acae6)

tests/integration/conftest.py
```diff
@@ -337,7 +337,8 @@ def make_batch(
         preprocess=preprocess,
     )
     batch = collate_fn(samples)
-    return {k: v.to(device) for k, v in batch.items()}
+    # We drop all non-tensors items, to mimic logic from Trainer class, see gpu_batch formation
+    return {k: v.to(device) for k, v in batch.items() if isinstance(v, torch.Tensor)}
```
fynnsu: The Trainer class doesn't drop the keys, they just only move them `.to(device)` if they're tensors:
```
gpu_batch = {
        k: v.to(self.local_rank, non_blocking=True)
        if isinstance(v, torch.Tensor)
        else v
        for k, v in batch.items()
    }
```

pavelgein: Do you mean, that comment is misleading or fix should be modified?

fynnsu: I think the fix should be modified. This line is currently filtering out non-tensors, but we should make the logic match the trainer class logic and just move the tensors to device while keeping the non-tensors.

pavelgein: But trainer class drops the keys in the sense that they will not be passed to model's forward, since not all keys from `batch` will be in `gpu_batch` (and gpu_batch will be passed to model's forward). 

As a solution, I can patch all tests that uses make_batch that they will pass only tensors to model's forward

pavelgein: As far as I can see, all the uses of make_batch directly passed to model's forward.
IMHO, if this "dropping" logic will be copied to all tests (well, there are not many of them), it will just decrease readability. But you're the maintainer, I will adjust code

fynnsu: Sorry, maybe I'm misunderstanding something but I'm pretty sure all the keys in `batch` end up in `gpu_batch` from this logic: https://github.com/vllm-project/speculators/blob/331a5f887f3a70bade6129fb7a5865f498f37d45/src/speculators/train/trainer.py#L546-L551

The way I understand this is if we have a batch with two kv pairs `{"key1": tensor([5]), "key2": 0}` where one is a tensor and the other is a non-tensor (in this case an `int`), we'll get a `gpu_batch` of `{"key1": tensor([5], device="cuda:0"), "key2": 0}`. i.e. the tensor will be copied to the gpu, and the non-tensor values will be left as is. And then `gpu_batch` gets passed directly to the model function.

In the version of the dict comprehension you have in conftest you'll instead get `{"key1": tensor([5])}`, with `key2` filtered out.

I just want the test batch creation logic to match the trainer logic, so that we're always keeping all of the keys.

pavelgein: Yeah, you are right and I am not, I don't know how I missed the else clause several times in a row.

pavelgein: Fixed in 85e38c9, thanks again

fynnsu (approved):
Thanks!

PR #1024, opened 2026-08-21
feat(cli): unified CLI surface

rahul-tuli:
## Purpose

Implements [CLI RFC #906](https://github.com/vllm-project/speculators/issues/906) — migrating standalone scripts to a unified `speculators` CLI surface using [typer](https://typer.tiangolo.com/).

**Landing strategy:** This umbrella PR tracks the full effort. Individual sub-PRs target `cli/infra` in a star topology — each is independently reviewable and mergeable. Once all sub-PRs are merged into `cli/infra` and then into `feat/cli`, this PR merges to `main` as a rubber stamp.

Each sub-PR is a mechanical migration — script logic moves verbatim, wired to typer, with a deprecation shim (removed in v0.9.0), tests, and doc updates. No core logic changes.

## PR topology

```
main
  └── feat/cli (this PR)
        └── cli/infra (#1018)
              ├── cli/stitch-mtp (#1019)
              ├── cli/prepare-data (#1020)
              ├── cli/generate-data (#1021)
              ├── cli/regenerate-responses (#1022)
              └── cli/train (#1023)
```

All command PRs target `cli/infra` independently (star topology). They can be reviewed and merged in any order.

## Sub-PRs

- [x] #1018 — CLI infrastructure + convert: typer app, `__main__.py`, help groups, `pyproject.toml` entry point
- [x] #1019 — `stitch-mtp`: move + wire + deprecation shim + tests
- [x] #1020 — `prepare-data`: move + wire + shim + tests
- [x] #1021 — `generate-data`: move + wire + shim + tests
- [x] #1022 — `regenerate-responses`: move + wire + shim + tests + aiohttp dep
- [x] #1023 — `train`: thin typer dispatch + `speculators.train.__main__` + tests

Note: All sub PRs have merged, one final task remains:

- [ ] Rebase the PR on latest main, and audit for any new changes that should be included in any scripts

## Naming

Per [feedback from @orestis-z](https://github.com/vllm-project/speculators/issues/906#issuecomment-5178115929):
- `stitch-mtp` (not `stitch`)
- Verb-first for data commands (`prepare-data`, `generate-data`)

## Scope

`evaluate` + `plot` commands are **excluded** from this effort per [feedback from @fynnsu](https://github.com/vllm-project/speculators/issues/906#issuecomment-5180629672) — they pull in additional dependencies (guidellm) that speculators doesn't currently depend on. Follow-up work, possibly via a `speculators[evaluate]` optional extra.

[...]

PR #1036, opened 2026-08-24
feat(train)!: remove hybrid training mode, rework the training tutorial

WindChimeRan:
## Summary

Two changes, one of them breaking.

### Removes hybrid training mode (breaking)

`--on-generate cache` kept freshly generated hidden states so later epochs could reuse them. It was broken on the Mooncake backend: `MooncakeTransfer` never overrode `cache()`, so it inherited the no-op hook on `HiddenStatesTransfer`, and because the cache branch bypassed the delete path entirely, every generated sample leaked into the store while nothing was ever cached — `MooncakeTransfer.get_cached()` returns `None` unconditionally, so each epoch regenerated everything from scratch.

`--on-generate` now accepts only `delete`, which has always been its default, so existing command lines are unaffected and `--on-generate cache` fails validation with a clear message. The flag is kept purely for backwards compatibility. The now-unused `cache()` hook is dropped from the `HiddenStatesTransfer` API and from `FileTransfer`.

Online and offline training are unchanged.

### Makes the training tutorial runnable end to end

- starts with the public [`inference-optimization/speculators-ci-datasets`](https://huggingface.co/datasets/inference-optimization/speculators-ci-datasets) `tutorial_regen` split, 5K on-policy Qwen3-8B rows -- the same dataset the e2e acceptance tests already use;
- replaces unrelated reference numbers with smoke-test results produced by the documented commands;
- records measured generation time and disk use, documents `hf:<id>[:<subset>:<split>]` inputs, and fixes the DFlash serving command;
- drops the hybrid walkthrough, matching the removal above;
- aligns the remaining DFlash2 and multi-node commands with the unified CLI from #1024, which had missed them.

## 5k tutorial results

4 H100s, `tutorial_regen`, 5 epochs, `checkpoint_best`, evaluated across all nine `RedHatAI/speculator_benchmarks` subsets:

| Algorithm | Draft tokens per step | Acceptance length | Position-0 acceptance | Training time |
| --------- | --------------------- | ----------------- | --------------------- | ------------- |
| Eagle-3   | 3                     | 1.89              | 52.4%                 | 15 min        |
| DFlash    | 15                    | 1.99              | 57.9%                 | 19 min        |
| DFlash2   | 7                     | 1.90              | 52.1%                 | 20 min        |
| DSpark    | 8                     | 2.02              | 54.2%                 | 17 min        |


fynnsu (commented):
> Dataset ownership. The dataset currently lives in a personal HF namespace. Happy to move it under RedHatAI/ or vllm-project/ if maintainers prefer; that changes only the URL in Step 1.

Yeah we should probably move this to RedHatAI/ HF account. 

> Epochs 5 -> 10 matches the measured runs. checkpoint_best still selects on validation loss, so for DFlash (which bottoms out around epoch 4 on this data) the extra epochs cost time, not quality.

I'd prefer we keep this at 5 (or even 3 epochs). At just 5k samples we're definitely overfitting and I think keeping the tutorial short/easy to run is more important than a slight improvement in final acc rates. We just need to make it clear this is just an example and real runs require more data.

WindChimeRan:
@fynnsu Thanks for the review! 
* I'll rerun the training on 5 epoch. 
* Could you please help me with the dataset donation? I think I don't have the permission from the redhat side. Or maybe you can just download the dataset and re-upload to redhat. (then I'll change the dataset name in this PR)

WindChimeRan:
@fynnsu ready for review!
new changes
* dataset now use the CI hf path
* remove the hybrid mode training code
* rerun the 5 epoch training for all models, including the new dflash2
* training tutorial now include both data paths (regen and renderer endpoint)

src/speculators/train/config/schema.py
```diff
@@ -270,11 +270,11 @@ class GenerationArgs(_Group):
         "Default 'generate' generates them on demand via the vLLM endpoint; the others "
         "skip the sample, skip with a warning, or raise.",
     )
-    on_generate: Literal["cache", "delete"] = Field(
+    on_generate: Literal["delete"] = Field(
```
fynnsu, line 273: Maybe we should raise a deprecation notice when the field is passed in explicitly. Then we could remove the field in a release or two.

docs/user_guide/tutorials/train.md
```diff
@@ -819,52 +686,38 @@ Check vLLM logs for speculative decoding metrics.
...
 # avg_seq_len × num_layers × hidden_size × dtype_bytes
-# 8192 × 4 × 4096 × 2 = ~268 MB per sample
+# 3000 × 4 × 4096 × 2 = ~98 MB per sample   (Qwen3-8B, 4 captured layers)
 
-# For 50K samples: ~13 TB
-# For 10K samples: ~2.6 TB
-# For 1K samples: ~260 GB
+# For 5K samples: ~460 GB
+# For 10K samples: ~960 GB
+# For 50K samples: ~4.8 TB
```
fynnsu: How were these obtained? Like the first one I get is measure. But then should 10k be 920 GB? and 50K 4.6 TB?

docs/user_guide/tutorials/train.md
fynnsu, line 1: I'd prefer to split this out into a separate pr because it seems independent from the hybrid training work.

fynnsu (commented):
Hi Ranran, thanks for the pr!

src/speculators/train/config/schema.py
WindChimeRan: Before adding the warning I want to check the bigger picture with you, because the UX becomes more spaghetti now. 

**How the modes are selected**

| | Online | Offline | Hybrid |
|---|---|---|---|
| Before this PR | `--on-missing generate --on-generate delete` | `--on-missing raise` | `--on-missing generate --on-generate cache` |
| This PR | `--on-missing generate` | `--on-missing raise` | removed |
| Ideally | `--training-mode online` | `--training-mode offline` | — |

docs/user_guide/tutorials/train.md
WindChimeRan: done

docs/user_guide/tutorials/train.md
WindChimeRan: I prefer 1 PR. because split will result in unnecessary merge conflicts. 

The change is small.

PR #1038, opened 2026-08-25
fix(loading): resolve LLM final norm for multimodal models

orestis-z:
## Summary

For multimodal models like Inkling, the verifier's final norm weights were resolved to the wrong tensor, causing training verifier targets to diverge from what the real model computes at inference — high training accuracy (~63%) but low speculative-decoding acceptance (~10%).

PR #989 already added Inkling weight aliases and shortest-suffix resolution in `_resolve_key`, which fixed most of this. However, the `model.norm.weight` alias list still lacked `llm.norm.weight`, so resolution fell back to the generic `norm.weight` suffix scan. That is fragile: any shorter spurious key ending in `norm.weight` would win, and it does not make the intended target explicit.

## Change

- Add `llm.norm.weight` as a higher-priority alias for `model.norm.weight`, so the Inkling LLM final norm (`model.llm.norm.weight`) is matched before the generic suffix scan.
- Add tests covering shortest-suffix preference and the full set of `llm.*` aliases.

## Testing

```
pytest tests/unit/utils/test_loading.py -k resolve_key
```

All 9 `_resolve_key` tests pass.

### DIFF
--- a/src/speculators/utils/loading.py
+++ b/src/speculators/utils/loading.py
@@ -11,18 +11,21 @@ from safetensors import safe_open
 _WEIGHT_ALIASES: dict[str, list[str]] = {
     "embed_tokens.weight": ["tok_embeddings.weight", "llm.embed.weight"],
     "lm_head.weight": ["output.weight", "llm.unembed.weight"],
-    "model.norm.weight": ["norm.weight"],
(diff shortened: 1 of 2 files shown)


fynnsu (approved):
Looks good!

Issue #952, opened 2026-08-06
[RFC]: Add native DeepSeek-V4-Flash DSpark support


Sawyer117:
## Summary

This RFC proposes adding native DeepSeek-V4-Flash support to `speculators`: a DSpark draft built on DSV4's own architecture. The implementation and the training run are both complete. The target is `deepseek-ai/DeepSeek-V4-Flash` (Preview), not the later `DeepSeek-V4-Flash-0731` release. The numbers below are measured against the released DeepSeek DSpark draft on the same serve (vLLM 0.23 + vllm-ascend, `num_speculative_tokens=5`, greedy, full `DATASET=all`):

| epoch | 0.5 | 1.0 | 1.5 | 2.0 | 2.5 | 3.0 | 3.5 | 4.0 | 4.5 | 5.0 |
|---|---|---|---|---|---|---|---|---|---|---|
| mean accept_len | 3.84 | 4.06 | 4.18 | 4.25 | 4.29 | 4.35 | 4.36 | 4.39 | 4.41 | 4.40 |
| % of released (4.42) | 87.0 | 91.8 | 94.6 | 96.3 | 97.2 | 98.4 | 98.7 | 99.4 | 99.7 | 99.5 |

![current best vs released draft](https://raw.githubusercontent.com/Sawyer117/speculators/feat/dsv4-dspark/docs/deployment/images/current_best_vs_released_final_ep5p0.png)

Per-dataset accept_len at the final checkpoint, on the same serve:

| | gsm8k | math500 | humaneval | mbpp | mt-bench |
|---|---|---|---|---|---|
| this draft, 5 epochs | 4.849 | 4.565 | 4.933 | 4.555 | 3.079 |
| released DSpark draft | 4.658 | 4.661 | 4.942 | 4.535 | 3.294 |

The curve is flat over the last three checkpoints: the five-dataset mean spans 0.014 across 4.0 / 4.5 /
5.0 epochs and gsm8k's per-checkpoint gain holds at +0.009. The learning rate is a cosine decay reaching
zero at 5 epochs, so the run ends fully annealed. Per-dataset and per-position numbers for all ten
checkpoints are in the evaluation ledger linked below.

## What is implemented

- A DSV4-native draft model: 3 × [MLA (q/o LoRA, per-head attention sink) + 256-routed/1-shared MoE
  (`sqrtsoftplus` scoring, `noaux_tc` top-k) + manifold-constrained hyper-connections], plus the
[...]

PR #1041, opened 2026-08-25
feat: DeepSeek-V4-Flash DSpark draft model

Sawyer117:
## Purpose

Follow-up to #952, split into two designs as @shanjiaz asked. This is the model
definition; the companion design (expert-parallel training) comes separately, and the two
are independent of each other.

Add a `dsv4_dspark` speculator: the DSpark algorithm (block drafting, a Markov transition
head, a confidence head) on a draft whose backbone mirrors DeepSeek-V4-Flash's decoder
layer. Three layers, ~21B total parameters, ~1.5B active.

One new directory, **2085 lines across 11 files**, plus 385 lines of tests. Zero deletions.

```
models/dsv4_dspark/
  __init__.py   19   config.py   130   core.py   796   checkpoint_mapping.py  259
  backbone/
    __init__.py 16   attention.py 219  moe.py    291   block.py   70
    hyper.py   124   rotary.py    112  norm.py    49
```

Three shared files, 34 lines between them, no behaviour change for any other model: three
lines of registration in `models/__init__.py`; two flags in `train/config/schema.py` that
only this model reads (`DraftArgs` already carries several of those); and in
`train/checkpointer.py` an optional hook so a model can translate its own on-disk layout
when a run resumes — a model that does not define it gets back the object it was passed,
which is a test rather than a claim.

**Why it needs its own model definition.** The draft is close to the target but not the
same, and the differences are in the attention: a per-head sink in the softmax denominator,
which no fused path takes, and a block drafter's access pattern — gamma-wide non-causal
queries over a sliding window of target context, no KV cache. Neither is reachable by
configuring an existing module.

**Not in this proposal.** Our tree carries four more files under `backbone/` — the
expert-parallel dispatch, a grouped GEMM, a kernel registry and a `torch.compile` wrapper.
None are here, and neither are the dispatch points: every heavy op calls its reference
implementation directly, so there is no plugin seam in this diff and no vendor call in it.
The dispatch belongs to the companion design; the rest is an accelerator story we would
rather bring separately.

**Three environment switches**, two of which we are happy to move: `DSPARK_RECOMPUTE`
(activation checkpointing per draft layer), `DSPARK_MOE_BALANCE_RATE` (the `noaux_tc` bias
update — the rate is also the switch), and `DSPARK_LOG_EXPERT_LOAD` (expert-utilisation
counters). The counters belong to the model, since routing collapse does not show in the
loss curve. The other two are recipe features that happen to live here — their own PR, or
config fields instead of environment reads, whichever you prefer. Either way, the numbers
below come from runs with both enabled.

## `--freeze-experts`

@zihanlin-ai's suggestion, and it removes the dependency on the expert-parallel design:
routed experts get `requires_grad=False` and drop out of the optimizer, so nothing is left
to shard and the two proposals can be reviewed in either order.

Two caveats. The memory does not go away, it stops being sharded — ~42 GB of frozen bf16
experts resident per rank, an 80 GB-class recipe. And we have not trained this way: every
number below is full-expert EP training, and we are not attaching an acceptance-length
claim to a regime we have not measured.

## `--init-from-target`

`--init-from-target attn moe hc norm` (or `all`) warm-starts those parts of each draft
layer from the matching verifier layer. Off by default: @zihanlin-ai found MTP lineage in
the released weights only in Flash-family layer 0, and we found the released draft's router
`gate.bias` balanced only in layer 0 and untouched elsewhere, yet scoring 4.42.

The router is never warm-started, only its experts. The verifier's `gate.weight` was fitted
to a much wider hidden distribution and concentrates the draft's routing on a few experts;
its balance bias solves for the verifier's own load. That is an argument, not a measurement
— we have not A/B'd inheriting the router — so it is stated here rather than left as a flag
whose default we could not defend.

@zihanlin-ai also asked for per-layer **and per-slot** expert-utilisation counters. This has
per-layer: used/dead counts, normalized entropy, effective expert count, and the hot-expert
set across steps, so a fixed collapsed subset can be told from a rotating one. Per-slot is
not there — the counters aggregate over a forward. Say the word and we will add it. Their
remaining points bear on expert-parallel training and are answered in the companion design.

## Performance

The expert compute is the reference implementation: correct, portable, not fast. We are not
quoting a speedup, because we have not benchmarked a fused grouped GEMM against it.

## Checkpoint layout — your call, not ours

Two conventions exist for a DSpark draft and they disagree:

| loader | reads |
|---|---|
| `vllm/models/deepseek_v4/nvidia/dspark.py` | `mtp.{0,1,2}.*` from the target checkpoint |
| `vllm_ascend/models/deepseek_v4/dspark.py` | `mtp.{i}.*` |
| `vllm/model_executor/models/qwen3_dspark.py` | speculators-native (`layers.*`, `d2t`/`t2d`) |

The split is by provenance, not hardware: drafts DeepSeek releases live in the target
checkpoint's `mtp.*` namespace, drafts speculators trains use this library's layout. We are
the first speculators-trained DSV4 draft, so the two have not collided before.

| | today | cost |
|---|---|---|
| **A** emit `mtp.*` | loads on GPU and Ascend unchanged | a speculators checkpoint that does not look like one |
| **B** emit speculators-native | nothing serves it | a loader branch in both vLLM and vllm-ascend |
| **C** emit native, ship an exporter | one documented step to serve | the step exists until B lands |

**We lean to A**: speculators and vLLM are one project, so a checkpoint this library trains
should be one the sibling engine can load. B does not exist, and C leaves the checkpoint
just as unloadable with an export step in front of it.

A is implemented as registered `WeightRenaming` / `WeightConverter` rules, the way
`transformers` handles Mixtral — no save override. It is also what keeps serving out of
scope: DeepSeek ships the draft *inside* the target checkpoint, 4,705 `mtp.*` tensors
alongside 67,606 target ones under a `config.json` that already carries the `dspark_*`
fields. A draft trained here writes exactly those 4,705, so nothing has to describe it.

Against A, and worth saying before someone finds it: the prefix is misleading and upstream
is already dealing with it. vLLM PR #52165 (open, closing issue #52111) puts it as
*"DeepSeek-V4-Flash-0731 and DeepSeek-V4-Pro-0813 do not have MTP heads — their `mtp.*`
tensors are DSpark drafters. vLLM routes them to `DeepSeekV4MTPModel` anyway and dies deep
in the weight loader."* If you would rather we emit the native layout and write the loader
changes as follow-up, we will.

One thing worth passing on either way: a round trip is not a sufficient check. A rule and
its own inverse agree whatever they emit, so a mapping can round-trip perfectly and still
write a layout no loader recognises. Ours did, until the key set was compared against the
release.

And the line we would draw: this proposal guarantees the checkpoint, not the engine. A run
that changes the recipe saves and reloads losslessly, and whatever the released layout has
no slot for keeps its module name rather than being forced into `mtp.*` under an invented
one. Whether an engine understands a recipe it has never seen is that engine's concern.
(Relatedly: `algos.py::update_dspark` falls back to `Qwen3DSparkModel`, so a
speculators-format DSV4 checkpoint reaches the wrong architecture. Choosing A is what keeps
this proposal independent of that.)

## Evidence

775,965 rows self-distilled from the target at temperature 0; 8 Ascend NPUs (EP8 + FSDP2),
hidden states supplied online by a live serving instance of the target. Scored on the same
stack, greedy, `num_speculative_tokens=5`:

| | gsm8k | math500 | humaneval | mbpp | mt-bench | 5-set mean | non-chat 4 |
|---|---:|---:|---:|---:|---:|---:|---:|
| released DeepSeek draft | 4.665 | 4.639 | 4.939 | 4.526 | 3.347 | **4.4232** | 4.6922 |
| this draft, 5 epochs | 4.845 | 4.565 | 4.971 | 4.546 | 3.154 | **4.4162** | **4.7317** |
| | | | | | | 99.84% | 100.84% |

Ahead on the four non-chat datasets; the remaining gap is multi-turn chat, which we read as
training-data coverage rather than architecture. The run is fully annealed and the last
three checkpoints span 0.014 on the five-set mean.

## Open questions

1. **What would you want to see before taking ~2000 lines in-tree?** We commit to
   maintaining it, keeping it working as the training stack moves, and testing it on the
   hardware we have. If there is a bar for a model this size — a test, a doc, a
   second-platform run — tell us and we will meet it.
2. **Is a reference expert implementation acceptable in-tree**, portable and correct but
   not fast? We think yes for a first landing, with performance as a separate conversation.
3. **Should `--freeze-experts` be the documented default** for anyone reproducing this
   without expert-parallel hardware, given we have not measured its quality?

## Tests

```
$ pytest tests/unit -q
798 passed, 5 skipped in 3:32
```

`make quality` passes as well, against the `ruff` and `mypy` versions `pyproject.toml`
pins.

Twelve of those tests are new and run on a CPU-sized model in about two seconds:

| test | what it pins |
|---|---|
| `test_saved_keys_match_the_released_draft` | the emitted key set equals the released draft's, exactly |
| `test_round_trip_is_bit_identical` | save then load returns the same tensors |
| `test_experts_are_stacked_in_the_module_and_per_expert_on_disk` | one stacked parameter in memory, one tensor per expert on disk |
| `test_resume_reads_back_the_layout_it_wrote` | the single-device resume path |
| `test_the_distributed_resume_path_loads_the_same_weights` | the same, through `set_model_state_dict` |
| `test_a_released_checkpoint_that_covers_nothing_is_refused` | `strict=False` cannot hide a total mismatch |
| `test_module_layout_checkpoints_still_load` | checkpoints written before the mapping existed |
| `test_the_checkpointer_hook_is_a_no_op_for_other_models` | the shared hook, on a Qwen3 DSpark draft |
| `test_freeze_routed_experts_leaves_the_rest_trainable` | `--freeze-experts` |
| `test_init_from_target_selects_parts` | `--init-from-target`, including `all` |
| `test_init_from_target_rejects_an_unknown_part` | an unknown part is refused, not ignored |
| `test_chunked_attention_matches_the_unchunked_path` | chunking the query axis changes only floating-point associativity |

## Checklist

I have filled in:

### DIFF
--- a/src/speculators/models/__init__.py
+++ b/src/speculators/models/__init__.py
@@ -2,6 +2,7 @@ from speculators.models.attention import ALL_ATTENTION_FUNCTIONS  # noqa: F401
 
 from .dflash import DFlashDraftModel, DFlashSpeculatorConfig
 from .dspark import DSparkDraftModel, DSparkSpeculatorConfig
+from .dsv4_dspark import DSV4DSparkConfig, DSV4DSparkDraftModel
 from .eagle3 import Eagle3DraftModel, Eagle3SpeculatorConfig
 from .mtp import MTPDraftModel, MTPSpeculatorConfig
 from .peagle import PEagleDraftModel, PEagleSpeculatorConfig
@@ -9,6 +10,8 @@ from .peagle import PEagleDraftModel, PEagleSpeculatorConfig
 __all__ = [
     "DFlashDraftModel",
     "DFlashSpeculatorConfig",
+    "DSV4DSparkConfig",
+    "DSV4DSparkDraftModel",
     "DSparkDraftModel",
     "DSparkSpeculatorConfig",
     "Eagle3DraftModel",
--- a/src/speculators/models/dsv4_dspark/__init__.py
+++ b/src/speculators/models/dsv4_dspark/__init__.py
@@ -1,0 +1,19 @@
+"""DeepSeek-V4-Flash DSpark draft speculator.
+
+A self-contained, backend-agnostic implementation of the DSpark draft *backbone* for
+DeepSeek-V4-Flash: multi-head latent attention with per-head attention sinks, 256 routed
+experts + 1 shared per layer, and hyper-connections in place of the residual stream --
+plain PyTorch throughout.
+
+The DSpark *method* (anchor-block sampling, the Markov and confidence heads, the
+compound loss, and the SpeculatorModel / trainer / data contract) is reused as-is from
+:mod:`speculators.models.dspark`; this package adds only the DSV4-native backbone and
+the thin subclass that swaps it in.
+"""
+
+from __future__ import annotations
+
+from .config import DSparkDraftConfig
+from .core import DSV4DSparkConfig, DSV4DSparkDraftModel
+
+__all__ = ["DSV4DSparkConfig", "DSV4DSparkDraftModel", "DSparkDraftConfig"]
--- a/src/speculators/models/dsv4_dspark/backbone/__init__.py
+++ b/src/speculators/models/dsv4_dspark/backbone/__init__.py
@@ -1,0 +1,16 @@
+"""Backend-agnostic DSV4 decoder backbone for the DSpark draft.
+
+A self-contained torch implementation of the DeepSeek-V4-Flash decoder shape:
+multi-head latent attention with per-head sinks, 256 routed experts + 1 shared, and
+hyper-connections. Plain PyTorch throughout, with no dependency on any accelerator
+package.
+"""
+
+from __future__ import annotations
+
+from .norm import RMSNorm, UnweightedRMSNorm
+
+__all__ = [
+    "RMSNorm",
+    "UnweightedRMSNorm",
+]
--- a/src/speculators/models/dsv4_dspark/backbone/attention.py
+++ b/src/speculators/models/dsv4_dspark/backbone/attention.py
@@ -1,0 +1,219 @@
+"""Multi-head latent attention with a per-head learnable sink (DSV4 draft).
+
+Two layers here:
+
+* :func:`sink_block_attention` — the numerical core: a dense, non-causal
+  sink-softmax over ``[context + block]`` KV. Vanilla SDPA cannot
+  express the sink (it is an extra term in the softmax denominator), so this
+  eager einsum form is the sink-correct path.
+
+* :class:`LatentAttention` — the MLA projection stack (low-rank query
+  ``wq_a→q_norm→wq_b`` + per-head RMS, single shared-KV ``wkv→kv_norm``,
+  grouped low-rank output ``wo_a→wo_b``) plus the learnable ``sink``. Its
+  training forward assembles per-anchor-block queries/keys/values and calls
+  :func:`sink_block_attention`; it does not use a KV cache (teacher-forced).
+
+Weight-name note: the released checkpoint stores these as ``wq_a`` / ``wq_b`` /
+``wkv`` / ``wo_a`` / ``wo_b`` / ``attn_sink``; the loader maps those onto the
+descriptive attribute names here.
+"""
+
+from __future__ import annotations
+
+import torch
+from torch import nn
+
+from .norm import RMSNorm, UnweightedRMSNorm
+from .rotary import apply_rotary_emb
+
+# k/v arrive as [N, Sk, D]: one shared KV head, no head axis.
+_SHARED_KV_RANK = 3
+# Cap on the fp32 logit block a single attention chunk may allocate. The dense
+# [N, Sq, H, Sk] intermediates are the largest tensors in the model: at 512 anchors over
+# a 22k-token context they are ~18 GB each, and three of them are live at once without
+# chunking. Chunking the query axis is exact -- the softmax normalizes per query row --
+# and leaves only the probabilities retained for backward, which activation
+# checkpointing then removes as well.
+_ATTN_CHUNK_BYTES = 1 << 30
+
+
+def sink_block_attention(
+    q: torch.Tensor,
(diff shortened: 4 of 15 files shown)


Sawyer117:
All six of @coderabbitai's findings are addressed in `c104fd2`. Incremental reviews are
off on this repo, so the threads stay open — listing them here instead.

1. **Uninitialized parameters.** `Router.weight` and the two mHC projections came from
   `torch.empty` and were only re-initialized when a degeneracy scan found NaN,
   non-finite or all-zero values. Finite garbage passes that check, so a run could start
   from uninitialized memory silently and non-reproducibly. Initialized where they are
   created now.
2. **Aliased expert-load counters.** The diagnostic held the tensor
   `update_load_balance_bias` all-reduces in place, so the per-rank histogram was
   reported as a global sum. Cloned.
3. **Per-rank expert slice with no dispatch.** Removed. Expert parallelism is the
   companion design; the slice had nothing here to match it.
4. **Conversion peak memory.** Layout conversion held the per-expert tensors and the
   stacked copies at once — a second full routed-expert allocation on resume. It streams
   by layer group and consumes its input now.
5. **Dense fp32 attention logits.** Chunked along the query axis; exact up to
   floating-point associativity (~2 ulp forward and backward, pinned by a test). Worth
   being precise about the size of the win: it takes roughly 3x off the peak, not an
   asymptotic factor, and `DSPARK_RECOMPUTE=1` is what removes the retained
   probabilities. A streaming softmax would be a kernel, which this PR deliberately does
   not carry.
6. **`DSPARK_RECOMPUTE`.** Accepts `true`/`yes`/`on` rather than raising.

`ruff`, `ruff format` and `mypy --check-untyped-defs` are clean against the versions
`pyproject.toml` pins; `pytest tests/unit` is 798 passed, 5 skipped. The `Quality`
workflow has not run on any of the three revisions — it is waiting on approval.

PR #1008, opened 2026-08-19
revert: remove speculators import from launch_vllm.py

rahul-tuli:
## Purpose

Reverts #958.

`launch_vllm.py` is executed via `VLLM_PYTHON`, which points at a vLLM-only virtualenv that does **not** have `speculators` installed. The top-level `from speculators.provenance import ...` added in #958 causes an immediate `ImportError`, crashing the vLLM server before it can start. This breaks every e2e smoke test that launches a vLLM server subprocess.

**Root cause:** `scripts/launch_vllm.py` must remain dependency-free with respect to `speculators` because it runs in the vLLM process, not the speculators process.

## What is reverted

- `scripts/launch_vllm.py` — provenance artifact writers and `speculators.provenance` imports
- `scripts/evaluate/evaluate.py` — `vllm_command.txt` hint
- `src/speculators/train/utils.py` — `find_repo_root` import and `cwd=` change
- `tests/unit/scripts/test_provenance_integration.py` — removed
- `tests/unit/scripts/test_vllm_provenance.py` — removed
- `AGENTS.md` — provenance docs reverted

## Test plan

- [ ] `test_mtp_finetuning_smoke` passes with vLLM 0.27.1 (cellA)
- [ ] `test_mtp_finetuning_smoke` passes with vLLM nightly (cellB)
- [ ] `make quality` passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1043, opened 2026-08-26
feat(launch_vllm): re-add vLLM provenance via generated sibling module

orestis-z:
## Summary

Re-applies the vLLM provenance feature from #958, which was reverted in #1008.

**Why #958 was reverted:** `launch_vllm.py` is executed via `VLLM_PYTHON` — a vLLM-only virtualenv that does **not** have `speculators` installed. The top-level `from speculators.provenance import ...` added in #958 raised an immediate `ImportError`, crashing the vLLM server before it could start and breaking every e2e smoke test that launches a server subprocess.

**Fix:** keep `src/speculators/provenance.py` as the single source of truth, but generate a byte-identical, stdlib-only sibling copy at `scripts/_provenance.py` and import from that. Because `scripts/` is `sys.path[0]` when the script is run directly (both the example shell scripts and the `VLLM_PYTHON` subprocess launch it that way), the import resolves with **zero dependency on `speculators`** — the exact constraint the revert called out.

A small `scripts/sync_provenance.py` drives the copy and is wired into the Makefile:
- `make style` → regenerates `scripts/_provenance.py`
- `make quality` → `--check` fails if the copy has drifted (so CI enforces parity)

## What's included

Exact restore of #958:
- `scripts/launch_vllm.py` — `--provenance-dir` / `--no-hash-checkpoints` flags + `vllm_command.txt` / `vllm.patch` / `checkpoint_sha256.txt` writers (best-effort; failures warn but never block launch). Only change vs #958: imports from `_provenance`.
- `scripts/evaluate/evaluate.py` — hint when `vllm_command.txt` is absent
- `src/speculators/train/utils.py` — `find_repo_root` + `cwd=repo_root` on the git subprocess
- `AGENTS.md` — provenance docs
- unit + integration tests

New (the drift-guard + regression coverage):
- `scripts/_provenance.py` (generated), `scripts/sync_provenance.py`, Makefile wiring
- `tests/unit/scripts/conftest.py` — puts `scripts/` on `sys.path`
- `TestVenvIsolation` in `test_provenance_integration.py` — imports `launch_vllm.py` with `speculators` stripped from `sys.path`, reproducing and guarding against the exact #1008 failure mode

## Test plan

- [x] `tests/unit/scripts/test_vllm_provenance.py` + `test_provenance_integration.py::TestVenvIsolation` pass (21 tests)
- [x] `ruff check` / `ruff format --check` pass on changed files
- [x] `scripts/sync_provenance.py --check` passes (copy in sync)
- [x] `mdformat --check AGENTS.md` passes
- [ ] e2e smoke tests (`test_mtp_finetuning_smoke`) pass with a vLLM-only venv

### DIFF
--- a/scripts/sync_provenance.py
+++ b/scripts/sync_provenance.py
@@ -1,0 +1,75 @@
+"""Sync src/speculators/provenance.py → scripts/_provenance.py.
+
+Run with no args to regenerate; run with --check to verify the copy is current.
+Wired into `make style` (regenerate) and `make quality` (check).
+"""
+
+from __future__ import annotations
+
+import argparse
+import sys
+import tempfile
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+SRC = REPO_ROOT / "src" / "speculators" / "provenance.py"
+DST = REPO_ROOT / "scripts" / "_provenance.py"
+
+BANNER = """\
+# ============================================================================
+# AUTO-GENERATED — do not edit directly.
+# Source of truth: src/speculators/provenance.py
+# Regenerate with: make style   (or: python scripts/sync_provenance.py)
+# ============================================================================
+"""
+
+
+def _generated(src_text: str) -> str:
+    return BANNER + src_text
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument(
+        "--check",
+        action="store_true",
+        help="Exit non-zero if scripts/_provenance.py is out of sync",
+    )
+    args = parser.parse_args()
+
+    src_text = SRC.read_text()
+    expected = _generated(src_text)
+
+    if args.check:
+        if not DST.exists():
+            print(
+                f"ERROR: {DST} does not exist. Run `make style` to generate it.",
+                file=sys.stderr,
+            )
+            sys.exit(1)
+        actual = DST.read_text()
+        if actual != expected:
+            print(
+                f"ERROR: {DST} is out of sync with {SRC}.\n"
+                "Run `make style` (or: python scripts/sync_provenance.py) to update.",
+                file=sys.stderr,
+            )
+            sys.exit(1)
+        print(f"OK: {DST} is in sync.")
+    else:
+        fd, tmp = tempfile.mkstemp(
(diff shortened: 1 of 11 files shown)


scripts/sync_provenance.py
```diff
@@ -0,0 +1,75 @@
...
+        )
+        tmp_path = Path(tmp)
+        try:
+            with open(fd, "w") as f:
+                f.write(expected)
+            tmp_path.replace(DST)
+        finally:
+            if tmp_path.exists():
+                tmp_path.unlink()
+        print(f"Synced {SRC} → {DST}")
```
rahul-tuli, line 71: Quick thing I noticed while testing this locally, the comment above the `sync-provenance` hook says it "fails the commit if the copy changed, prompting a re-stage," but I don't think that's what actually happens? I tried editing `src/speculators/provenance.py` and running `python scripts/sync_provenance.py` directly, and it regenerated `scripts/_provenance.py` correctly but still exited `0`. Since pre-commit only blocks a commit on a non-zero exit code, it looks like this hook can never actually stop a commit, it'll just quietly update the file in the working tree, and whatever was already staged (the old, out-of-sync version) is what gets committed. (this would ofcourse get caught in our CI just confirming if this is the intended behavior)

rahul-tuli (approved):
LGTM! just a couple of questions:
1) About the intended behavior posted as a comment on sync_provenance.py
2) The PR looks good for now, and we can probably do this as a follow up, I was thinking if a symlink would achieve the same goal with less ongoing maintenance than the generated-copy + sync script?

scripts/sync_provenance.py
orestis-z: Good catch, but it does block — just not via the exit code. pre-commit fails any run where a hook *modifies a tracked file*, independent of exit status (same mechanism as `ruff format` / `end-of-file-fixer`). Standalone the script exits 0, but under pre-commit the regenerated `_provenance.py` is detected as modified and the commit is rejected. Verified locally (edit `provenance.py`, stage only it, commit):

```
Sync scripts/_provenance.py ...Failed
- hook id: sync-provenance
- files were modified by this hook
```

exit 1, with `_provenance.py` left modified for re-staging — so the stale staged copy can't land. `make quality --check` is the CI backstop regardless.

PR #1050, opened 2026-08-27
feat(dflash): support multimodal (VL) training with plain-rope drafts

curnane-lab:
## Purpose

Enable DFlash drafts to train against multimodal (VL) such as Qwen3.5-4B and Qwen3-VL.

The data pipeline is already algorithm-agnostic and multimodal-capable: image rows ride the Chat Completions path (vLLM is served with `--allowed-local-media-path` and runs the full VLM forward), and the captured verifier hidden states already encode the image content - the draft itself never sees pixels. The training forward uses plain 1D position ids for every speculator.

The one gap was draft config construction: `create_transformer_layer_config` inherited `mrope_section` from the verifier's `text_config` into the draft config (the EAGLE-3 mechanism), while

- the training forward is plain 1D, and
- vLLM's DFlash serving path rejects MRoPE draft configs outright.

This PR strips `mrope_section` (and, via the existing guard, the coupled `partial_rotary_factor`) from the draft config for DFlash speculators (`PLAIN_ROPE_DRAFT_SPECULATORS`). The draft's rope is only an internal distance metric for its own attention - the verifier's (3, N) MRoPE positions carry no signal for the draft - so plain rope keeps train/serve consistent for free. This also matches the official `Qwen/Qwen3.5-4B-DFlash` checkpoint, whose draft config ships `rope_scaling: null`.

Also adds:

- an end-to-end online example, `examples/train/dflash_qwen3_5_4b_sharegpt4v_online_5k.sh`, mirroring the official Qwen3.5-4B-DFlash draft geometry (block size 16, 5 draft layers, aux target layers [1, 8, 15, 22, 29], vocab 248320);
- a "Multimodal (VL) Verifiers" section in `docs/user_guide/algorithms/dflash.md`.

The same plain-rope design was validated end-to-end in the SpecForge implementation, see [sgl-project/SpecForge#730](https://github.com/sgl-project/SpecForge/pull/730) for details.

## Tests

Unit tests for the new rope-config behavior: `dflash` strips `mrope_section` on both the transformers>=5 `rope_parameters` path and the pre-5 `rope_scaling` path (the coupled `partial_rotary_factor` is dropped by the existing guard, `rope_theta` preserved), while `eagle3` keeps `mrope_section` when `speculator_type` is passed.

```bash
python -m pytest tests/unit/train/test_rope_config.py tests/unit/train/test_draft_config_init.py tests/unit/train/config -q
# 129 passed
```

E2E: a new multimodal dflash case in `tests/e2e/smoke/test_offline_training.py` (Qwen3-VL-2B-Instruct + `sharegpt4v_coco`) covers render -> offline hidden-state extraction with images -> training -> vLLM inference validation of the trained checkpoint; a leaked `mrope_section` would surface at the vLLM engine step. The multimodal online path is already covered by the existing MM case in `test_online_training.py`.

```bash
python -m pytest tests/e2e/smoke/test_offline_training.py --collect-only -q
# 6 tests collected
bash -n examples/train/dflash_qwen3_5_4b_sharegpt4v_online_5k.sh
# no syntax errors
```

## Checklist

I have filled in:

````


fynnsu:
@curnane-lab does vLLM have a reason for rejecting the mrope drafter configs? If not, perhaps it would make more sense to add support for mrope drafters on the vllm side, rather than removing the support on the speuclators training side?

curnane-lab:
> @curnane-lab does vLLM have a reason for rejecting the mrope drafter configs? If not, perhaps it would make more sense to add support for mrope drafters on the vllm side, rather than removing the support on the speuclators training side?

MRoPE support is unimplemented upstream; adding it for DFlash would mean position materialization, MRoPE rope application in the proposer, and position-buffer headroom (cf. vllm-project/vllm#48725) - a large engine feature for one algorithm.

More importantly, plain-rope is the correct form for DFlash rather than a workaround: the draft's inputs are the verifier's aux hidden states - which already fuse the vision content and the verifier's own MRoPE effect - plus token embeddings and its own position encoding; vision never flows through the draft's rope, and there is no KV/attention sharing between draft and verifier (the draft's K is an fc projection of hidden features). The draft's rope is an internal distance metric whose only hard requirement is train/serve consistency. The official z-lab/Qwen3.5-4B-DFlash checkpoint ships rope_scaling: null, and the same design was validated end-to-end in sgl-project/SpecForge#730.

Issue #1051, opened 2026-08-27
[RFC]: Add XPress training pipeline support


ZKBig:
### Motivation.

A block-diffusion drafter such as DFlash proposes a whole block of draft tokens in one forward pass, which makes drafting cheap. The cost is accuracy: the block is decoded from per-position marginals, so token *k* is drawn without seeing what the drafter chose at *k-1*. Nothing ties the positions together, while the target model verifies conditionally, so blocks that are individually plausible but jointly improbable get rejected early. That caps the achievable acceptance length.

We would like to contribute XPress, which restores the missing causality without a token-by-token loop, as a first-class speculator alongside DFlash and DSpark rather than as a fork. On Qwen3-8B it raises acceptance length over the Markov head on every benchmark we measured, at both temperatures, and over the plain DFlash drafter it is worth about +30% acceptance and ~1.3x end-to-end throughput on average. Full tables are in the third field.

Paper: xPress: Parallel Refinement for Diffusion Drafters in Speculative Decoding (https://arxiv.org/abs/2608.02438).

### Proposed Change.

A new speculator, `xpress`, registered next to the existing ones.

**The refiner head.** Per position it reads the drafter's hidden state, a block-global summary and the previous token's embedding, fuses them into a low-rank space (r = 256), mixes across positions with a lower-triangular (causal) linear mixer, passes through an r-space SwiGLU MLP, and reads back out as a bias on the target LM head's logits. Everything except the two vocabulary projections lives in the bottleneck (~80M params, 96% of which are those two matrices). The whole refiner is:

```python
lat = self.token_embed(prev_token_ids)                    # [N, B, r]
x   = self.in_proj(torch.cat([hcache, lat], dim=-1))      # 3r -> r
l_eff = self.mix_l * self.causal_tril                     # lower-triangular, per channel
x = x + torch.einsum("ckj,njc->nkc", l_eff, x)            # causal mix across the block
x = x + self.mlp_down(F.silu(self.mlp_gate(x)) * self.mlp_up(x))
return self.readout(x)                                    # r -> V, a bias on the base logits
```

`hcache` is the pass-invariant `[down_h(h) ; down_g(g)]`, computed once per block and reused. `mix_l` is initialised to zeros, so at init the mixer is the identity and the head starts as a pure previous-token bias.

[...]

PR #1061, opened 2026-08-30
Add XPress: a causal refiner over the DFlash block drafter

ZKBig:
## Purpose

Adds **XPress**, a causal refiner over the DFlash block-parallel draft backbone, as a first-class algorithm alongside DFlash, DFlash2 and DSpark. Implements RFC #1051. Paper: https://arxiv.org/abs/2608.02438

A block-parallel drafter proposes a whole block in one forward pass, which makes drafting cheap, and decodes each position from its own marginal, which makes it inaccurate: token *k* is drawn without seeing what the drafter chose at *k-1*. The verifier scores conditionally, so blocks that are individually plausible but jointly improbable are rejected early, capping acceptance length.

XPress adds a small head that reads each position's hidden state, a block-global summary and the previous token, fuses them into a low-rank bottleneck, mixes across the block with a lower-triangular (causal) mixer, and reads back out as a bias on the base logits. Because the mixer is truly causal rather than a local smoother, the block is resolved by **Jacobi iteration** rather than a left-to-right loop: the same call runs K times, each pass re-feeding the previous pass's argmax, so the serial depth at inference is a constant K instead of the block size.

Training co-trains the head with the backbone over free-running Jacobi rounds. Each round conditioned on the previous round's argmax, which is the exact recurrence inference performs, with stop-grad between rounds, and anchors the backbone with a base-logits term so its standalone drafting quality does not erode.

`XPressDraftModel` subclasses `DFlashDraftModel`, so the backbone, data pipeline, verifier support and checkpointing are inherited unchanged.

### Shared code touched

*   `loss_function` / `compound_loss` gain `denom_decay` (default off), which normalizes a position-decayed loss by the decayed weight sum instead of scaling only the numerator. With the default the code path is unchanged.
*   `xpress` joins the DFlash backbone algorithm group and the DFlash-family defaults, so an XPress run without `--num-draft-layers` gets 5 layers rather than 1, and unset `--loss-fn` / `--block-size` resolve to the validated recipe (`{"ce": 0.1, "tv": 0.9}`, 16) rather than the generic defaults.
*   `XPressDraftModel` / `XPressSpeculatorConfig` are exported from the package root, which `tests/unit/test_package_imports.py` requires of every algorithm.

### Removed since the first revision

Following review, the following were taken out of this PR rather than fixed in place:

*   **fp32 master weights.** FSDP2's `MixedPrecisionPolicy(param_dtype=bf16, reduce_dtype=fp32)` already keeps the sharded masters and the optimizer step in fp32, so the shadow copies were redundant, and they detached from the sharded parameters, so the FSDP case was not handled correctly either.
*   **`--val-data-path`, `--eval-interval` / `--eval-max-batches`.** Mid-epoch validation and a separate validation corpus are trainer features unrelated to XPress and are planned as their own PR.
*   **`--no-packing`.** It fed the packer a constant length so each batch held one conversation per rank, which is how our reference recipe specifies its batch. Useful, misleadingly named, and unrelated to XPress. 
*   **`anchor_cap_to_max_valid`, `prefix_valid_mask`.** Both existed only to match our reference implementation. `prefix_valid_mask` fires only when a block spills into the next packed document *and* that document has supervised tokens within `block_size`, which effectively never happens. `models/dflash/utils.py` is back to upstream.
*   **The Jacobi-rollout accept-length metric** and its `eval_jacobi_passes` flag. Measuring it needed `anchor_valid` out of the shared `_backbone_forward`, which meant changing that signature and every caller. DSpark already reports an equivalent `accept_len` computed from the loss mask, so the metric was dropped.
*   **Annealing `base_anchor_weight` toward a floor.** Algorithm-specific logic does not belong in the shared trainer; happy to adopt the general weight scheduler when it lands.
*   **The metric-reduce ordering guard.** Defensive hardening of `dist.reduce` call order, unrelated to XPress.

## Tests

```
# new
pytest tests/unit/models/test_xpress_head.py \
       tests/unit/models/test_denom_decay.py \
       tests/unit/train/config/test_xpress_registration.py

# regression across the models and trainer this PR touches 

pytest tests/unit
make quality
```

The unit tests pin the invariants Jacobi iteration depends on: the mixer being causal across the block (a settled prefix cannot be disturbed by a later slot, which is what makes the fixed point exist), the pass-invariant hidden cache matching the full path, and the training-time mixer `x + (L * tril) x` being the same function as the folded `L * tril + I` an inference engine loads — so an exported checkpoint means the same thing at serving time as it did in training. The registration tests pin the lookup the trainer actually performs, `SpeculatorModel.registry["xpress"]`, and that the `xpress` group is reachable on `TrainConfig` (registering it in `_GROUPS` alone leaves every `--xpress-*` flag silently resolving to nothing).

### Results

```
$ pytest tests/unit/test_package_imports.py tests/unit/models/test_xpress_head.py \
         tests/unit/models/test_denom_decay.py \
         tests/unit/train/config/test_xpress_registration.py -q
12 passed

$ pytest tests/unit/train/config -q
78 passed

$ pytest tests/unit -q   # deselecting the environment failures below
945 passed, 27 deselected

$ make quality
ruff check ..................... All checks passed!
ruff format --check ............ 251 files already formatted
mdformat --check ............... ok
mypy --check-untyped-defs ...... Success: no issues found in 229 source files
sync_provenance --check ........ OK
```

I have filled in:

### DIFF
--- a/examples/train/xpress_morph_config.py
+++ b/examples/train/xpress_morph_config.py
@@ -1,0 +1,69 @@
+"""Re-type a converted DFlash checkpoint as an XPress starting point.
+
+``convert_model()`` knows nothing about XPress, so a warm start is two steps:
+convert the DFlash backbone normally, then run this to switch the config over and
+stamp the fields that live in the MODEL CONFIG rather than on the CLI. Keeping it
+separate leaves the upstream converter untouched.
+
+    python examples/train/xpress_morph_config.py <converted_dir> [--shift]
+
+Idempotent, and meant to be re-run: a checkpoint morphed by an older revision
+would otherwise silently miss any field added since.
+
+``--rank`` / ``--mlp-ratio`` size the refiner head. They matter here and nowhere
+else: a ``--from-pretrained`` run reads its architecture from this config, so the
+CLI flags of the same name are ignored on that path. The defaults (256 / 2, an MLP
+hidden of 512) are what the released checkpoints use.
+
+``--shift`` selects the DeepSeek/DSpark block convention
+(``sample_from_anchor=True``, every slot predicts the next token). The default is
+fill-in, which is what z-lab checkpoints are native to: slot 0 carries the known
+anchor and slots 1..B-1 predict.
+"""
+
+import argparse
+import json
+import logging
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument("checkpoint", type=Path, help="converted checkpoint dir")
+    parser.add_argument("--shift", action="store_true")
+    parser.add_argument("--rank", type=int, default=256)
+    parser.add_argument("--mlp-ratio", type=int, default=2)
+    args = parser.parse_args()
+
+    path = args.checkpoint / "config.json"
+    cfg = json.loads(path.read_text())
+    model_type = cfg.get("speculators_model_type")
+    if model_type not in ("dflash", "xpress"):
+        raise SystemExit(
+            f"expected a converted dflash/xpress checkpoint, got {model_type!r}"
+        )
+
+    cfg["speculators_model_type"] = "xpress"
+    cfg["architectures"] = ["XPressDraftModel"]
+    cfg["sample_from_anchor"] = args.shift
+    cfg["xpress_rank"] = args.rank
+    cfg["xpress_mlp_ratio"] = args.mlp_ratio
+    cfg.setdefault("num_jacobi_passes", 6)
+    cfg.setdefault("mask_token_id", 151669)
+    path.write_text(json.dumps(cfg, indent=2))
+
+    logging.basicConfig(level=logging.INFO, format="%(message)s")
+    logger.info(
+        "morphed %s: xpress, %s, block_size=%s, rank=%d, mlp_hidden=%d",
+        path,
+        "shift" if args.shift else "fill-in",
+        cfg.get("block_size"),
+        args.rank,
+        args.rank * args.mlp_ratio,
+    )
+
+
+if __name__ == "__main__":
+    main()
--- a/examples/train/xpress_qwen3_8b_regen_online.sh
+++ b/examples/train/xpress_qwen3_8b_regen_online.sh
@@ -1,0 +1,202 @@
+#!/usr/bin/env bash
+# XPress on Qwen3-8B, one node / 8 GPUs: vLLM verifier on GPU 0, 7 trainer ranks on
+# GPUs 1-7. Trains on a REGENERATED corpus (on-policy responses from the target).
+#
+# The validated XPress b16 recipe. Every training flag below was audited rather than
+# guessed -- see XPRESS-TRAINING.md before changing any of them.
+#
+#   bash examples/train/xpress_qwen3_8b_regen_online.sh
+#
+# Known deviations from the published recipe (see XPRESS-TRAINING.md):
+#   * hidden states are produced online by vLLM; data order differs.
+#   * no fp32 master weights: the bf16 params are clipped (at the same 1.0) and
+#     stepped directly, rather than fp32 copies of them.
+#
+# EVAL protocol: the held-out split of --data-path (--train-data-ratio), accept length
+# averaged over conversations, block_size-1 Jacobi passes.
+set -Eeuo pipefail
+
+PY=${PY:-python}
+cd "$(cd "$(dirname "$0")/../.." && pwd)"
+
+if [[ -z "${VLLM_PY:-}" ]]; then
+    if $PY -c 'import vllm' 2>/dev/null; then
+        VLLM_PY=$PY
+    elif [[ -x ./.venv_vllm/bin/python ]] && ./.venv_vllm/bin/python -c 'import vllm' 2>/dev/null; then
+        VLLM_PY=$(cd .venv_vllm/bin && pwd)/python
+        echo "note: \$PY has no vllm; using ./.venv_vllm/bin/python for the verifier"
+    else
+        echo "FATAL: no interpreter with vllm found." >&2
+        echo "  \$PY ($PY) cannot import vllm, and ./.venv_vllm/bin/python is absent or lacks it." >&2
+        echo "  Fix: uv venv .venv_vllm --python 3.12 && VIRTUAL_ENV=.venv_vllm uv pip install 'vllm>=0.22.0'" >&2
+        echo "  Or set VLLM_PY=/path/to/python-with-vllm" >&2
+        exit 1
+    fi
+fi
+
+$VLLM_PY -c 'import vllm' 2>/dev/null || {
+    echo "FATAL: VLLM_PY ($VLLM_PY) cannot import vllm." >&2
+    exit 1
+}
+
+$VLLM_PY -c '
+from vllm.distributed.kv_transfer.kv_connector.factory import KVConnectorFactory as F
+import sys
+if "ExampleHiddenStatesConnector" not in getattr(F, "_registry", {}):
+    sys.exit("this vllm build has no ExampleHiddenStatesConnector "
+             "(needed for hidden-state extraction); install a newer vllm")
+' || { echo "FATAL: see above ($VLLM_PY)" >&2; exit 1; }
+$PY -c 'import speculators, hs_connectors' 2>/dev/null || {
+    echo "FATAL: \$PY ($PY) cannot import speculators/hs_connectors." >&2
+    echo "  Fix: uv pip install -e ./hs_connectors -e .   (hs_connectors FIRST)" >&2
+    exit 1
+}
+
+export FLASHINFER_DISABLE_VERSION_CHECK=1
+export HF_HOME=${HF_HOME:-/root/.cache/huggingface}
+
+_want_tmp=${TMPDIR:-/dev/shm/spec-tmp}
+mkdir -p "$_want_tmp" 2>/dev/null || true
+_exec_ok=0
+if [[ -d "$_want_tmp" ]]; then
+    _probe="$_want_tmp/.execprobe.$$"
+    printf '#!/bin/sh\nexit 0\n' > "$_probe" 2>/dev/null && chmod +x "$_probe" 2>/dev/null \
+        && "$_probe" 2>/dev/null && _exec_ok=1
+    rm -f "$_probe"
+fi
+if [[ "$_exec_ok" != 1 ]]; then
+    _fallback=${XDG_CACHE_HOME:-$HOME/.cache}/spec-tmp
+    echo "note: $_want_tmp is not exec-capable (noexec mount?); using $_fallback instead" >&2
+    echo "      -- torch.compile dlopens .so files it builds under TMPDIR" >&2
+    _want_tmp=$_fallback
+    mkdir -p "$_want_tmp"
+fi
+export TMPDIR=$_want_tmp
+# Pin the compile caches explicitly so they follow TMPDIR even if something else
+# repoints TMPDIR later; these are the dirs that actually need exec permission.
+export TORCHINDUCTOR_CACHE_DIR=${TORCHINDUCTOR_CACHE_DIR:-$TMPDIR/torchinductor}
+export TRITON_CACHE_DIR=${TRITON_CACHE_DIR:-$TMPDIR/triton}
+mkdir -p "$TORCHINDUCTOR_CACHE_DIR" "$TRITON_CACHE_DIR"
+# wandb: export WANDB_API_KEY yourself (never hardcode it) and set the project.
+export WANDB_PROJECT=${WANDB_PROJECT:-ripple-dspark}
+
+# Self-diagnosis for a multi-week run. A wedged rank previously showed up only as a
+# frozen progress bar with no way to see WHERE it was wedged (py-spy needs CAP_SYS_PTRACE,
+# which containers usually withhold). With these, a stuck collective aborts with every
+# rank's stack plus an NCCL flight-recorder dump instead of hanging silently forever.
+export PYTHONFAULTHANDLER=${PYTHONFAULTHANDLER:-1}
+export TORCH_NCCL_TRACE_BUFFER_SIZE=${TORCH_NCCL_TRACE_BUFFER_SIZE:-2000}
+export TORCH_NCCL_DUMP_ON_TIMEOUT=${TORCH_NCCL_DUMP_ON_TIMEOUT:-1}
+export TORCH_NCCL_HEARTBEAT_TIMEOUT_SEC=${TORCH_NCCL_HEARTBEAT_TIMEOUT_SEC:-1800}
+
+MODEL="Qwen/Qwen3-8B"
+OUT_ROOT=${OUT_ROOT:-/root/speculators_out}
+BACKBONE=$OUT_ROOT/dflash_b16_zlab_converted
+DATA_DIR=${DATA_DIR:-/root/DeepSpec/data}
+TRAIN_JSONL=${TRAIN_JSONL:-$DATA_DIR/qwen3_8B_refiner_train_nothink.jsonl}
+OUTPUT_DIR=$OUT_ROOT/xpress_b16_zlab_8gpu
+TRAIN_DATA=$OUTPUT_DIR/train
+RUN_NAME=${RUN_NAME:-xpress-b16-speculators-8gpu}
+VLLM_PORT=${VLLM_PORT:-8300}
+RDZV_PORT=${RDZV_PORT:-29610}
+SEQ_LENGTH=4096
+TARGET_LAYER_IDS="1 9 17 25 33"
+MAX_SAMPLES=${MAX_SAMPLES:-1311126}
+LOG_FREQ=${LOG_FREQ:-50}
+mkdir -p "$OUT_ROOT" "$OUTPUT_DIR"
+
+if [[ -t 1 ]]; then
+    echo "WARNING: stdout is a TTY. rich/tqdm and vLLM per-request logs share one" >&2
+    echo "  console lock; if the terminal (ssh/tmux) cannot drain them fast enough" >&2
+    echo "  the writer blocks while holding it and rank 0 stalls -- the other ranks" >&2
+    echo "  then spin in NCCL forever. A TTY run also dies with your ssh session." >&2
+    echo "  Prefer:  nohup bash $0 > run.log 2>&1 &" >&2
+    sleep 5
+fi
+
+echo "=== Step 0: convert z-lab b16 -> speculators format (idempotent) ==="
+if [ ! -f "$BACKBONE/config.json" ]; then
+    $PY - <<PYEOF
+from speculators.convert.entrypoints import convert_model
+convert_model(
+    model="z-lab/Qwen3-8B-DFlash-b16",
+    verifier="$MODEL",
+    algorithm="dflash",
+    output_path="$BACKBONE",
+    aux_hidden_state_layer_ids=[1, 9, 17, 25, 33],
+)
+PYEOF
+fi
+
+$PY examples/train/xpress_morph_config.py "$BACKBONE" --rank 256 --mlp-ratio 2
+
+echo "=== Step 1: prepare data (one-time; marker records WHAT was prepared) ==="
+prepare_split () {
+    local stamp="$2/.data_ready" want
+    want="$1|${3:-all}|$SEQ_LENGTH|$(stat -c %s "$1")"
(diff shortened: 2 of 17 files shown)


src/speculators/models/dflash/utils.py
```diff
@@ -53,6 +54,11 @@ def select_anchors(
     anchor_valid = torch.zeros(num_anchors, dtype=torch.bool, device=device)
 
     k = min(num_anchors, valid_indices.numel())
+    if cap_to_max_valid:
+        # Cap at min(num_anchors, valid_candidates - 1): drop one anchor whenever
```
fynnsu: Why?

src/speculators/models/dflash/core.py
```diff
@@ -478,7 +488,31 @@ def _backbone_forward(
         if not self.config.sample_from_anchor:
             aligned_loss_mask[:, :: self.block_size] = 0
 
-        return hidden, logits, targets, aligned_loss_mask, anchored_block_indices
+        if getattr(self.config, "prefix_valid_mask", False):
```
fynnsu: Have you observed this causing an issue for your training?

We are aware of this bug, but so far have ignored it because it seems like it would occur very rarely and therefore may not be worth the extra complexity / computation required to handle it.

More concretely, this is only an issue when:
1. We sample an anchor within the last `block_size` elements of a sequence (which must therefore be trainable)
2. The next sequence has non-zero loss mask elements within `block_size` of the anchor.

However, most of our training sequences begin with a system prompt plus a user prompt (which have loss mask 0). And for a typical `block_size` of `8` or even `16`, the system prompt and user prompt are pretty much always going to be longer than the `block_size`, meaning no spillover actually occurs.

We can certainly add this option if you think its worthwhile, or if there's a use case I'm not seeing, but if the above makes sense to you, I'd prefer to avoid the extra complexity/computation running on every batch.

src/speculators/models/xpress/core.py
```diff
@@ -0,0 +1,361 @@
...
+            # raw mixer starts at zero (validated recipe: mixer_init=zeros);
+            # the causal tril + identity are applied functionally in forward.
+            module.mix_l.data.zero_()
+            # non-persistent buffer: not in any checkpoint, garbage after a
+            # meta-device load — recompute.
+            module.causal_tril.data.copy_(
+                torch.tril(torch.ones_like(module.causal_tril))
+            )
+
+    def _reinit_nonpersistent_buffers(self) -> None:
```
shanjiaz, line 73: Is this needed? What was the issue you ran into?

src/speculators/models/xpress/config.py
```diff
@@ -0,0 +1,83 @@
...
+    anchor_cap_to_max_valid: bool = Field(
+        default=False,
+        description=(
+            "Anchor sampling: cap the sampled anchors at "
+            "min(max_anchors, valid_candidates - 1) rather than at "
+            "valid_candidates. Worth ~0.3% of supervised blocks."
+        ),
+    )
+
+    prefix_valid_mask: bool = Field(
```
fynnsu: I don't think this needs to be in the config.

src/speculators/models/xpress/config.py
```diff
@@ -0,0 +1,83 @@
...
+        "XPress checkpoints use 2 (r=256 -> hidden 512); 4 doubles the head's MLP\n"
+        "and is a different architecture.",
+    )
+    num_jacobi_passes: int = Field(
+        default=6,
+        description="Inference-time parallel refine passes K (engine-side knob; "
+        "stored so converted checkpoints carry the validated default).",
+    )
+
+    anchor_cap_to_max_valid: bool = Field(
```
fynnsu: Probably doesn't need to be in config

src/speculators/models/xpress/config.py
```diff
@@ -0,0 +1,83 @@
...
+    prefix_valid_mask: bool = Field(
+        default=False,
+        description=(
+            "Loss mask: make the per-block validity mask a PREFIX (cumprod) "
+            "mask -- once a slot is invalid every later slot in the block is "
+            "masked too, because a left-to-right accept can never reach it."
+        ),
+    )
+
+    eval_jacobi_passes: int | None = Field(
```
fynnsu: If this is just a training time value then it shouldn't go in the model config, should just go into training time args. We try to ensure the model config only stores inference time settings.

scripts/train.py
```diff
@@ -687,6 +697,9 @@ def main(cfg: TrainConfig):  # noqa: C901
         val_call_kwargs=val_call_kwargs,
         optimizer=args.optimizer,
         weight_decay=args.weight_decay,
+        fp32_master_weights=args.fp32_master_weights,
```
fynnsu: We shouldn't need the `fp32_master_weights` logic from this pr. Speculators already initializes master weights to fp32 and uses autocast to run forward/backward in lower precision while accumulating/updating weights in fp32.

scripts/train.py
```diff
@@ -673,6 +680,9 @@ def main(cfg: TrainConfig):  # noqa: C901
         prefetch_factor=args.prefetch_factor,
         preprocess=preprocess,
         train_data_ratio=args.train_data_ratio,
+        no_packing=args.no_packing,
```
fynnsu: Why do you want to disable sequence packing?

scripts/train.py
```diff
@@ -673,6 +680,9 @@ def main(cfg: TrainConfig):  # noqa: C901
...
+        val_data_path=args.val_data_path,
+        val_transfer=val_transfer,
```
fynnsu: I'm for adding support for a separate validation set but I think it would've been better to do in a separate pr. We can leave it this time, but it may slow down review from xpress.

scripts/train.py
```diff
@@ -687,6 +697,9 @@ def main(cfg: TrainConfig):  # noqa: C901
         val_call_kwargs=val_call_kwargs,
         optimizer=args.optimizer,
         weight_decay=args.weight_decay,
+        fp32_master_weights=args.fp32_master_weights,
+        eval_interval=args.eval_interval,
+        eval_max_batches=args.eval_max_batches,
```
fynnsu: Maybe should be named `val_*` to match other `val_*` arguments?

src/speculators/models/xpress/core.py
```diff
@@ -0,0 +1,361 @@
...
+        model = cls(config=config)
+        model.load_vocab_mappings(t2d, d2t)
+        model.load_verifier_weights()
+        return model
+
+    @staticmethod
+    def get_trainer_kwargs(**kwargs) -> tuple[dict, dict]:
+        """Resolve XPress's compound loss and consistency knobs."""
+        shared = {
+            "loss_config": resolve_loss_config(kwargs["loss_fn"]),
```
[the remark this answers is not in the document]
fynnsu: Yes, this would be good to include.

src/speculators/train/optimizers.py
fynnsu, line 1: As mentioned above, I don't think the fp32 changes are required to support fp32 master weights. It also seems like these changes might not handle fsdp correctly.

src/speculators/train/trainer.py
```diff
@@ -378,15 +394,93 @@ def make_scheduler(opt: torch.optim.Optimizer):
...
+        )
+
+    def _sync_from_masters(self):
+        """Copy the stepped fp32 masters back into the model's parameters."""
+        if not self._fp32_masters:
+            return
+        with torch.no_grad():
+            for param, master in self._fp32_masters:
+                param.data.copy_(master.data)
+
```
fynnsu: Remove.

src/speculators/train/trainer.py
```diff
@@ -378,15 +394,93 @@ def make_scheduler(opt: torch.optim.Optimizer):
...
+            # object identity (and its shape/dtype/device guards) stable.
+            self._anchor_buf = torch.zeros(
+                (), device=self.local_rank, dtype=torch.float32
+            )
+        progress = min(1.0, self.global_step / self._anchor_total)
+        self._anchor_now = self._anchor_start * (1 - progress) + float(floor) * progress
+        self._anchor_buf.fill_(self._anchor_now)  # type: ignore[union-attr]
+        call_kwargs["base_anchor_weight"] = self._anchor_buf
+
+    def _reduce_metrics(self, metrics: dict[str, torch.Tensor]) -> None:
```
fynnsu: We already have logic for this. Metrics named `*_sum` and `*_total` will be summed and then `*_sum` / `*_total` will be computed to get the final metric value.

src/speculators/train/trainer.py
```diff
@@ -378,15 +394,93 @@ def make_scheduler(opt: torch.optim.Optimizer):
...
+            # stops applying, after one warning in the first minutes. A tensor carries
+            # the same number with no value guard, and reusing one buffer keeps the
+            # object identity (and its shape/dtype/device guards) stable.
+            self._anchor_buf = torch.zeros(
+                (), device=self.local_rank, dtype=torch.float32
+            )
+        progress = min(1.0, self.global_step / self._anchor_total)
+        self._anchor_now = self._anchor_start * (1 - progress) + float(floor) * progress
+        self._anchor_buf.fill_(self._anchor_now)  # type: ignore[union-attr]
+        call_kwargs["base_anchor_weight"] = self._anchor_buf
```
fynnsu: I don't think we currently changing weighting values during runs with the exception of the lr scheduler. I can see the value in adding this as a feature but I'd prefer not to add it here. It will need to be implemented in a general way. The `trainer.py` file is used by all of the drafter algorithms and therefore shouldn't implement algorithm specific logic.

I can help add support for this as a follow up pr.

src/speculators/train/trainer.py
```diff
@@ -507,6 +608,17 @@ def train_epoch(self, epoch: int):
...
+            # the gradients are already applied and zeroed. global_step is identical
+            # on every rank, so all ranks enter val_epoch's collectives together.
+            if (
+                self.config.eval_interval
+                and self.val_loader is not None
+                and self.global_step % self.config.eval_interval == 0
+            ):
+                self.val_epoch(epoch, max_batches=self.config.eval_max_batches)
+                self.model.train()  # val_epoch left it in eval mode
+
```
fynnsu: Please remove.

We can add mid-epoch validation runs in a follow up pr, but it shouldn't be included here.

fynnsu (changes requested):
Thank you for opening the pr! I've left a number of comments below.

The overarching feedback is that this pr currently makes too many unrelated changes. While some of these changes have merit, it would be better to open them as separate prs so that they can be evaluated/reviewed in isolation. For now, please clean up some of these changes and we can work on landing the main xpress components.

Thank you, and let me know if you have any questions or need help with anything.

src/speculators/models/xpress/core.py
ZKBig: Yes, this is needed when we warm-starting XPress from an existing DFlash checkpoint using from_pretrained. The problem comes from the rotary_emb.inv_freq. Because it is registered with persistent=False so it is in no checkpoint, and Qwen3RotaryEmbedding has no parameters so it is never in missing_keys. As a result, from_pretrained neither loads nor re-initialize it. So buffer actually keeps unintialized memory, and this will cause NaN numerical issue, which is met on my side when I try to use speculator to do the training.

src/speculators/models/xpress/core.py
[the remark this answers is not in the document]
ZKBig: Good catch! Agree to fix it to match DFlash/DSpark

src/speculators/models/xpress/core.py
```diff
@@ -0,0 +1,361 @@
...
+        # honest inference proxy (the analytical accept_rate above is
+        # teacher-forced). Shares the exact recurrence of the consistency loop.
+        if not self.training:
+            with torch.no_grad():
+                # The in-training accept-length rollout uses block_size - 1 passes;
+                # num_jacobi_passes (default 6) is the OFFLINE package-eval setting
+                # that converted checkpoints export, so it must not be reused here.
+                _eval_passes = getattr(self.config, "eval_jacobi_passes", None)
+                if not _eval_passes:
+                    _eval_passes = self.block_size - 1
```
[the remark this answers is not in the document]
ZKBig: Good catch! Agree to fix it

src/speculators/train/optimizers.py
```diff
@@ -59,25 +59,86 @@ def split_named_params_for_muon(
...
+    directly rounds every update to bf16's ~8-bit mantissa. Early in training an
+    update is large enough to survive that; once the LR schedule decays it falls
+    below the representable step at the weight's magnitude and is silently
+    rounded away -- training flattens while the gradients stay healthy.
+    """
+    masters: dict[str, Tensor] = {}
+    for name, param in model.named_parameters():
+        if not param.requires_grad or param.dtype == torch.float32:
+            continue
+        masters[name] = param.detach().clone().float().requires_grad_(True)
```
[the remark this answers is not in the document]
ZKBig: Good Catch! Both are real, but I plan to drop the fp32-master change entirely. As @fynnsu noted, FSDP2's MixedPrecisionPolicy(param_dtype=bf16) already keeps the sharded masters and the optimizer step in fp32, so the shadow copies were redundant. The same solution for the following 'Preserve FP32 master state across checkpoint resume.' and 'Exercise the trainer master-weight path.'.

src/speculators/models/dflash/utils.py
ZKBig: Good Catch! It is used to bit-match an off-by-one in our reference training implementation, so that loss curves can match each other when we compared the two codebases. In this case, I agree to remove it and revert dflash/utils.py to upstream.

src/speculators/models/dflash/core.py
ZKBig: That is also a good point. My original thinking is because the loss mask is still sliced across document boundaries, a block spilling into the next document gets supervised on tokens the drafter cannot even attend to. However, as you say, this would occur very rarely, so I agree to revert it.

src/speculators/models/xpress/config.py
ZKBig: Agree!

src/speculators/models/xpress/config.py
ZKBig: Sure. As I replied in previous section, agree to remove it

src/speculators/models/xpress/config.py
ZKBig: Thanks for the suggestion! Agree to move it to the training args

scripts/train.py
ZKBig: Thanks for this remind. Let me remove it.

scripts/train.py
ZKBig: It actually feeds the packer a constant length equal to the token budget, so each bin fits exactly one conversation (Sorry for the misleading  name). That makes the global batch a count of conversations rather than a token budget, which is what our reference recipe specifies. I needed it only to reproduce that recipe faithfully so the loss curves were comparable across the two codebases. So I think we can remove this one to avoid the complexity.

scripts/train.py
ZKBig: Sure, I agree to remove it. Thanks  for telling me this.

scripts/train.py
ZKBig: Same as the previous one, let me remove it.

src/speculators/train/optimizers.py
ZKBig: Sure. As I replied previosuly, I agree to remove all codes about fp32 changes.

src/speculators/train/trainer.py
ZKBig: Agree to remove！

src/speculators/train/trainer.py
ZKBig: Thanks for this important catch. Hardcoding base_anchor_weight / base_anchor_floor into a shared file is not the idea way for this PR, agree to remove this.

While for XPress, it does help in our runs. The anchor term keeps the co-trained backbone from being dragged by the untrained refiner early on, but pulls against the real objective later. So a general weight scheduler would be useful to us.

src/speculators/train/trainer.py
ZKBig: Agree to remove it. This one is also used for the faithful reproduce so that I can match each other when we compared the two codebases.

src/speculators/models/xpress/core.py
shanjiaz: Thanks! Very weird. Not a blocker but we might want to create a issue on huggingface for this.

examples/train/xpress_morph_config.py
```diff
@@ -0,0 +1,69 @@
+"""Re-type a converted DFlash checkpoint as an XPress starting point.
```
shanjiaz, line 1: I think we should make a todo here, instead of having a morph config script, we have a unified warm start feature that allows warm starting xpress and dspark from a dflash/dflash2 model.

src/speculators/models/xpress/core.py
```diff
@@ -0,0 +1,309 @@
...
+        verifier_last_hidden_states: torch.Tensor,
+        document_ids: torch.Tensor,
+        position_ids: torch.Tensor | None = None,
+        loss_config: LossConfig | None = None,
+        gamma: float = 4.0,
+        max_anchors: int = 3072,
+        per_position_loss_weight: str = "fixed-exp-decay",
+        dpace_alpha: float = 0.5,
+        consistency_weight: float = 0.3,
+        consistency_passes: int = 3,
```
shanjiaz, line 188: Just curious, does `consistency_passes` here have to match what we use during inference?

shanjiaz (commented):
I think it looks good to me, just pending regression test.

examples/train/xpress_qwen3_8b_regen_online.sh
fynnsu, line 1: Could you simplify this script to align more with the other examples we have. e.g. https://github.com/vllm-project/speculators/blob/main/examples/train/dflash2_qwen3_8b_ultrachat_online_5k.sh

src/speculators/models/xpress/core.py
```diff
@@ -0,0 +1,309 @@
...
+    Training runs one teacher-forced refine pass plus ``consistency_passes``
+    free-running rounds (each conditioned on the previous round's argmax, the
+    exact recurrence of inference-time Jacobi decoding, stop-grad between
+    rounds), and anchors the co-trained backbone with a base-logits term.
+    Everything else is inherited from DFlash.
+    """
+
+    config_class: ClassVar[type[XPressSpeculatorConfig]] = XPressSpeculatorConfig  # type: ignore[misc]
+
+    def _init_weights(self, module) -> None:
```
shanjiaz, line 48: I'm anticipating as we land future support for warm starting dspark & xpress from dflash & dflash2, we'll run into similar issues. We should probably move _init_weights and _reinit_nonpersistent_buffers to [SpeculatorModel](https://github.com/vllm-project/speculators/blob/main/src/speculators/model.py#L213) for better model loading support. Seems like that's how transformers handle their nonpersistent buffers, see this [example](https://github.com/huggingface/transformers/blob/main/src/transformers/models/deepseek_v4/modeling_deepseek_v4.py#L1245). @fynnsu How do you feel?

src/speculators/losses/utils.py
```diff
@@ -292,13 +297,20 @@ def loss_function(
...
         elementwise_loss = elementwise_loss * decay_mult
 
-    denominator = loss_mask.sum(dim=1) + _LOSS_REDUCTION_EPS
+    if denom_decay and decay_mult is not None:
+        # A decay that scales the numerator but not the denominator lowers the
+        # reported loss without changing the gradient direction; dividing by the
+        # decayed weight sum makes the term a true weighted mean.
+        denominator = (loss_mask * decay_mult).sum(dim=1) + _LOSS_REDUCTION_EPS
+    else:
+        denominator = loss_mask.sum(dim=1) + _LOSS_REDUCTION_EPS
```
fynnsu, line 313: I'd prefer to have this in a follow up pr.

src/speculators/models/xpress/core.py
fynnsu: Yeah we should probably standardize this. Since this is primarily related to loading pretrained dflash checkpoints, you could leave it out of this pr and we'll resolve it in a general way in a future pr.

fynnsu (commented):
Thank you for updating. Added a few more comments below.

src/speculators/train/trainer.py
fynnsu: I've got an initial implementation here: https://github.com/vllm-project/speculators/pull/1130. Note: this is currently just targeting loss weight scheduling across different loss functions but for the same target. 

So it doesn't yet handle your use case well because it doesn't look at the auxiliary loss sources. That being said, you could already implement the loss weight scheduling with this because it will scale the primary loss, while keep the other losses fixed.

PR #1064, opened 2026-08-31
feat(eval): add MRCR mode for long-context acceptance rate benchmarks

reed-meyerson:
## Summary

Adds a **`long-context`** evaluation mode that measures speculative-decoding acceptance as context length grows, using [OpenAI's MRCR dataset](https://huggingface.co/datasets/openai/mrcr) as a source of long, multi-turn prompts. Selection is **deterministic** and **monotonic in `max_model_len`** (a larger context window runs a superset of a smaller one), cost is bounded by construction, and the per-request machinery is factored into a dataset-agnostic harness that future benchmarks can reuse.

Correctness is intentionally not graded — MRCR is used only for its long prompts, not to measure retrieval accuracy.

### How it works

- **Deterministic superset selection.** Records are stratified into MRCR's native power-of-2 token bins and, within each bin, ranked by a stable content hash (`sha256(seed + prompt)`); the lowest-ranked `--samples-per-bin` per bin form the candidate set. Bin assignment uses an `n_chars / 4.9` token estimate (MRCR carries no token count; the ratio is 4.9 ± 0.1 across the whole dataset, validated offline against o200k) — zero runtime cost, no `tiktoken` dependency. The candidate set is independent of `max_model_len`; the only context-dependent step is the exact render fit-test, which is monotonic, so `Kept(ctx1) ⊆ Kept(ctx2)` for `ctx1 ≤ ctx2`.
- **Exact lengths + fit via the server's own render endpoint.** Each candidate is rendered through `/v1/chat/completions/render`, applying the chat template and tokenizing exactly as the server would — giving an exact prompt length and naturally rejecting anything that doesn't fit, with no separate tokenizer. Requires the server started with `VLLM_ENABLE_SCALE_OUT_ENDPOINTS=1`.
- **Bounded cost.** Rendering is capped to ~`samples_per_bin × n_bins`, done concurrently smallest-first, and stops early once a whole bin renders oversized (all larger bins are strictly bigger). Generation runs only on what fits.
- **Per-request detailed metrics.** Generation uses `--per-request-spec-decode-metrics detailed`, so each response carries its own acceptance stats plus per-verify-step arrays. Every raw result is streamed to `raw_requests.jsonl` as the source of truth; the aggregated CSVs are one slicing of it, rebuildable offline with different bin edges (no re-inference).
- **Two report slicings.** Acceptance by prompt length at request start, and by running token position within each request.

### Code organization

- **`spec_acceptance.py`** (new) — dataset-agnostic per-request acceptance harness: stratified selection, render/generate, raw-log persistence, binning, reporting. Reusable by `throughput`/`sweep` in a future PR.
- **`mrcr_bench.py`** — thin MRCR adapter (record → messages / selection key / length proxy); all inference/binning/IO delegates to the harness.
- **`rebin_mrcr.py`** — re-slice a saved `raw_requests.jsonl` with different `--context-bin-edges` / `--position-bin-size`, no server needed.
- **`evaluate.py`** — dispatch + new flags; `throughput`/`sweep` unchanged.
- **`plot.py`** — MRCR acceptance-vs-length scatter/binned-mean plot.
- **`_fetch_model_info()`** — returns `id` + `max_model_len` from `/v1/models`.

### CLI args (long-context mode)

- `--samples-per-bin` (default 20) — records per native token bin; density/cost knob
- `--selection-seed` (default 0) — mixed into the content hash to pick a different deterministic sample
- `--position-bin-size` (default 256) — token-position bin width for the by-position report

### Dependencies

Long-context mode needs only `pandas`, `pyarrow`, `huggingface_hub` (dataset loading); requests go straight to the target server's HTTP API. Drops the previous `inspect-ai`, `openai`, `tiktoken`, and `scikit-learn` requirements.

### Usage

\`\`\`bash
# Server must expose the render endpoint and per-request detailed metrics:
VLLM_ENABLE_SCALE_OUT_ENDPOINTS=1 vllm serve <model> \
    --per-request-spec-decode-metrics detailed \
    --speculative-config '{"model":"<draft>","num_speculative_tokens":N}'

python evaluate.py --target http://localhost:8000/v1 long-context --samples-per-bin 20

# Re-slice results offline with different bins (no server):
python rebin_mrcr.py --raw-log <out>/raw_requests.jsonl --position-bin-size 512
\`\`\`

### DIFF
--- a/scripts/evaluate/evaluate.py
+++ b/scripts/evaluate/evaluate.py
@@ -4,7 +4,7 @@
 Modes:
     throughput   Max throughput run for acceptance rates
     sweep        Full pipeline (gen-len, sweep, CSV)
-    mrcr         Long-context acceptance rates via OpenAI's MRCR dataset,
+    long-context Long-context acceptance rates via OpenAI's MRCR dataset,
                  bucketed by context length (uses Inspect AI as the harness)
 
 Examples:
@@ -21,8 +21,8 @@ Examples:
         --dataset speedbench/qualitative/coding \\
         --speedbench-data-dir ./speedbench_data
 
-    # MRCR (long-context acceptance rate by bucket, capped at 128k tokens):
-    python evaluate.py --target http://localhost:8000/v1 mrcr \\
+    # Long-context (acceptance rate by context-length bucket, capped at 128k tokens):
+    python evaluate.py --target http://localhost:8000/v1 long-context \\
         --mrcr-max-context 131072 --mrcr-max-samples-per-bucket 20
 """
 
@@ -303,7 +303,7 @@ def run_benchmark(args: argparse.Namespace) -> None:
 
     save_eval_provenance(output_dir)
 
-    if args.mode == "mrcr":
+    if args.mode == "long-context":
         from mrcr_bench import run_mrcr
         model_info = _fetch_model_info(args.target)
         run_mrcr(
@@ -412,11 +412,11 @@ def main() -> None:
     )
     parser.add_argument(
         "mode",
-        choices=["throughput", "sweep", "mrcr"],
+        choices=["throughput", "sweep", "long-context"],
         help=(
             "throughput: max-rate run for acceptance rates; "
             "sweep: full benchmarking pipeline; "
-            "mrcr: long-context acceptance rates via OpenAI's MRCR dataset"
+            "long-context: acceptance rates across context-length buckets"
         ),
     )
     parser.add_argument(
--- a/scripts/evaluate/evaluate.py
+++ b/scripts/evaluate/evaluate.py
@@ -293,85 +293,6 @@ def _run_subset(
     return acceptance_csv, perf_csv, max_tokens if is_sweep else None
 
 
-def _run_mrcr(args: argparse.Namespace, metrics_url: str, output_dir: Path) -> None:
-    """Drive MRCR long-context prompts through the target server, bucket by
-    bucket, diffing spec-decode Prometheus counters around each bucket's run.
-
-    Uses Inspect AI as the eval harness (dataset loading + concurrent
-    generation against the target's OpenAI-compatible endpoint via the
-    ``vllm`` model provider). Correctness is not graded here — only
-    acceptance metrics are collected.
-    """
-    from inspect_ai import eval as inspect_eval
-
-    from mrcr_bench import build_task, load_mrcr_buckets, warn_if_oversized
-
-    model_info = _fetch_model_info(args.target)
-    if model_info is None:
-        logger.error("Could not determine served model info from %s/models", args.target)
-        sys.exit(1)
-    model_name = model_info["id"]
-    max_model_len = model_info.get("max_model_len")
-    logger.info(
-        "Server reports model=%s max_model_len=%s", model_name, max_model_len
-    )
-    logger.info(
-        "Loading MRCR 2-needle dataset (max_context=%d, max_samples_per_bucket=%s)...",
-        args.mrcr_max_context,
-        args.mrcr_max_samples_per_bucket,
-    )
-    buckets = load_mrcr_buckets(
-        max_context=args.mrcr_max_context,
-        max_samples_per_bucket=args.mrcr_max_samples_per_bucket,
-    )
-    if not buckets:
-        logger.error(
-            "No MRCR buckets available at or below --mrcr-max-context=%d",
-            args.mrcr_max_context,
-        )
-        sys.exit(1)
-
-    root_url = args.target.rstrip("/").removesuffix("/v1")
-    log_dir = output_dir / "artifacts" / "mrcr_logs"
-    acceptance_csv: CsvWriter | None = None
-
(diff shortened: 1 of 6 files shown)


scripts/evaluate/evaluate.py
```diff
@@ -78,7 +84,8 @@
 )
 
 
-def _fetch_model_name(target: str) -> str | None:
+def _fetch_model_info(target: str) -> dict | None:
```
orestis-z: nit: if you know more about the structure of models[0] maybe you can include that (e.g. `dict[str, Any]`

fynnsu (approved):
Tested locally and works well

reed-meyerson:
Pushed a substantial rework of this mode (force-updated the branch). The high-level goal is unchanged — long-context spec-decode acceptance via MRCR — but the approach changed enough that I've rewritten the PR description above. Summary of the pivot:

- **Deterministic, superset selection** replaces the old `--mrcr-max-context` / `--mrcr-max-samples-per-bucket` knobs. The selected set is now a pure function of the dataset, flags, and the server's `max_model_len`, and is monotonic: raising `max_model_len` runs a superset of a smaller run. New knobs: `--samples-per-bin`, `--selection-seed`, `--position-bin-size`.
- **Direct HTTP instead of Inspect AI.** Requests go straight to the server's `/v1/chat/completions/render` (exact lengths + fit-test) and `/v1/chat/completions` (with `--per-request-spec-decode-metrics detailed`). Drops the `inspect-ai` / `openai` / `tiktoken` / `scikit-learn` deps.
- **Per-request detailed metrics** streamed to `raw_requests.jsonl` (source of truth), replacing the Prometheus-counter diffing; reports are rebuildable offline via `rebin_mrcr.py`.
- **Generic harness extracted** to `spec_acceptance.py` so `throughput`/`sweep` can adopt the same per-request reporting later.

Validated end-to-end against Qwen3-8B + a DFlash speculator at `--max-model-len 40960`: bounded render count with the bin early-stop firing, both CSV reports produced, detailed per-step metrics flowing through, and offline rebin + plot both reproducing from the raw log.

scripts/evaluate/evaluate.py
```diff
@@ -459,6 +489,37 @@ def main() -> None:
...
     )
+    lc_group = parser.add_argument_group("long-context mode")
+    lc_group.add_argument(
+        "--samples-per-bin",
+        type=int,
+        default=DEFAULT_SAMPLES_PER_BIN,
+        help=(
+            "long-context: records to run per native MRCR token bin. Larger "
+            "context windows fill more bins, so the selected set at a bigger "
+            f"--max-model-len is a superset of a smaller one (default: "
```
fynnsu, line 500: I'm not sure I'm completely following this.

scripts/evaluate/evaluate.py
```diff
@@ -459,6 +489,37 @@ def main() -> None:
...
+            "rank within each bin; changing it picks a different deterministic "
+            "sample (default: 0)"
+        ),
+    )
+    lc_group.add_argument(
+        "--position-bin-size",
+        type=int,
+        default=DEFAULT_POSITION_BIN_SIZE,
+        help=(
+            "long-context: token-position bin width for the by-position report "
```
fynnsu, line 519: I also find this a little confusing because this is a different kind of bin than the `--samples-per-bin` arg. 

Maybe we can make it clearer that `--samples-per-bin` is the number of samples per **preset** MRCR bin vs this value which is more for graphing?

scripts/evaluate/plot.py
```diff
@@ -367,6 +374,273 @@ def run_speedup(args: argparse.Namespace) -> None:
...
+    _set_log2_ticks(ax)
+    ax.set_title(title, fontsize=13, fontweight="bold")
+    ax.set_xlabel("Context length (tokens, token position)", fontsize=11)
+    ax.set_ylabel("Acceptance length", fontsize=11)
+    ax.grid(True, alpha=0.3, zorder=0)
+    ax.spines["top"].set_visible(False)
+    ax.spines["right"].set_visible(False)
+
+
+def run_mrcr(args: argparse.Namespace) -> None:
```
fynnsu, line 583: Kinda confusing name given the other `run_mrcr` fn, and this seems to be just plotting logic.

fynnsu (changes requested):
I've listed a couple smaller things below but I think we need to clean up the UX a bit. (Also please fix DCO / merge conflict)

1. I think we need to make the `--per-request-spec-decode-metrics detailed` requirement clearer. There seems to be some logic for triggering this on a test sample, but when I tried locally it didn't raise an error and instead just iterated through the benchmark failing on every request.
2. This is adding 3 new files. I feel like we should be able to consolidate some of these. 

3. x-axis on this should maybe be in non-exponent form to match the context length at request start chart.
4. I find the "acceptance by context length at request start" and "acceptance by context length at token position" nomenclature a little confusing. Maybe something like "acceptance by input sequence length" and "acceptance at token position" is a bit clearer?


Image

PR #1068, opened 2026-09-01
Update kv cache extraction following restructure

fynnsu:
## Purpose

Following https://github.com/vllm-project/vllm/pull/51718, which changed the kv cache layout the connector uses, hidden states extraction (using mooncake connector) was failing with a shape issue. This pr updates the `extract_from_kv_cache` function to handle both kv cache layout options (since the new one is only on the nightly release). 

## Tests

Added tests for both layout structures. Will validate with CI + smoke tests. 

## Checklist

I have filled in:


shanjiaz (approved):
Thanks for fixing!

PR #1070, opened 2026-09-01
Add job checks for Unit, Integration, and Smoke tests

fynnsu:
## Purpose

This adds status results for each subclass of tests (unit, integration, smoke) which should appear in the github ui.

I've intentionally used the same context key (e.g. "buildkite/speculators/unit-tests") for both the H100 and L4 tests. This will allow more seamless switching between runner types (since we won't have to update branch protection rules). However, it also means that we should make sure we don't run both the L4 and H100 tests together. 

## Tests

Will check if this runs in the pr, but I'm guessing it will have to land first to go into effect. 

## Checklist

I have filled in:

### DIFF
--- a/.buildkite/gpu-tests/gpu-tests-H100.yml
+++ b/.buildkite/gpu-tests/gpu-tests-H100.yml
@@ -1,5 +1,8 @@
 steps:
   - group: ":test_tube: Unit Tests (H100)"
+    notify:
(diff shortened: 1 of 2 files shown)


deepak-kumar-neu (approved):
lgtm!

shanjiaz (approved):
Nice

PR #995, opened 2026-08-14
set default max anchor to 512

shanjiaz:
## Purpose
Our max anchor is too high and makes each step slow

## Tests
current dflash runs have been using 512

## Checklist

I have filled in:

- [x] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [x] The test plan/results, such as providing test command and pasting the results.
- [ ] (Optional) The necessary documentation update.
- [x] I (a human) have written or reviewed the code in this pr to the best of my ability.

PR #1071, opened 2026-09-02
fix(training): Fix regression on DFlash/DFlash2/DSpark: --max-anchors is now per packed document, not per packed sequence

WindChimeRan:
## Summary

`--max-anchors` now means anchors **per packed document** instead of per packed sequence. This is the contract `num_anchors` has in SpecForge and TorchSpec, which sample `[B, num_anchors]` with one conversation per row. It applies to DFlash, DFlash2 and DSpark, which share the one `select_anchors` call in `DFlashDraftModel._backbone_forward`. P-EAGLE is untouched.

## Why

Speculators packs several conversations into one `total_seq_len` row. `select_anchors` drew `min(max_anchors, valid)` positions across the whole row and ignored `document_ids`, so the 512 default copied from the reference trainers in #995 gave 2-8x fewer anchors per token than those trainers, and short conversations were starved: one with 187 supervised tokens got about 28 anchors here versus all 187 in SpecForge.

On the tutorial dataset (`windchimeran/speculator-tutorial:tutorial_regen`, mean 3046 tokens, 54% supervised) packed to 8192, per-row 512 covered 14.7% of supervised tokens; per-document 512 covers 30.8%.

## How it works

- Every document in the row gets exactly `max_anchors` slots, filled with a uniform sample of its supervised positions and padded with `anchor_valid == False` when it has fewer (`select_anchors`, `models/dflash/utils.py`). Short documents are fully covered, long ones capped.
- Query length is `num_documents * max_anchors * block_size`, so it depends only on the document count. The family forwards and the block-mask builder compile with `dynamic=False`, giving one static compile per document count. Counts above 8 round up to a power of two, bounding the total at 18 variants.
- Nothing in the compiled forward has a data-dependent shape, so the recompile limits cannot be reached by data. Hitting one means a bug, and dynamo's default reaction (silently running eager) would surface later as an OOM on this memory-heavy path, so `fail_on_recompile_limit_hit` makes it fail immediately instead.
- DSpark derives `num_blocks` from the sampled tensor rather than the flag; its stale 3072 fallback default becomes 512, matching the schema.

## Validation

DSpark trained on 4x H100 80GB, `main` vs this branch, identical data, cache and seed. Qwen3-8B verifier, `tutorial_regen` (4993 samples), offline hidden states, `--num-layers 5 --epochs 5 --total-seq-len 8192 --lr 3e-4 --loss-fn '{"ce":0.1,"tv":0.9}' --draft-vocab-size 32000 --save-best --seed 42`, `torchrun --nproc_per_node 4`. torch 2.13.0+cu130, vLLM 0.28.0, default `simple_flex_attention`, `--max-anchors` left at the 512 default on both arms. Single seed.

Acceptance length from `evaluate.py throughput` over all nine `RedHatAI/speculator_benchmarks` subsets, serving `checkpoint_best`:

| subset | main | this PR | delta |
|---|---|---|---|
| HumanEval | 2.1580 | 2.2203 | +0.0623 |
| math_reasoning | 2.1896 | 2.2927 | +0.1031 |
| qa | 2.0384 | 2.1583 | +0.1199 |
| question | 2.0694 | 2.1306 | +0.0612 |
| rag | 2.0866 | 2.1793 | +0.0927 |
| summarization | 1.8577 | 1.9345 | +0.0768 |
| tool_call | 2.1117 | 2.1678 | +0.0562 |
| translation | 1.8318 | 1.8930 | +0.0612 |
| writing | 2.0643 | 2.1297 | +0.0654 |
| **mean** | **2.0453** | **2.1229** | **+0.0776 (+3.8%)** |

Acceptance length improves on all nine subsets. The `main` mean of 2.0453 reproduces the 2.02 reported for DSpark in #1036.

Training cost and convergence over the same runs:

| | main | this PR |
|---|---|---|
| wall time, 5 epochs | 17m38s | 22m07s (+25%) |
| median step | 340 ms | 411 ms (+21%) |
| peak allocated | 28.79 GiB | 43.21 GiB (+50%) |
| peak reserved | 29.95 GiB | 51.06 GiB (+71%) |
| best val loss | 0.640 | 0.634 |
| val loss per epoch | .733 .690 .666 .648 .640 | .718 .673 .648 **.634** .638 |
| `checkpoint_best` | epoch 4 | epoch 3 |

Validation loss is lower at every epoch. Peak memory includes optimizer state and gradients.

Compiled-variant count on this dataset, from the sampler log:

```
Epoch 0: 471 packed batches holding 1-4 samples each (4 distinct counts)  [train]
Epoch 0:  52 packed batches holding 1-4 samples each (4 distinct counts)  [val]
```

Four document counts, about eight variants across train and val, against a limit of 64. The power-of-two rounding never fires on this data; it is insurance for datasets with many short documents.

Per-document scaling, measured separately as a microbenchmark (GB10, sdpa, compiled forward+backward at Qwen3-8B draft size, activation memory only, no optimizer state):

| docs in row | anchors | step | peak activation memory |
|---|---|---|---|
| 1 | 512 | 303 ms | 4.3 GiB |
| 2 | 1024 | 922 ms | 5.9 GiB |
| 4 | 2048 | 1840 ms | 9.2 GiB |
| 6 | 3072 | 3355 ms | 12.7 GiB |

## Pitfalls

- **512 is now per document, not per row.** If you tuned `--max-anchors` as a per-row budget, divide it by your typical documents per row. Example recipes move from 3072 to the reference default 512; the Kimi K3 recipe from 1024 to 512.
- **Memory and step time scale with documents per row**, the way they scale with batch size in the reference trainers. Budget from the end-to-end numbers above, not the activation-only microbenchmark.
- **Published results measured under the old semantics shift when rerun**, including the tutorial in #1036.
- **Recipes may now be one epoch too long.** More supervision per step converges faster, so `checkpoint_best` moved from epoch 4 to epoch 3 on the 5-epoch tutorial recipe.

## Tests

- `tests/unit/models/test_dflash_utils.py`: slots per document, short document covered and padded, only supervised positions, empty row.
- `tests/integration/models/test_model_forward.py::test_no_recompile_across_loss_masks`: a fixed document count stays one compiled graph as the loss mask varies, guarding the static-shape invariant that `fail_on_recompile_limit_hit` relies on.
- CPU: `tests/unit/models`, `tests/unit/train/config` (283 passed).
- GB10: DFlash/DSpark/DFlash2 integration and parity tests (69 passed); recompile log shows static specializations only.
- 4x H100: the end-to-end DSpark runs in Validation above.

## Checklist

- [x] The purpose of the PR
- [x] The test plan/results
- [x] The necessary documentation update


fynnsu:
I'm not sold on this change. This will significantly increase the number of padded/non-valid tokens we compute, and although the pr may cap the number of forward compilations, this will still increase it significantly and add another source of failure / slowdown.

We also introduce new rank divergence with this change, and some different ranks will now have batches of varying sizes, which the current reduction logic doesn't account for (not a big deal, mostly affects logged metric correctness but still). 

I also believe the ablation improvement can largely be explained by the increase in average anchors per batch that the pr branch sees. It is being trained on strictly more data and also in turn seeing additional runtime/hardward usage consistent with this. Likely training the main branch with a slightly higher `max-anchors` to match the average `max-anchors` of the pr, would yield a similar result.

WindChimeRan:
@fynnsu 

I think the regression issue is true (default from 3072 to 512). This is actually far smaller than the papers' setting. 

I also believe our doc packing training is the right way, but it's just make this fix non-trivial. (You cannot have fixed tokens, fixed anchors, and per-document quotas at once. Main kept the first two by giving up the third.)


* **option 1**: keep the current 512 default (I prefer this one today)
* option 2: revert to 3072 (it still doesn't match the papers, it's just an approximation)
* previous attempts: just use a for loop to do 512 per doc, the code looks clean but it results in OOM, because of dynamic compilations
* this PR: I agree with you that the current solution is not ideal (that's also why I mark it as a draft). 

My plan is to do some training data analysis to see the valid anchor distribution first (we also need to consider loss mask), then maybe there will be some obvious solutions. 

WDYT?

dsikka:
@mergifyio refresh

PR #1072, opened 2026-09-02
Update hs_connector to use upstream extract_from_kv_cache

fynnsu:
## Purpose

Following issues with getting the num hidden states from vllm attributes which maybe change between versions, I've decided to go a different approach from #1068. Instead we make use of the example_hidden_states_connector `extract_from_kv_cache` fn, which does get updated when the kv cache layout changes and should have a consistent api. 

## Tests

Ran hs_connectors e2e tests locally. 

## Checklist

I have filled in:


rahul-tuli (approved):
LGTM!

PR #1081, opened 2026-09-03
Make vocab mapping usage more explicit

fynnsu:
## Purpose

Fixes a few issues with vocab mappings.

Currently we cache vocab mappings to `t2d.npy` and `d2t.npy` in the data folder
and automatically load them if they're there. This can result in reduced vocabs
being used unintentionally (without specifying --draft-vocab-size). This commit 
changes it so we:
1. Only look for the files if `--draft-vocab-size` is explicitly passed.
2. Caches the files to `t2d-32000.npy` (where 32000 is the vocab size)

Together, these changes ensure we only ever look for and load vocab mappings that match
the provided `--draft-vocab-size`. This should also fix a nightly failure we've seen with mismatched vocab mapping sizes. This was caused by multiple runs creating/reusing cached vocab mappings. 

Also removes old `build_vocab_mapping.py` script as vocab mapping creation is handled implicitly by `speculators train`. 

## Tests

Added some tests. Will validate using CI. 

## Checklist

I have filled in:


src/speculators/train/config/schema.py
```diff
@@ -197,9 +197,9 @@ class DraftArgs(_Group):
 class DataArgs(_Group):
     data_path: str = Field(
         default="./output",
-        description="Root data directory with the preprocessed dataset, vocab "
-        "mappings (d2t.npy, t2d.npy), token frequencies (token_freq.pt), and hidden "
-        "states.",
+        description="Root data directory with the preprocessed dataset, size-keyed "
```
shanjiaz, line 200: what does `size-keyed` mean?

shanjiaz (approved):

src/speculators/train/config/schema.py
fynnsu: Sorry the wording is maybe a little claude-y (codex-y?). It just means using `d2t-32000.npy` and `t2d-32000.npy` where `32000` is the size.

PR #1083, opened 2026-09-04
fix(utils): isolate inherited class registries

edgeai1:
## Purpose

  Prevent a subclass from mutating its parent class registry. A subclass inherits the parent's registry dictionary until
  its first registration, so registering on the subclass previously added entries to the parent registry as well.

  ## Tests

  - `tests/unit/utils/test_registry.py`
  - `tests/unit/utils/test_pydantic_utils.py`

  Both test modules pass: 22 tests.

## Checklist

I have filled in:


tests/unit/utils/test_registry.py
```diff
@@ -169,6 +169,26 @@ class TestClass2:
...
+    class ParentClass:
+        pass
+
+    class ChildRegistry(ParentRegistry):
+        pass
+
+    @ChildRegistry.register()
+    class ChildClass:
+        pass
+
```
rahul-tuli, line 189: nit: Could we also add a `ParentRegistry.registry is not ChildRegistry.registry`

rahul-tuli (approved):
Thanks for catching this!

fynnsu (approved):

PR #1088, opened 2026-09-06
feat(metrics): add shared reference-prefix metrics across drafters

WindChimeRan:
## What changes

Add `reference_acc_at_pos_i` across DFlash, DSpark, DFlash2, EAGLE3, P-EAGLE, and MTP.

**Existing training metrics use different references, horizons, and denominators, making comparisons between drafters difficult.** This PR adds one shared definition for reference-prefix agreement.

Existing metrics stay available. Removing the legacy accuracy and acceptance reports is a separate draft: #1102.

## Metric definition

`reference_acc_at_pos_i` measures how often all draft tokens through zero-based position i match the stored continuation.

For example, `[match, mismatch, match]` succeeds at position 0 and fails at positions 1 and 2.

A start is counted only when every prediction through i is available and every reference token is supervised within the same non-padding document. Counts are pooled across batches and ranks before computing the rate.

Reference matching and serving acceptance differ, especially with sampling. See the [training-metrics guide](https://github.com/WindChimeRan/speculators/blob/codex/shared-acceptance-metrics/docs/user_guide/training_metrics.md) for metric semantics and limitations.

## Cost and tests

The metric adds no teacher forward or projection. GPU training-throughput overhead remains unmeasured.

## Checklist

- [x] Purpose and metric definition are documented.
- [x] Test results are included.
- [x] Existing metrics remain available during migration.
- [x] I (a human) have written or reviewed the revised code in this PR to the best of my ability.


src/speculators/train/trainer.py
```diff
@@ -530,7 +530,13 @@ def train_epoch(self, epoch: int):
 
                 metrics = {k: v.item() for k, v in metrics.items()}
                 world_size = dist.get_world_size() if self.is_distributed else 1
+                reference_counts = {
+                    k: int(v)
+                    for k, v in metrics.items()
+                    if k.startswith("reference_acc_at_pos_")
+                }
                 metrics = normalize_counted_metrics(metrics, world_size)
+                metrics.update(reference_counts)
```
shanjiaz, line 550: Not a blocker, just curious, when running the experiments, have we observed any slowdown? Potentially we don't have to calculate every step if we're not logging every step.

shanjiaz (approved):
I think the logic looks correct to me. One small comment. Thanks for adding this! @WindChimeRan Can you resolve quality please.

src/speculators/train/trainer.py
WindChimeRan: image

from my previous exp

shanjiaz: Nice!

WindChimeRan: > Potentially we don't have to calculate every step if we're not logging every step.

I think we should calculate every step, and log them in an accumulated way. (same pattern in other training libs)

* Before: report only the current batch
* Now: Training metrics pool all batches since the previous log across ranks. Any remaining batches are logged at epoch end or the step limit. Validation metrics are pooled across all validation batches and ranks.

The cost should be the same.

src/speculators/train/trainer.py
fynnsu, line 1: How are the changes in this file necessary for the shared metric?

docs/user_guide/training_metrics.md
fynnsu, line 1: This isn't very clear and is kind of strange to be talking about just this one metric plus details on internal things like `_sum` and `_total` counts. 

Let's keep this diff simple and remove and docs changes. If we're going to add a training metrics doc page, I'd much rather it be done in a separate pr, in a systematic way rather than included here alongside minor touchups to the dflash2 alg page.

src/speculators/models/dflash2/core.py
```diff
@@ -167,12 +168,24 @@ def forward(
...
         )
+        candidate_logits = scored_logits[..., :-1]
+        with torch.no_grad():
+            original_scores = torch.cat(
+                [candidate_logits.detach()[..., :-1], scored_logits.detach()[..., -1:]],
+                dim=-1,
+            )
+            pred_ids = candidate_ids.gather(
+                -1, original_scores.argmax(dim=-1, keepdim=True)
+            ).squeeze(-1)
```
fynnsu, line 189: This needs to be simplified somehow. I'd prefer to have a slightly biased prediction (with the injected teacher target) than this extra logic which is very tied into the implementation of score candidates and the score processing.

src/speculators/models/eagle3/core.py
```diff
@@ -358,6 +359,17 @@ def forward(  # noqa: C901
             # shape: [1, total_seq_len]
 
         if return_loss:
+            if draft_tokens:
```
fynnsu, line 362: Since we require `ttt_steps>0` and `draft_tokens is always appended to on each iteration of the loop above, this condition should always be true. Let's remove the `if`

src/speculators/models/mtp/core.py
```diff
@@ -217,7 +221,20 @@ def forward(
         # Cap steps so short sequences still produce partial results.
         effective_steps = min(num_steps, max(0, seq_len - 2))
         valid_len = seq_len - effective_steps - 1
+        horizon = num_steps
+        prediction_ids = input_ids.new_zeros(batch_size, max(valid_len, 0), horizon)
+        observed = torch.arange(horizon, device=device) < effective_steps
         if valid_len <= 0 or effective_steps == 0:
```
fynnsu, line 227: These are defined here but no used till later?

src/speculators/models/metrics.py
```diff
@@ -94,3 +94,120 @@ def compute_accuracy_multi_step(
     total_per_pos.scatter_add_(0, pos_idx, torch.ones_like(correct, dtype=torch.float))
 
     return correct_per_pos, total_per_pos  # shape: [num_pos], [num_pos]
+
+
+@torch.no_grad()
+def compute_reference_prefix_metrics(
+    pred_ids: torch.Tensor,  # [batch, starts, horizon], in draft vocabulary
```
fynnsu, line 101: What are `starts` and `horizon`? Do these correspond to `num_anchors` and `block_size`? If so, can we reuse those for consistency?

src/speculators/models/metrics.py
```diff
@@ -94,3 +94,120 @@ def compute_accuracy_multi_step(
...
+    input_ids: torch.Tensor,  # [batch, sequence], in verifier vocabulary
+    first_target: torch.Tensor,  # [starts] or [batch, starts]
```
fynnsu, line 103: Why are there two shape options here?

fynnsu: Also not clear what `first_target` is.

src/speculators/models/metrics.py
```diff
@@ -94,3 +94,120 @@ def compute_accuracy_multi_step(
...
+) -> dict[str, torch.Tensor]:
+    """Count complete matching prefixes at every configured draft position.
+
+    Zero-based position i uses starts with i+1 observed, supervised references
+    in the anchor's document. Eligibility is independent of prediction matches.
+    ``d2t`` stores offsets: verifier_id = draft_id + d2t[draft_id].
+    """
+    horizon = pred_ids.shape[-1]
+    seq_len = input_ids.shape[1]
+    if seq_len == 0 or horizon == 0:
```
fynnsu, line 118: What is this case handling? When would this occur?

src/speculators/models/metrics.py
```diff
@@ -94,3 +94,120 @@ def compute_accuracy_multi_step(
...
+@torch.no_grad()
+def compute_reference_prefix_metrics(
+    pred_ids: torch.Tensor,  # [batch, starts, horizon], in draft vocabulary
+    input_ids: torch.Tensor,  # [batch, sequence], in verifier vocabulary
+    first_target: torch.Tensor,  # [starts] or [batch, starts]
+    loss_mask: torch.Tensor | None,
+    document_ids: torch.Tensor | None,
+    *,
+    d2t: torch.Tensor | None = None,
+    observed: torch.Tensor | None = None,
```
fynnsu, line 108: What is observed?

src/speculators/models/metrics.py
```diff
@@ -94,3 +94,120 @@ def compute_accuracy_multi_step(
...
+            valid &= loss_mask.gather(1, safe_positions).reshape_as(pred_ids).bool()
+        if document_ids is not None:
+            docs = document_ids.gather(1, safe_positions).reshape_as(pred_ids)
+            anchor_positions = (positions[..., 0] - 1).clamp(0, seq_len - 1)
+            anchor_docs = document_ids.gather(1, anchor_positions).unsqueeze(-1)
+            valid &= (docs == anchor_docs) & (anchor_docs >= 0)
+        if observed is not None:
+            valid &= observed
+        if d2t is not None:
+            pred_ids = pred_ids + d2t[pred_ids]
```
fynnsu, line 138: This is handling a lot of cases. It might be better to just decide what format we want to support and require that for all the callers.

fynnsu (changes requested):
Thanks for the pr, however I'd like to see this further simplified. Please limit the changes to those required for implementing the metric.

Issue #797, opened 2026-07-16
[Bug]: DFlash training performs worse on Qwen3.5 than on Qwen3


chenfengt:
### Your current environment

Please provide the following information about your environment if applicable:

- vLLM
- Speculators 0.6.0
- CUDA
- PyTorch 2.8.0
- Transformers 5.5.0
- Hardware
- Model

### 🐛 Describe the bug

Speculators version: 0.6.0.
When training DFlash with the same corpus, questions, and training scripts, Qwen3.5‑4B performs worse than Qwen3‑4B during the training phase. This can be clearly observed from the TensorBoard statistics.
Is this due to inherent differences between the models, or is it an issue with Speculators?
Below are the data plots for both models during training.

<img width="401" height="425" alt="Image" src="https://github.com/user-attachments/assets/9c781b3f-59ac-4238-b3ac-39f63d9f6d93" />
<img width="401" height="427" alt="Image" src="https://github.com/user-attachments/assets/b3b9e182-0059-46b8-9281-eb9052f5c5d8" />
<img width="398" height="430" alt="Image" src="https://github.com/user-attachments/assets/fa79ea46-e843-4931-9e72-8262f59fc8d6" />

train scripts:
python scripts/train.py --verifier-name-or-path /llm/model/Qwen3-4B-Instruct-2507 --data-path ./output/dflash_qwen3_4b_medium_10k --hidden-states-path ./output/dflash_qwen3_4b_medium_10k/hidden_states --save-path ./output/dflash_qwen3_4b_medium_10k/checkpoints --speculator-type dflash --block-size 8 --num-layers 5 --target-layer-ids 2 10 18 26 34 --epochs 30 --lr 5e-4 --total-seq-len 8192 --on-missing raise --seed 42 --draft-arch qwen3 --scheduler-type cosine --max-anchors 512 --prefetch-factor 2 --num-workers 8 --logger tensorboard --run-name dflash_qwen3_4b_medium_10k --log-freq 4
[...]

PR #1090, opened 2026-09-07
fix(models): resolve verifier_norm class by verifier family

minziyu:
Some verifier families store the final RMSNorm weight in the Gemma convention — applied gain is (1 + w) — while verifier_norm was always constructed as a plain Qwen3RMSNorm (gain w). This includes the Gemma models (as covered for Gemma3 by #892) and the Qwen3.5/Qwen3.8 family, whose Qwen3_5RMSNorm is an alias of vLLM's GemmaRMSNorm (vllm/model_executor/models/qwen3_5.py); transformers' Qwen3_5RMSNorm.forward likewise computes output * (1.0 + weight).

The mismatch silently mis-scales the reconstructed verifier targets: on a Qwen3.8-27B verifier (V=248320), feeding real model.norm.weight and real lm_head rows through both variants gives rel-L2 0.51 vs the verifier's own final hidden state (KL ~0.48 nat on the reconstructed target distribution, top-1 prob 0.008 vs 0.071 true). Training converges regardless — argmax mostly survives the ~2x scale error — so nothing flags it. Likely (part of) what #797 observed as "DFlash performs worse on Qwen3.5 than Qwen3".

Generalizes #892's Gemma3 norm-class selection into a shared resolve_verifier_norm_class helper: detection keys off the verifier's model_type / text_config.model_type (read from its config.json when resolvable) with an architectures fallback — not weight statistics, since in the Qwen3.8 checkpoint per-layer norm weights sit near zero (gemma-style storage) while the final norm weight sits near 1.0, so value-based heuristics would misfire on exactly the layer that matters. The checkpoint weight loads verbatim; the convention is applied in the norm class's forward, matching the verifier's own numerics.

Verified end-to-end on the DSpark-on-Qwen3.8 pipeline (hidden-states extraction through serving) with the equivalent weight-fold variant of this fix; a unit test asserts Gemma3RMSNorm(w) == Qwen3RMSNorm(w+1).

Refs #892, #797.

### DIFF
--- a/src/speculators/models/dflash/core.py
+++ b/src/speculators/models/dflash/core.py
@@ -124,9 +124,6 @@ class DFlashDraftModel(DraftVocabMixin, SpeculatorModel):
             config.transformer_layer_config.hidden_size,
             eps=config.transformer_layer_config.rms_norm_eps,  # type: ignore[arg-type]
         )
-        # Must apply the verifier's own final-norm convention (`x * (1 + w)`
-        # for the Gemma/Qwen3.5 families, `x * w` otherwise) or the
-        # reconstructed verifier targets are silently mis-scaled.
         self.verifier_norm = resolve_verifier_norm_class(config)(
             config.transformer_layer_config.hidden_size,
             eps=config.transformer_layer_config.rms_norm_eps,  # type: ignore[arg-type]
--- a/src/speculators/models/dflash/core.py
+++ b/src/speculators/models/dflash/core.py
@@ -25,6 +25,7 @@ from speculators.models.utils import (
     conditional_torch_compile,
     flatten_rope_parameters,
     resolve_target_layer_ids,
+    resolve_verifier_norm_class,
 )
 
 logger = logging.getLogger(__name__)
@@ -123,7 +124,10 @@ class DFlashDraftModel(DraftVocabMixin, SpeculatorModel):
             config.transformer_layer_config.hidden_size,
             eps=config.transformer_layer_config.rms_norm_eps,  # type: ignore[arg-type]
         )
-        self.verifier_norm = Qwen3RMSNorm(
+        # Must apply the verifier's own final-norm convention (`x * (1 + w)`
+        # for the Gemma/Qwen3.5 families, `x * w` otherwise) or the
+        # reconstructed verifier targets are silently mis-scaled.
+        self.verifier_norm = resolve_verifier_norm_class(config)(
             config.transformer_layer_config.hidden_size,
             eps=config.transformer_layer_config.rms_norm_eps,  # type: ignore[arg-type]
         )
--- a/src/speculators/models/utils.py
+++ b/src/speculators/models/utils.py
@@ -1,9 +1,7 @@
-import json
 import logging
 import warnings
 from copy import deepcopy
-from functools import partial
-from pathlib import Path
+from functools import lru_cache, partial
 
 import torch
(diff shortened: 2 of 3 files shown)


src/speculators/models/dflash/core.py
```diff
@@ -123,7 +124,10 @@ def __init__(
             config.transformer_layer_config.hidden_size,
             eps=config.transformer_layer_config.rms_norm_eps,  # type: ignore[arg-type]
         )
-        self.verifier_norm = Qwen3RMSNorm(
+        # Must apply the verifier's own final-norm convention (`x * (1 + w)`
+        # for the Gemma/Qwen3.5 families, `x * w` otherwise) or the
+        # reconstructed verifier targets are silently mis-scaled.
+        self.verifier_norm = resolve_verifier_norm_class(config)(
```
fynnsu: Please move this comment into the `resolve_verifier_norm_class` helper fn

src/speculators/models/utils.py
```diff
@@ -1,13 +1,67 @@
...
+    """Whether the verifier's final norm is `x_norm * (1 + w)` rather than `x_norm * w`."""
+    verifier = getattr(getattr(config, "speculators_config", None), "verifier", None)
+    if verifier is None:
+        return False
+
+    candidates: list[str] = []
+
+    # Prefer the verifier's own config.json (model_type is the reliable signal,
+    # and the multimodal wrapper keeps the family under text_config).
+    name_or_path = getattr(verifier, "name_or_path", None)
```
fynnsu, line 51: You can use this helper function to get the verifier config:

 https://github.com/vllm-project/speculators/blob/8e0892042335bb76e35e799b7bef464eb97bd567/src/speculators/models/utils.py#L20

src/speculators/models/utils.py
```diff
@@ -1,13 +1,67 @@
...
 logger = logging.getLogger(__name__)
 
+# Verifier families whose final norm follows the Gemma convention
+# `x_norm * (1 + w)` instead of the plain RMSNorm `x_norm * w`. Matched as a
+# prefix against the verifier's HF `model_type` / `text_config.model_type`
+# (read from its config.json when resolvable) with a fallback to lowercased
+# `architectures`. Qwen3.5 is in here because `Qwen3_5RMSNorm` is an alias of
+# vLLM's `GemmaRMSNorm` (see vllm/model_executor/models/qwen3_5.py) and
+# transformers' `Qwen3_5RMSNorm.forward` computes `output * (1.0 + weight)`.
+GEMMA_STYLE_FINAL_NORM_PREFIXES = ("gemma", "qwen3_5")
```
fynnsu: It seems like gemma3n and gemma4 don't use this format, and therefore shouldn't be included. 

https://github.com/huggingface/transformers/blob/main/src/transformers/models/gemma3n/modeling_gemma3n.py#L2400-L2428

https://github.com/huggingface/transformers/blob/main/src/transformers/models/gemma4/modeling_gemma4.py#L2499-L2527

Maybe we could match `"gemma", "gemma2", "gemma3", "gemma3_text", "recurrent_gemma"` for model types, and match exactly instead of prefixes? We'd probably also have to drop the architecture prefix matching then too. What do you think?

fynnsu (commented):
Thank you for finding this. Added a few comments.

fynnsu (approved):
Thanks for the fix!

PR #1097, opened 2026-09-08
fix(train): restore signal handlers on every exit from the decorator

### LINKED ISSUE #1089: [Bug]: Training shutdown decorator leaves signal handlers installed after return or error
### Bug

At `3419401a380db305ed6493ba4680ad8a3c3e0e45`, `with_graceful_shutdown()` installs SIGINT/SIGTERM handlers but restores them only when it catches `TrainingInterruptedError`. After a normal return or an unrelated exception, the installed handlers still belong to the completed training call.

This matters when invoking training from a longer-lived Python process: a later interrupt enters the stale training handler, rather than the handler that was active before training. Repeated/nested decorated calls can also capture and restore stale handlers.

### Minimal reproduction

From a checkout, this uses only Python's standard library and loads the actual production file. It avoids unrelated package initialization and restores the process's handlers before exiting.

```python
import runpy
import signal

module = runpy.run_path("src/speculators/train/graceful_shutdown.py")
before = {sig: signal.getsignal(sig) for sig in (signal.SIGINT, signal.SIGTERM)}

@module["with_graceful_shutdown"]()
def train(self):
    return "finished"

try:
    print(train(object()))
    print({sig.name: signal.getsignal(sig) == handler
           for sig, handler in before.items()})
finally:
    for sig, handler in before.items():
        signal.signal(sig, handler)
```

Observed on Windows / Python 3.13:

```text
finished
{'SIGINT': False, 'SIGTERM': False}
```

Expected: both values are `True`. Raising an ordinary `RuntimeError` inside the decorated call also leaves the handlers installed.

### Proposed scope and verification

A minimal fix is to add an outer `finally: handler.restore()` to the existing try/except, preserving the early restore before interrupt-checkpoint saving. A local two-line fix plus six regression cases covers return values, ordinary exceptions, `KeyboardInterrupt`, checkpoint success/failure, and nested calls.

Isolated tests loading the real stdlib-only module: **4 failed / 2 passed before**, **6 passed after**. A tiny CUDA linear-model workload also reproduces the leak and verifies restoration after the fix for FP32/BF16 and return/error paths; this is **not** a full Speculators Trainer or distributed-training test.

The standard pytest entry point is currently blocked on this Windows environment by the package's `fcntl` import. A follow-up PR still needs the contributor's human review/sign-off and supported-environment checks. This report was prepared with coding-agent assistance; no human-review checklist is being asserted here.

dchaudhari7177:
Closes #1089.

Reproduced with the script from the issue, on `origin/main`:

```
return value: finished
handler restored? {'SIGINT': False, 'SIGTERM': False}
```

`restore()` was reachable only from the `except TrainingInterruptedError` branch, so a normal return or an unrelated exception left the call's handlers installed.

## The change

A `finally` block, so every exit path restores. The `except` branch keeps its own `restore()` — that one has to happen *before* the checkpoint save, so a deliberate second Ctrl+C during the save still forces exit.

That makes `restore()` run twice on the interrupted path, which pulls in two related fixes the issue also points at with "repeated/nested decorated calls can also capture and restore stale handlers":

- **`restore()` is idempotent.** Without it, the second call would undo a handler installed in between — for instance one set up by `maybe_save_checkpoint()`.
- **`install()` is idempotent**, for the mirror-image reason: a second `install()` while already installed would capture *this* handler as the "original", and `restore()` would then put it back rather than remove it.

One more thing found while writing the fix: `restore()` skipped a signal whose original handler was `None`, which `getsignal()` returns when the previous handler was not set from Python. There is nothing to put back in that case, and leaving this handler installed is precisely the bug, so it now falls back to `SIG_DFL`.

## Verification

Seven cases, run against the branch and against `origin/main`:

```
                          with fix    on main
normal return             PASS        FAIL: handlers outlived a successful call
unrelated exception       PASS        FAIL: handlers outlived a failed call
interruption              PASS        PASS
save fails                PASS        PASS
nested calls              PASS        FAIL
restore idempotent        PASS        FAIL
install idempotent        PASS        FAIL
```

The two that pass on `main` are the interruption paths, which already restored — which is the right control, since this must not change them. `interruption` also asserts the checkpoint save still runs (`saved == ["interrupted"]`).

`ruff` 0.15.20 (the pin in `pyproject.toml`) check and format: clean.

### A note on how I ran them

`tests/unit/train/test_graceful_shutdown.py` imports `speculators.train.graceful_shutdown` normally, which is right for CI. It cannot execute on Windows, because importing the package pulls in `hs_connectors.transfer`, which does `import fcntl`:

```
hs_connectors/src/hs_connectors/transfer.py:6: in <module>
    import fcntl
E   ModuleNotFoundError: No module named 'fcntl'
```

That is pre-existing and not specific to this change — `tests/unit/train/test_logger.py` and plain `import speculators` fail the same way here. So I ran the seven test bodies against a directly-loaded module via `runpy.run_path`, the same technique the issue used to avoid unrelated package initialization, which is where the table above comes from. On Linux CI the committed test runs as written.

### DIFF
--- a/src/speculators/train/graceful_shutdown.py
+++ b/src/speculators/train/graceful_shutdown.py
@@ -52,25 +52,46 @@ class GracefulShutdownHandler:
         self._original_sigterm: Any = None
         self._timeout = timeout
         self._owner_pid: int | None = None
+        self._installed = False
 
     def install(self):
-        """Register signal handlers for SIGINT and SIGTERM."""
+        """Register signal handlers for SIGINT and SIGTERM.
+
+        Idempotent: a second call while already installed would capture this
+        handler as the "original" and restore() would then reinstall it.
+        """
+        if self._installed:
+            return
         self._owner_pid = os.getpid()
         self._original_sigint = signal.getsignal(signal.SIGINT)
         self._original_sigterm = signal.getsignal(signal.SIGTERM)
         signal.signal(signal.SIGINT, self._handler)
         signal.signal(signal.SIGTERM, self._handler)
+        self._installed = True
 
     def restore(self):
-        """Restore original signal handlers.
+        """Restore the handlers that were active before install().
 
         Called before the checkpoint save attempt so that a deliberate
-        second Ctrl+C during the save causes immediate exit.
+        second Ctrl+C during the save causes immediate exit, and again from
+        the decorator's finally block so the handlers never outlive the call.
+
+        Idempotent, which is what makes calling it from both places safe: the
+        second call must not undo a handler installed in between, such as one
+        set up by maybe_save_checkpoint().
         """
-        if self._original_sigint is not None:
-            signal.signal(signal.SIGINT, self._original_sigint)
-        if self._original_sigterm is not None:
-            signal.signal(signal.SIGTERM, self._original_sigterm)
+        if not self._installed:
+            return
+        # getsignal() returns None when the previous handler was not set from
+        # Python. There is nothing to put back in that case, and leaving this
+        # handler installed is the bug being fixed, so fall back to the default.
+        sigint = self._original_sigint
+        sigterm = self._original_sigterm
+        signal.signal(signal.SIGINT, sigint if sigint is not None else signal.SIG_DFL)
+        signal.signal(
+            signal.SIGTERM, sigterm if sigterm is not None else signal.SIG_DFL
+        )
+        self._installed = False
 
     def _handler(self, signum, frame):  # noqa: ARG002
         # Only handle in the process that installed the handler.
@@ -145,6 +166,13 @@ def with_graceful_shutdown(
                     logger.exception("Failed to save interrupt checkpoint")
                 finally:
                     timer.cancel()
+            finally:
+                # The except branch above restores before the save, but a normal
+                # return or an unrelated exception never reached it, so this
+                # call's handlers stayed installed after it finished. In a
+                # longer-lived process a later interrupt then entered a handler
+                # belonging to a training run that had already ended.
+                handler.restore()
 
         return wrapper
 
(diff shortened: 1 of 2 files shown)


LOGO127:
Thanks for picking up #1089. I independently checked head `161eefe1a882d81a1036e60b406fa150f9ef9a66` against the original `3419401` baseline using the issue's production-module approach, extended to actual POSIX signal delivery in isolated WSL subprocesses.

The probe covers normal return, an unrelated exception, and training interruption, each with SIGINT and SIGTERM. It verifies the original handlers are active before the checkpoint callback and sends a subsequent signal to verify restored dispatch. Baseline: **2 passed / 4 expected failures**. Your head: **6 passed / 0 failures**.

This is agent-assisted independent verification of the original report, not a full Trainer/distributed-training run or blanket review approval. The production module is loaded via `runpy`; checkpoint saving is a callback rather than a real write. Native non-Python prior handlers (`getsignal() is None`) and watchdog expiry were not exercised. I have paused my unpublished duplicate candidate in favor of collaborating on this PR.

fynnsu (approved):
Thanks for the fix! Please update DCO and I can land!

dchaudhari7177:
Rebased onto current `main` with `--signoff`, so the DCO check is satisfied and the branch is no longer behind. Thanks for the review — no code change in this push, the diff is identical.

dchaudhari7177:
@fynnsu the red Buildkite run on `5dd2b4d` (#1498) is the agent's disk filling up, not this change. `unit-tests` ends with `3 failed, 1006 passed`, and all three are `tests/unit/scripts/test_response_regeneration.py::test_load_input_dataset_from_local_file` failing with:

```
OSError: [Errno 28] No space left on device: '/model-cache/datasets/_model-cache_datasets_json_default-...lock'
```

This PR only touches `graceful_shutdown.py` and its tests. A retry of the build should clear it. I can't trigger one from a fork, but I'm happy to push an empty rebase if that's easier for you.

fynnsu:
@dchaudhari7177 Thanks for updating. Yes we had an issue with our runners, but I think it should be resolved now. I've restarted the failing tests.

PR #1100, opened 2026-09-08
add gradient checkpointing for dflash/dspark

walterchenchn:
## Purpose
DSpark/DFlash training on long sequences (32K+) consumes excessive activation memory — intermediate MLP and attention activations for all decoder layers are retained for backward, driving peak memory beyond NPU limits.
This PR enables gradient checkpointing: each decoder layer's forward is checkpointed so only the layer input is saved for backward, and intermediate activations are recomputed during the backward pass. This reduces activation memory by ~10 GB for a 5-layer DSpark model at 32K sequence length, at the cost of ~30-50% slower backward.
Two changes were required: (1) setting supports_gradient_checkpointing = True on DFlashDraftModel (the class-level gate that was missing — Qwen3DFlashDecoderLayer already inherits GradientCheckpointingLayer, so the mechanism was always available, only the flag was absent), and (2) wiring gradient_checkpointing through TrainerConfig, the trainer's setup_model(), and the CLI schema so users can opt in via --gradient-checkpointing.

## Tests
Testing on the qwen3-8b model, with a sequence length of 8k and a max-anchor of 2048, the GPU memory usage decreased from 50G to 25G.

## Checklist

I have filled in:

### DIFF
--- a/src/speculators/models/dflash/core.py
+++ b/src/speculators/models/dflash/core.py
@@ -37,6 +37,7 @@ _compiled_create_block_mask = torch.compile(create_block_mask)
 @SpeculatorModel.register("dflash")
 class DFlashDraftModel(DraftVocabMixin, SpeculatorModel):
     config_class: ClassVar[type[DFlashSpeculatorConfig]] = DFlashSpeculatorConfig  # type: ignore[misc]
+    supports_gradient_checkpointing = True  # noqa: D003  # Qwen3DFlashDecoderLayer inherits GradientCheckpointingLayer
     _no_split_modules = ["Qwen3DFlashDecoderLayer"]
     _keys_to_ignore_on_load_missing: ClassVar[list[str]] = [  # type: ignore[misc]
         "embed_tokens.weight",
--- a/src/speculators/train/config/schema.py
+++ b/src/speculators/train/config/schema.py
@@ -416,6 +416,12 @@ class TrainerArgs(_Group):
         "parameters are fully replicated (DDP-like). Enable when the model does not "
         "fit in a single GPU's memory.",
     )
+    gradient_checkpointing: bool = Field(
+        default=False,
+        description="Enable gradient checkpointing on decoder layers to save "
+        "activation memory at the cost of ~30-50 percent slower backward. "
+        "Recommended for 32K+ sequence lengths or large max_anchors.",
+    )
     max_steps: int | None = Field(
         default=None,
         ge=1,
--- a/src/speculators/train/trainer.py
+++ b/src/speculators/train/trainer.py
@@ -138,6 +138,7 @@ class TrainerConfig(NamedTuple):
     hidden_states_dtype: torch.dtype = torch.bfloat16
     log_freq: int = 1
     fsdp_shard: bool = False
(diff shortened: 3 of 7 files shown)


src/speculators/train/trainer.py
```diff
@@ -297,6 +298,15 @@ def setup_model(self):
         # Verify model is compatible with training infrastructure
         SpeculatorModel.verify_training_compatible(self.model)
 
+        # Enable gradient checkpointing BEFORE FSDP/DDP wrapping to save
+        # activation memory at the cost of recomputation during backward.
+        # Each decoder layer's forward is checkpointed: only the layer input
+        # is saved for backward; intermediate activations (MLP, attention)
+        # are recomputed. Saves ~10 GB for 5-layer DSpark with 32K seq.
+        if self.config.gradient_checkpointing:
+            self.model.gradient_checkpointing_enable()
```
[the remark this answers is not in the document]
fynnsu: @walterchenchn I think the suggested change here is a good idea.

src/speculators/models/dflash/core.py
```diff
@@ -37,6 +37,7 @@
 @SpeculatorModel.register("dflash")
 class DFlashDraftModel(DraftVocabMixin, SpeculatorModel):
     config_class: ClassVar[type[DFlashSpeculatorConfig]] = DFlashSpeculatorConfig  # type: ignore[misc]
+    supports_gradient_checkpointing = True  # noqa: D003  # Qwen3DFlashDecoderLayer inherits GradientCheckpointingLayer
```
fynnsu, line 40: This covers `dflash, dspark, dflash2` as they all use this base class. Could we also add support to P-Eagle?

src/speculators/train/config/schema.py
fynnsu, line 1: Could you update the train cli docs at `docs/cli/train.md` to include the new arg?

fynnsu (commented):
Thanks for the pr! Left a few small comments below, ptal.

src/speculators/models/dflash/core.py
walterchenchn: Code has been updated

src/speculators/train/config/schema.py
walterchenchn: done

src/speculators/train/trainer.py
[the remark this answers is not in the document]
walterchenchn: Code has been updated

src/speculators/models/dflash/core.py
fynnsu: Actually slight clarification. We should add support to P-Eagle but **not** to Eagle-3 because Eagle-3 does some dynamic cache stuff which I believe is incompatible.

fynnsu:
Added a comment, also could you please fix DCO

walterchenchn:
> Added a comment, also could you please fix DCO

done

src/speculators/models/dflash/core.py
walterchenchn: done

fynnsu (approved):
Thanks for the pr!

PR #1107, opened 2026-09-09
feat(eval): long-context acceptance benchmark on a recorded request table

reed-meyerson:
## Summary

Adds a **`long-context`** evaluation mode that measures speculative-decoding acceptance as context length grows, using [OpenAI's MRCR dataset](https://huggingface.co/datasets/openai/mrcr) as a source of long, multi-turn prompts.

The design principle: **a benchmark is just a set of requests to make (in some order, with an optional stop rule); everything is recorded to a table, and reports are pure functions over that table.** So the pieces are decoupled into three layers, none of which is MRCR-specific except the last:

1. **`request_runner.py` — generic runner.** Given an ordered set of requests (chat messages + sampling budget + opaque `metadata`) and an optional per-batch stop predicate, it renders each request through the server's own render endpoint (exact prompt length + context fit-test), generates whatever fits, and streams one flat row per request to a **Parquet table**: request identity, metadata, exact prompt length, response text, and the *raw* server metrics blob — all verbatim. It knows nothing about bins, context length, or spec-decode. Each batch is a separate Parquet part file, so a crash only loses the batch in flight.
2. **`acceptance_report.py` — spec-decode analysis over the table.** Context-length bucket edges live here as *analysis parameters*, not harness constants. Reports acceptance (a) by prompt length at request start and (b) by running token position within each request.
3. **`mrcr.py` — thin MRCR adapter.** Loads the dataset and turns it into a deterministically-selected, length-ordered request set.

Because the raw table is the source of truth, reports can be re-sliced with different bin edges and **no server** via `rebin.py`, and plotted with `plot.py acceptance`.

Correctness is intentionally not graded — MRCR is used only for its long prompts.

### Deterministic superset selection

Records are stratified into MRCR's native power-of-2 token bins and, within each bin, ranked by a stable content hash (`sha256(seed + prompt)`); the lowest-ranked `--samples-per-bin` per bin form the candidate set. Bin assignment uses `n_chars / 4.9` (MRCR carries no token count; the ratio is 4.9 ± 0.1 across the whole dataset, validated offline against o200k) — zero runtime cost, no `tiktoken` dependency. The candidate set is independent of `max_model_len`; the only context-dependent step is the render fit-test, which is monotonic, so `Kept(ctx1) ⊆ Kept(ctx2)` for `ctx1 ≤ ctx2`. Batches run smallest-first and stop once a whole bin renders oversized.

### CLI

```bash
# Server must expose the render endpoint and per-request detailed metrics:
VLLM_ENABLE_SCALE_OUT_ENDPOINTS=1 vllm serve <model> \
    --per-request-spec-decode-metrics detailed \
    --speculative-config '{"model":"<draft>","num_speculative_tokens":N}'

python evaluate.py --target http://localhost:8000/v1 long-context --samples-per-bin 20

# Re-slice results offline with different bins (no server):
python rebin.py --table <out>/raw_table --context-bin-edges 0,8192,32768,131072

# Plot the acceptance figure offline (no server):
python plot.py acceptance --table <out>/raw_table --output <out>/acceptance.png
```

New flags (long-context only): `--samples-per-bin`, `--selection-seed`, `--position-bin-size`. `throughput`/`sweep` are unchanged and don't import the long-context deps.

### Visualization

`plot.py` gains an **`acceptance`** subcommand that renders the recorded table into one figure — another pure slicing of the same source of truth as the CSVs, so it needs no server:

Long-context acceptance plot

Three panels share one context-length x-axis (example: Qwen3-8B + a DFlash speculator, 7 speculative tokens, MRCR 2-needle):

- **Mean accept len** (line) — mean committed tokens per verify step in each context bin; the headline trend.
- **Tokens generated** (bars) — committed tokens per bin, i.e. how much data backs each column. Deliberately *decoupled* from acceptance (a step-count bar would inflate low-acceptance bins), so it reads cleanly as sample size.
- **Acceptance-length distribution** (heatmap) — within each context bin, the share of verify **steps** that committed exactly *k* tokens (each column sums to ~100%). The integer in a cell is just the precise readout of its color.

The x-axis is the *running* context length at each verify step (prompt length + everything committed so far), not per-request — so it stays meaningful even when generation is long relative to the bin width. Bins default to ~2 log2 bins per octave and can be overridden with `--context-bin-edges`.

Reading the example: acceptance degrades monotonically as context grows — mean length falls 2.73 → 1.68 tokens/step from ~4k to ~26k, and the share of single-token steps climbs 28% → 52%.

### Dependencies

Long-context mode adds `pandas`, `pyarrow`, `huggingface_hub` (dataset loading + Parquet table); requests go straight to the target server's HTTP API, so there's no separate eval harness. Plotting reuses the existing `matplotlib`/`numpy` deps — nothing new.

### Validation

Ran end-to-end against Qwen3-8B + a DFlash speculator at `--max-model-len 40960`: deterministic bounded selection with the bin early-stop firing, a Parquet table written (one part file per bin), both acceptance reports produced, detailed per-step metrics flowing through, `rebin.py` reproducing the reports offline with different bin edges, and `plot.py acceptance` rendering the figure above from the recorded table. Recording is small — a full MRCR run is single-digit MB (the raw prompts are *not* stored, only a reference; the per-step arrays are bounded by the generation cap, not the context length).

### Note

Supersedes #1064 — this replaces the earlier Inspect-AI-based MRCR mode with the recorded-table architecture. Opening fresh rather than force-pushing that branch so the two approaches can be compared. Draft while I gather more validation.

The example image is hosted on the fork's `pr-assets` branch (kept out of this PR's diff), since `gh` can't upload to GitHub's attachment CDN directly.

---

Reusability: `throughput`/`sweep` could later adopt `request_runner` + `acceptance_report` to get the same per-request table and reporting (out of scope here).

### DIFF
--- a/scripts/evaluate/evaluate.py
+++ b/scripts/evaluate/evaluate.py
@@ -91,6 +91,20 @@ _SPEEDBENCH_COLUMN_MAPPER = (
 )
 
 
+def _positive_int(value: str) -> int:
+    """argparse type: accept only integers >= 1.
+
+    Guards flags whose non-positive values fail late and expensively -- e.g. a
+    negative --samples-per-bin becomes a negative list slice that runs nearly the
+    whole dataset, and a non-positive --position-bin-size only blows up at the
+    reporting step, after the entire inference run.
+    """
+    parsed = int(value)
+    if parsed < 1:
+        raise argparse.ArgumentTypeError(f"must be >= 1, got {parsed}")
+    return parsed
+
+
 def _fetch_model_info(target: str) -> dict | None:
     """Return the first ``/v1/models`` entry (has ``id`` and ``max_model_len``)."""
     base = target.rstrip("/")
@@ -505,7 +519,7 @@ def main() -> None:
     lc_group = parser.add_argument_group("long-context mode")
     lc_group.add_argument(
         "--samples-per-bin",
-        type=int,
+        type=_positive_int,
         default=DEFAULT_SAMPLES_PER_BIN,
         help=(
             "long-context: records to run per native MRCR token bin. Larger "
@@ -525,7 +539,7 @@ def main() -> None:
     )
     lc_group.add_argument(
         "--position-bin-size",
-        type=int,
+        type=_positive_int,
         default=DEFAULT_POSITION_BIN_SIZE,
         help=(
             "long-context: token-position bin width for the by-position report "
--- a/scripts/evaluate/evaluate.py
+++ b/scripts/evaluate/evaluate.py
@@ -4,12 +4,22 @@
 Modes:
     throughput   Max throughput run for acceptance rates
     sweep        Full pipeline (gen-len, sweep, CSV)
+    long-context Per-request acceptance vs. context length via OpenAI's MRCR
+                 dataset. Deterministically samples a superset-monotonic set
+                 stratified across MRCR's native token bins, records every
+                 request to a Parquet table, and reports acceptance binned by
+                 context length. Requires the server started with
+                 --per-request-spec-decode-metrics detailed and the render
+                 endpoint enabled (VLLM_ENABLE_SCALE_OUT_ENDPOINTS=1).
 
 Examples:
     python evaluate.py --target http://localhost:8000/v1 throughput
     python evaluate.py --target http://localhost:8000/v1 sweep
     python evaluate.py --target http://localhost:8000/v1 sweep \\
         --subsets "HumanEval,qa" --gen-kwargs '{"temperature":0.6}'
+
+    python evaluate.py --target http://localhost:8000/v1 long-context \\
+        --samples-per-bin 20
 
     # SPEED-Bench (run prepare_speedbench.py once first to split data):
     python evaluate.py --target http://localhost:8000/v1 throughput \\
@@ -63,6 +73,9 @@ DEFAULT_SUBSETS = (
 )
 DEFAULT_MAX_CONCURRENCY = 128
 DEFAULT_MAX_REQUESTS = 200
+# long-context: records selected per native MRCR token bin (density/cost knob).
+DEFAULT_SAMPLES_PER_BIN = 20
+DEFAULT_POSITION_BIN_SIZE = 256
 DEFAULT_GEN_LEN_RATE = 128
 DEFAULT_SWEEP_RATE = 10
 DEFAULT_DATA_COLUMN_MAPPER = (
@@ -78,7 +91,8 @@ _SPEEDBENCH_COLUMN_MAPPER = (
 )
 
 
-def _fetch_model_name(target: str) -> str | None:
+def _fetch_model_info(target: str) -> dict | None:
+    """Return the first ``/v1/models`` entry (has ``id`` and ``max_model_len``)."""
     base = target.rstrip("/")
     if not base.endswith("/v1"):
         base += "/v1"
@@ -88,10 +102,15 @@ def _fetch_model_name(target: str) -> str | None:
             data = json.loads(resp.read())
         models = data.get("data", [])
         if models:
-            return models[0].get("id")
+            return models[0]
     except (URLError, json.JSONDecodeError, OSError) as e:
-        logger.warning("Could not fetch model name from %s: %s", url, e)
+        logger.warning("Could not fetch model info from %s: %s", url, e)
     return None
+
+
+def _fetch_model_name(target: str) -> str | None:
+    info = _fetch_model_info(target)
+    return info.get("id") if info else None
 
 
 def _sanitize_dir_name(name: str) -> str:
@@ -281,16 +300,41 @@ def _run_subset(
     return acceptance_csv, perf_csv, max_tokens if is_sweep else None
 
 
+def _run_long_context(args: argparse.Namespace, output_dir: Path) -> None:
+    # Imported lazily so throughput/sweep runs don't pull the long-context deps
+    # (pandas, pyarrow, huggingface_hub).
+    from mrcr import run_mrcr  # noqa: PLC0415
+
+    run_mrcr(
+        target=args.target,
+        model_info=_fetch_model_info(args.target),
+        output_dir=output_dir,
+        max_concurrency=args.max_concurrency,
+        samples_per_bin=args.samples_per_bin,
+        selection_seed=args.selection_seed,
+        position_bin_size=args.position_bin_size,
+    )
+
+
 def run_benchmark(args: argparse.Namespace) -> None:
+    output_dir = Path(args.output_dir)
+    output_dir.mkdir(parents=True, exist_ok=True)
+    save_eval_provenance(output_dir)
+
+    if args.mode == "long-context":
+        _run_long_context(args, output_dir)
+    else:
+        _run_guidellm_modes(args, output_dir)
+    logger.info("Benchmarking complete! Results: %s", output_dir)
+
+
+def _run_guidellm_modes(args: argparse.Namespace, output_dir: Path) -> None:
     check_dependencies()
     is_sweep = args.mode == "sweep"
 
     metrics_url = args.target.rstrip("/").removesuffix("/v1") + "/metrics"
-    output_dir = Path(args.output_dir)
     artifacts_dir = output_dir / "artifacts"
     artifacts_dir.mkdir(parents=True, exist_ok=True)
-
-    save_eval_provenance(output_dir)
 
     if not (output_dir / "vllm_command.txt").exists():
         logger.info(
@@ -365,8 +409,6 @@ def run_benchmark(args: argparse.Namespace) -> None:
         with (output_dir / "max_tokens.json").open("w") as f:
             json.dump(all_max_tokens, f, indent=2)
 
-    logger.info("Benchmarking complete! Results: %s", output_dir)
-
 
 def main() -> None:
     logging.basicConfig(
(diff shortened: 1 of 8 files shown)


LOGO127:
Thanks for adding durable per-request recording! I independently reproduced a run-isolation edge case in `run_requests`: reusing a table directory for a shorter run leaves old Parquet parts in the new report.

I tested the unmodified runner at stacked PR #1109 commit `f8329aef33ad535eb7dba2d26647c6e00bbda8bd`; its `request_runner.py` has no diff against this PR's current head `2e60ec2decbef2115e71b0af401b9de0e41b4d28`. Posting here because this is shared recording behavior, not throughput-specific wiring.

Reproduction: using a fresh temporary directory and a localhost HTTP fixture returning successful chat-completion responses, with `fit_test=False` and concurrency 1:

```python
def req(i):
    return Request(messages=[{"role": "user", "content": i}], request_id=i)

run_requests(root_url, model, [[req("old-0")], [req("old-1")]],
             table_dir, max_concurrency=1, fit_test=False)
run_requests(root_url, model, [[req("new-0")]],
             table_dir, max_concurrency=1, fit_test=False)
print(load_table(table_dir)["request_id"].to_pylist())
# Actual: ['new-0', 'old-1']; expected for a new run: ['new-0']
```

`part-00000.parquet` is overwritten, but `part-00001.parquet` survives and `load_table` reads both. A clean-directory control passes. This is separate from the existing CSV append feedback: the raw request table itself mixes runs. Reusing an explicit output directory appears allowed by the CLI's `mkdir(exist_ok=True)`.

Would rejecting an existing nonempty table directory before sending requests, or allocating a unique per-run table directory, fit the intended behavior? Either would preserve old artifacts without silently merging them. If resume is intended instead, it needs an explicit part/manifest policy.

Validation scope: CPU/WSL, Python 3.11.15, PyArrow 23.0.1, pytest 9.1.1; real runner, HTTP and Parquet I/O, with a synthetic localhost response (no model or GPU). The two-case isolation test gives one passing clean-directory control and one failing reuse case. This is a correctness probe, not a model-performance benchmark or a full project test run.

scripts/evaluate/request_runner.py
```diff
@@ -0,0 +1,309 @@
...
+        n_oversized,
+        n_error,
+        table_dir,
+    )
+    return table_dir
+
+
+def load_table(table_dir: Path) -> pa.Table:
+    """Read a recorded table (a directory of Parquet part files) back in."""
+    return pq.read_table(table_dir, schema=TABLE_SCHEMA)
```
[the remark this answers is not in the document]
LOGO127: I checked this specific case with the unmodified runner, and could not reproduce the empty-directory exception on PyArrow 23.0.1 / Python 3.11.15 (WSL/Linux). Passing `schema=TABLE_SCHEMA` returns a zero-row table with that schema.

From `scripts/evaluate`, this is a small CPU-only check (no model/server needed):

```python
from pathlib import Path
from tempfile import TemporaryDirectory
from request_runner import TABLE_SCHEMA, load_table, run_requests

with TemporaryDirectory() as root:
    path = Path(root) / "raw_table"
    run_requests("http://127.0.0.1:1", "unused", [], path,
                 max_concurrency=1, fit_test=False)
    assert list(path.iterdir()) == []
    table = load_table(path)
    assert table.num_rows == 0
    assert table.schema == TABLE_SCHEMA
```

Both `batches=[]` and `batches=[[]]` passed; a separate nonexistent-path control raises `FileNotFoundError`. Three focused tests passed. The runner tested at #1109 `f8329aef` is identical to this PR's `2e60ec2d` version. I have not checked other PyArrow versions, so a failing version/reproducer would be helpful before adding a special case. This is AI-assisted local correctness verification, not model evaluation or full-project validation.

WindChimeRan (commented):
Previous work: https://github.com/vllm-project/speculators/pull/919/
@shanjiaz

reed-meyerson:
Thanks for the careful repro. This was a real bug, and your two-case (clean control + failing reuse) framing made it easy to confirm. Fixed in `7a86e76`.

On the design question: rather than rejecting a nonempty directory or allocating a unique one, `run_requests` now clears stale `part-*.parquet` files at the start of a run:

```python
table_dir.mkdir(parents=True, exist_ok=True)
for stale in table_dir.glob("part-*.parquet"):
    stale.unlink()
```

The reasoning is consistency with the rest of the harness rather than a new isolation policy. The default output directory is already timestamped and unique per run (`evaluate.py`), so reusing one is an explicit choice; and when a directory *is* reused, the derived CSV reports in it already replace rather than append (a sibling fix in the same commit). The raw table silently merging runs was the odd one out, and clearing stale parts makes the table authoritative for the current run, matching the reports written alongside it. No benchmark in these scripts does resume or checkpointing (runs re-execute wholesale), so there's nothing a part/manifest policy would serve. Your exact reproduction is now a regression check (asserts a shorter rerun yields only the new rows).

shanjiaz (changes requested):
Thanks for adding this! 
My main concern is scope. This PR touches eight files and introduces a generic request-running/persistence framework, but that framework is currently used only by MRCR. SPEED-Bench, the original datasets, throughput, and sweep all continue through the existing GuideLLM path. As a result, we pay the complexity cost of a second evaluation framework without actually unifying the evaluation paths. Would really appreciate it if we can break this down into two phase:

1. A minimal diff enabling MRCR benchmarking.
2. Full refactor to include the generic request framework that supports a more pluggable backend. 

Thank you thank you.

scripts/evaluate/request_runner.py
```diff
@@ -0,0 +1,328 @@
...
+                )
+                for req in batch
+            ]
+            rows = [f.result() for f in as_completed(futures)]
+
+            pq.write_table(
+                pa.Table.from_pylist(rows, schema=TABLE_SCHEMA),
+                table_dir / f"part-{part_idx:05d}.parquet",
+                compression="zstd",
+            )
```
rahul-tuli, line 296: Suggestion: Since the PR description says that if something fails midflight we would only only loose a batch

But a crash/SIGKILL/OOM *during* the write leaves a truncated part file, and `load_table` (lines 326–328) then raises on the corrupt footer, making **every** completed batch unreadable, not just the one in flight

Would it make sense to write to a temp path in the same directory and `os.rename` it into place (atomic on POSIX), so a dead run leaves at worst an orphan temp file that `load_table` never sees?

scripts/evaluate/request_runner.py
```diff
@@ -0,0 +1,328 @@
...
+    granted: int | None = None
+
+    if fit_test:
+        rendered = _render(root_url, model, req)
+        if isinstance(rendered, str):  # "oversized" | "error"
+            return _row(req, rendered)
+        prompt_tokens, granted = rendered
+        if granted < min_room:
+            # Fits, but too little generation room left to be worth measuring.
+            return _row(req, "oversized", prompt_tokens=prompt_tokens)
```
rahul-tuli, line 210: A prompt that fits but leaves under 32 tokens of generation room gets saved as `oversized`  even though it fit. That mislabels the table, and `_all_oversized` (`mrcr.py:95-101`) can stop the whole run on a bin where nothing actually failed to fit. Could this get its own status (e.g. `insufficient_room`) that doesn't count as oversized? Or am I misunderstanding this?

scripts/evaluate/plot.py
```diff
@@ -367,6 +374,220 @@ def run_speedup(args: argparse.Namespace) -> None:
...
+            file=sys.stderr,
+        )
+        sys.exit(1)
+
+    if args.context_bin_edges:
+        edges = np.array(
+            sorted(int(e) for e in args.context_bin_edges.split(",") if e.strip())
+        )
+    else:
+        edges = _auto_log2_edges(positions)
```
rahul-tuli, line 577: `rebin.py` now validates `--context-bin-edges` (sorts, dedupes, requires ≥2), but this copy of the same flag only sorts;  Could both scripts share one parsing helper?

scripts/evaluate/evaluate.py
```diff
@@ -63,6 +73,9 @@
 )
 DEFAULT_MAX_CONCURRENCY = 128
 DEFAULT_MAX_REQUESTS = 200
+# long-context: records selected per native MRCR token bin (density/cost knob).
+DEFAULT_SAMPLES_PER_BIN = 20
+DEFAULT_POSITION_BIN_SIZE = 256
```
rahul-tuli, line 78: nit: Since this is also defined in acceptance_report.py. Could both import it from one small constants module, keeping the lazy-import boundary intact?

scripts/evaluate/evaluate.py
```diff
@@ -389,10 +445,11 @@ def main() -> None:
     )
     parser.add_argument(
         "mode",
-        choices=["throughput", "sweep"],
+        choices=["throughput", "sweep", "long-context"],
```
rahul-tuli, line 448: Could we also update `docs/user_guide/tutorials/evaluating_performance.md` to reflect the new mode

rahul-tuli (changes requested):
Thanks for the careful work here, @reed-meyerson the recorded-table design is genuinely good, but I second @shanjiaz's comment that landing a minimal MRCR-only diff first would get the benchmark into people's hands quickly, and the runner/report/plot generalization would be a much easier second review.

I'd be glad to re-review promptly in either shape. The core ideas here are solid; it's mostly a question of sequencing and a few comments questions/suggestions

PR #1110, opened 2026-09-09
feat(benchmark): add detailed profiling, visualization, and statistical tests

orestis-z:
## Summary

Turns `scripts/benchmark.py` into a harness that can answer "did this change make training faster, and where did the time go" with numbers you can defend in review.

**Trainer instrumentation** (`src/speculators/train/trainer.py`, ~20 lines)
- Finer timing marks (`queue`, `pre_clip`) decompose `fetch_ms` into DataLoader wait + H2D transfer, and `bwd_ms` into backward + grad clip
- Record `torch.cuda.memory_allocated()` at each phase boundary, so memory is attributed per phase rather than as one peak
- A `Trainer.profiler` hook the harness drives with `profiler.step()`, so a `torch.profiler` schedule lines up with training steps

**`run`** — adds `--profile`, which attaches `torch.profiler` (warmup steps become profiler warmup, measured steps are actively profiled), writes a Chrome trace, and aggregates `key_averages()` into a per-kernel device-time breakdown stored in the result JSON. Also `--profile-dir` and `--profile-stacks`.

**`report`** (renamed from `visualize`) — interactive Plotly HTML plus a markdown twin. Three sections under a shared title/subtitle pattern: step timing breakdown, per-phase GPU memory, and a GPU kernel pie with a top-20 table in an appendix. Kernels under 2% of device time are merged into one "Other" slice.

**`compare`** — Welch's t-test p-value and Cohen's d alongside the means, plus 95% CIs from `scipy.stats`. When both runs were profiled it also diffs the kernel tables, ranked by absolute change, which is what localizes a regression to the launch that caused it.

**`comment`** (new) — renders a result, or a before/after pair, as markdown for a GitHub PR comment. GitHub strips scripts so the Plotly report cannot be embedded; this emits the same findings as tables, with a mermaid pie on single-run comments. Carries a collapsed provenance block (invocation, commit, GPU, torch/CUDA/transformers versions, config) and a hidden marker so a job can update its own comment in place instead of stacking a new one per push. `--label` names a variant when several comments share a PR.

### Notes on the numbers

- Kernel times are device-side self time, so the synchronisation `--profile` adds does not distort them — but the wall-clock timings from the same run are inflated and should not be compared against an unprofiled run.
- The recorded invocation mirrors `train_command.txt`: train args are forwarded verbatim after `--` and cannot be reconstructed from the config dump.

## Test plan

Worked example on #1111, which changes only the fused-loss tile. Benchmarked as two commits differing in nothing else (`ab757675` → `bd39c804`, Qwen3-0.6B, seq_len 2048, A100-80GB, 10 warmup + 50 measured):

```bash
python scripts/benchmark.py run --synthetic --profile --output before.json \
  -- --verifier-name-or-path Qwen/Qwen3-0.6B --total-seq-len 2048
python scripts/benchmark.py compare before.json after.json
python scripts/benchmark.py report after.json
python scripts/benchmark.py comment after.json --baseline before.json --output c.md
```

2.01x faster, 145.3 -> 72.3 ms/step (p<0.001, d=-71.4); the kernel diff attributes it to `loss_forward_kernel` at 73.8 -> 4.3 ms/step while the gemms move under 0.1%, which is the control you want to see. Rendered comments: [before](https://github.com/vllm-project/speculators/pull/1111#issuecomment-5633916880), [after](https://github.com/vllm-project/speculators/pull/1111#issuecomment-5633917507), [diff](https://github.com/vllm-project/speculators/pull/1111#issuecomment-5633765532).

- [x] `run --synthetic`, with and without `--profile`
- [x] `report` on both, profiled (3 sections) and not (2 sections); Plotly bundle loaded exactly once either way
- [x] `compare` with and without kernel data on either side
- [x] `comment` single-run and before/after, posted and edited in place on a real PR
- [x] Old result JSONs predating `ci_95` and the kernel breakdown still load in `report`, `compare` (mixed kernel availability degrades to no kernel section) and `comment`

Not yet done: no CI workflow, since GitHub-hosted runners have no GPU. The marker is emitted but nothing consumes it yet.


orestis-z:
see also generated comments in https://github.com/vllm-project/speculators/pull/1111

orestis-z:
This PR has been split into a 4-PR stack using `gh stack` for targeted, self-contained review:

1. **#1122**: `feat(benchmark): add trainer timing marks, memory tracking, and statistical tests`
2. **#1123**: `feat(benchmark): add torch.profiler support and GPU kernel breakdown`
3. **#1124**: `feat(benchmark): add HTML and Markdown report generation`
4. **#1125**: `feat(benchmark): add 'comment' subcommand for GitHub PR summaries`

PR #1117, opened 2026-09-14
Add launch vllm enable scale out flag

shanjiaz:
## Purpose
VLLM_ENABLE_SCALE_OUT_ENDPOINTS is deprecated upstream in vLLM, switching to --enable-scale-out. More context see https://github.com/vllm-project/vllm/pull/55176

## Tests
Tested locally.

## Checklist

I have filled in:


scripts/launch_vllm.py
```diff
@@ -115,12 +118,25 @@ def render_throughput_defaults(cpus: int | None = None) -> tuple[int, int]:
     return api_servers, DEFAULT_RENDERER_NUM_WORKERS
 
 
+def _vllm_supports_scale_out_flag() -> bool:
+    """Return whether the installed vLLM accepts ``--enable-scale-out``."""
+    from packaging.version import InvalidVersion, Version  # noqa: PLC0415
```
fynnsu: My concern here is that packaging isn't actually part of the python stdlib and so isn't guaranteed to be available here. I also checked and it isn't a vllm dep (just a vllm build dep) so might not be installed. 

Could we use a different approach. Maybe something like this (from codex):

```python

  VLLM_SCALE_OUT_VERSION_BOUNDARY = (0, 29, 1)


  def _vllm_supports_scale_out_flag() -> bool:
      try:
          from vllm import __version_tuple__

          release = tuple(int(part) for part in __version_tuple__[:3])
      except (ImportError, AttributeError, TypeError, ValueError):
          return False

      return release > VLLM_SCALE_OUT_VERSION_BOUNDARY
```

`__version_tuple__` has been available since before `vllm v0.12.0` which is the limit of what we officially support.

fynnsu (approved):

PR #1121, opened 2026-09-15
Add spec-dec demo with identical text and proper metrics

eldarkurtic:
This PR adds a small TUI demo showing side-by-side model generation with and without speculative decoding.

The setup ensures that both runs produce identical text, as they should in theory, while also reporting relevant performance metrics.

Use it as a template when building or showcasing speculative decoding demos in vLLM.

demo_fast


fynnsu (approved):
Thanks for the demo!

I pushed one small change to correct an off-by-one issue, but otherwise looks good.

PR #1130, opened 2026-09-16
Loss weight schedule

fynnsu:
## Purpose

We already support using a combination of loss functions (e.g. `--loss-fn '{"ce": 0.1, "tv": 0.9}'`) which will use both CE and TV losses and weight them based on the float values. This pr adds support for scheduled (based on global step) loss function weighting. You can now specify

```
--loss-fn '{"ce": 0.1, "tv": {"type": "linear", "start": 0.9, "end": 0.1, "start_step": 0, "end_step": 10000}}'
```

Which would use a linear schedule for the tv weight. Note that the format is backwards compatible with the previous implementation.

This doesn't fully solve the issue of shifting loss weights when there are multiple different loss targets (e.g. auxiliary losses for DSpark or DFlash2). I'll add support for that in a future pr. 

## Tests

Added unit tests. 

## Checklist

I have filled in:


src/speculators/losses/utils.py
```diff
@@ -262,7 +378,17 @@ def compound_loss(
...
+        if isinstance(weight, LinearWeightSchedule):
+            if loss_step is None:
+                raise ValueError(
+                    f"Loss term '{name}' has a scheduled weight but no loss_step "
+                    "was provided."
+                )
+            weight_value = weight.value(loss_step)
+        else:
+            weight_value = weight
+        total = total + weight_value * term
```
rahul-tuli, line 391: nit: should we also add something like `term_losses[f"{name}_weight"] = weight_value` (purely for logging)

rahul-tuli (approved):
LGTM!

PR #1133, opened 2026-09-17
fix(hs_connectors): support non-CUDA accelerators in the Mooncake bac…

ys2025-AI:
## Summary
- Add hs_connectors/device.py with accelerator_module() (torch.accelerator-first; CUDA/XPU/NPU/MUSA/MTIA fallback; stream-capability check) and ensure_accelerator_context() (per-process init, fork-safe PID guard).
- Use the shared helper in the Mooncake connector for stream/event creation (replaces hardcoded torch.cuda.*), and initialize the device context before MooncakeHiddenStatesStore.setup() so the Ascend transport can allocate its local segment.
- Fall back to unpinned host memory when pin_memory is unsupported.

## Motivation
On Ascend NPU the Mooncake connector could not run: torch.cuda.Stream() raises without CUDA, and the Mooncake Ascend transport fails with "AscendDirectTransport: cannot allocate local segment" when no device context exists (e.g. in dataloader workers).

## Testing
- tests/unit/hs_connectors: 46 passed
- NPU MooncakeHiddenStatesStore put/get round-trip: OK
- Connector resolves torch_npu.npu; no torch.cuda/_stream_module references

## Checklist

I have filled in:


hs_connectors/src/hs_connectors/device.py
```diff
@@ -0,0 +1,124 @@
...
+        return
+
+    try:
+        module = accelerator_module(need_stream=False)
+    except RuntimeError:
+        # CPU-only process: nothing to initialize (a plain TCP store may still
+        # work, so this is not fatal).
+        return
+
+    device = 0
```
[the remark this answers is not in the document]
ys2025-AI: Fixed in 29dc951. We now prefer `torch.accelerator.current_device_index()`, fall back to `module.current_device()`, and skip `set_device()` when no index is available, so the worker's device is preserved instead of being reset to 0.

hs_connectors/src/hs_connectors/device.py
```diff
@@ -0,0 +1,130 @@
...
+        if module is getattr(torch, name, None):
+            return name
+    return None
+
+
+@lru_cache(maxsize=2)
+def accelerator_module(*, need_stream: bool = True) -> Any:
+    """Return the active accelerator module.
+
+    Prefers ``torch.accelerator`` (torch>=2.6) to detect the active device type
```
fynnsu: speculators requires `torch>=2.9` right now, and recent versions of vllm will similarly require newer versions of pytorch.

I think we can update the pyproject.toml dep for hs_connectors to require `torch>=2.6` and simplify this logic a bit.

hs_connectors/src/hs_connectors/device.py
```diff
@@ -0,0 +1,130 @@
...
+        return
+
+    device: int | None = None
+    acc = getattr(torch, "accelerator", None)
+    if acc is not None and getattr(acc, "is_available", lambda: False)():
+        with contextlib.suppress(Exception):
+            device = acc.current_device_index()
+    if device is None and hasattr(module, "current_device"):
+        with contextlib.suppress(Exception):
+            device = module.current_device()
```
fynnsu: This also seems like very defensive programming. I think setting a minimum torch version could help here too.

fynnsu (commented):
Thanks for the pr! I think we can simplify the logic a good bit by setting a minimum torch version and relying more on the `torch.accelerator` api, but otherwise looks good.

fynnsu (approved):
Thanks for updating!

ys2025-AI:
> Thanks for updating!

@fynnsu Thanks for the review. Could you please help approve and run the workflows when you get a chance?

fynnsu (approved):

PR #1140, opened 2026-09-18
Check DFlash2 layers by capability, not by exact type

### LINKED ISSUE #1137: DFlash2 asserts an exact layer class, making the documented _make_decoder_layer override unusable
`DFlashDraftModel._make_decoder_layer` is documented as the override point for DFlash-family variants, but `DFlash2SpeculatorModel.__init__` then asserts every layer is exactly `Qwen3DFlash2DecoderLayer`, so overriding it raises `AssertionError`.

[`dflash2/core.py#L51-L53`](https://github.com/vllm-project/speculators/blob/main/src/speculators/models/dflash2/core.py#L51-L53)

```python
for layer_ in self.layers:
    assert isinstance(layer_, Qwen3DFlash2DecoderLayer)
    layer_.reset_convolutions()
```

The assertion is stronger than the next line needs — that only requires a layer able to reset its convolutions — and as an `assert` it is stripped under `python -O`.

### Fix

```python
if not hasattr(layer_, "reset_convolutions"):
    raise TypeError(
        f"{type(layer_).__name__} has no reset_convolutions(); a DFlash2 "
        "decoder layer must wrap its sublayers in resettable convolutions."
    )
layer_.reset_convolutions()
```

Same protection, and the documented override works.

### Context

A generic decoder layer with your unmodified `GroupedDynamicCausalConv` as its two sublayer wrappers reproduces `Qwen3DFlash2DecoderLayer` to 4.0e-06 on a 12.6 scale. (The convolutions have to be randomised first — `reset_parameters()` initialises them as identity, so an init-only comparison proves nothing.) This assertion is the only thing blocking it.

Branch ready with DCO sign-off; opening the issue first per CONTRIBUTING.

Hyungkeun-Park:
## Purpose

Fixes #1137.

`DFlashDraftModel._make_decoder_layer` is documented as the override point for DFlash-family variants, but `DFlash2SpeculatorModel.__init__` then asserts every layer is exactly `Qwen3DFlash2DecoderLayer`, so that override raises. The assertion is also stronger than the next line needs — it only requires a layer able to reset its convolutions — and as an `assert` it is stripped under `python -O`.

The `hasattr` check keeps the same protection, names the offending class, and lets the documented override work.

## Tests

```python
import torch

class Resettable(torch.nn.Module):
    def reset_convolutions(self): ...

class NotResettable(torch.nn.Module):
    pass
```

A layer exposing `reset_convolutions()` now passes where it previously raised `AssertionError`; one without it raises `TypeError` naming the class, instead of a bare `AssertionError` that disappears under `python -O`.

`ruff check` / `ruff format --check` show no new violations in the touched file.

## Checklist

I have filled in:


fynnsu (approved):
Thanks for the pr!

rahul-tuli:
Hi @Hyungkeun-Park Thanks for this PR; could you fix quality and rebase on latest main so we can merge this in quickly

